PR #6, opened 2019-04-04
Allow dynamic page titles with a sensible default

mec:
Pages should have unique page titles and this gives us a way to set the
page title for any view by setting the value of page_title in the view
file.

If none is set we will use the service name from the locales file
which this commit also adds.

Wherever the service name appears we should us t(:service_name) so
we can change the service name at any point from a single place.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -30,4 +30,4 @@
 # available at http://guides.rubyonrails.org/i18n.html.
 
 en:
-  hello: "Hello world"
+  service_name: "Teacher’s Payment Service"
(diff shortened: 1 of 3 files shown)


config/locales/en.yml
```diff
@@ -30,4 +30,4 @@
 # available at http://guides.rubyonrails.org/i18n.html.
 
 en:
-  hello: "Hello world"
+  service_name: "Teachers: claim back student loan repayments"
```
robbiepaul: Just asked the team here what service name we should start with, it was suggested we go with `Teacher’s Payment Service`

PR #16, opened 2019-04-04
Rubocop-rails gem has been renamed

robbiepaul:
This will get rid of a deprecation warning when running bundle.


tekin:
Closing as outdated now we've moved to standardrb.

PR #23, opened 2019-04-09
README.md tweaks

tekin:


tekin:
> Just wondering if it is worth mentioning Bullet in the test section, as failing Bullet checks will cause the tests to fail and you may wonder why if you didn't know to look for Bullet and have an N+1?

Yep, definitely worth mentioning that. If you add the mention is added as part of the Bullet PR then we avoid the dependency between the two PRs.

PR #24, opened 2019-04-18
Import schools and local authority data

robbiepaul:
- Add a rake task that imports the required school and local authority data
- Add migrations and models needed for importing the data
- Enables the pgcrypto extension for Postgres so we can use UUIDs instead of incrementing integers as IDs
- Add WebMock, FactoryBot and Shoulda-Matchers to aid with testing

~TBC: We're clarifying the required fields to assert an eligible school with the policy team~


robbiepaul:
We've had a response from the policy team regarding which data from GIAS we need to capture to determine if a teacher teaches at a school which is eligible for the student loan reimbursement claim. 

This PR is ready for review 👍

app/services/school_data_importer.rb
```diff
@@ -0,0 +1,70 @@
...
+    school.county = row["County (name)"]
+    school.postcode = row["Postcode"]
+    school.phase = row["PhaseOfEducation (code)"].to_i
+    school.school_type_group_code = row["EstablishmentTypeGroup (code)"]
+    school.school_type_code = row["TypeOfEstablishment (code)"]
+    school
+  end
+
+  def download_and_save_csv
+    File.open(csv_file_name, "wb") do |saved_file|
```
tekin: Rather than writing to disk explicitly we could simply return the file handle that `open` returns. That way we don't have to worry about cleaning up tempfiles ourselves as Ruby will do it for us.

app/models/school.rb
```diff
@@ -0,0 +1,13 @@
+class School < ApplicationRecord
+  belongs_to :local_authority, required: true
+
+  enum phase: {not_applicable: 0, nursery: 1, primary: 2,
```
tekin: Do we need a presence validation for this field also?

app/models/school.rb
```diff
@@ -0,0 +1,13 @@
...
+               middle_deemed_primary: 3, secondary: 4,
+               middle_deemed_secondary: 5, sixteen_plus: 6, all_through: 7,}
+
+  validates :urn, presence: true
+  validates :name, presence: true
+  validates :school_type_group_code, presence: true
```
tekin: Do we want to have `enum`s defined on these types also to provide some level of validation for the values?

app/models/school.rb
```diff
@@ -0,0 +1,13 @@
+class School < ApplicationRecord
+  belongs_to :local_authority, required: true
```
tekin: Since Rails 5 `belongs_to` associations are required by default so I think the `required` and the validation below are redundant.

app/services/school_data_importer.rb
```diff
@@ -0,0 +1,70 @@
+require "open-uri"
+require "csv"
+
+class SchoolDataImporter
+  def run
+    download_and_save_csv
+
+    CSV.foreach(csv_file_name, headers: true, encoding: "windows-1251:utf-8") do |row|
+      local_authority = LocalAuthority.find_or_create_by(
+        code: row["LA (code)"],
```
tekin: By using `fetch` when accessing this hash where we expect the keys to be present, e.g. `row.fetch("LA (code)")` we can fail early with `Key error` instead of getting `nil` and failing lower down.

app/services/school_data_importer.rb
```diff
@@ -0,0 +1,70 @@
+require "open-uri"
+require "csv"
+
+class SchoolDataImporter
+  def run
+    download_and_save_csv
+
+    CSV.foreach(csv_file_name, headers: true, encoding: "windows-1251:utf-8") do |row|
+      local_authority = LocalAuthority.find_or_create_by(
```
tekin: `LocalAuthority.find_or_create_by!` here so we fail if creation fails?

app/services/school_data_importer.rb
```diff
@@ -0,0 +1,70 @@
...
+        code: row["LA (code)"],
+        name: row["LA (name)"]
+      )
+
+      school = row_to_school(row)
+
+      school.local_authority = local_authority
```
tekin: how about moving the finding/creating of the school and this assignment into the `row_to_school` method to make things more encapsulated.

app/services/school_data_importer.rb
```diff
@@ -0,0 +1,70 @@
...
+
+      unless school.valid?
+        Rails.logger.warn("Failed to import school: #{school.urn} - #{school.errors.full_messages}") && next
+      end
+      school.save
```
tekin: I would be tempted to call `save!` here and not warn on validation failures. If there is data that doesn't validate in our code we definitely want a noisy warning about that. And if it's the source data that is dodgy, we we also probably want to know about that and then make a call on whether we want to skip over schools that don't validate.

app/services/school_data_importer.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+
+    delete_csv_file
+  rescue OpenURI::HTTPError => e
```
tekin: Let's not catch these for now. If the download is failing, we want a noisy failure. and if our intention is to make this a background job then failing noisily will mean that the retry mechanism can kick in and intermittent failures are handled automatically.

spec/services/school_data_importer_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+        "EstablishmentName" => "Penistone Grammer School",
+        "Street" => "123 High Street",
+        "PhaseOfEducation (code)" => "4",
+        "LA (code)" => "40",
+        "LA (name)" => "Blackpool",
+        "EstablishmentTypeGroup (code)" => "10",
+        "TypeOfEstablishment (code)" => "22",
+      }
+    end
+    let!(:request) { stub_request(:get, gias_csv_url).to_return(body: "foo", status: 200) }
```
tekin: We can make this spec more of an integation spec and less brittle to changes of the internal implementation if we return a subset of actual CSV data here from the request. That would remove the need to stub `File` and `CSV` and make it so all the refactoring suggested above could be performed without having to change this spec.

spec/services/school_data_importer_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+        expect(imported_school.name).to eql("Penistone Grammer School")
+        expect(imported_school.street).to eql("123 High Street")
+        expect(imported_school.phase).to eql("secondary")
+        expect(imported_school.school_type_code).to eql(22)
+        expect(imported_school.school_type_group_code).to eql(10)
+        expect(imported_school.local_authority.code).to eql(40)
+        expect(imported_school.local_authority.name).to eql("Blackpool")
+      end
+
+      context "when the school already exists" do
```
tekin: I think a similar spec for when the local authority exists would also be helpful in terms of coverage and confidence the code is doing what we want.

robbiepaul:
I've made all the changes discussed and the tests are passing. However mid-way through running the task I get this error:
```
ActiveRecord::StatementInvalid (PG::CharacterNotInRepertoire: ERROR:  invalid byte sequence for encoding "UTF8": 0xe0 0x20 0x42)
```

It fails on the school with URN `126416` on it's name: `St Thomas \xE0 Becket Church of England Aided Primary School`

db/migrate/20190416104000_create_schools.rb
```diff
@@ -0,0 +1,22 @@
...
+  def change
+    create_table :schools, id: :uuid do |t|
+      t.integer :urn, null: false
+      t.string :name, null: false
+      t.string :street
+      t.string :locality
+      t.string :town
+      t.string :county
+      t.string :postcode
+      t.integer :phase
```
tekin: we should make this `null: false`

PR #27, opened 2019-04-30
A Teacher can start a TSLR claim

tekin:
This is the first step in the claim process for Teacher Student Loan Reimbursements. The aim is to:

- Allow the user to consent before continuing
- Capture the answer to the first question
- Give us a foundation to build out the rest of the journey

For now there is little modelling and abstraction of the multi-page application process as this only implements the first two steps, so the specific URLs/slugs (i.e. 'qts-year' and 'claim-school') are hardcoded for now. I expect this modelling to begin emerging when the next story is worked on (specifying the school that the teacher taught at).

The foundation of this is a RESTful `ClaimsController` that behaves like a [singular resource](https://guides.rubyonrails.org/v5.2/routing.html#singular-resources) for the current in-progress claim. The `show` action is used for the various stages of the claim, with the `slug` representing the current stage.

### DIFF
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -1,4 +1,6 @@
 class ClaimsController < ApplicationController
+  before_action :send_unstarted_claiments_to_the_start, only: [:show, :update]
+
   def new
   end
 
@@ -25,7 +27,11 @@ class ClaimsController < ApplicationController
   end
 
   def current_claim
-    @current_claim ||= TslrClaim.find(session[:tslr_claim_id])
+    @current_claim ||= TslrClaim.find(session[:tslr_claim_id]) if session.key?(:tslr_claim_id)
   end
   helper_method :current_claim
+
+  def send_unstarted_claiments_to_the_start
+    redirect_to root_url unless current_claim.present?
+  end
 end
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -3,8 +3,31 @@ class ClaimsController < ApplicationController
   end
 
   def create
-    TslrClaim.create!
+    claim = TslrClaim.create!
+    session[:tslr_claim_id] = claim.to_param
 
     redirect_to qts_year_claim_path
   end
+
+  def update
+    current_claim.update_attributes(claim_params)
+    redirect_to claim_school_claim_path
+  end
+
+  def qts_year
+  end
+
+  def claim_school
+  end
+
+  private
+
+  def claim_params
+    params.require(:tslr_claim).permit(:qts_award_year)
+  end
+
+  def current_claim
+    @current_claim ||= TslrClaim.find(session[:tslr_claim_id])
+  end
+  helper_method :current_claim
 end
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -1,0 +1,10 @@
+class ClaimsController < ApplicationController
+  def new
+  end
+
+  def create
+    TslrClaim.create!
+
+    redirect_to qts_year_claim_path
+  end
+end
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -8,22 +8,26 @@ class ClaimsController < ApplicationController
     claim = TslrClaim.create!
     session[:tslr_claim_id] = claim.to_param
 
-    redirect_to claim_path(:qts_year)
+    redirect_to claim_path("qts-year")
   end
 
   def show
-    render params[:slug]
+    render claim_page_template
   end
 
   def update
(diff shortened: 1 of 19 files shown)


app/views/claims/qts_year.html.erb
```diff
@@ -1,8 +1,22 @@
-<h1 class="govuk-heading-xl">
-  Which academic year were you awarded qualified teacher status (QTS)?
-</h1>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_for current_claim, url: claim_path do |form| %>
+      <h1 class="govuk-heading-xl">
```
mec: If we wrap the question in a `<fieldset>` and the title in a `<legend>` we make a few accessibility wins. See https://design-system.service.gov.uk/components/fieldset/ for details and the class names.

app/views/claims/qts_year.html.erb
```diff
@@ -1,8 +1,22 @@
...
-  You are only eligable to claim back student loan repayments if you qualified
-  after September 1st 2013.
-</p>
+      <span id="tslr_claim_qts_award_year-hint" class="govuk-hint">
+        You are only eligable to claim back student loan repayments if you qualified
+        on or after September 1st 2013.
+      </span>
+
+      <div class="govuk-form-group">
+        <%= form.select :qts_award_year, options_for_qts_award_year, { include_blank: true }, class: "govuk-select" %>
```
mec: The blank default selection feels odd to me, but we have already recognised that a select here may not be the best approach to answer this question, but that we need more/better user research so we should continue with what we have.

app/controllers/claims_controller.rb
```diff
@@ -3,8 +3,31 @@ def new
   end
 
   def create
-    TslrClaim.create!
+    claim = TslrClaim.create!
+    session[:tslr_claim_id] = claim.to_param
```
mec: Is there anything we should consider setting about our session cookies for the app? Is there a defualt expiry time that may not meet our needs? 

As I understand it, Rails stores the cookies securely so that is not an issue?

app/controllers/claims_controller.rb
```diff
@@ -3,8 +3,31 @@ def new
...
+  private
+
+  def claim_params
+    params.require(:tslr_claim).permit(:qts_award_year)
+  end
+
+  def current_claim
+    @current_claim ||= TslrClaim.find(session[:tslr_claim_id])
+  end
+  helper_method :current_claim
```
mec: Is this a prefered style for the place to declare these `helper_methods` I see them alot at the top of the file?

config/routes.rb
```diff
@@ -3,5 +3,7 @@
   get "start/index"
   root "claims#new"
 
-  resources :claims, only: [:new, :create, :show, :update], param: :slug
+  constraints slug: /qts_year|claim_school/ do
```
mec: The `contraints` is a regex so this is saying 'qts_year or claim_school' only, just sanity checking my understanding

app/controllers/claims_controller.rb
tekin: Yep, so rather than implement this as part of this story I've added a post-it to the user story map specifically for this:

IMG_3423

It will need planning/prioritising and probably some service design work to define the exact flow (e.g. what content do we show the user that has their session expire on them? Do we pre-warn the user about the expiring session?)

app/controllers/claims_controller.rb
tekin: It's not a rigid convention, but generally folks place them together somewhere near the top off the controller code. Personally I do like having them near the method definition but I'm not wedded to it.

config/routes.rb
tekin: yep, that's correct.

app/views/claims/qts_year.html.erb
tekin: Hmmm, so my reading of that is that the fieldset is to be used when _"you need to show a relationship between multiple form inputs"_. Given that this is a single form input, does it make sense to do that?

app/views/claims/qts_year.html.erb
tekin: Yep, I expect us to iterate on this question in the future.

db/migrate/20190416090236_create_tslr_claims.rb
```diff
@@ -0,0 +1,9 @@
+class CreateTslrClaims < ActiveRecord::Migration[5.2]
+  def change
+    create_table :tslr_claims do |t|
```
mec: We talked about this having a UID.

app/views/claims/qts_year.html.erb
mec: 👍

PR #31, opened 2019-05-09
Fix sequential migration dependency error

robbiepaul:
When running `rails db:migrate` I got the error `PG::UndefinedFunction: ERROR:  function gen_random_uuid() does not exist`

After inspection it looks like the the `CreateTslrClaims` migration happens before we enable the pgcrypto extension causing the error.

The fix puts the `CreateTslrClaims` migration in the correct place that it occurred sequentially. As it is the most recent migration we need to point the schema to it as well.


robbiepaul:
On second thoughts, after speaking to Tekin this may cause issues on staging with the way Rails tracks migrations in a separate table.

We'll just need to bare this in mind if future developers run in to this error they should run `rails db:schema:load`

PR #32, opened 2019-05-09
(39) Teacher specifies claim school

mec:
So that a teacher can locate a school easily we allow them to search by school name.

We limit the search string to 4 characters on more to limit the number of returned schools.  Once a list of possible schools is displayed the teacher selects their school from the list by choosing a radio button. 

The radio buttons are displayed with the school name and the school address to allow the teacher to distinguise between schools with similar or identical names.

If the teacher is unable to locate their desired school, they can restart the search process to try again.

Our ambition is to use a Javascript autocomplete to progressively enhance this feature, this PR does not include this work.

## Updated Gem
This PR includes a version change on the `dxw_govuk_frontend_rails` gem to introduce the new secondary button pattern that we now use.

Story card: 

https://trello.com/c/W9G3hkfy
## Screenshots

School search:
Screenshot 2019-05-09 at 14 38 38

School search name length validation: 
Screenshot 2019-05-09 at 14 38 45

School search selection:
Screenshot 2019-05-09 at 14 38 50

School search no selection:
Screenshot 2019-05-09 at 14 40 33

School search no results:
Screenshot 2019-05-09 at 14 39 09

### DIFF
--- a/app/views/claims/claim_school.html.erb
+++ b/app/views/claims/claim_school.html.erb
@@ -33,8 +33,9 @@
             <div class="govuk-radios">
               <%= form.collection_radio_buttons :claim_school_id, @schools, :id, :name do |b| %>
                 <div class="govuk-radios__item">
-                  <%= b.radio_button( class: "govuk-radios__input") %>
-                  <%= b.label class: "govuk-label govuk-radios__label"  %>
+                  <%= b.radio_button class: "govuk-radios__input" %>
+                  <%= b.label class: "govuk-label govuk-radios__label govuk-label--s" %>
+                  <span class="govuk-hint govuk-radios__hint"><%= b.object.address %></span>
                 </div>
               <% end %>
             </div>
--- a/app/views/claims/claim_school.html.erb
+++ b/app/views/claims/claim_school.html.erb
@@ -4,6 +4,23 @@
       Which school were you employed at between 6 April XXXX and 5 April XXXX?
     </h1>
 
-    <p class="govuk-body">Coming next...</p>
+    <%= form_tag(claim_path("claim-school"), method: :get) do %>
+      <div class="govuk-form-group">
+        <%= label_tag :school_search, "School name", class: "govuk-label" %>
+        <%= text_field_tag :school_search, params[:school_search], class: "govuk-input" %>
+      </div>
+
+      <div class="govuk-form-group">
+        <%= submit_tag "Search", class: "govuk-button" %>
+      </div>
+    <% end %>
+
+    <% if @schools.present? %>
+      <ul>
+        <% @schools.each do |school| %>
+          <li><%= school.name %></li>
+        <% end %>
+      </ul>
+    <% end %>
   </div>
 </div>
--- a/app/views/claims/claim_school.html.erb
+++ b/app/views/claims/claim_school.html.erb
@@ -4,23 +4,43 @@
       Which school were you employed at between 6 April XXXX and 5 April XXXX?
     </h1>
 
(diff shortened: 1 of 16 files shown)


app/views/claims/claim_school.html.erb
```diff
@@ -1,9 +1,59 @@
...
+      </span>
+      <%= form_for current_claim, url: claim_path do |form| %>
+        <%= hidden_field_tag :school_search, params[:school_search] %>
+
+        <div class="govuk-form-group">
+          <fieldset class="govuk-fieldset" aria-describedby="school-search-result-hint">
+            <div class="govuk-radios">
+              <%= form.collection_radio_buttons :claim_school_id, @schools, :id, :name do |b| %>
+                <div class="govuk-radios__item">
+                  <%= b.radio_button( class: "govuk-radios__input") %>
```
robbiepaul: We could lose the perens here to make it consistent with the rest of the markup?

app/views/claims/claim_school.html.erb
```diff
@@ -1,9 +1,59 @@
...
     <h1 class="govuk-heading-xl">
       Which school were you employed at between 6 April XXXX and 5 April XXXX?
     </h1>
 
-    <p class="govuk-body">Coming next...</p>
+    <% if @schools.blank? %>
+
+      <% if params[:school_search].present? %>
+        <p class="govuk-body">
+          <strong>No results match that search term. Try again.</strong>
```
robbiepaul: If I perform a search for 'pen' I see the validation errors but also 'No results match that search term' - this could be confusing to users. Could we check that they did a valid search before showing this message?

robbiepaul:
Another issue I've thought of which we may want to address in a separate story but if you search for a school that has been converted in to an academy it will show up twice. Once as the closed school and once as the new open school.

For example, search for 'marlborough junior school' - two schools with the same name and address show up

app/views/claims/claim_school.html.erb
mec: 👍

tekin:
> Another issue I've thought of which we may want to address in a separate story but if you search for a school that has been converted in to an academy it will show up twice. Once as the closed school and once as the new open school.
> 
> For example, search for 'marlborough junior school' - two schools with the same name and address show up

Great catch. Let's capture this on the story map and mention it to the rest of the team today.

spec/fixtures/schools.yml
```diff
@@ -0,0 +1,13 @@
...
+  urn: 1234
+  phase: :secondary
+  school_type_group: :la_maintained
+  school_type: :community_school
+
+hampstead_school:
+  name: "Hampstead School"
+  urn: 4567
+  phase: :secondary
+  school_type_group: la_maintained
```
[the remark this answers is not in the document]
tekin: They should probably all be strings actually. The YAML file should map attributes to database columns, and so any symbols will end up converted to strings anyway.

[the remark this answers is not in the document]
mec: 👍

app/views/claims/claim_school.html.erb
mec: Great spot! My thinking here is because the validation for this search is in the controller and if the validation fails we add an error to the current claim I can do:

`<% if params[:school_search].present? && current_claim.errors.empty? %>`

Rather than duplicating the length check in the view, what do you think?

mec:
👍

app/views/claims/claim_school.html.erb
tekin: @mec that sounds like a sensible approach to me.

PR #33, opened 2019-05-10
Determine if a given school meets the TSLR eligibility criteria

robbiepaul:
To be eligible for TSLR payments a teacher must've taught in a school that meets the following criteria:

- State funded school in England
- Local Authority maintained secondary (including middle deemed secondary)
- Local Authority maintained or non-maintained special schools
- Secondary phase academy/free school
- In one of the 25 participating local authorities

Using the data from the Get Information About Schools database and after talking with the Financial Incentives policy team we can make these assertions about school eligibility:

- State funded school means any school that isn't independent, a university, a welsh school or 'Other'
- Local Authority maintained, Colleges, Academies and Free schools are only eligible if they have a phase of secondary or middle deemed secondary
- All special schools are eligible except independent special schools.
  Note: The only way to filter out independent special schools is with the 'school type' field (they will have a 'school type group' of 'special_schools')

We may also want to capture the date a school closed and add a check to see if the school closed before the year the teacher is claiming for, however this wasn't part of the official criteria so I haven't included it in this PR. This means that it will be possible for a teacher to select a school that otherwise would be eligible but closed last decade.

### DIFF
--- a/spec/services/tslr/school_eligibility_spec.rb
+++ b/spec/services/tslr/school_eligibility_spec.rb
@@ -1,0 +1,103 @@
+require "rails_helper"
+
+RSpec.describe Tslr::SchoolEligibility do
+  describe "#check" do
+    subject { Tslr::SchoolEligibility.new(school).check }
+    let(:school) { build(:school, school_attributes.merge({local_authority: local_authority})) }
+
+    context "when it is in an eligible area" do
+      let(:local_authority) { build(:local_authority, :eligible) }
+
+      context "and it is an academy" do
+        let(:academy_school_attributes) { {school_type_group: :academies} }
+
+        context "and it has an education phase of secondary or middle deemed secondary" do
+          let(:school_attributes) { academy_school_attributes.merge({phase: :secondary}) }
+          it { is_expected.to be true }
+        end
+
+        context "and it has an education phase of nursery, primary, middle deemed primary, 16+ or all through" do
+          let(:school_attributes) { academy_school_attributes.merge({phase: :primary}) }
+          it { is_expected.to be false }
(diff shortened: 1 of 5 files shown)


robbiepaul:
@tekin I've made some changes since our discussion earlier:

- Put the Local Authority names in comments next to the codes
- Simplified the `#check` method to be more expressive about eligibility criteria.
- Added more permutations of schools to the specs
- Squashed the commits and reworded the assertions in the commit message to make more sense

spec/services/tslr/school_eligibility_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+
+      context "and it is not a state funded school" do
+        let(:school_attributes) { {school_type_group: :independent_schools} }
+        it { is_expected.to be false }
+      end
+    end
+
+    context "when it is not in an eligible area" do
+      let(:local_authority) { build(:local_authority, :ineligible) }
+      let(:school_attributes) { {} }
```
tekin: For this spec to be a bit more legit, I think the school attributes should be for an otherwise eligible school, i.e. state funded in an eligible phase. Otherwise it might be failing eligibility for any one of the other checks.

tekin: Ah, sorry, just noticed these specs are using the factory that already has eligible attributes, so that isn't necessary.

robbiepaul:
I've spoken to Chris Cuff from the policy team. It turns out that special schools also have to have the education phase of 'secondary' or 'middle deemed secondary' as well. That simplifies our code slightly, I've updated the specs to reflect the change.

PR #34, opened 2019-05-14
A teacher can say whether they are still employed at a school

mec:
So a teacher can confirm their eligibility, we need to know they are still teaching and we would like to know at which shcool.

We've added a postgres enum to the database to store the teachers 'employment status', this has meant switching schema formats to sql rather than the usual rb.

We decided to duplicate the still teaching question view for now to keep our approach as simple as possible and our views clearly named, we like the idea of one view per question.

The sequence of pages shown to a teacher is stored in the model and we've laid the groundwork to enable us to skip pages based on certain criteria.

If a teacher is still teaching at the same school as they are claiming for, we still wanted to store that school in the `current_school` association, we've done this in the model with a before_save callback, this led to a problem with Bullet that we've not resolved yet, so the `unused_eager_loading` is now disabled on the current school association in the testing environment, see 170450385ab2582e896d720bf9e037ae6054c951

This work does not include checking an eligibility, this will come later.

## I'm concerned about
- We should look at the Bullet issue and confirm we are happy to ignore the error it it raising.
- Whilst I am confident that disabling the 'field_with_error' css class is the right call now, I'm worried about impact elsewhere or that we forget it's been switched off, perhaps makeing a card in the backlog is a good idea?

## Screenshots
Question:
Screenshot_2019-05-14 Teachers claim back student loan repayments

Searching for current school:
Screenshot_2019-05-14 Teachers claim back student loan repayments(2)
Screenshot_2019-05-14 Teachers claim back student loan repayments(3)

Stubbed next question:
Screenshot_2019-05-14 Teachers claim back student loan repayments(1)

### DIFF
--- a/app/models/tslr_claim.rb
+++ b/app/models/tslr_claim.rb
@@ -23,6 +23,7 @@ class TslrClaim < ApplicationRecord
   }, _prefix: :employed_at
 
   belongs_to :claim_school, optional: true, class_name: "School"
+  belongs_to :current_school, optional: true, class_name: "School"
 
   validates :claim_school,      on: :"claim-school", presence: {message: "Select a school from the list"}
   validates :qts_award_year,    on: :"qts-year", inclusion: {in: VALID_QTS_YEARS, message: "Select the academic year you were awarded qualified teacher status"}
--- a/app/models/tslr_claim.rb
+++ b/app/models/tslr_claim.rb
@@ -9,6 +9,12 @@ class TslrClaim < ApplicationRecord
     "2019-2020",
   ].freeze
 
+  enum employment_status: {
+    claim_school: 0,
+    different_school: 1,
+    no_school: 2,
+  }, _prefix: :employed_at
+
   belongs_to :claim_school, optional: true, class_name: "School"
   validates :claim_school, presence: {message: "Select a school from the list"}, on: :"claim-school"
 
--- a/app/models/tslr_claim.rb
+++ b/app/models/tslr_claim.rb
@@ -30,11 +30,17 @@ class TslrClaim < ApplicationRecord
   validates :qts_award_year,    on: :"qts-year", inclusion: {in: VALID_QTS_YEARS, message: "Select the academic year you were awarded qualified teacher status"}
   validates :employment_status, on: :"still-teaching", presence: {message: "Choose the option that describes your current employment status"}
 
+  before_save :update_current_school, if: :employment_status_changed?
+
   delegate :name, to: :claim_school, prefix: true, allow_nil: true
 
   def page_sequence
     PAGE_SEQUENCE.dup.tap do |sequence|
       sequence.delete("current-school") if employed_at_claim_school?
     end
+  end
+
+  def update_current_school
+    self.current_school = employed_at_claim_school? ? claim_school : nil
   end
 end
--- a/app/models/tslr_claim.rb
+++ b/app/models/tslr_claim.rb
@@ -3,6 +3,7 @@ class TslrClaim < ApplicationRecord
     "qts-year",
     "claim-school",
     "still-teaching",
+    "current-school",
     "subject-percentage",
   ].freeze
 
@@ -30,4 +31,10 @@ class TslrClaim < ApplicationRecord
   validates :employment_status, on: :"still-teaching", presence: {message: "Choose the option that describes your current employment status"}
 
   delegate :name, to: :claim_school, prefix: true, allow_nil: true
+
+  def page_sequence
+    PAGE_SEQUENCE.dup.tap do |sequence|
+      sequence.delete("current-school") if employed_at_claim_school?
+    end
+  end
 end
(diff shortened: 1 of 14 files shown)


spec/features/tslr_claim_spec.rb
```diff
@@ -22,7 +22,52 @@
...
+    click_on "Agree and continue"
+
+    claim = TslrClaim.order(:created_at).last
+
+    expect(page).to have_text("Which academic year were you awarded qualified teacher status")
+    select "September 1 2014 - August 31 2015", from: :tslr_claim_qts_award_year
+    click_on "Continue"
+
+    expect(claim.reload.qts_award_year).to eql("2014-2015")
+    expect(page).to have_text("Which school were you employed at between")
```
robbiepaul: As this spec is testing the different school question, do you think making the assertions above make our tests more brittle? We already cover them in our main 'happy path' spec above this one.
If for some reason we weren't on the right page at this point we would fail anyway as the form fields would be missing.

app/models/tslr_claim.rb
```diff
@@ -9,8 +17,30 @@ class TslrClaim < ApplicationRecord
     "2019-2020",
   ].freeze
 
+  enum employment_status: {
```
robbiepaul: Is there a reason why we're using these Postgres enums as opposed to the standard ActiveRecord enums? The only advantage I can think of is that we're storing a string instead of an integer but when we get the data back from our model ActiveRecord will convert the integer back to a string anyway.

For example, with the code:
```ruby
enum employment_status: [:claim_school, :different_school, :no_school]
```
You could then do
```ruby
claim.employment_status = "claim_school"
```
So `claim.employment_status` would return the string `claim_school` even though it's stored as an integer.

It feels like we're going against the grain slightly with Postgres enums and changing the schema format to SQL could cause issues if we're on slightly different versions of Postgres? For example if I run any migrations a few unrelated lines get removed from the `structure.sql`:

```sql
--
-- Name: plpgsql; Type: EXTENSION; Schema: -; Owner: -
--

CREATE EXTENSION IF NOT EXISTS plpgsql WITH SCHEMA pg_catalog;


--
-- Name: EXTENSION plpgsql; Type: COMMENT; Schema: -; Owner: -
--

COMMENT ON EXTENSION plpgsql IS 'PL/pgSQL procedural language';
```
I'm on Postgres 11.1

mec: It's a fair point. I think there are three things we wanted to achieve:

1. use a hash to free the values from the array order
2. use a prefix so we can write `employed_at_claim_school?`, `employed_at_no_school?` and `employed_at_different_school?` which feels more expressive
3. add meaning and security at the db level by storing the values instead of meaningless integers 

I agree the tech overhead seems high, perhaps @tekin can speak to that call?  

This is the prior art we looked at, which is worth a read:

https://naturaily.com/blog/ruby-on-rails-enum

tekin: I think this is a good point worth discussing.

> It's a fair point. I think there are three things we wanted to achieve:
> 
> 1. use a hash to free the values from the array order
> 2. use a prefix so we can write `employed_at_claim_school?`, `employed_at_no_school?` and `employed_at_different_school?` which feels more expressive

I think this point is moot, we can still use `prefix` and have those expressive boolean methods without using a custom postgres enum type.

> 3. add meaning and security at the db level by storing the values instead of meaningless integers

In my mind, this is the main advantage: that it makes it so the actual value in the database makes more sense. e.g. `"claim_school"` has more meaning and communicates more than `1`. This would be useful say in a situation say where we were exporting all the claims and wanted to do so without the overhead of going through the active record models (@mec as we did [recently on the CCS project](https://github.com/dxw/DataSubmissionServiceAPI/blob/develop/app/models/export/submissions/extract.rb).

> 
> I agree the tech overhead seems high, perhaps @tekin can speak to that call?
> 
> This is the prior art we looked at, which is worth a read:
> 
> https://naturaily.com/blog/ruby-on-rails-enum

As with everything, it's all about tradeoffs! On balance perhaps it's safer to avoid the custom enum type for now. The cost to us is we end up with an ordered array of numbers mapping to a logical value/concept in the database. Perhaps that's OK for now to avoid tying ourselves to a database-specific extension. We can meet halfway by at least being explicit with the integer values we map to:

```ruby
 enum employment_status: {  
  claim_school: 0,
  different_school: 1,
  no_school: 2,
}, _prefix: :employed_at
```

spec/features/tslr_claim_spec.rb
mec: Rather than testing the question like a unit, I feel like this test is to give us confindence that the journey for a user with a different school is working end to end.

We talked about removing the duplication in these tests but decided to let it stand for now as the feature test suite grows we feel like we'll reach a point that this will get refactored.

app/models/tslr_claim.rb
mec: I'm happy to drop the DB changes. :+1:

spec/features/tslr_claim_spec.rb
robbiepaul: Yeah I agree with testing it end-to-end, I was just curious about duplicating the `expect` parts. Happy to leave it for now and look at refactoring later if it needs it

PR #37, opened 2019-05-16
An administrative user can authenticate themselves to access an "admin" area of the service

robbiepaul:
So that DfE staff in the Financial Incentives Team will be able to access the data relating to claims and process them, we need an authentication mechanism that will grant them access to a secure part of the application.

This story is to do the initial plumbing that will allow a user to sign in using DfE Sign-in to access an "/admin" part of the application. For now, this will simply contain a holding page with a "sign-out" link that ends the session.

The identifier and secret are stored in Heroku. Place them in a `.env.local` file and run `bin/start` to start the dev server with https.

Note: Anyone with access to the 'Teacher Payments' service in DfE Sign In will be able to sign in, we currently don't check which organisation is returned. We may want to do this check in the future depending on how access to services is managed.

https://trello.com/c/A7865IBw


config/environments/development.rb
```diff
@@ -52,10 +52,26 @@
   # Suppress logger output for asset requests.
   config.assets.quiet = true
 
+  # Force https
+  config.force_ssl = true
```
erbridge: Does this also enable HSTS on local machines? If so, it makes it hard to run other things on `localhost` that don't use HTTPS.

docs/architecture-decisions/0007-use-dfe-sign-in-for-admin-authentication.md
```diff
@@ -0,0 +1,26 @@
...
+
+We will use DfE's single sign-on service [DfE Sign In](https://services.signin.education.gov.uk/)
+and authenticate through OpenID Connect.
+
+## Consequences
+
+DfE Sign In accounts can only be created by a small number of 'approvers' in
+DfE. 
+
+Developers will be required to run development with https as DfE Sing In
```
mec: Type here, 'Sing'

config/environments/development.rb
mec: Looks like we can disable the HSTS part, by setting:

`config.ssl_options = { hsts: false }`

https://api.rubyonrails.org/v5.2.3/classes/ActionDispatch/SSL.html

@erbridge do you fancy exlpaining why HSTS could be an issue? I've never come accross it before?

erbridge: HSTS basically tells your browser to always redirect that URL to HTTPS. This is fine in real environments, since once you add SSL, you rarely remove it, but for `localhost`, it means that your browser will try to upgrade all `http://localhost` requests to `https://localhost`, which might not be there. And it can be a real pain to get your browser to forget that once it's happened.

robbiepaul: @erbridge Good catch, I've just checked and local https will still work without this config setting so I've removed it

config/initializers/omniauth.rb
```diff
@@ -0,0 +1,24 @@
+OmniAuth.config.logger = Rails.logger
+
+dfe_sign_in_issuer_uri    = URI(ENV.fetch("DFE_SIGN_IN_ISSUER"))
```
tekin: I guess calling `fetch` here means the app will fall over in environments where these variables are not defined, e.g. the current temporary prod environment on Heroku. Just noting that here as it means we'll have to add the variables to the prod environment before this code is promoted there.

bin/start
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
tekin: For the sake of following a more Rails-ish convention, maybe we should use a `Procfile` with foreman instead of a custom bash script for this. We can create a `Procfile.dev` and a [`.foreman`](https://ddollar.github.io/foreman/#DEFAULT-OPTIONS) file to point to this file as the default to use locally without confusing Heroku with local-development only stuff in a generic`Procfile`.

README.md
```diff
@@ -20,6 +20,24 @@ Architecture decision records can be found in the
 2. Run `bundle exec rails db:setup` to set up the database development
 3. Run `bundle exec rails server` to launch the app on http://localhost:3000
 
+### DfE Sign In credentials
+
+By default in development OmniAuth will run in test mode. This means that you
```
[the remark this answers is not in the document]
robbiepaul: Ah, I have that setting on in Rubymine but it seems a bit inconsistent. I'll have a look if there's a way I can highlight trailing whitespace

bin/start
robbiepaul: Thanks for this, I've learnt something new today 👍 I've amended the commit with the change

robbiepaul:
I've added the refactor as a separate commit for now to make reviewing easier. If you're happy with it I'll squash it together with the other commit.

The main side effect of the change is that we'll need to request to get our callback URLs changed with DfE Sign In, which is fine I just thought I'd note it here.

README.md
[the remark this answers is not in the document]
robbiepaul: It turns out the official Rubymine Markdown plugin was overriding this setting, after disabling the plugin it's now stripping the whitespace off the end of the line in markdown files 👍

README.md
```diff
@@ -16,9 +16,36 @@ Architecture decision records can be found in the
...
+   development certificate.
+   - Run `openssl req -x509 -sha256 -nodes -newkey rsa:2048 -days 365 -keyout localhost.key -out localhost.crt`
+   - Open Keychain Access on your Mac and go to the Certificates category in
+     your System keychain. Once there, import the `localhost.crt`
+     `File > Import Items`. Double click the imported certificate and change the
+     “When using this certificate:” dropdown to `Always Trust` in the Trust
+     section.
+2. Run `bundle install` to install the gem dependencies
+3. Run `bundle exec rails db:setup` to set up the database development
+4. Run `foreman start` to launch the app on https://localhost:3000
```
erbridge: Any chance we could binstub `foreman`?

config/environments/development.rb
erbridge: We should be forcing SSL in production, though. How about something like:
```
config.force_ssl = true
config.ssl_options = { hsts: Rails.env.production? }
```

README.md
tekin: I've suggested we add it to the `Gemfile` `development` group so we can run it with bundler instead.

robbiepaul: I've updated the readme to say `bundle exec foreman start`

robbiepaul:
I'm gonna hold off merging this PR until the callback URL has changed in DfE Sign In, I've requested it via email (30/5). I'll also add the missing environment vars to Heroku

robbiepaul:
I thought I would do one last test of this locally and it seems that `foreman` ignores the `.env.local` file, preferring `.env`. 

So I've made some changes to reflect that: 
- added `.env` to the gitignore and updated readme
- added `.env.development` (for running the app without DfE Sign In) and `.env.test` (for CI)

The redirect URL has been changed on DfE Sign In and I've added the environment vars to Heroku staging, so this is ready to go

PR #39, opened 2019-05-16
A teacher can provide their teacher reference number

mec:
So that we can pay a teacher their teacher student loan re-imbursement payment we need to make sure they are really a teacher. Collecting their TRN is one of the ways we will achieve this.

At this stage we are validating the presence of the TRN and normalising it to digits only before saving it to the database.

At the moment we think that storing only digits is the right thing to do based on our understanding of TRNs, the DQT uses only digits to sign teacher in. We have contacted a subject matter expert at DfE to gain more understanding.

https://trello.com/c/01yTwht6

### DIFF
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -63,7 +63,8 @@ class ClaimsController < ApplicationController
       :address_line_3,
       :address_line_4,
       :postcode,
-      :date_of_birth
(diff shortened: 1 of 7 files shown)


mec:
We had a response back from someone at DfE, they weren't the right person but we have the name of someone who may be, we'll close this PR while we see what they say about TRNs.

PR #40, opened 2019-05-20
A teacher can provide their Teacher Reference Number (TRN)

robbiepaul:
So that we can pay a teacher their teacher student loan reimbursement payment we need to make sure they are really a teacher. Collecting their TRN is one of the ways we will achieve this.

At this stage we are validating the presence of the TRN and normalising it to digits only before saving it to the database.

We have contacted a subject matter expert at DfE and  now understand that TRNs are 7 digits long.

https://trello.com/c/01yTwht6

### DIFF
--- a/app/models/tslr_claim.rb
+++ b/app/models/tslr_claim.rb
@@ -41,6 +41,7 @@ class TslrClaim < ApplicationRecord
   validates :teacher_reference_number, on: :"teacher-reference-number", presence: {message: "Enter your teacher reference number"}
 
   before_save :update_current_school, if: :employment_status_changed?
+  before_save :normalise_teacher_reference_number, if: :teacher_reference_number_changed?
 
   delegate :name, to: :claim_school, prefix: true, allow_nil: true
 
@@ -66,5 +67,9 @@ class TslrClaim < ApplicationRecord
 
   def update_current_school
     self.current_school = employed_at_claim_school? ? claim_school : nil
+  end
+
+  def normalise_teacher_reference_number
+    teacher_reference_number.gsub!(/\D/, "")
   end
 end
--- a/app/models/tslr_claim.rb
+++ b/app/models/tslr_claim.rb
@@ -7,6 +7,7 @@ class TslrClaim < ApplicationRecord
     "full-name",
     "address",
     "date-of-birth",
+    "teacher-reference-number",
     "complete",
   ].freeze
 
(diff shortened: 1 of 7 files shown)


app/models/tslr_claim.rb
```diff
@@ -20,6 +21,8 @@ class TslrClaim < ApplicationRecord
     "2019-2020",
   ].freeze
 
+  TRN_LENGTH = 7
```
tekin: We should explicitly `freeze` this constant to make it immutable.

app/models/tslr_claim.rb
```diff
@@ -65,4 +71,16 @@ def ineligible_claim_school?
...
+
+  def normalise_trn
+    self.teacher_reference_number = normalised_trn
+  end
+
+  def normalised_trn
+    teacher_reference_number.gsub(/\D/, "")
+  end
+
+  def normalised_trn_length
```
tekin: I think this method could better explain what it is doing. At the moment it kinda reads like it returns the length of a normalised TRN, and also the "normalised" bit feels more like a side effect of the implementation, rather than the behaviour we are trying to capture. How about something like `trn_must_be_seven_digits` instead?

app/models/tslr_claim.rb
```diff
@@ -65,4 +71,16 @@ def ineligible_claim_school?
...
+    errors.add(:teacher_reference_number, "Teacher reference number should contain seven digits") unless teacher_reference_number.present? && normalised_trn.length == TRN_LENGTH
```
tekin: Small language tweak: I think the error message should use `must` instead of `should` as the length of the TRN is a hard requirement, not something that is suggestive or optional.

app/models/tslr_claim.rb
```diff
@@ -37,8 +40,11 @@ class TslrClaim < ApplicationRecord
   validates :address_line_3,    on: :address, presence: {message: "Enter your town or city"}
   validates :postcode,          on: :address, presence: {message: "Enter your postcode"}
   validates :date_of_birth,     on: :"date-of-birth", presence: {message: "Enter your date of birth"}
+  validates :teacher_reference_number, on: :"teacher-reference-number", presence: {message: "Enter your teacher reference number"}
+  validate :normalised_trn_length, on: :"teacher-reference-number"
```
tekin: This has me wondering if we _always_ want to validate the format of the TRN, rather than just in the `"teacher-reference-number"` scope. i.e. what happens if it gets changed elsewhere for whatever reason? would the system be safer if we validate the format of the TRN as long as it isn't `nil`?

app/models/tslr_claim.rb
mec: Phew, that's what we thought!

app/models/tslr_claim.rb
mec: Great call, we talk this through and think this is the right approach.

PR #41, opened 2019-05-20
A teacher can provide their National Insurance number

robbiepaul:
We need to capture the teacher's National Insurance number so that we can operate PAYE on their behalf.

We are validating the NI number is present and conforms to the correct format. We are removing whitespace before saving it to the database.

https://design-system.service.gov.uk/patterns/national-insurance-numbers/
https://trello.com/c/rwg4j8qH/

### DIFF
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -64,7 +64,8 @@ class ClaimsController < ApplicationController
       :address_line_4,
       :postcode,
       :date_of_birth,
-      :teacher_reference_number
+      :teacher_reference_number,
+      :national_insurance_number,
     )
   end
 
--- a/spec/models/tslr_claim_spec.rb
+++ b/spec/models/tslr_claim_spec.rb
@@ -90,6 +90,13 @@ RSpec.describe TslrClaim, type: :model do
     end
   end
 
+  context "when saving in the “national-insurance-number” validation context" do
+    it "validates the presence of national_insurance_number" do
+      expect(TslrClaim.new).not_to be_valid(:"national-insurance-number")
+      expect(TslrClaim.new(national_insurance_number: "QQ123456C")).to be_valid(:"national-insurance-number")
+    end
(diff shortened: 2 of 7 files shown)


app/controllers/claims_controller.rb
```diff
@@ -64,7 +64,8 @@ def claim_params
       :address_line_4,
       :postcode,
       :date_of_birth,
-      :teacher_reference_number
+      :teacher_reference_number,
+      :national_insurance_number
```
tekin: We should probably get that trailing comma in to stop the red/green diff juggling. I wonder why Standard doesn't complain about this one? Because it's not explicitly wrapped in square brackets?

spec/models/tslr_claim_spec.rb
```diff
@@ -90,6 +90,39 @@
...
 
+  context "when saving in the “national-insurance-number” validation context" do
+    it "validates the presence of national_insurance_number" do
+      expect(TslrClaim.new).not_to be_valid(:"national-insurance-number")
+      expect(TslrClaim.new(national_insurance_number: "QQ123456C")).to be_valid(:"national-insurance-number")
+    end
+  end
+
+  context "when saving a record that has a National Insurance number" do
+    it "validates that the first two characters are letters" do
```
tekin: It looks like these `it` blocks may have been iteratively written as the code was implemented in a TDD-style perhaps? That's 💯! That said, once the code is written and we have green, it can sometimes make sense to perform a refactor step on the tests too, as TDD can result in more tests than is strictly necessary. We want just enough to give us the confidence the code is doing what we want whilst keeping the number of tests to a minimum to avoid test bloat. Perhaps the refactor step here is to simplify this context down to a single `it "validates the National Insurance number is the correct format"` and bung a handful of valid and invalid NI numbers in there. Because the format for an NI number is established and well documented out on the internet, we don't necessarily need our tests to document it too.

PR #42, opened 2019-05-20
Disable unused ActiveStorage feature

robbiepaul:
It's possible to directly upload files with ActiveStorage, we should disable this as we'll probably never use it. We can always re-enable it later.


tekin:
👍

PR #51, opened 2019-05-21
Limit the school search to 50 results

robbiepaul:
So that we can help avoid any smart DDoS attempts by bad actors we limit the search results to only return the first 50 results.

https://trello.com/c/KvIcPsnw

### DIFF
--- a/spec/models/school_spec.rb
+++ b/spec/models/school_spec.rb
@@ -13,6 +13,11 @@ RSpec.describe School, type: :model do
     it "return schools matching the search term" do
       expect(School.search("Penistone")).to match_array([schools(:penistone_grammar_school)])
     end
+
+    it "limits the results" do
+      stub_const("School::SEARCH_RESULTS_LIMIT", 1)
+      expect(School.search("a").count).to eql(1)
+    end
   end
 
   describe "#address" do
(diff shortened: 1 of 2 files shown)


spec/models/school_spec.rb
```diff
@@ -13,6 +13,11 @@
     it "return schools matching the search term" do
       expect(School.search("Penistone")).to match_array([schools(:penistone_grammar_school)])
     end
+
+    it "limits the results" do
+      stub_const("School::SEARCH_RESULTS_LIMIT", 1)
+      expect(School.search("a")).to match_array([schools(:penistone_grammar_school)])
```
mec: I think I would have gone `expect(School.search("a").count).to eql(1)` does that work as well?

robbiepaul: Yeah that works as well and is probably a bit more robust against ordering changes. I've changed it

spec/models/school_spec.rb
```diff
@@ -13,6 +13,11 @@
     it "return schools matching the search term" do
       expect(School.search("Penistone")).to match_array([schools(:penistone_grammar_school)])
     end
+
+    it "limits the results" do
+      stub_const("School::SEARCH_RESULTS_LIMIT", 1)
```
tekin: woah, I had no idea `stub_const` was a thing...

PR #53, opened 2019-05-22
Add length validation to the postcode field

mec:
We have a database length limit of 11 characters on the postcode field but no validation to check it when submitting. Whilst unlikely this would cause an error in our app, so this adds a length validator to control the problem.

I was looking at production and hit this which raised a rollbar error, happy to not merge this one as collecting the address is temporary, but it's here if we want to.


mec:
I'm closing this for now, I'll add length validation to all the fields with a db limit.

PR #54, opened 2019-05-23
Add length validation to contact fields

mec:
We have a database length limit on the following fields but no
validation to check it when submitting:

- name (200)
- address_line_1 (100)
- address_line_2 (100)
- address_line_3 (100)
- postcode (11)

Users trying to submit values longer than this would cause an error in our app, so this adds a length validators to stop the problem.

This commit also adds some white space around the validation definitions as
they are getting large and hard to read.

### DIFF
--- a/app/models/tslr_claim.rb
+++ b/app/models/tslr_claim.rb
@@ -36,22 +36,41 @@ class TslrClaim < ApplicationRecord
   belongs_to :current_school, optional: true, class_name: "School"
 
   validates :claim_school,              on: :"claim-school", presence: {message: "Select a school from the list"}
+
   validates :qts_award_year,            on: :"qts-year", inclusion: {in: VALID_QTS_YEARS, message: "Select the academic year you were awarded qualified teacher status"}
+
   validates :employment_status,         on: :"still-teaching", presence: {message: "Choose the option that describes your current employment status"}
+
   validates :mostly_teaching_eligible_subjects, on: :"subjects-taught", inclusion: {in: [true, false], message: "Select either Yes or No"}
+
   validates :full_name,                 on: :"full-name", presence: {message: "Enter your full name"}
+  validates :full_name,                 length: {maximum: 200, message: "Full name must be 200 characters or less"}
+
   validates :address_line_1,            on: :address, presence: {message: "Enter your building and street address"}
+  validates :address_line_1,            length: {maximum: 100, message: "Address lines must be 100 characters or less"}
+
+  validates :address_line_2,            length: {maximum: 100, message: "Address lines must be 100 characters or less"}
+
   validates :address_line_3,            on: :address, presence: {message: "Enter your town or city"}
+  validates :address_line_3,            length: {maximum: 100, message: "Address lines must be 100 characters or less"}
+
   validates :postcode,                  on: :address, presence: {message: "Enter your postcode"}
+  validates :postcode,                  length: {maximum: 11, message: "Postcode must be 11 characters or less"}
+
   validates :date_of_birth,             on: :"date-of-birth", presence: {message: "Enter your date of birth"}
+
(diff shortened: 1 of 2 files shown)


app/models/tslr_claim.rb
```diff
@@ -37,18 +37,36 @@ class TslrClaim < ApplicationRecord
...
+                                        length: {maximum: 100, message: "Address lines must be 100 characters or less"}
+
+  validates :address_line_2,            on: :address,
+                                        length: {maximum: 100, message: "Address lines must be 100 characters or less"}
+
+  validates :address_line_3,            on: :address, presence: {message: "Enter your town or city"},
+                                        length: {maximum: 100, message: "Address lines must be 100 characters or less"}
+
+  validates :postcode,                  on: :address, presence: {message: "Enter your postcode"},
+                                        length: {maximum: 11, message: "Postcode must be 11 characters or less"}
```
robbiepaul: Should we always validate these lengths, not just in the specific validation contexts? In the same way we did with the email address?

At the moment we could still get the DB error if we do something like:
```
TslrClaim.create!(postcode: '123456789123')
```

app/models/tslr_claim.rb
```diff
@@ -37,18 +37,36 @@ class TslrClaim < ApplicationRecord
...
-  validates :full_name,                 on: :"full-name", presence: {message: "Enter your full name"}
-  validates :address_line_1,            on: :address, presence: {message: "Enter your building and street address"}
-  validates :address_line_3,            on: :address, presence: {message: "Enter your town or city"}
-  validates :postcode,                  on: :address, presence: {message: "Enter your postcode"}
+
+  validates :full_name,                 on: :"full-name",
+                                        presence: {message: "Enter your full name"},
+                                        length: {maximum: 200, message: "Full name must be 200 characters or less"}
+
+  validates :address_line_1,            on: :address, presence: {message: "Enter your building and street address"},
```
robbiepaul: We should probably be consistent with the line breaks here so it's easier to scan. On the validation rule above you have `presence` on a separate line

app/models/tslr_claim.rb
```diff
@@ -37,18 +37,36 @@ class TslrClaim < ApplicationRecord
...
   validates :teacher_reference_number,  on: :"teacher-reference-number", presence: {message: "Enter your teacher reference number"}
+
   validate :trn_must_be_seven_digits
+
   validates :national_insurance_number, on: :"national-insurance-number", presence: {message: "Enter your National Insurance number"}
+
   validate  :ni_number_is_correct_format
+
   validates :email_address, on: :"email-address", presence: {message: "Enter an email address"}
-  validates :email_address, format: {with: URI::MailTo::EMAIL_REGEXP, message: "Enter an email address in the correct format, like name@example.com"}, \
```
robbiepaul: 👍

spec/models/tslr_claim_spec.rb
```diff
@@ -38,6 +38,10 @@
       expect(TslrClaim.new).not_to be_valid(:"full-name")
       expect(TslrClaim.new(full_name: "John Kimble")).to be_valid(:"full-name")
     end
+    it "validates the length of name is 200 characters or less" do
```
robbiepaul: Can we have an empty line between `it` blocks?

app/models/tslr_claim.rb
mec: I've had another stab at this, beause a lot have two validation contexts now, the standard 'save' and the custom one. I think it reads better now?

mec:
@robbiepaul I had a go at re-ordering the specs as I found it hard to navigate. I ended up adding a 'saving' validation context above all the custom ones and grouping the standard save specs there - let me know what you think and sorry for all the change!

PR #55, opened 2019-05-23
Claimants say if they taught 50% of their time in an eligible subject

robbiepaul:
So that we can determine if a teacher is eligible for this policy or not, we need to know if they taught the eligible subjects at least 50% of their time.

The content comes from the [TSLR eligibility checker](https://github.com/DFE-Digital/tslr/blob/f01d2713eab4698e11ef55323530bd0e1ca922fa/TSLR.UI/pages/question/subjects-taught.vue) from alpha, we may want to improve it at a later date.

### DIFF
--- a/app/models/tslr_claim.rb
+++ b/app/models/tslr_claim.rb
@@ -66,17 +66,25 @@ class TslrClaim < ApplicationRecord
   end
 
   def ineligible?
-    ineligible_claim_school? || employed_at_no_school?
+    ineligible_claim_school? || employed_at_no_school? || not_taught_eligible_subjects_enough?
   end
 
   def ineligibility_reason
-    [:ineligible_claim_school, :employed_at_no_school].find { |eligibility_check| send("#{eligibility_check}?") }
+    [
+      :ineligible_claim_school,
+      :employed_at_no_school,
+      :not_taught_eligible_subjects_enough,
+    ].find { |eligibility_check| send("#{eligibility_check}?") }
   end
 
   private
 
   def ineligible_claim_school?
     claim_school.present? && !claim_school.eligible_for_tslr?
+  end
+
+  def not_taught_eligible_subjects_enough?
+    mostly_teaching_eligible_subjects == false
   end
 
   def update_current_school
--- a/app/models/tslr_claim.rb
+++ b/app/models/tslr_claim.rb
@@ -4,6 +4,7 @@ class TslrClaim < ApplicationRecord
     "claim-school",
     "still-teaching",
     "current-school",
+    "subjects-taught",
     "full-name",
     "address",
     "date-of-birth",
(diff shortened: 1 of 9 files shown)


robbiepaul:
@tekin I've made some changes since we spoke:
- Made the question a Yes/No question
- Changed the column to be a boolean instead of an enum
- Renamed the page and the database column

app/models/tslr_claim.rb
```diff
@@ -77,6 +83,10 @@ def ineligible_claim_school?
     claim_school.present? && !claim_school.eligible_for_tslr?
   end
 
+  def taught_not_enough_eligible_subjects?
```
tekin: How about `not_taught_enough_eligible_subjects?` for the method name? "tought not" feels like the words are in the wrong order.r

app/models/tslr_claim.rb
```diff
@@ -77,6 +83,10 @@ def ineligible_claim_school?
...
+    !mostly_teaching_eligible_subjects.nil? && !mostly_teaching_eligible_subjects
```
tekin: `mostly_teaching_eligible_subjects == false`?

app/views/claims/subjects_taught.html.erb
```diff
@@ -0,0 +1,41 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render("shared/error_summary", instance: current_claim) if current_claim.errors.any? %>
+    <%= form_for current_claim, url: claim_path do |form| %>
+      <%= form.hidden_field :mostly_teaching_eligible_subjects %>
+      <div class="govuk-form-group">
+        <fieldset class="govuk-fieldset">
+          <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
+            <h1 class="govuk-fieldset__heading">Did you teach for more than 50% of your teaching time in one of the eligible subjects?</h1>
```
tekin:
```suggestion
            <h1 class="govuk-fieldset__heading">Did you teach eligible subjects for more than 50% of your teaching time between 6 April XXXX and 5 April XXXX?</h1>
```

app/views/claims/subjects_taught.html.erb
```diff
@@ -0,0 +1,41 @@
...
+          </legend>
+          <p class="govuk-body">
+            You can only claim your student loan payments back if you've been teaching one or more of these subjects since 6 April 2018:
```
tekin:
```suggestion
            Eligible subjects are:
```

app/models/tslr_claim.rb
robbiepaul: 🙈 ah yes!

robbiepaul:
@tekin Thanks for the suggestions, I've made some changes.

PR #57, opened 2019-05-29
Bank account details capture page

mec:
We are not capturing the bank 'account holder name' at this stage as we hope to get that data from elsewhere.

Followed the layout from @titlescreen (ta!),left out the optional 'SSL notice', we can always add it in later if user research shows we need it.

Screenshot_2019-05-29 Teachers claim back student loan repayments

Trello: https://trello.com/c/ayr6F6OV

## Queries I had 
This seemed to be as DRY as I could ge the normalisation methods, I couldn't see a way to pass which attribute had changed in the `before_save` hook.

At first I set the fields to integers in the DB but this meant constantly coercing the values to strings all over the place, so in the end I changed the DB to a string, I figured the form submits a string anyway. Wonder if I could've used a `to_i` at the end of the normalisation, be interested in your thoughts on this one?

### DIFF
--- a/app/models/tslr_claim.rb
+++ b/app/models/tslr_claim.rb
@@ -12,6 +12,7 @@ class TslrClaim < ApplicationRecord
     "national-insurance-number",
     "student-loan-amount",
     "email-address",
+    "bank-details",
     "check-your-answers",
     "confirmation",
   ].freeze
@@ -74,6 +75,13 @@ class TslrClaim < ApplicationRecord
   validates :email_address,             on: [:"email-address", :submit], presence: {message: "Enter an email address"}
   validates :email_address,             format: {with: URI::MailTo::EMAIL_REGEXP, message: "Enter an email address in the correct format, like name@example.com"},
                                         length: {maximum: 256, message: "Email address must be 256 characters or less"},
+                                        allow_nil: true
+
+  validates :bank_sort_code,            on: :"bank-details", presence: {message: "Enter a sort code"}
+  validates :bank_sort_code,            length: {is: 6, message: "Sort code must be 6 digits"},
+                                        allow_nil: true
+  validates :bank_account_number,       on: :"bank-details", presence: {message: "Enter an account number"}
+  validates :bank_account_number,       length: {is: 8, message: "Account number must be 8 digits"},
                                         allow_nil: true
 
   before_save :update_current_school, if: :employment_status_changed?
--- a/app/models/tslr_claim.rb
+++ b/app/models/tslr_claim.rb
@@ -78,15 +78,16 @@ class TslrClaim < ApplicationRecord
                                         allow_nil: true
 
   validates :bank_sort_code,            on: :"bank-details", presence: {message: "Enter a sort code"}
-  validates :bank_sort_code,            length: {is: 6, message: "Sort code must be 6 digits"},
-                                        allow_nil: true
   validates :bank_account_number,       on: :"bank-details", presence: {message: "Enter an account number"}
-  validates :bank_account_number,       length: {is: 8, message: "Account number must be 8 digits"},
-                                        allow_nil: true
+
+  validate  :bank_account_number_must_be_eight_digits
+  validate  :bank_sort_code_must_be_six_digits
 
   before_save :update_current_school, if: :employment_status_changed?
   before_save :normalise_trn, if: :teacher_reference_number_changed?
   before_save :normalise_ni_number, if: :national_insurance_number_changed?
+  before_save :normalise_bank_account_number, if: :bank_account_number_changed?
+  before_save :normalise_bank_sort_code, if: :bank_sort_code_changed?
 
   delegate :name, to: :claim_school, prefix: true, allow_nil: true
   delegate :name, to: :current_school, prefix: true, allow_nil: true
@@ -176,5 +177,27 @@ class TslrClaim < ApplicationRecord
   def ni_number_is_correct_format
     errors.add(:national_insurance_number, "Enter a National Insurance number in the correct format") \
       if national_insurance_number.present? && !normalised_ni_number.match(/\A[a-z]{2}[0-9]{6}[a-d]{1}\Z/i)
+  end
+
+  def normalise_bank_account_number
+    self.bank_account_number = normalised_bank_detail(bank_account_number)
+  end
+
+  def normalise_bank_sort_code
+    self.bank_sort_code = normalised_bank_detail(bank_sort_code)
+  end
+
+  def normalised_bank_detail(bank_detail)
+    bank_detail.gsub(/\s|-/, "")
+  end
+
+  def bank_account_number_must_be_eight_digits
+    errors.add(:bank_account_number, "Bank account number must contain eight digits") \
+      if bank_account_number.present? && normalised_bank_detail(bank_account_number) !~ /\A\d{8}\z/
+  end
+
+  def bank_sort_code_must_be_six_digits
+    errors.add(:bank_sort_code, "Sort code must contain six digits") \
+      if bank_sort_code.present? && normalised_bank_detail(bank_sort_code) !~ /\A\d{6}\z/
   end
 end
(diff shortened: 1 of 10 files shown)


db/migrate/20190529101022_add_bank_details_to_tslr_claim.rb
```diff
@@ -0,0 +1,6 @@
+class AddBankDetailsToTslrClaim < ActiveRecord::Migration[5.2]
+  def change
+    add_column :tslr_claims, :account_sort_code, :string, limit: 6
+    add_column :tslr_claims, :account_number, :string, limit: 8
```
tekin: Can we rename these fields to `bank_account_number` and `bank_sort_code` to be slightly more precise?

app/models/tslr_claim.rb
```diff
@@ -142,4 +143,26 @@ def ni_number_is_correct_format
...
+  def normalise_account_number
+    self.account_number = normalised_bank_detail(account_number)
+  end
+
+  def normalise_account_sort_code
+    self.account_sort_code = normalised_bank_detail(account_sort_code)
+  end
+
+  def normalised_bank_detail(bank_detail)
+    bank_detail.gsub(/\D/, "")
```
tekin: This is effectively doing the same normalisation as the TRN normalisation. Is there a way to combine the two methods into a single one that normalises a numeric value?

Thinking about this a bit more, I think in reality we could have probably been more specific with the normalisation (of the TRN and these fields) and only strip characters that we think users would legitimately enter, i.e. whitespace and '-', and then validate what they've entered is actually positive numeric integer (i.e. `/\A\d+\z/`). Otherwise the normalisation means that a user could enter `"ABC1234XX?12"` and we'd accept that as a sort code, normalising it to `123412`. I don't think there's a huge risk or danger here but something to keep in mind.

app/models/tslr_claim.rb
```diff
@@ -134,4 +143,26 @@ def ni_number_is_correct_format
     errors.add(:national_insurance_number, "Enter a National Insurance number in the correct format") \
       if national_insurance_number.present? && !normalised_ni_number.match(/\A[a-z]{2}[0-9]{6}[a-d]{1}\Z/i)
   end
+
+  def normalise_account_number
+    self.account_number = normalised_bank_detail(account_number)
+  end
+
+  def normalise_account_sort_code
+    self.account_sort_code = normalised_bank_detail(account_sort_code)
```
erbridge: Don't think it needs to change, but I think the methods you were looking for are `account_number_changed?` and `account_sort_code_changed?`, same as you used in the callback declaration.

app/models/tslr_claim.rb
mec: I've had a second attempt at this, we should validate againse more now as per your example, see what you think?

app/models/tslr_claim.rb
mec: Yeah, I had those. I was thinking I might be able to do something like:

`before_save :normalise_bank_details, if: :bank_sort_code_changed? || :bank_account_number_changed?`

But looking at it now I'm not sure it's as clear.

app/views/claims/bank_details.html.erb
```diff
@@ -0,0 +1,42 @@
...
+            <span  id="account-number-hint" class="govuk-hint">For example: 70 87 24 90</span>
+            <%= form.text_field :bank_account_number,
+              class: "govuk-input govuk-input--width-20",
+              "aria-describedby" => "account-number-hint"
+            %>
+          </div>
+
+        </fieldset>
+        <button type="submit" class="govuk-button">
+          Continue
```
robbiepaul: We could use `<%= form.submit "Continue", class: "govuk-button" %>` so we're consistent with the other pages

spec/models/tslr_claim_spec.rb
```diff
@@ -55,12 +55,21 @@
...
-        expect(TslrClaim.new(account_number: "123456789")).not_to be_valid
-        expect(TslrClaim.new(account_sort_code: "10111213")).not_to be_valid
-
-        expect(TslrClaim.new(account_number: "12345678")).to be_valid
-        expect(TslrClaim.new(account_sort_code: "101112")).to be_valid
+      it "validates the format of bank_account_number" do
+        expect(TslrClaim.new(bank_account_number: "ABC12 34 56 789")).not_to be_valid
+        expect(TslrClaim.new(bank_account_number: "12-34-56-78")).to be_valid
+      end
+      it "validates the length of bank_account_number" do
```
robbiepaul: I wonder if these two examples could be combined in to one `it` block? If the length of the account number is wrong it isn't in the correct format.

spec/models/tslr_claim_spec.rb
```diff
@@ -55,12 +55,21 @@
...
+        expect(TslrClaim.new(bank_account_number: "12 34 56 78 90")).not_to be_valid
+        expect(TslrClaim.new(bank_account_number: "12-34-56-78")).to be_valid
+      end
+      it "validates the format of bank_sort_code" do
+        expect(TslrClaim.new(bank_sort_code: "ABC12 34 567")).not_to be_valid
+        expect(TslrClaim.new(bank_sort_code: "12 34 56")).to be_valid
+      end
+      it "validates the length of bank_sort_code" do
```
robbiepaul: Same as above. Also could we have an empty line between specs?

spec/models/tslr_claim_spec.rb
```diff
@@ -146,11 +155,27 @@
...
-        expect(TslrClaim.new(account_sort_code: "123456", account_number: "87654321")).to be_valid(:"bank-details")
-      end
+    it "validates that the bank_account_number and bank_sort_code are present" do
+      expect(TslrClaim.new).not_to be_valid(:"bank-details")
+      expect(TslrClaim.new(bank_sort_code: "12 34 56", bank_account_number: "87 65 43 21")).to be_valid(:"bank-details")
+    end
+  end
+
+  describe "#bank_account_number" do
+    it "strips out whitespace and the “-” character" do
```
robbiepaul: 💯 I like this approach over removing all non-digit characters

spec/helpers/claims_helper_spec.rb
```diff
@@ -44,6 +61,8 @@
         ["Teacher reference number", "1234567"],
         ["National Insurance number", "QQ123456C"],
         ["Email address", "test@email.com"],
+        ["Account number", "12345678"],
+        ["Sort code", "123456"],
```
robbiepaul: We may want to format the sort code in the standard way when we show it back to the user. However it wasn't an AC on this card so we can probably leave it for a future iteration

app/views/claims/bank_details.html.erb
mec: 👍

spec/models/tslr_claim_spec.rb
mec: 👍

spec/helpers/claims_helper_spec.rb
mec: Yep, I think we have this issue to tackle for a lot of the fields.

spec/models/tslr_claim_spec.rb
mec: 👍

spec/models/tslr_claim_spec.rb
mec: 👍

spec/models/tslr_claim_spec.rb
```diff
@@ -134,6 +145,31 @@
...
   end
 
+  context "when saving in the “bank-details” validation context" do
+    it "validates that the bank_account_number and bank_sort_code are present" do
+      expect(TslrClaim.new).not_to be_valid(:"bank-details")
+      expect(TslrClaim.new(bank_sort_code: "12 34 56", bank_account_number: "87 65 43 21")).to be_valid(:"bank-details")
+    end
+  end
+
+  describe "#bank_account_number" do
```
tekin: As discussed, the two specs in these describe blocks are really testing the before save normalisation and not the attribute reader, so more accurate to move the `it` blocks up into the `save` context at the top and ditch the extra `describes.

mec: I've gone ahead and re-factored the TSLRClaim spec as we discussed, as well as moving the `it` blocks mentioned above, felt like it was a natural time to pick up on it.

PR #58, opened 2019-05-29
A user can review and complete their claim

erbridge:
> So that a teacher can review their claim details before submitting (and so that we know they have completed their claim) we need to remind them of their answers and ask them to confirm they are correct.

Things to note:
- We've switched to using i18n for question translations to reduce repetition.
- The "complete" view is now called "confirmation" to allow "complete" to be the action rather than the result.
- Validations all run on `:complete` as well as during their given stage. **New validations should also run in the `:complete` context.**
- The validation error on completing an ineligible claim probably needs looking at again when it's possible to get in that state (e.g. after changing eligibility values).
- Dates now default to e.g. 1 May 2019, in line with GOV.UK guidance.


spec/models/tslr_claim_spec.rb
```diff
@@ -291,4 +298,46 @@
       end
     end
   end
+
+  describe "#complete!" do
+    context "when claim is valid" do
```
tekin: Here's where the "valid" wording used in the trait ends up being potentially misleading/confusing to the reader.

spec/models/tslr_claim_spec.rb
```diff
@@ -291,4 +298,46 @@
...
+
+      it "sets completed_at to now" do
+        freeze_time do
+          tslr_claim.complete!
+
+          expect(tslr_claim.reload.completed_at).to eq Time.zone.now
+        end
+      end
+
+      context "and the claim has already been completed" do
```
tekin: I'm not sure that this is something we want to set in stone as required behaviour by capturing it as a spec. Are there circumstances where we would want the user to be able to re-"complete" their claim and change the timestamp for when it was completed? I can think of a reason why we would probably want to prevent this, i.e. it could potentially mess around with the batching process if the timestamp for a claim changed after it was "completed".

db/migrate/20190529103851_add_completed_at_to_tslr_claims.rb
```diff
@@ -0,0 +1,5 @@
+class AddCompletedAtToTslrClaims < ActiveRecord::Migration[5.2]
+  def change
+    add_column :tslr_claims, :completed_at, :datetime
```
tekin: let's add an index on this whilst we're here, I can foresee this being the thing that we use to sort claims when displaying them to administrative users.

app/controllers/claims_controller.rb
```diff
@@ -5,6 +5,8 @@ class ClaimsController < ApplicationController
   before_action :end_expired_sessions
   before_action :update_last_seen_at
 
+  after_action :clear_claim_session, if: proc { params[:slug] == "confirmation" && current_claim&.completed? }
```
tekin: Worth pulling the proc to a named method? Also wondering if the safe-navigation operator is required. Would the code ever get as far as a `current_claim` that is `nil` with the user on the "confirmation" page?

spec/requests/claims_spec.rb
```diff
@@ -61,6 +61,23 @@
         expect(response).to redirect_to(root_path)
       end
     end
+
+    context "when a claim is completed" do
```
tekin: I think it would be good to be more specific with this context/behaviour. It's more like "when the user reaches the confirmation page after completing their claim"

config/locales/en.yml
```diff
@@ -30,4 +30,20 @@
 # available at http://guides.rubyonrails.org/i18n.html.
 
 en:
-  service_name: "Teachers: claim back student loan repayments"
+  date:
+    formats:
+      default: "%d %B %Y"
```
tekin: 👍

spec/factories/local_authorities.rb
```diff
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :local_authority do
-    sequence(:code) { |n| 1000 + n }
-    name { "York" }
+    sequence(:code) { |n| Tslr::SchoolEligibility::ELIGIBLE_LOCAL_AUTHORITY_CODES[n % Tslr::SchoolEligibility::ELIGIBLE_LOCAL_AUTHORITY_CODES.count] }
```
tekin: So things get a little tricky here... Because this updated sequence is working from a fixed set of LA codes and we have a uniqueness constraint on the column, it introduces the potential for spec flakiness due to clashes. That's why the current version intentionally skips the range of known LA codes to avoid potential clashes.

One of the ways we're getting around this problem of having essentially a fixed set of identifiers for the local authorities is by using fixtures to pre-seed the test database with eligible and ineligible local authorities (and schools): https://github.com/DFE-Digital/dfe-teachers-payment-service/commit/7f3fa389ff58f7732c4136079dfe8991b0e02f35.

It can get a bit tricky when trying to mix factories and fixtures (mostly because the fixture helpers aren't available in the `FactoryBot.define` scope), but it is possible. For example, here's how the fixture for an eligible school (`:penistone_grammar_school`) could be assigned to a claim in a factory:

```ruby
FactoryBot.define do
  factory :tslr_claim_with_eligible_school, class: TslrClaim do
    claim_school { School.find(ActiveRecord::FixtureSet.identify(: penistone_grammar_school, :uuid)) }
  end
end
```

app/views/claims/email_address.html.erb
```diff
@@ -4,7 +4,7 @@
 
     <%= form_for current_claim, url: claim_path do |form| %>
       <h1 class="govuk-heading-xl">
-        <%= form.label :email_address, "What is your email address?"  %>
+        <%= form.label :email_address, t("tslr.questions.email_address")  %>
```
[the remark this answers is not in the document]
erbridge: Yeah. I figured we could refactor them as we need them elsewhere.

app/models/tslr_claim.rb
```diff
@@ -77,13 +78,24 @@ class TslrClaim < ApplicationRecord
...
 
   def page_sequence
     PAGE_SEQUENCE.dup.tap do |sequence|
       sequence.delete("current-school") if employed_at_claim_school?
     end
   end
 
+  def complete!
+    if ineligible?
+      errors.add(:base, "You can only apply for this payment if you meet the eligibility criteria.")
```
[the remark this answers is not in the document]
erbridge: We thought about this, but it seemed strange to pull the reason text into the model, as it's a view concern. I'll think about it some more.

app/models/tslr_claim.rb
```diff
@@ -77,13 +78,24 @@ class TslrClaim < ApplicationRecord
...
+      return false
+    end
+
+    self.completed_at = Time.zone.now
+    save(context: :complete)
```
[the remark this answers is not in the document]
erbridge: I didn't know about `touch`. Sounds good.

spec/factories/tslr_claims.rb
```diff
@@ -0,0 +1,28 @@
...
+      national_insurance_number { "QQ123456C" }
+      email_address { "test@email.com" }
+    end
+
+    trait :valid_and_ineligible do
+      valid_and_eligible
+      mostly_teaching_eligible_subjects { false }
+    end
+
+    trait :invalid do
```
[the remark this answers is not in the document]
erbridge: If we rename to `submitted`, we can use `submittable` and `unsubmittable`.

spec/features/tslr_claim_spec.rb
```diff
@@ -72,6 +72,14 @@
 
     expect(claim.reload.email_address).to eq("name@example.tld")
 
+    expect(page).to have_text("Check your answers before sending your application")
+
+    freeze_time do
+      click_on "Confirm and send"
```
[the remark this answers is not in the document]
erbridge: This was a change requested by @titlescreen to bring it inline with MPS. The example from the design system is "Accept and send".

spec/models/tslr_claim_spec.rb
erbridge: I took this spec as more informative of the actual behaviour than as a requirement. We decided not to worry about preventing resubmission at this point, but do you think it needs addressing now?

app/controllers/claims_controller.rb
erbridge: Ah, that's a relic from before we relied on the slug. I'll remove it.

app/controllers/claims_controller.rb
```diff
@@ -34,6 +33,15 @@ def ineligible
 
   private
 
+  def update_current_claim!
+    if params[:slug] == "check-your-answers"
```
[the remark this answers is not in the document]
erbridge: Oops. Missed that in a rebase. Thanks!

app/models/tslr_claim.rb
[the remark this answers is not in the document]
erbridge: Decided to pull it into the model as a `full_ineligibility_reason` method, alongside the existing `ineligibility_reason`.

spec/models/tslr_claim_spec.rb
tekin: Yeah, the problem with a spec is that it signals to someone reading it that the behaviour is important enough to fail the build if the behaviour is removed. It makes it harder to refactor the underlying code without knowing which specs are hard routes and which ones are simply informative.

erbridge: Ok, I will remove this spec.

spec/factories/local_authorities.rb
erbridge: I did wonder about that, but didn't see any good reason to have ineligible LAs by default, so thanks for that.

erbridge:
Ready for another review @tekin.

spec/models/tslr_claim_spec.rb
```diff
@@ -291,4 +298,50 @@
       end
     end
   end
+
+  describe "#submit!" do
```
tekin: One last suggestion to improve the readability of these specs a bit:

```diff
diff --git a/spec/models/tslr_claim_spec.rb b/spec/models/tslr_claim_spec.rb
index c274a36..2416c62 100644
--- a/spec/models/tslr_claim_spec.rb
+++ b/spec/models/tslr_claim_spec.rb
@@ -300,15 +300,17 @@ RSpec.describe TslrClaim, type: :model do
   end
 
   describe "#submit!" do
+    around do |example|
+      freeze_time { example.run }
+    end
+
+    before { tslr_claim.submit! }
+
     context "when the claim is eligible and submittable" do
       let(:tslr_claim) { create(:tslr_claim, :eligible_and_submittable) }
 
       it "sets submitted_at to now" do
-        freeze_time do
-          tslr_claim.submit!
-
-          expect(tslr_claim.reload.submitted_at).to eq Time.zone.now
-        end
+        expect(tslr_claim.submitted_at).to eq Time.zone.now
       end
     end
 
@@ -316,11 +318,7 @@ RSpec.describe TslrClaim, type: :model do
       let(:tslr_claim) { create(:tslr_claim, :eligible_but_unsubmittable) }
 
       it "doesn't set submitted_at" do
-        freeze_time do
-          tslr_claim.submit!
-
-          expect(tslr_claim.reload.submitted_at).to be_nil
-        end
+        expect(tslr_claim.submitted_at).to be_nil
       end
     end
 
@@ -328,19 +326,11 @@ RSpec.describe TslrClaim, type: :model do
       let(:tslr_claim) { create(:tslr_claim, :eligible_and_submittable, mostly_teaching_eligible_subjects: false) }
 
       it "doesn't set submitted_at" do
-        freeze_time do
-          tslr_claim.submit!
-
-          expect(tslr_claim.reload.submitted_at).to be_nil
-        end
+        expect(tslr_claim.submitted_at).to be_nil
       end
 
       it "adds an error" do
-        freeze_time do
-          tslr_claim.submit!
-
-          expect(tslr_claim.errors.messages[:base]).to include("You must have spent at least half your time teaching an eligible subject.")
-        end
+        expect(tslr_claim.errors.messages[:base]).to include("You must have spent at least half your time teaching an eligible subject.")
       end
     end
   end
```

PR #59, opened 2019-06-03
Claimants can enter the amount of student loan they repaid last tax year

robbiepaul:
So that we know how much to pay the teacher, we need them to tell us how much student loan they paid in the last tax year.

Includes a non-standard GOV.UK pattern for capturing monetary amounts.

https://trello.com/c/3iY89vAu/

# Screenshot

Screenshot 2019-06-03 at 11 36 15

### DIFF
--- a/app/helpers/claims_helper.rb
+++ b/app/helpers/claims_helper.rb
@@ -17,6 +17,7 @@ module ClaimsHelper
       [t("tslr.questions.claim_school"), claim.claim_school_name],
       [t("tslr.questions.current_school"), claim.current_school_name],
       [t("tslr.questions.mostly_teaching_eligible_subjects"), claim.mostly_teaching_eligible_subjects? ? "Yes" : "No"],
+      [t("tslr.questions.student_loan_amount", claim_school_name: claim.claim_school_name), number_to_currency(claim.student_loan_repayment_amount)],
     ]
   end
 
--- a/app/views/claims/student_loan_amount.html.erb
+++ b/app/views/claims/student_loan_amount.html.erb
@@ -1,0 +1,23 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render("shared/error_summary", instance: current_claim) if current_claim.errors.any? %>
+
+    <%= form_for current_claim, url: claim_path do |form| %>
+      <h1 class="govuk-heading-xl">
+        <%= form.label :student_loan_repayment_amount, t("tslr.questions.student_loan_amount", claim_school_name: current_claim.claim_school_name)  %>
+      </h1>
+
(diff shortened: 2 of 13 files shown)


app/views/claims/student_loan_amount.html.erb
```diff
@@ -0,0 +1,24 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render("shared/error_summary", instance: current_claim) if current_claim.errors.any? %>
+
+    <%= form_for current_claim, url: claim_path do |form| %>
+      <h1 class="govuk-heading-xl">
+        How much student loan did you repay while you were at <%= current_claim.claim_school_name %> between 6 April XXXX and 5 April XXXX?
+      </h1>
+
+      <%= form.label :student_loan_repayment_amount, t("tslr.questions.student_loan_amount"), class: "govuk-label"  %>
```
tekin: I feel like this label is possibly redundant, and the question in the heading above should be the actual form label. Also making the question the label would make this page and the way it is played back on the confirmation page more consistent with the rest of the questions.

robbiepaul: @tekin As discussed I've made the question the label and used it on the "check your answers" page. Ready for another look

app/helpers/claims_helper.rb
```diff
@@ -17,6 +17,7 @@ def claim_answers(claim)
       [t("tslr.questions.claim_school"), claim.claim_school_name],
       [t("tslr.questions.current_school"), claim.current_school_name],
       [t("tslr.questions.mostly_teaching_eligible_subjects"), claim.mostly_teaching_eligible_subjects? ? "Yes" : "No"],
+      [t("tslr.questions.student_loan_amount", claim_school_name: claim.current_school_name), number_to_currency(claim.student_loan_repayment_amount)],
```
tekin: This should be `claim_school_name`?

robbiepaul: Thanks, fixed

spec/models/tslr_claim_spec.rb
```diff
@@ -168,6 +183,14 @@
     end
   end
 
+  describe "#student_loan_repayment_amount" do
+    it "saves with monetary characters removed" do
```
tekin: I'm approving the PR, but spotted one more thing: this spec is a little off. The behaviour of the code isn't to do with saving, but is more specifically the cleanup that happens on assignment. e.g. something like: 

```ruby
claim = TslrClaim.new
claim.student_loan_repayment_amount = "£ 5,000.40"
expect(claim.student_loan_repayment_amount).to eql(5000.40)
```

feel free to correct and merge.

robbiepaul: I've corrected the spec 👍

PR #66, opened 2019-06-05
Session timeout modal warning

robbiepaul:
To help users who may be taking some time to fill in their claim, we want to warn them before their session times out so that they don't lose the information they've entered so far.

After 28 minutes of inactivity the user will see a dialog explaining that their session is about to timeout and they will lose any progress. If they dismiss the modal or click `Continue session` the session will be refreshed for another 28 minutes.

# Screenshot

Screenshot 2019-06-06 at 16 43 53

https://trello.com/c/MXz7Rowo


robbiepaul:
Build is failing due to it not being able to find Chrome. There's an issue on the buildpack repo: https://github.com/heroku/heroku-buildpack-google-chrome/issues/26

I've tried a few of the recommendations, but no joy. Will look at it in more detail tomorrow

README.md
```diff
@@ -25,7 +25,7 @@ Architecture decision records can be found in the
      `File > Import Items`. Double click the imported certificate and change the
      “When using this certificate:” dropdown to `Always Trust` in the Trust
      section.
-2. Run `bundle install` to install the gem dependencies
+2. Run `bundle install` and `yarn install` to install the dependencies
```
pezholio: Might be worth pointing out that you need yarn installed too

robbiepaul:
It turns out there were two issues with the CI build:

- Chromedriver couldn't find the executable so we have to explicitly tell it to use the one in the environment variables.
- After this the browser was becoming unresponsive, because we are now using the bin and not the shim we have to tell chrome that, with the arguments `disable-dev-shm-usage no-sandbox`

This is ready for review now

robbiepaul:
[Session timeout modal warning](https://trello.com/c/MXz7Rowo/149-session-timeout-modal-warning)

app/assets/javascripts/timeout_dialog.js.erb
```diff
@@ -0,0 +1,43 @@
...
+    xhr.open("GET", "<%= Rails.application.routes.url_helpers.claim_refresh_session_path %>");
+    xhr.onreadystatechange = function() {
+      if (xhr.readyState > 3 && xhr.status === 200) {
+        clearInterval(win.timeoutTimer);
+        win.timeoutTimer = setInterval(showTimeoutDialog, timeoutInMilliseconds);
+      }
+    };
+    xhr.setRequestHeader("X-Requested-With", "XMLHttpRequest");
+    xhr.send();
+  }
```
erbridge: I think you can use `Rails.ajax` for this, unless you're doing something special.

robbiepaul: Nice, I didn't know that existed

app/assets/stylesheets/govuk_frontend_rails.scss
```diff
@@ -12,6 +12,7 @@ $govuk-image-url-function: 'font-url';
 
 @import "components/flash";
 @import "components/currency_input";
+@import "components/a11y-dialog";
```
erbridge: This stops being `govuk_frontend_rails` now, I think. Does it make sense to rename it?

robbiepaul: I've added a commit that extracts our application specific components out to a separate `govuk_frontend_custom.scss`

PR #67, opened 2019-06-06
A user can change their answers when reviewing their claim

pezholio:
Trello card: https://trello.com/c/OhhpWDA2/175-a-user-can-change-their-answers-when-reviewing-their-claim

This adds links for a user to change their responses, as per the pattern in the [GDS design system](https://design-system.service.gov.uk/patterns/check-answers/)

image

### DIFF
--- a/app/models/tslr_claim.rb
+++ b/app/models/tslr_claim.rb
@@ -104,7 +104,7 @@ class TslrClaim < ApplicationRecord
       return false
     end
 
-    touch(:submitted_at) if valid?(:submit)
+    touch(:submitted_at) if can_be_submitted?
   end
 
   def submitted?
--- a/app/models/tslr_claim.rb
+++ b/app/models/tslr_claim.rb
@@ -77,8 +77,8 @@ class TslrClaim < ApplicationRecord
                                         length: {maximum: 256, message: "Email address must be 256 characters or less"},
                                         allow_nil: true
 
-  validates :bank_sort_code,            on: :"bank-details", presence: {message: "Enter a sort code"}
-  validates :bank_account_number,       on: :"bank-details", presence: {message: "Enter an account number"}
+  validates :bank_sort_code,            on: [:"bank-details", :submit], presence: {message: "Enter a sort code"}
+  validates :bank_account_number,       on: [:"bank-details", :submit], presence: {message: "Enter an account number"}
 
   validate  :bank_account_number_must_be_eight_digits
   validate  :bank_sort_code_must_be_six_digits
@@ -109,6 +109,10 @@ class TslrClaim < ApplicationRecord
 
   def submitted?
     submitted_at.present?
+  end
+
+  def can_be_submitted?
+    valid?(:submit)
   end
 
   def ineligible?
--- a/spec/features/tslr_claim_spec.rb
+++ b/spec/features/tslr_claim_spec.rb
@@ -159,4 +159,19 @@ RSpec.feature "Teacher Student Loan Repayments claims" do
     visit claim_path("qts-year")
     expect(page).to have_current_path(root_path)
   end
+
+  scenario "Teacher can edit their response" do
+    claim = create(:tslr_claim, :eligible_and_submittable)
+    allow_any_instance_of(ClaimsController).to receive(:current_claim) { claim }
+
+    visit claim_path("check-your-answers")
+
+    find("a[href='#{claim_path("national-insurance-number")}']").click
+
+    fill_in "National Insurance number", with: "AB123456C"
+    click_on "Continue"
+
+    expect(claim.reload.national_insurance_number).to eq("AB123456C")
+    expect(page).to have_content("Check your answers before sending your application")
+  end
 end
--- a/spec/features/tslr_claim_spec.rb
+++ b/spec/features/tslr_claim_spec.rb
@@ -203,16 +203,16 @@ RSpec.feature "Teacher Student Loan Repayments claims" do
       expect(find("input[name='school_search']").value).to eq(current_school.name)
     end
 
-    context "Teacher goes to change their claim school" do
+    context "When changing claim school" do
       before do
         find("a[href='#{claim_path("claim-school")}']").click
       end
 
-      scenario "sees their original claim school" do
+      scenario "Teacher sees their original claim school" do
         expect(find("input[name='school_search']").value).to eq(claim.claim_school.name)
       end
 
-      context "changes their claim school" do
+      context "When choosing a new claim school" do
         before do
           choose_school schools(:penistone_grammar_school)
         end
@@ -225,22 +225,22 @@ RSpec.feature "Teacher Student Loan Repayments claims" do
           expect(current_path).to eq(claim_path("still-teaching"))
         end
 
-        context "is still teaching at the same school" do
+        context "When still teaching at the claim school" do
           before do
             choose_still_teaching
           end
 
-          it "current school is set correctly" do
+          scenario "current school is set correctly" do
             expect(claim.reload.employment_status).to eql("claim_school")
             expect(claim.reload.current_school).to eql schools(:penistone_grammar_school)
           end
 
-          it "Teacher is redirected to the check your answers page" do
+          scenario "Teacher is redirected to the check your answers page" do
             expect(current_path).to eq(claim_path("check-your-answers"))
           end
         end
 
-        context "is still teaching at another school" do
+        context "When still teaching but at a different school" do
           before do
             choose_still_teaching "Yes, at another school"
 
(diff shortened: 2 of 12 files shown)


pezholio:
[A user can change their answers when reviewing their claim](https://trello.com/c/OhhpWDA2/175-a-user-can-change-their-answers-when-reviewing-their-claim)

app/views/claims/_school_search.html.erb
```diff
@@ -0,0 +1,19 @@
...
+
+<%= form_tag(path, method: :get) do %>
+  <div class="govuk-form-group">
+    <%= label_tag :school_search, "School name", class: "govuk-label" %>
+    <%= text_field_tag :school_search, params[:school_search], class: "govuk-input", value: school&.name %>
+  </div>
+  <div class="govuk-form-group">
+    <%= submit_tag "Search", class: "govuk-button" %>
+  </div>
+<% end %>
```
erbridge: Did anything change when refactoring these? If not, would be a massive pain to roll the refactor back? The changes to the autocomplete involve a different refactor that will be a pain to merge with this.

db/migrate/20190605153707_add_complete_to_tslr_claims.rb
```diff
@@ -0,0 +1,5 @@
+class AddCompleteToTslrClaims < ActiveRecord::Migration[5.2]
+  def change
+    add_column :tslr_claims, :complete, :boolean
```
erbridge: Let's not use `complete` for this, as the claim still has some processing to be made before it's complete. Also, do we need a new field? Isn't completeness a derived property, based on the status of the other claim attributes? I think we can use `claim.valid?(:submit)` instead, which runs all the validators to establish if the claim is submittable.

app/views/claims/_school_search.html.erb
pezholio: Not a huge change. Happy to roll it back 👍

db/migrate/20190605153707_add_complete_to_tslr_claims.rb
pezholio: Yeah, that makes sense. Would save having to add another migration too.

spec/features/tslr_claim_spec.rb
```diff
@@ -159,4 +159,116 @@
...
+    let(:claim_school) { create(:school, :tslr_eligible, name: "Claim School") }
+    let(:current_school) { create(:school, :tslr_eligible, name: "Current School") }
+
+    let(:claim) do
+      claim = create(:tslr_claim,
+        :eligible_and_submittable,
+        complete: true,
+        employment_status: :different_school,
+        claim_school: claim_school)
+      claim.current_school = current_school
```
erbridge: Could this be simplified by including it in the `create`?

spec/features/tslr_claim_spec.rb
```diff
@@ -159,4 +159,116 @@
...
+      visit claim_path("check-your-answers")
+    end
+
+    scenario "Teacher can edit a field" do
+      find("a[href='#{claim_path("national-insurance-number")}']").click
+
+      fill_in "National Insurance number", with: "AB123456C"
+      click_on "Continue"
+
+      expect(claim.reload.national_insurance_number).to eq("AB123456C")
```
erbridge: Could we either explicitly set or check the existing value, so we know this is actually a change?

app/models/tslr_claim.rb
```diff
@@ -111,6 +111,10 @@ def submitted?
     submitted_at.present?
   end
 
+  def can_be_submitted?
+    valid?(:submit)
+  end
```
erbridge: Since this exists now, could we use it for `submit!`, too?

app/models/tslr_claim.rb
```diff
@@ -77,8 +77,8 @@ class TslrClaim < ApplicationRecord
                                         length: {maximum: 256, message: "Email address must be 256 characters or less"},
                                         allow_nil: true
 
-  validates :bank_sort_code,            on: :"bank-details", presence: {message: "Enter a sort code"}
-  validates :bank_account_number,       on: :"bank-details", presence: {message: "Enter an account number"}
+  validates :bank_sort_code,            on: [:"bank-details", :submit], presence: {message: "Enter a sort code"}
+  validates :bank_account_number,       on: [:"bank-details", :submit], presence: {message: "Enter an account number"}
```
erbridge: :+1:

spec/features/tslr_claim_spec.rb
```diff
@@ -159,4 +159,115 @@
...
+      expect(page).to have_content("Check your answers before sending your application")
+    end
+
+    scenario "Teacher sees their original current school when editing" do
+      find("a[href='#{claim_path("current-school")}']").click
+
+      expect(find("input[name='school_search']").value).to eq(current_school.name)
+    end
+
+    context "Teacher goes to change their claim school" do
```
erbridge: Could we rephrase these so they read as the type of thing they are, do you think? i.e. `context "When changing claim school"`, `scenario "Teacher sees their original claim school"`, etc. Sorry to be nitpicky, but I think it's good to be consistent about how tests are written to help with reading them.

spec/features/tslr_claim_spec.rb
```diff
@@ -159,4 +159,115 @@
...
+      before do
+        find("a[href='#{claim_path("claim-school")}']").click
+      end
+
+      scenario "sees their original claim school" do
+        expect(find("input[name='school_search']").value).to eq(claim.claim_school.name)
+      end
+
+      context "changes their claim school" do
```
erbridge: "When choosing a new claim school"

spec/features/tslr_claim_spec.rb
```diff
@@ -159,4 +159,115 @@
...
+
+        scenario "school is changed correctly" do
+          expect(claim.reload.claim_school).to eql schools(:penistone_grammar_school)
+        end
+
+        scenario "Teacher is redirected to the are you still employed screen" do
+          expect(current_path).to eq(claim_path("still-teaching"))
+        end
+
+        context "is still teaching at the same school" do
```
erbridge: "When still teaching at the claim school"

spec/features/tslr_claim_spec.rb
```diff
@@ -159,4 +159,115 @@
...
+          before do
+            choose_still_teaching
+          end
+
+          it "current school is set correctly" do
```
erbridge: Should these be `scenario`s rather than `it`s?

spec/features/tslr_claim_spec.rb
```diff
@@ -159,4 +159,115 @@
...
+            expect(claim.reload.employment_status).to eql("claim_school")
+            expect(claim.reload.current_school).to eql schools(:penistone_grammar_school)
+          end
+
+          it "Teacher is redirected to the check your answers page" do
+            expect(current_path).to eq(claim_path("check-your-answers"))
+          end
+        end
+
+        context "is still teaching at another school" do
```
erbridge: "When still teaching but at a different school"

spec/features/tslr_claim_spec.rb
```diff
@@ -159,4 +159,115 @@
...
+            choose_still_teaching "Yes, at another school"
+
+            fill_in "School name", with: "Hampstead"
+            click_on "Search"
+
+            choose "Hampstead School"
+            click_on "Continue"
+          end
+
+          it "sets school and status correctly" do
```
erbridge: "School and employment status are set correctly"

spec/features/tslr_claim_spec.rb
```diff
@@ -159,4 +159,115 @@
...
+            expect(claim.reload.employment_status).to eql("different_school")
+            expect(claim.reload.current_school).to eql schools(:hampstead_school)
+          end
+
+          it "Teacher is redirected to the check your answers page" do
+            expect(current_path).to eq(claim_path("check-your-answers"))
+          end
+        end
+
+        context "is no longer teaching" do
```
erbridge: "When no longer teaching"

spec/features/tslr_claim_spec.rb
```diff
@@ -159,4 +159,115 @@
...
+          before do
+            choose_still_teaching "No"
+          end
+
+          it "sets the status" do
```
erbridge: "Employment status is set correctly"

spec/features/tslr_claim_spec.rb
```diff
@@ -159,4 +159,115 @@
...
+            expect(claim.reload.employment_status).to eq("no_school")
+          end
+
+          it "tells the teacher they are not eligible" do
```
erbridge: "Teacher is told they are not eligible"

spec/features/tslr_claim_spec.rb
pezholio: This gave me all sorts of bother when I tried to do it in the create - for some reason the `current_school` wasn't taking. I couldn't figure out why, so decided to do it this way.

erbridge: I see. That will be because of `before_save :update_current_school, if: :employment_status_changed?` in the model. Maybe that should guard against this situation, though it's possible that it will never come up in real situations.

spec/features/tslr_claim_spec.rb
pezholio: They're functionally the same, but I'd say `it` is more expressive when you're only checking the state of something

pezholio: On further inspection, it seems like the guidance is to use `scenario` in all cases, so happy to change this for consistency 👍

spec/features/tslr_claim_spec.rb
pezholio: Yeah, I don't think this will ever come up in real life. I can use `create` and skip callbacks/validations, but this will result in more code/hacks, so I think this is fine for now. I'll add a comment to point out why I'm doing it the way I'm doing for others

spec/features/tslr_claim_spec.rb
```diff
@@ -159,4 +159,120 @@
...
+      old_number = claim.national_insurance_number
+      new_number = "AB123456C"
+
+      expect {
+        find("a[href='#{claim_path("national-insurance-number")}']").click
+        fill_in "National Insurance number", with: new_number
+        click_on "Continue"
+      }.to change {
+        claim.reload.national_insurance_number
+      }.from(old_number).to(new_number)
```
erbridge: Nice! Didn't know you could do this.

pezholio: Yeah, it's pretty cool. I've never used it in features before, but seems to work 😄

PR #75, opened 2019-06-10
Ensure current school is selected if needed

robbiepaul:
If the user selects "still teaching at another school", we need to know which school they are teaching at. We don't add this validation to the "submit" context as it isn't always required.

https://trello.com/c/ciJxfVuu


robbiepaul:
[User can submit no current school without JS](https://trello.com/c/ciJxfVuu/211-user-can-submit-no-current-school-without-js)

PR #77, opened 2019-06-11
Highlight errors in form inputs as described in the design system

mec:
We needed a few things to display these errors so there is a `FormHelper` class to wrap up the methods.

Read about the errors and markup required:

https://design-system.service.gov.uk/

The FormHelper implements:

`form_group_tag` is used to render a form group for either the entire form or an individual element.

`errors_tag` is used to render error messages next to the appropriate element.

`css_classes_for_input` adds the appropriate classes to any `<input>` element.

`css_classes_for_select` adds the appropriate classes for a `<select>` element.

`form_field_id_for_errored_attribute` is used to override a error summary link

Becasue the GOVUK errors call for the red indicator bar to include the whole form group I had to change the two school search views to accomadate this. I took the oportunity to re-factor them to use the 'label as heading' and 'legend as heading' patterns:

https://design-system.service.gov.uk/get-started/labels-legends-headings/

After updating the views some of the tests were broken, Capybara could no longer select an element by the label text so this was fixed by changing to `fill_in(:school_search)`

## Trello

Trello: https://trello.com/c/Y73FJPN9

### Before
Screenshot_2019-06-10 Teachers claim back student loan repayments 

Screenshot_2019-06-10 Teachers claim back student loan repayments(1)

### After
Screenshot_2019-06-10 Teachers claim back student loan repayments(2)

Screenshot_2019-06-10 Teachers claim back student loan repayments(3)

### DIFF
--- a/app/helpers/govuk_form_helper.rb
+++ b/app/helpers/govuk_form_helper.rb
@@ -1,0 +1,127 @@
+module GovukFormHelper
+  # = GOVUK Form Tag Helpers
+  #
+  # For usage, components and patterns see:
+  # https://design-system.service.gov.uk/
+
+  # Can be used to wrap multiple or single components.
+  #
+  # Creates a div with the 'govuk-form-group' css class, if the provided object has errors, adds
+  # the 'govuk-form-group--error' css class.
+  #
+  # Passing the attribute symbol will only add the error class if the object contains errors for
+  # the matching symbol.
+  #
+  # Generally pass only an object when wrapping multiple components and pass an object and attribute when
+  # wrapping a single component.
+  #
+  # === Examples
+  # ==== Wrap multiple components
+  #
+  #   <%= form_group_tag object do %>
+  #     <fieldset class="govuk-fieldset">
+  #       <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
+  #         <h1 class="govuk-fieldset__heading">
+  #           Where do you live?
+  #         </h1>
+  #       </legend>
+  #     ...
+  #     </fieldset>
+  #   <% end %>
+  #
+  #   <div class="govuk-form-group">
+  #     <fieldset class="govuk-fieldset">
+  #       <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
+  #         <h1 class="govuk-fieldset__heading">
+  #           Where do you live?
+  #         </h1>
+  #       </legend>
+  #        ...
+  #     </fieldset>
+  #   </div>
+  #
+  # ==== Wrap a single component
+  #
+  #   <%= form_group_tag object, :event_name do %>
+  #     ...
+  #   <% end %>
+  #
+  #   <div class="govuk-form-group">
+  #     <label class="govuk-label" for="object_event_name">
+  #       Event name
+  #     </label>
+  #     <input class="govuk-input" id="object_event_name" name="event_name" type="text">
+  #   </div>
+  #
+  def form_group_tag(object, attribute = nil, &block)
+    css_classes = ["govuk-form-group"]
+    if attribute.present? && object.errors.key?(attribute)
+      css_classes << "govuk-form-group--error"
+    elsif attribute.nil? && object.errors.any?
+      css_classes << "govuk-form-group--error"
+    end
+    content_tag(:div, class: css_classes) do
+      yield
+    end
+  end
+
+  # Returns a set of +<span>+ tags that contains the error messages in _object_ that match the
+  # _attribute_ symbol.
+  #
+  # === Examples
+  #   errors_tag(object, :attribute)
+  #
+  #   <span id="object_attibute_name-error" class="govuk-error-message">
+  #     <span class="govuk-visually-hidden">Error: </span>Error message one<br>
+  #     <span class="govuk-visually-hidden">Error: </span>Error message two
+  #   </span>
+  #
+  def errors_tag(object, attribute)
+    return unless object.errors.messages[attribute].present?
+
+    messages = object.errors.messages[attribute].map { |message| content_tag(:span, "Error:", class: "govuk-visually-hidden") + " " + message }
+    content_tag(:span, messages.join("<br>").html_safe, {class: "govuk-error-message", id: error_id(object.class.model_name.singular, attribute)})
+  end
+
+  # Returns a string of the GOVUK css classes for the +<select>+ html element along with any
+  # existing css classes provided.
+  #
+  # When _object_ contains errors for _attribute_ the 'govuk-input--error' css class is added.
+  #
+  # === Example
+  #
+  # css_classes_for_select(object, :attribute, "another-css-class")
+  #
+  #   "govuk-select another-css-class"
+  #
+  def css_classes_for_select(object, attribute, css_classes = "")
+    css_classes = css_classes.split
+    css_classes << "govuk-select"
+    css_classes << "govuk-select--error" if object.errors.key?(attribute)
+    css_classes.join(" ")
+  end
+
+  # Returns a string of the GOVUK css classes for the +<input>+ html element along with any
+  # existing css classes provided.
+  #
+  # When _object_ contains errors for _attribute_ the 'govuk-input--error' css class is added.
+  #
+  # === Example
+  #
+  # css_classes_for_input(object, :attribute, "another-css-class")
+  #
+  #   "govuk-select another-css-class"
+  #
+  def css_classes_for_input(object, attribute, css_classes = "")
+    css_classes = css_classes.split
+    css_classes << "govuk-input"
+    css_classes << "govuk-input--error" if object.errors.key?(attribute)
+    css_classes.join(" ")
+  end
+
+  private
+
+  def error_id(object_name, attribute)
+    "#{object_name}_#{attribute}-error"
+  end
+end
--- a/app/views/claims/address.html.erb
+++ b/app/views/claims/address.html.erb
@@ -10,46 +10,53 @@
           </h1>
         </legend>
 
-        <div class="govuk-form-group">
+        <%= form_group_tag current_claim, :address_line_1 do %>
           <%= form.label :address_line_1, class: "govuk-label" do %>
             Building and street <span class="govuk-visually-hidden">line 1 of 2</span>
           <% end %>
-          <%= form.text_field :address_line_1, autocomplete: "address-line1", class: "govuk-input" %>
-        </div>
+          <%= errors_tag current_claim, :address_line_1 %>
+          <%= form.text_field :address_line_1, autocomplete: "address-line1",
+            class: css_classes_for_input(current_claim, :address_line_1) %>
+        <% end %>
 
-
-        <div class="govuk-form-group">
+        <%= form_group_tag current_claim, :address_line_2 do %>
           <%= form.label :address_line_2, class: "govuk-label" do %>
             <span class="govuk-visually-hidden">Building and street line 2 of 2</span>
           <% end %>
-          <%= form.text_field :address_line_2, autocomplete: "address-line2", class: "govuk-input" %>
-        </div>
+          <%= errors_tag current_claim, :address_line_2 %>
+          <%= form.text_field :address_line_2, autocomplete: "address-line2",
+            class: "govuk-input" %>
+        <% end %>
 
-        <div class="govuk-form-group">
+        <%= form_group_tag current_claim, :address_line_3 do %>
           <%= form.label :address_line_3, class: "govuk-label" do %>
             Town or city
           <% end %>
-          <%= form.text_field :address_line_3, autocomplete: "address-line3", class: "govuk-input govuk-!-width-two-thirds" %>
-        </div>
+          <%= errors_tag current_claim, :address_line_3 %>
+          <%= form.text_field :address_line_3, autocomplete: "address-line3",
+            class: css_classes_for_input(current_claim, :address_line_3, "govuk-!-width-two-thirds") %>
+        <% end %>
 
-        <div class="govuk-form-group">
+        <%= form_group_tag current_claim, :address_line_4 do %>
           <%= form.label :address_line_4, class: "govuk-label" do %>
             County
           <% end %>
-          <%= form.text_field :address_line_4, autocomplete: "region", class: "govuk-input govuk-!-width-two-thirds" %>
-        </div>
+          <%= errors_tag current_claim, :address_line_4 %>
+          <%= form.text_field :address_line_4, autocomplete: "region",
+            class: "govuk-input govuk-!-width-two-thirds" %>
+        <% end %>
 
-        <div class="govuk-form-group">
+        <%= form_group_tag current_claim, :postcode do %>
           <%= form.label :postcode, class: "govuk-label" do %>
             Postcode
           <% end %>
-          <%= form.text_field :postcode, autocomplete: "postal-code", class: "govuk-input govuk-input--width-10" %>
-        </div>
-      </fieldset>
+          <%= errors_tag current_claim, :postcode %>
+          <%= form.text_field :postcode, autocomplete: "postal-code",
+            class: css_classes_for_input(current_claim, :postcode, "govuk-input--width-10") %>
+        <% end %>
 
-      <div class="govuk-form-group">
-        <%= form.submit "Continue", class: "govuk-button" %>
-      </div>
+      </fieldset>
+      <%= form.submit "Continue", class: "govuk-button" %>
     <% end %>
   </div>
 </div>
--- a/app/views/claims/bank_details.html.erb
+++ b/app/views/claims/bank_details.html.erb
@@ -11,33 +11,29 @@
             </h1>
           </legend>
 
-          <p class="govuk-body govuk-hint">
+          <span class="govuk-hint">
             We need to collect this information to pay you.
-          </p>
+          </span>
 
-          <div class="govuk-form-group">
+          <%= form_group_tag current_claim, :bank_sort_code do %>
             <%= form.label :bank_sort_code, "Sort code", class: "govuk-label" %>
-            <span  id="sort-code-hint" class="govuk-hint">For example: 44 00 26</span>
+            <span id="sort-code-hint" class="govuk-hint">For example: 44 00 26</span>
+            <%= errors_tag current_claim, :bank_sort_code %>
             <%= form.text_field :bank_sort_code,
-              class: "govuk-input govuk-!-width-one-quarter",
-              "aria-describedby" => "sort-code-hint"
-            %>
-          </div>
+              class: css_classes_for_input(current_claim, :bank_sort_code, "govuk-!-width-one-quarter"),
+              "aria-describedby" => "sort-code-hint" %>
+          <% end %>
 
-          <div class="govuk-form-group">
+          <%= form_group_tag current_claim, :bank_account_number do %>
             <%= form.label :bank_account_number, "Account number", class: "govuk-label" %>
-            <span  id="account-number-hint" class="govuk-hint">For example: 70 87 24 90</span>
+            <span id="account-number-hint" class="govuk-hint">For example: 70 87 24 90</span>
+            <%= errors_tag current_claim, :bank_account_number %>
             <%= form.text_field :bank_account_number,
-              class: "govuk-input govuk-input--width-20",
-              "aria-describedby" => "account-number-hint"
-            %>
-          </div>
-
+              class: css_classes_for_input(current_claim, :bank_account_number, "govuk-input--width-20"),
+              "aria-describedby" => "account-number-hint" %>
+          <% end %>
         </fieldset>
-
-        <div class="govuk-form-group">
-          <%= form.submit "Continue", class: "govuk-button" %>
-        </div>
+        <%= form.submit "Continue", class: "govuk-button" %>
       <% end %>
   </div>
 </div>
--- a/app/views/claims/claim_school.html.erb
+++ b/app/views/claims/claim_school.html.erb
@@ -3,27 +3,27 @@
 
     <%= render("shared/error_summary", instance: current_claim) if current_claim.errors.any? %>
 
-    <h1 class="govuk-heading-xl">
-      <%= t("tslr.questions.claim_school") %>
-    </h1>
+
+      <h1 class="govuk-heading-xl">
+        <%= t("tslr.questions.claim_school") %>
+      </h1>
 
     <% if @schools.blank? %>
-
-      <% if params[:school_search].present? && current_claim.errors.empty? %>
-        <p class="govuk-body">
-          <strong>No results match that search term. Try again.</strong>
-        </p>
-      <% else %>
-        <span id="school-search-hint" class="govuk-hint">
-          Search for the school name, use at least four characters.
-        </span>
-      <% end %>
-
+        <% if params[:school_search].present? && current_claim.errors.empty? %>
+          <p class="govuk-body">
+            <strong>No results match that search term. Try again.</strong>
+          </p>
+        <% else %>
+          <span id="school-search-hint" class="govuk-hint">
+            Search for the school name, use at least four characters.
+          </span>
+        <% end %>
       <%= form_tag(claim_path("claim-school"), method: :get) do %>
-        <div class="govuk-form-group">
-          <%= label_tag :school_search, "School name", class: "govuk-label" %>
-          <%= text_field_tag :school_search, params[:school_search], class: "govuk-input", value: current_claim.claim_school&.name %>
-        </div>
+          <%= form_group_tag current_claim, :school_search do %>
+            <%= label_tag :school_search, "School name", class: "govuk-label" %>
+            <%= errors_tag current_claim, :school_search %>
+            <%= text_field_tag :school_search, params[:school_search], id: :tslr_claim_school_search, class: css_classes_for_input(current_claim, :school_search), value: current_claim.claim_school&.name  %>
+          <% end %>
         <div class="govuk-form-group">
           <%= submit_tag "Search", class: "govuk-button" %>
         </div>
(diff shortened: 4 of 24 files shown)


app/helpers/govuk_form_helper.rb
```diff
@@ -0,0 +1,131 @@
...
+  #   <div class="govuk-form-group">
+  #     <label class="govuk-label" for="object_event_name">
+  #       Event name
+  #     </label>
+  #     <input class="govuk-input" id="object_event_name" name="event_name" type="text">
+  #   </div>
+  #
+  def form_group_tag(object, attribute = nil, &block)
+    css_classes = ["govuk-form-group"]
+    case attribute
```
tekin: Given there is only a single branch is this `case` I'd be tempted to write this as an `if` instead. Maybe even without the nested conditionals too:

```ruby
if attribute.present? && object.errors.key?(attribute)
  css_classes += ["govuk-form-group--error"]
elsif object.errors.any?
  css_classes += ["govuk-form-group--error"]
end
```

app/helpers/govuk_form_helper.rb
```diff
@@ -0,0 +1,131 @@
...
+    when Symbol
+      css_classes += ["govuk-form-group--error"] if object.errors.key?(attribute)
```
pezholio: I think `<<` is preferred over `+=`, becuase it's faster. `+=` creates a new object, while `<<` appends onto the existing object

app/helpers/govuk_form_helper.rb
```diff
@@ -0,0 +1,131 @@
...
+  #   </span>
+  #
+  def errors_tag(object, attribute)
+    return if object.errors.empty?
+    messages = object.errors.messages[attribute]
+
+    messages.map! do |message|
+      content_tag(:span, "Error: ", class: "govuk-visually-hidden") + message
+    end
+    content_tag(:span, messages.join("</br>").html_safe, class: "govuk-error-message", id: error_id(object.class.model_name.singular, attribute))
```
tekin: If the object has errors, but not on the specific attribute, won't we end up with an empty `<span id="object_attibute_name-error" class="govuk-error-message">`?

tekin: The break tag should be either `<br>` or `<br />`. I don't believe `</br>` is a valid tag.

app/helpers/govuk_form_helper.rb
```diff
@@ -0,0 +1,131 @@
...
+  #
+  # === Example
+  #
+  # css_classes_for_select(object, :attribute, "another-css-class")
+  #
+  #   "govuk-select another-css-class"
+  #
+  def css_classes_for_select(object, attribute, css_classes = "")
+    css_classes = css_classes.split
+    css_classes += ["govuk-select"]
```
pezholio: Again, I'd use the `<<` operator here

app/helpers/govuk_form_helper.rb
```diff
@@ -0,0 +1,131 @@
...
+  #
+  # === Example
+  #
+  # css_classes_for_input(object, :attribute, "another-css-class")
+  #
+  #   "govuk-select another-css-class"
+  #
+  def css_classes_for_input(object, attribute, css_classes = "")
+    css_classes = css_classes.split
+    css_classes += ["govuk-input"]
```
pezholio: And another `<<`

spec/features/school_search_spec.rb
```diff
@@ -5,7 +5,7 @@
     claim = start_tslr_claim
     choose_qts_year
 
-    fill_in "School name", with: "Penistone"
```
pezholio: As the change you made in the previous commit broke the tests, I'd probably squash this commit into the previous one, with a note in the description about why you had to alter the tests

spec/helpers/form_helper_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+        end
+
+        it "adds the error class when there are errors on the attribute" do
+          expect(helper.form_group_tag(claim, :attribute) {}).to have_css(".govuk-form-group--error")
+        end
+      end
+    end
+  end
+
+  describe "#css_classes_for_input" do
```
tekin: We can be more precise for all these specs, right? i.e. `eq` instead of `include`?

app/views/claims/address.html.erb
```diff
@@ -3,50 +3,54 @@
...
-        <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
-          <h1 class="govuk-fieldset__heading">
-            <%= t("tslr.questions.address") %>
-          </h1>
-        </legend>
-
-        <div class="govuk-form-group">
-          <%= form.label :address_line_1, class: "govuk-label" do %>
-            Building and street <span class="govuk-visually-hidden">line 1 of 2</span>
+      <%= form_group_tag current_claim do %>
```
tekin: These outer groups weren't here before, and now we have nested form groups. Is that right? Also is it an issue that the outer group will get the errors class applied to it?

app/helpers/govuk_form_helper.rb
```diff
@@ -0,0 +1,131 @@
+module GovukFormHelper
```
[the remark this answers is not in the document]
mec: Yeah, I'd like to do more, but for this I just wanted to meet the ACs.

spec/features/school_search_spec.rb
mec: Good call. Thank you.

app/views/claims/claim_school.html.erb
```diff
@@ -3,27 +3,27 @@
...
+        <% end %>
       <%= form_tag(claim_path("claim-school"), method: :get) do %>
-        <div class="govuk-form-group">
-          <%= label_tag :school_search, "School name", class: "govuk-label" %>
-          <%= text_field_tag :school_search, params[:school_search], class: "govuk-input", value: current_claim.claim_school&.name %>
-        </div>
+          <%= form_group_tag current_claim, :school_search do %>
+            <%= label_tag :school_search, "School name", class: "govuk-label" %>
+            <%= errors_tag current_claim, :school_search %>
+            <%= text_field_tag :school_search, params[:school_search], id: :tslr_claim_school_search, class: css_classes_for_input(current_claim, :school_search), value: current_claim.claim_school&.name  %>
```
tekin: For some reason the error message link doesn't move focus into the text field for the school search in the same way it does for the other text fields when there are errors. Any ideas?

app/helpers/govuk_form_helper.rb
mec: Nice, thanks! And looks nicer!

app/views/claims/bank_details.html.erb
```diff
@@ -19,22 +20,22 @@
             <%= form.label :bank_sort_code, "Sort code", class: "govuk-label" %>
             <span  id="sort-code-hint" class="govuk-hint">For example: 44 00 26</span>
             <%= form.text_field :bank_sort_code,
-              class: "govuk-input govuk-!-width-one-quarter",
+              class: css_classes_for_input(current_claim, :bank_sort_code, "govuk-!-width-one-quarter"),
```
tekin: Do we also need a `<%= errors_tag current_claim, :bank_sort_code %>` to go with this text field?

app/views/claims/bank_details.html.erb
```diff
@@ -19,22 +20,22 @@
...
               "aria-describedby" => "sort-code-hint"
             %>
           </div>
 
           <div class="govuk-form-group">
             <%= form.label :bank_account_number, "Account number", class: "govuk-label" %>
             <span  id="account-number-hint" class="govuk-hint">For example: 70 87 24 90</span>
-            <%= form.text_field :bank_account_number,
-              class: "govuk-input govuk-input--width-20",
+            <%= form.text_field :bank_account_number,  class: css_classes_for_input(current_claim, :bank_account_number, "
```
tekin: As above, do we also need a `<%= errors_tag current_claim, :bank_account_number %>` to go with this text field?

app/views/claims/bank_details.html.erb
```diff
@@ -4,6 +4,7 @@
       <%= render("shared/error_summary", instance: current_claim) if current_claim.errors.any? %>
 
       <%= form_for current_claim, url: claim_path do |form| %>
+        <%= form_group_tag current_claim do %>
```
tekin: Because of this outer group both fields bank fields end up with the red line, even when only one of the fields have an error. That may be OK but thought I'd point it out.

app/helpers/govuk_form_helper.rb
```diff
@@ -0,0 +1,131 @@
...
+  #   <span id="object_attibute_name-error" class="govuk-error-message">
+  #     <span class="govuk-visually-hidden">Error: </span>Error message one<br>
+  #     <span class="govuk-visually-hidden">Error: </span>Error message two
+  #   </span>
+  #
+  def errors_tag(object, attribute)
+    return if object.errors.empty?
+    messages = object.errors.messages[attribute]
+
+    messages.map! do |message|
```
erbridge: Does `!`ing this mean we're modifying `object.errors.messages[attribute]`, since we're interacting with a reference?

app/helpers/govuk_form_helper.rb
```diff
@@ -0,0 +1,131 @@
...
+      content_tag(:span, "Error: ", class: "govuk-visually-hidden") + message
+    end
+    content_tag(:span, messages.join("</br>").html_safe, class: "govuk-error-message", id: error_id(object.class.model_name.singular, attribute))
```
erbridge: Is it safe to `html_safe` this? Do errors ever contain the content users input, and if so, are we sanitizing it somewhere?

app/views/claims/claim_school.html.erb
```diff
@@ -3,57 +3,62 @@
```
erbridge: This is going to be fun to merge with the autocomplete changes...

app/views/claims/claim_school.html.erb
```diff
@@ -3,57 +3,62 @@
 
     <%= render("shared/error_summary", instance: current_claim) if current_claim.errors.any? %>
 
-    <h1 class="govuk-heading-xl">
-      <%= t("tslr.questions.claim_school") %>
-    </h1>
-
     <% if @schools.blank? %>
```
erbridge: At this point, these might as well be two separate views, since the two cases share next to nothing.

app/views/claims/email_address.html.erb
```diff
@@ -3,19 +3,18 @@
...
-      <div class="govuk-form-group">
-        <%= form.email_field :email_address, autocomplete: "email", spellcheck: "false", class: "govuk-input" %>
-      </div>
-
-      <div class="govuk-form-group">
-        <%= form.submit "Continue", class: "govuk-button" %>
-      </div>
+        <span class="govuk-hint">We will only use your email address to update you about your claim.</span>
+        <%= errors_tag current_claim, :email_address %>
+        <div class="govuk-form-group">
```
erbridge: Should this be another `form_group_tag`? Or should we lose this div entirely like happened with `full_name`?

app/helpers/govuk_form_helper.rb
mec: Much better, thanks. Ended up here:

```ruby
if attribute.present? && object.errors.key?(attribute)
  css_classes << "govuk-form-group--error"
elsif attribute.nil? && object.errors.any?
  css_classes << "govuk-form-group--error"
end
```

app/helpers/govuk_form_helper.rb
mec: Absolutely! I've changed the guard:

`return unless object.errors.messages[attribute].present?`

app/views/claims/national_insurance_number.html.erb
```diff
@@ -3,19 +3,18 @@
...
 
-      <div class="govuk-form-group">
-        <%= form.text_field :national_insurance_number, spellcheck: "false", class: "govuk-input" %>
-      </div>
-
-      <div class="govuk-form-group">
-        <%= form.submit "Continue", class: "govuk-button" %>
-      </div>
+        <div class="govuk-form-group">
+          <%= form.text_field :national_insurance_number, spellcheck: "false", class: css_classes_for_input(current_claim, :national_insurance_number) %>
```
erbridge: Does this need a matching `errors_tag`?

app/helpers/govuk_form_helper.rb
mec: I thought I'd changed that <br>…

spec/helpers/form_helper_spec.rb
mec: We can!

app/views/claims/bank_details.html.erb
mec: Yeah, it's a good point. I think I need to revisting the GOVUK design system…

app/views/claims/bank_details.html.erb
mec: Yes, thank you.

app/views/claims/bank_details.html.erb
mec: oh…

app/helpers/govuk_form_helper.rb
erbridge: Is `attribute.nil?` necessary? What should happen if e.g. `""` is passed in as the `attribute`?

app/helpers/govuk_form_helper.rb
mec: Added a spec to check this.

app/helpers/govuk_form_helper.rb
mec: Yes, I wasn't sure about that, I've got the `dup` now do you think that is better?

erbridge: I think that works. You could also reassign with a plain `map`, e.g. `messages = message.map do ...`

app/helpers/govuk_form_helper.rb
```diff
@@ -0,0 +1,130 @@
...
+  #     <span class="govuk-visually-hidden">Error: </span>Error message one<br>
+  #     <span class="govuk-visually-hidden">Error: </span>Error message two
+  #   </span>
+  #
+  def errors_tag(object, attribute)
+    return unless object.errors.messages[attribute].present?
+    messages = object.errors.messages[attribute].dup
+
+    messages.map! do |message|
+      content_tag(:span, "Error:", class: "govuk-visually-hidden") + " " + message
```
erbridge: Shouldn't the `" "` also be hidden?

app/views/claims/national_insurance_number.html.erb
mec: Yes! thank you, it's like I have `error_tag` blindness!

app/views/claims/teacher_reference_number.html.erb
```diff
@@ -3,19 +3,18 @@
...
 
-      <div class="govuk-form-group">
-        <%= form.text_field :teacher_reference_number, spellcheck: "false", class: "govuk-input" %>
-      </div>
-
-      <div class="govuk-form-group">
-        <%= form.submit "Continue", class: "govuk-button" %>
-      </div>
+        <div class="govuk-form-group">
+          <%= form.text_field :teacher_reference_number, spellcheck: "false", class: css_classes_for_input(current_claim, :teacher_reference_number) %>
```
erbridge: Missing `errors_tag`?

app/views/claims/student_loan_amount.html.erb
```diff
@@ -3,21 +3,20 @@
...
-      <span class="govuk-hint">Only include repayments made through your teaching wages. You can find this information on your payslips.</span>
-      <div class="govuk-form-group govuk-currency-input">
-        <div class="govuk-currency-input__inner">
-          <span class="govuk-currency-input__inner__unit">&pound;</span>
-          <%= form.text_field :student_loan_repayment_amount, class: "govuk-currency-input__inner__input govuk-input govuk-input--width-5" %>
+        <span class="govuk-hint">Only include repayments made through your teaching wages. You can find this information on your payslips.</span>
+        <div class="govuk-form-group govuk-currency-input">
+          <div class="govuk-currency-input__inner">
+            <span class="govuk-currency-input__inner__unit">&pound;</span>
+            <%= form.text_field :student_loan_repayment_amount, class: css_classes_for_input(current_claim, :student_loan_repayment_amount, "govuk-currency-input__inner__input govuk-input--width-5") %>
```
erbridge: Missing `errors_tag`?

app/views/claims/claim_school.html.erb
mec: Yup. I think it's on me?

app/views/claims/date_of_birth.html.erb
```diff
@@ -3,39 +3,40 @@
...
+          <span id="date-of-birth-hint" class="govuk-hint">
+            For example, 31 3 1980. We need this information to verify your identity.
+          </span>
 
-        <div class="govuk-date-input">
-          <div class="govuk-date-input__item">
-            <div class="govuk-form-group">
-              <%= label_tag :"tslr_claim_date_of_birth_3i", "Day", class: "govuk-label govuk-date-input__label" %>
-              <%= text_field_tag :"tslr_claim[date_of_birth(3i)]", form.object.date_of_birth.try(:day), id: "tslr_claim_date_of_birth_3i", class: "govuk-input govuk-date-input__input govuk-input--width-2", type: "number", autocomplete: "bday-day", pattern: "[0-9]*" %>
+          <div class="govuk-date-input">
```
erbridge: Missing `errors_tag`?

app/views/claims/current_school.html.erb
```diff
@@ -3,53 +3,57 @@
...
+                <%= t("tslr.questions.current_school") %>
+              </h1>
+            </legend>
+            <span id="school-search-result-hint" class="govuk-hint">
+              Select your school from the search results.
+            </span>
+
+            <%= hidden_field_tag :school_search, params[:school_search] %>
+
+              <%= errors_tag current_claim, :claim_school %>
```
erbridge: Indentation looks out here.

app/views/claims/claim_school.html.erb
```diff
@@ -3,57 +3,62 @@
...
+                <%= t("tslr.questions.claim_school") %>
+              </h1>
+            </legend>
+            <span id="school-search-result-hint" class="govuk-hint">
+              Select your school from the search results.
+            </span>
+
+            <%= hidden_field_tag :school_search, params[:school_search] %>
+
+              <%= errors_tag current_claim, :claim_school %>
```
erbridge: Indentation.

app/views/claims/email_address.html.erb
mec: yep, the second one is redundant!

app/views/claims/claim_school.html.erb
mec: Does that have any impact on the autocomplete?

app/views/claims/claim_school.html.erb
erbridge: Autocomplete might be held up for other reasons. I'll happily work out how to merge it if this goes in first.

app/helpers/govuk_form_helper.rb
mec: Yeah, I prefer your simpler approach.

app/views/claims/claim_school.html.erb
erbridge: Well, it involves changing the controller, so yes, some.

app/helpers/govuk_form_helper.rb
mec: :D amazing! All the examples have it there, I did have it the other way but decided to go with the letter of the law:

https://design-system.service.gov.uk/components/error-message/

app/views/claims/teacher_reference_number.html.erb
mec: I must have messed something up. I can't have missed all these… worrying.

app/helpers/govuk_form_helper.rb
erbridge: Seems wrong to me, as it leaves a potentially visible whitespace, depending on what it's adjacent to, but sure.

app/views/claims/teacher_reference_number.html.erb
erbridge: Suggest looking through everywhere you've added the new tags and double checking?

mec: yep

mec: I think I have them all, the DOB one is a bit odd, I'll have to look at it tomorrow.

app/helpers/govuk_form_helper.rb
mec: I set `attribute` to `nil` by default (line 56). If you pass "" there would be no match and the error css class would not get added, which is my expected behaviour.

The idea was that not passing `attribute` implies you are using the form group for the whole 'question' (e.g. fullname) whereas passing `attribute` implies use in a multi-field form (e.g. address).

I did have two methods `form_group_tag` and `field_group_tag` perhaps that would be more explicit?

app/views/claims/claim_school.html.erb
mec: Great spot, it turned out to be the `id` of the label `for` not matching, I've declated both `id`s now and it is working.

app/helpers/govuk_form_helper.rb
tekin: Yeah it should be safe: if we ever interpolated the value into the error message (e.g. `"%{value} is not a valid email"`, Rails escapes `value`.

app/views/claims/address.html.erb
mec: Looking at this, the outer group is wrong, I've updated it, thank you.

app/helpers/govuk_form_helper.rb
mec: I don't believe error messages ever contain user input in this context.

I can set content_tag to not escape which makes the `html_safe` unnescessary, do we think that is a better approach?

mec: I'll stick with what it was, it feels like setting `content_tag` to not escape maybe be a worse in the long run.

app/helpers/govuk_form_helper.rb
erbridge: I guess I'm just trying to work out what a `blank?` attribute that wasn't `nil` would mean, but honestly, it's unlikely to happen anyway.

app/helpers/govuk_form_helper.rb
erbridge: @tekin I didn't know that. Thanks.

@mec I think that would have the same ultimate effect. I think `html_safe` is clearer.

app/views/claims/bank_details.html.erb
```diff
@@ -11,33 +11,29 @@
...
             <%= form.label :bank_sort_code, "Sort code", class: "govuk-label" %>
             <span  id="sort-code-hint" class="govuk-hint">For example: 44 00 26</span>
+            <%= errors_tag current_claim, :bank_sort_code %>
             <%= form.text_field :bank_sort_code,
-              class: "govuk-input govuk-!-width-one-quarter",
-              "aria-describedby" => "sort-code-hint"
-            %>
-          </div>
+              class: css_classes_for_input(current_claim, :bank_sort_code, "govuk-!-width-one-quarter"),
+                "aria-describedby" => "sort-code-hint" %>
```
erbridge: Indentation nitpick (it's off by a space-ish).

app/views/claims/student_loan_amount.html.erb
```diff
@@ -3,21 +3,21 @@
...
+        <h1 class="govuk-form-wrapper">
+          <%= form.label :student_loan_repayment_amount, t("tslr.questions.student_loan_amount", claim_school_name: current_claim.claim_school_name), class: "govuk-label govuk-label--xl" %>
+        </h1>
+
+        <span class="govuk-hint">Only include repayments made through your teaching wages. You can find this information on your payslips.</span>
+
+        <%= errors_tag current_claim, :student_loan_repayment_amount %>
 
-      <span class="govuk-hint">Only include repayments made through your teaching wages. You can find this information on your payslips.</span>
-      <div class="govuk-form-group govuk-currency-input">
```
erbridge: Is it ok that we've lost `govuk-currency-input`?

mec: Yes, but I've tied up the styles to match now as well.

app/views/claims/claim_school.html.erb
mec: I think I will leave this for now, we kept it all in the same view for simplicity and we can always seperate it later.

app/views/claims/address.html.erb
```diff
@@ -10,46 +10,51 @@
...
+        <% end %>
 
         <div class="govuk-form-group">
           <%= form.label :address_line_4, class: "govuk-label" do %>
             County
           <% end %>
-          <%= form.text_field :address_line_4, autocomplete: "region", class: "govuk-input govuk-!-width-two-thirds" %>
+          <%= form.text_field :address_line_4, autocomplete: "region",
+            class: "govuk-input govuk-!-width-two-thirds" %>
         </div>
```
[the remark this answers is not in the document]
mec: As far as my understanding goes, this is correct. We only validate that values are present so the two values that are not required should not be highlighted, which is what I believe the function of the 'red bar' is:

Screenshot_2019-06-12 Teachers claim back student loan repayments

mec:
[Highlight errors in form inputs as described in the design system](https://trello.com/c/Y73FJPN9/125-highlight-errors-in-form-inputs-as-described-in-the-design-system)

app/views/claims/address.html.erb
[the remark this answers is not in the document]
erbridge: Sorry, what I meant was that they should get their own `errors_tag` and `form_group_tag` so if we did start validating them, they would have the correct treatment. I don't see any reason not to include them on every form, and it sets a precedent to always use them. I can imagine this being a thing that is easily missed when changing validations.

mec:
@pezholio @tekin @erbridge thank you for all the support and patience!

PR #79, opened 2019-06-11
Set up Docker and first-pass Azure web app

erbridge:
The `Dockerfile` builds 3 images: a `dependencies` image containing the JS and Ruby dependencies and things needed to build and install them, a `web` image which contains the app (but not the specs) copying dependencies from the `dependencies` image, and a `base` image both of them depend on. `web` is the image that actually gets uploaded and run. We have opted for explicit copies over relying on the `.dockerignore`. The entries in `.dockerignore` are to speed up the build as Docker makes a copy of the build context (the target directory) before running the build, which can be quite large including the temporary files.

The app can be built in development or production mode using `--build-arg`s (defaulting to production), and there's a `docker-compose.yml` set up to build and run a functioning development environment.

Only the development environment is set up. The general template is `azure/template.json` and the parameters belong in `azure/parameters/$ENVIRONMENT_NAME.json`.

The template uses templates from the `master` branch of [DFE-Digital/bat-platform-building-blocks](https://github.com/DFE-Digital/bat-platform-building-blocks). Probably we should pin the version of that at some point.

Parameters are currently included in the repo, but include secrets that should be reset, ignored, and shared via whatever secret management system we end up using. We're currently using a manually built and pushed docker image in @pezholio's DockerHub account.

We are currenly using a new, throwaway resource group to avoid hitting the groups that appear to have been set up for us, until we better understand why they're there.

Note, this changes the production database password envirnment variable to conform with Azure's expected format (Azure has many of those).

To deploy to Azure, run `bin/azure-deploy development`. The script depends on having the Azure CLI installed (`brew install azure-cli`).

### DIFF
--- a/azure/parameters/development.json
+++ b/azure/parameters/development.json
@@ -1,0 +1,36 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "appServiceDockerImage": {
+      "value": "pezholio/teacher-payment-service"
+    },
+    "databaseName": {
+      "value": "development"
+    },
+    "databaseUsername": {
+      "value": "user"
+    },
+    "databasePassword": {
+      "value": "8dXuq*jvAt-4beU"
+    },
+    "securityAlertEmailAddress": {
+      "value": "foo@example.com"
+    },
+    "SECRET_KEY_BASE": {
+      "value": "827a46887bf03dfbe280c10dc00110ef44c7608e0dadbb16f2fbdd1cace12c005801d847c04db0a132f2f2b9de861eee5dd5b43a979613cb985ff857bdbb7a6b"
+    },
+    "DFE_SIGN_IN_ISSUER": {
+      "value": ""
+    },
+    "DFE_SIGN_IN_REDIRECT_URL": {
+      "value": ""
+    },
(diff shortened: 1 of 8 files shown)


erbridge:
[MVP Cloud Infrastructure Platform deployment](https://trello.com/c/0WuN4WDX/187-mvp-cloud-infrastructure-platform-deployment)

misaka:
Looks pretty solid and well thought-out! Definitely will be stealing some of this if we ever re-work our build system.

Have you considered using the [`config` gem](https://github.com/railsconfig/config) to standardise use of app settings with env var overrides?

azure/parameters/development.json
```diff
@@ -0,0 +1,36 @@
...
+      "value": "pezholio/teacher-payment-service"
+    },
+    "databaseName": {
+      "value": "development"
+    },
+    "databaseUsername": {
+      "value": "user"
+    },
+    "databasePassword": {
+      "value": "8dXuq*jvAt-4beU"
```
robbiepaul: Just to confirm, this is a placeholder and will be reset?

erbridge: Yes. Definitely. All of the secrets in here will be reset when we decide what to do with secrets.

erbridge:
@robbiepaul No. We did that with TVS, and it slowed everything down in a very frustrating way. The `docker-compose.json` was put together for local testing, and should probably be kept up to date and working, but let's stick to running things natively.

PR #80, opened 2019-06-11
Validate that `current_school` is set on submit

erbridge:
It's set by the `update_current_school` callback if the user selects same school, otherwise the user should be setting it manually.

We now run `update_current_school` on validation as well as saving. This is fine to do as the value of it is invisible to the user. The other callbacks only run on save so we show the user the values they entered if there's an error with them, not a normalized version of them.

### DIFF
--- a/app/models/tslr_claim.rb
+++ b/app/models/tslr_claim.rb
@@ -39,7 +39,7 @@ class TslrClaim < ApplicationRecord
   belongs_to :current_school, optional: true, class_name: "School"
 
   validates :claim_school,              on: [:"claim-school", :submit], presence: {message: "Select a school from the list"}
-  validates :current_school,            on: [:"current-school"], presence: {message: "Select a school from the list"}
+  validates :current_school,            on: [:"current-school", :submit], presence: {message: "Select a school from the list"}
 
   validates :qts_award_year,            on: [:"qts-year", :submit], inclusion: {in: VALID_QTS_YEARS, message: "Select the academic year you were awarded qualified teacher status"}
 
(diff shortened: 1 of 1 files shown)


erbridge:
[User can submit no current school without JS](https://trello.com/c/ciJxfVuu/211-user-can-submit-no-current-school-without-js)

app/models/tslr_claim.rb
```diff
@@ -84,11 +84,11 @@ class TslrClaim < ApplicationRecord
   validate  :bank_account_number_must_be_eight_digits
   validate  :bank_sort_code_must_be_six_digits
 
-  before_save :update_current_school, if: :employment_status_changed?
-  before_save :normalise_trn, if: :teacher_reference_number_changed?
-  before_save :normalise_ni_number, if: :national_insurance_number_changed?
-  before_save :normalise_bank_account_number, if: :bank_account_number_changed?
-  before_save :normalise_bank_sort_code, if: :bank_sort_code_changed?
+  before_validation :update_current_school, if: :employment_status_changed?
+  before_validation :normalise_trn, if: :teacher_reference_number_changed?
```
robbiepaul: When we were originally working on this we briefly discussed doing these normalisations earlier in the lifecycle. However it would mean that if the user made an error, they would see the normalised version in the form which might be a bit concerning. Happy to discuss it

app/models/tslr_claim.rb
```diff
@@ -84,11 +84,11 @@ class TslrClaim < ApplicationRecord
   validate  :bank_account_number_must_be_eight_digits
   validate  :bank_sort_code_must_be_six_digits
 
-  before_save :update_current_school, if: :employment_status_changed?
-  before_save :normalise_trn, if: :teacher_reference_number_changed?
-  before_save :normalise_ni_number, if: :national_insurance_number_changed?
-  before_save :normalise_bank_account_number, if: :bank_account_number_changed?
-  before_save :normalise_bank_sort_code, if: :bank_sort_code_changed?
+  before_validation :update_current_school, if: :employment_status_changed?
```
robbiepaul: 👍

app/models/tslr_claim.rb
erbridge: Ah. I see what you mean. I guess the current school one is invisible to users, so ok to do. I'll edit to put the rest of them back.

erbridge:
@robbiepaul Ready for another review.

PR #82, opened 2019-06-12
Admin can download submitted claims

pezholio:
This allows an admin to download a CSV version of all submitted claims with all the relevant information. It also generates a unique reference number for each claim which can be presented to the user later on in case of any queries.


pezholio:
[Admin can download submitted claims](https://trello.com/c/hvestFqR/186-admin-can-download-submitted-claims)

app/models/tslr_claim.rb
```diff
@@ -93,6 +93,8 @@ class TslrClaim < ApplicationRecord
   delegate :name, to: :claim_school, prefix: true, allow_nil: true
   delegate :name, to: :current_school, prefix: true, allow_nil: true
 
+  scope :complete, -> { where.not(submitted_at: nil) }
```
tekin: I think "submitted" rather than "complete" here.

config/locales/en.yml
```diff
@@ -53,3 +53,23 @@ en:
...
+      address_line_4: "Address 4"
+      postcode: "Postcode"
+      date_of_birth: "Date of Birth"
+      teacher_reference_number: "Teacher Reference"
+      national_insurance_number: "NI Number"
+      email_address: "Email"
+      mostly_teaching_eligible_subjects: "Mostly Teaching Eligible Subjects?"
+      bank_sort_code: "Sort Code"
+      bank_account_number: "Account Number"
+      student_loan_repayment_amount: "Repayment Amount"
```
tekin: missing end-of-line here.

app/presenters/tslr_claims_presenter.rb
```diff
@@ -0,0 +1,18 @@
+class TslrClaimsPresenter < BasePresenter
+  attr_accessor :decorated_collection
+
+  def initialize(claims)
+    self.decorated_collection = claims.map { |v| TslrClaimPresenter.new(v) }
```
tekin: mapping here means that _all_ the records will be loaded into memory, and then all of them will get a presenter, which could leave us open to a DOS attack if someone spams us with lots of submissions - memory would be eaten up without the garbage collector able to clean anything up. Instead we should assign the collection and then we can use `find_each` when looping through at the point of generation to load things in batches and ensure the garbage collector can do its work.

app/presenters/tslr_claims_presenter.rb
```diff
@@ -0,0 +1,18 @@
...
+    super(claims)
+  end
+
+  def csv_headers
+    TslrClaim.csv_fields.map { |h| translation_for_field(h) }
+  end
+
+  private
+
+  def translation_for_field(header)
```
tekin: How about `header_string_for_field` or something so that it says what it does, not how it does it.

app/presenters/tslr_claims_presenter.rb
```diff
@@ -0,0 +1,29 @@
...
+  def initialize(claims)
+    self.decorated_collection = claims.map { |v| TslrClaimPresenter.new(v) }
+    super(claims)
+  end
+
+  def csv_headers
+    TslrClaim.csv_fields.map { |h| translation_for_field(h) }
+  end
+
+  def csv_string
```
tekin: Again, potential DOS if there are lots of submissions - memory usage will grow linearly with the number of submissions as the entire CSV data will be held in memory as a string. Can this class write to and return a `Tempfile` instead of a string so we keep memory usage to a fixed size?

app/models/tslr_claim.rb
```diff
@@ -205,4 +234,17 @@ def bank_sort_code_must_be_six_digits
...
+  def generate_reference!
+    reference = loop {
+      ref = nanoid
+      break ref unless self.class.exists?(reference: ref)
+    }
+
+    update_attribute(:reference, reference)
+  end
+
+  def nanoid
```
tekin: This method name could be more helpful - it's named after the underlying implementation, not for what it actually does. As discussed, perhaps we can abstract this to a `Reference.generate` class so that it's encapsulated somewhere and we can still stub it in the specs where necessary.

app/presenters/tslr_claim_csv_row.rb
```diff
@@ -0,0 +1,29 @@
+require "delegate"
+
+class TslrClaimCSVRow < SimpleDelegator
```
erbridge: Can we be consistent with `CSV` vs `Csv` in class names, please?

pezholio: Ah, good call. I made the decision to go with `Csv`

app/presenters/tslr_claim_csv_row.rb
```diff
@@ -0,0 +1,45 @@
+require "delegate"
+
+class TslrClaimCsvRow < SimpleDelegator
+  def csv_row
```
tekin: rename this method to `data`? calling `tslr_claim_csv_row.data` feels more natural

spec/features/admin_csv_download_spec.rb
```diff
@@ -0,0 +1,43 @@
+require "rails_helper"
+
+RSpec.feature "Download CSV of claims" do
+  context "User is logged in" do
+    before do
+      visit admin_path
+      click_on "Sign in"
+    end
+
+    context "Completed claims exist" do
```
tekin: "submitted" not "completed". same for the variables below.

spec/features/admin_csv_download_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      let!(:complete_claims) { create_list(:tslr_claim, 5, :eligible_and_submittable, submitted_at: DateTime.now) }
+      let!(:incomplete_claims) { create_list(:tslr_claim, 2, :eligible_and_submittable) }
+
+      context "User downloads CSV" do
```
tekin: Reckon this context and the scenarios below could be merged into a single `scenario "user downloads claims CSV"`. Nested contexts add cognitive overhead and we should aim to avoid them where we can.

spec/features/admin_csv_download_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+        scenario "CSV has the correct number of rows" do
+          expect(csv.count).to eq(complete_claims.count + 1)
+        end
+      end
+    end
+  end
+
+  context "User is not logged in" do
+    context "User tries to download claims" do
```
tekin: here too: we can avoid the nested contexts and have a single scenario: "cannot download claims CSV".

spec/models/tslr_claim_spec.rb
```diff
@@ -415,4 +415,13 @@
       end
     end
   end
+
+  describe "submitted" do
+    let!(:complete_claims) { create_list(:tslr_claim, 5, :eligible_and_submittable, submitted_at: DateTime.now) }
```
tekin: s/complete/submitted

robbiepaul:
I don’t think it’s on the ACs, but I can imagine the school URNs being an important piece of of information when checking claims. It’s just an assumption though

app/views/admin/page/index.html.erb
```diff
@@ -3,5 +3,7 @@
     <h1 class="govuk-heading-xl">
       Admin
     </h1>
+
+    <%= link_to 'Download claims', admin_claims_path(format: :csv) %>
```
tekin:
```suggestion
    <%= link_to 'Download claims', admin_claims_path(format: :csv), class: "govuk-button" %>
```

spec/presenters/tslr_claim_csv_row_spec.rb
```diff
@@ -0,0 +1,54 @@
+require "rails_helper"
+require "csv"
```
tekin: This require needs to go in the file for the class itself, otherwise the `CSV` library isn't loaded outside of the test environment.

app/controllers/admin/claims_controller.rb
```diff
@@ -0,0 +1,10 @@
+class Admin::ClaimsController < Admin::BaseAdminController
+  def index
+    claims = TslrClaim.includes(:claim_school, :current_school).submitted
```
tekin:
```suggestion
    claims = TslrClaim.includes(:claim_school, :current_school).submitted.order(:submitted_at)
```

spec/presenters/tslr_claims_csv_spec.rb
```diff
@@ -0,0 +1,27 @@
+require "rails_helper"
+require "csv"
```
tekin: As above, the dependency on `CSV` exists on the `TslrClaimsCsv` class so the require needs to go in that class' file.

app/views/admin/page/index.html.erb
pezholio: I did consider this, but `govuk-button` is rarely used for links (apart from on start pages), so I decided against it. I guess it's not a public facing page, so is probably OK here

tekin:
> I don’t think it’s on the ACs, but I can imagine the school URNs being an important piece of of information when checking claims. It’s just an assumption though

@robbiepaul possibly, but lets worry about adding that down the road if it turns out we definitely need it

spec/models/tslr_claim_spec.rb
```diff
@@ -415,4 +415,13 @@
       end
     end
   end
+
+  describe "submitted" do
+    let!(:submitted_claims) { create_list(:tslr_claim, 5, :eligible_and_submittable, submitted_at: DateTime.now) }
```
erbridge: This can use the `:submitted` trait instead, can't it?

pezholio: Ah, right you are

PR #84, opened 2019-06-12
Add a reference number for a TSLR Claim

pezholio:
This generates a human-friendly, difficult to guess reference number for a claim, so a user can easily refer to their claim if there are any queries.

### DIFF
--- a/spec/models/tslr_claim_spec.rb
+++ b/spec/models/tslr_claim_spec.rb
@@ -367,6 +367,24 @@ RSpec.describe TslrClaim, type: :model do
       it "sets submitted_at to now" do
         expect(tslr_claim.submitted_at).to eq Time.zone.now
       end
+
+      it "generates a reference" do
+        expect(tslr_claim.reference).to_not eq nil
+      end
+
+      context "when a claim with the same reference already exists" do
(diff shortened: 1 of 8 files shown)


pezholio:
[Add a reference number for a TSLR Claim](https://trello.com/c/q3ZX91Yr/230-add-a-reference-number-for-a-tslr-claim)

spec/models/tslr_claim_spec.rb
```diff
@@ -365,7 +365,26 @@
       let(:tslr_claim) { create(:tslr_claim, :eligible_and_submittable) }
 
       it "sets submitted_at to now" do
-        expect(tslr_claim.submitted_at).to eq Time.zone.now
+        expect(tslr_claim.reload.submitted_at).to eq Time.zone.now
+      end
+
+      it "generates a reference" do
+        expect(tslr_claim.reload.reference).to_not eq nil
```
erbridge: I don't think these should need reloading because they're already saved?

pezholio: Ah, you're right. In a different permutation I'd run `update_columns`, which didn't update the attributes in place, but I've since binned that off

PR #86, opened 2019-06-12
Notify the claimant of their application

robbiepaul:
So that a user has confidence that they have finished a thing, and so they have a record of having completed their application, we will send the user an email confirming their claim.

We want to include a reference number so that it is easy to find their claim if they need support.

- [delayed_job](https://github.com/collectiveidea/delayed_job) to process background jobs
- [GOV.UK Notify](https://www.notifications.service.gov.uk/) for sending ActionMailer emails

The content used is just temporary and is from the confirmation page.

## Preview of the email

Screenshot 2019-06-13 at 15 13 44

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -72,6 +72,11 @@ GEM
     crack (0.4.3)
       safe_yaml (~> 1.0.0)
     crass (1.0.4)
+    delayed_job (4.1.5)
+      activesupport (>= 3.0, < 5.3)
+    delayed_job_active_record (4.1.3)
+      activerecord (>= 3.0, < 5.3)
+      delayed_job (>= 3.0, < 5)
     diff-lcs (1.3)
     dotenv (2.7.2)
     dotenv-rails (2.7.2)
@@ -307,6 +312,7 @@ DEPENDENCIES
   bullet
   byebug
   capybara
+  delayed_job_active_record
   dotenv-rails
   dxw_govuk_frontend_rails
   factory_bot_rails
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -102,6 +102,7 @@ GEM
       activesupport (>= 4.2)
       aes_key_wrap
       bindata
+    jwt (2.2.1)
     launchy (2.4.3)
       addressable (~> 2.3)
     listen (3.1.5)
@@ -113,6 +114,9 @@ GEM
       nokogiri (>= 1.5.9)
     mail (2.7.1)
       mini_mime (>= 0.1.1)
+    mail-notify (0.1.0)
+      actionmailer (~> 5.0)
+      notifications-ruby-client (~> 3.1)
     marcel (0.3.3)
       mimemagic (~> 0.3.2)
     method_source (0.9.2)
@@ -126,6 +130,8 @@ GEM
     nio4r (2.3.1)
     nokogiri (1.10.3)
       mini_portile2 (~> 2.4.0)
+    notifications-ruby-client (3.1.0)
+      jwt (>= 1.5, < 3)
     omniauth (1.9.0)
       hashie (>= 3.4.6, < 3.7.0)
       rack (>= 1.6.2, < 3)
@@ -308,6 +314,7 @@ DEPENDENCIES
   jbuilder (~> 2.9)
   launchy
   listen (>= 3.0.5, < 3.2)
+  mail-notify
   nanoid
   omniauth
   omniauth_openid_connect
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -42,7 +42,7 @@ class ClaimsController < ApplicationController
 
   def update_current_claim!
     if params[:slug] == "check-your-answers"
-      current_claim.submit!
+      mail_claim_submitted if current_claim.submit!
     else
       current_claim.attributes = claim_params
       current_claim.save(context: params[:slug].to_sym)
@@ -104,6 +104,10 @@ class ClaimsController < ApplicationController
       :bank_sort_code,
       :bank_account_number,
     )
+  end
+
+  def mail_claim_submitted
+    ClaimMailer.submitted(current_claim).deliver_now
   end
 
   def claim_page_template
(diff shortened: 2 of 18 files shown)


robbiepaul:
[Notify the claimant of their application](https://trello.com/c/0VdIxBKg/185-notify-the-claimant-of-their-application)

robbiepaul:
I need to add some instructions to the readme. If you need an invite to Notify let me know

Gemfile.lock
```diff
@@ -113,6 +114,9 @@ GEM
       nokogiri (>= 1.5.9)
     mail (2.7.1)
       mini_mime (>= 0.1.1)
+    mail-notify (0.1.0)
```
erbridge: Looks like this is an old version of `mail-notify`. I wonder why your Bundler is pulling in old dependencies.

app/controllers/claims_controller.rb
```diff
@@ -106,6 +106,10 @@ def claim_params
     )
   end
 
+  def claim_submitted
```
erbridge: Maybe this could be named after what it does rather than when it happens? e.g. `mail_claim_submitted` or something.

spec/requests/claims_spec.rb
```diff
@@ -178,6 +181,10 @@
             expect(in_progress_claim.submitted_at).to be_present
           end
 
+          it "sends an email" do
+            expect(claim_mailer_spy).to have_received(:deliver_later)
+          end
+
```
erbridge: Think we could also test the negative case: we don't send an email if submission fails.

Gemfile.lock
robbiepaul: Hmm, strange. I'll have a look

robbiepaul: It looks like this is the most recent version: https://rubygems.org/gems/mail-notify

erbridge: Oh. Hah. Sorry. Read it as `0.0.1`.

robbiepaul:
@erbridge I've made those changes, ready for another look

config/application.rb
```diff
@@ -32,5 +32,8 @@ class Application < Rails::Application
     config.generators do |g|
       g.orm :active_record, primary_key_type: :uuid
     end
+
+    config.active_job.queue_adapter = :que
+    config.action_mailer.deliver_later_queue_name = ""
```
mec: I read that this should be `"default"` is that implied somehow?

robbiepaul: Yeah, you're right. `:default` works as well, I've changed it so it's more obvious.

robbiepaul:
@mec I must've missed that first time round reading the docs, it isn't as obvious on the pre 1.0 branch. 1.0 as still in beta: https://github.com/chanks/que/tree/0.x

It says it's highly recommended that we switch. We'll have to make sure we're all on the same version of Postgres or we'll get issues like we saw before. Maybe Que isn't the right solution?

mec:
@robbiepaul we should all talk, like we did last time. It seems like we will always use PG so it's a given, but the version problems might be an issue, I'd want to know more before we made any jump?

mec:
It does say 'You can use schema_format as :ruby if you want but keep in mind this is highly advised against, as some parts of Que will not work.' – worth understanding which 'parts' first?

db/schema.rb
```diff
@@ -28,6 +28,17 @@
     t.index ["code"], name: "index_local_authority_districts_on_code", unique: true
   end
 
+  create_table "que_jobs", primary_key: ["queue", "priority", "run_at", "job_id"], comment: "3", force: :cascade do |t|
```
mec: None of this schema looks too scary?

robbiepaul: Yeah I agree, I'll switch it to `sql` locally just to see what we'd miss by keeping as `ruby`

mec: If we stick with `ruby` be good to note this in the commit message for future us?

robbiepaul: This is what the change that is produced by Que looks like when switching to `sql`:

```
CREATE TABLE public.que_jobs (
    priority smallint DEFAULT 100 NOT NULL,
    run_at timestamp with time zone DEFAULT now() NOT NULL,
    job_id bigint NOT NULL,
    job_class text NOT NULL,
    args json DEFAULT '[]'::json NOT NULL,
    error_count integer DEFAULT 0 NOT NULL,
    last_error text,
    queue text DEFAULT ''::text NOT NULL
);

COMMENT ON TABLE public.que_jobs IS '3';

CREATE SEQUENCE public.que_jobs_job_id_seq
    START WITH 1
    INCREMENT BY 1
    NO MINVALUE
    NO MAXVALUE
    CACHE 1;

ALTER SEQUENCE public.que_jobs_job_id_seq OWNED BY public.que_jobs.job_id;
```
The only difference I see is the `CREATE SEQUENCE public.que_jobs_job_id_seq...` bit

robbiepaul: Yeah, looks like the `job_id` is always 0 with `:ruby` https://github.com/chanks/que/issues/148

robbiepaul: Not sure what to do about this, I'll start a thread on Slack

robbiepaul: Opted for [`delayed_job`](https://github.com/collectiveidea/delayed_job) instead

spec/requests/claims_spec.rb
```diff
@@ -165,6 +165,12 @@
       end
 
       context "when updating from check-your-answers" do
+        let(:claim_mailer_spy) { spy(:mailer) }
+        before do
+          allow(ClaimMailer).to receive(:submitted).with(in_progress_claim)
```
tekin: Rather than stubbing the mailer I think it's better to check the contents of `ActionMailer::Base.deliveries` as part of the expectation: tests the expected outcome, not the implementation.

robbiepaul:
@tekin I've changed the request spec to check the contents of `ActionMailer::Base.deliveries`

I've set up a new API key for staging that will only send to 'Team and whitelist' and I've put the environment variables in to Heroku

PR #89, opened 2019-06-13
Set up basic auth on Azure

erbridge:


erbridge:
[Set up Basic Auth on CIP](https://trello.com/c/G8i7ISW0/244-set-up-basic-auth-on-cip)

PR #92, opened 2019-06-13
Adding Architecture Decision Records

mec:
- 8 Use GOVUK Verify for claimant identity assurance
- 9 Capture a Teacher Reference Number to validate a claimants qualifications
- 10 DfE can download claim data during private beta phase


mec:
@tekin I didn't attempt 'ADR for using NAV as the payment solution', not sure it's a decision yet or that I have enought context.

PR #95, opened 2019-06-17
The asset pre-compile includes autocomplete.css

mec:
When using the asset pipeline, non-js assets from the node_modules directory
must be included explicitly, otherwise they are not pre-compiled in production.

### DIFF
--- a/config/initializers/assets.rb
+++ b/config/initializers/assets.rb
@@ -12,3 +12,4 @@ Rails.application.config.assets.paths << Rails.root.join("node_modules")
 # application.js, application.css, and all non-JS/CSS in the app/assets
 # folder are already added.
 # Rails.application.config.assets.precompile += %w( admin.js admin.css )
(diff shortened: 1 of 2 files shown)


config/initializers/assets.rb
```diff
@@ -12,3 +12,4 @@
 # application.js, application.css, and all non-JS/CSS in the app/assets
 # folder are already added.
 # Rails.application.config.assets.precompile += %w( admin.js admin.css )
+Rails.application.config.assets.precompile += %w(accessible-autocomplete/src/autocomplete.css)
```
robbiepaul: Should this be `accessible-autocomplete/dist/accessible-autocomplete.min.css`?

mec: Sorry, this is a draft, just wanted to try it in a review app on heroku!

mec: Yep, but as the assets are compiled anyway the src makes more sense to me and to be honest, I cannot get the asset pipeline to use the minified version!

PR #99, opened 2019-06-17
238 bug users who have javascript disabled cant click anything

titlescreen:
Resolves the issue of not being able to click anything if JS is disabled. Also fixes related bug of modal overlay loading in and out on page refresh.


robbiepaul:
Prettier is failing on the build, I'm gonna pick this up with Rich in the morning

PR #100, opened 2019-06-17
Only reset inferred current school when persisted

tekin:
The build broke because the before save callback was changed to a before
validation callback. Skipping and setting callbacks in the factories
makes the code brittle. Instead of doing that, this reworks the callback
so that it only resets the current school when the employment status
changes on a persisted record, as it's only on persisted records
(created as part of the claim journey) that we expect to automatically
set/reset the current school based on the answer to the employment
question.

### DIFF
--- a/app/models/tslr_claim.rb
+++ b/app/models/tslr_claim.rb
@@ -84,7 +84,7 @@ class TslrClaim < ApplicationRecord
   validate  :bank_account_number_must_be_eight_digits
   validate  :bank_sort_code_must_be_six_digits
 
-  before_validation :update_current_school, if: :employment_status_changed?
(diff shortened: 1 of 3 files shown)


app/models/tslr_claim.rb
```diff
@@ -84,7 +84,7 @@ class TslrClaim < ApplicationRecord
   validate  :bank_account_number_must_be_eight_digits
   validate  :bank_sort_code_must_be_six_digits
 
-  before_validation :update_current_school, if: :employment_status_changed?
+  before_validation :reset_inferred_current_school, if: ->(record) { record.persisted? && record.employment_status_changed? }
```
pezholio: Cool, I didn't know about `persisted?`. That's a nice solution 👍

tekin: yep, it's the opposite of `new_record?`

PR #104, opened 2019-06-18
Address CVE-2015-9284

tekin:
There is a known vulnerability in OmniAuth that could potentially be used in combination with a vulnerable OAuth provider to gain a signed-in user's session. Whilst DfE Sign-in uses OpenID, not OAuth, the substance of this patch (adding CSRF protection to the "Sign in" link on the
service) is still worth doing.

For more background see:

https://github.com/DFE-Digital/dfe-teachers-payment-service/network/alert/Gemfile.lock/omniauth/open
and https://github.com/omniauth/omniauth/pull/809
which leads to https://github.com/cookpad/omniauth-rails_csrf_protection

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,6 +33,7 @@ gem "rollbar"
 # Use OmniAuth with OpenIDConnect
 gem "omniauth"
 gem "omniauth_openid_connect"
+gem "omniauth-rails_csrf_protection"
 
 # ActionMailer support for GOV.UK Notify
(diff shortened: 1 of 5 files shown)


tekin:
[Ensure the OmniAuth Ruby gem is patched and secure](https://trello.com/c/BkojHzaV/221-ensure-the-omniauth-ruby-gem-is-patched-and-secure)

PR #105, opened 2019-06-18
Fix student loan repayment errors

pezholio:
If a user enters an amount of £99999 for their student loan repayment, the database raises an error, as there is a hard precision of 7 and a scale of 2 set in the database (which translates to a 5 figure number with two decimal places). I've altered this so the validations take place in the model, but also limited it on the frontend too, by changing the field to an `<input type="number">` with a max value of `99999`. Not sure if this is worth doing, but is an extra belt and braces step.

### DIFF
--- a/app/models/tslr_claim.rb
+++ b/app/models/tslr_claim.rb
@@ -72,6 +72,7 @@ class TslrClaim < ApplicationRecord
   validates :student_loan_repayment_amount, on: [:"student-loan-amount", :submit], presence: {message: "Enter your student loan repayment amount"}
   validates_numericality_of :student_loan_repayment_amount, message: "Enter a valid monetary amount",
                                                             allow_nil: true,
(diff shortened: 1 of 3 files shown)


pezholio:
[Fix Student Loan Repayment error](https://trello.com/c/DcYyrjCt/297-fix-student-loan-repayment-error)

tekin:
> I wonder if it makes sense to make other numeric value fields such as those for bank details and telephone number fields too to improve the mobile experience?

I wasn't suggesting that be done as part of this PR by the way!

mec:
I vote good. https://www.w3.org/WAI/tutorials/forms/validation/

tekin:
Actually @pezholio, do you mind updating the validation to verify it's also a positive number? That'll dust off one of the things that Giles highlighted during his security review.

PR #108, opened 2019-06-20
Set up secret management in Azure

pezholio:
This sets up secrets mangement, so we can confidently store, set and share production secrets in one place. 

We've set up an Azure Key Vault in a different resource group as part of the deployment process, we then have a template parameters file with placeholders that, once the Key Vault has been set up, we replace with real parameters via `sed`. Once the deploy is kicked off, Azure then knows where to fetch the secrets from.

The only issue is on a fresh deploy, the initial deploy will fail because the secrets aren't in place. We could address this by prompting for any missing secrets, but I think this may be overkill as we're only doing this a few times at the absolute most.


pezholio:
[Set up secret management in Azure](https://trello.com/c/0nLxwePz/247-set-up-secret-management-in-azure)

bin/azure-deploy
```diff
@@ -12,6 +13,8 @@ case $ENVIRONMENT_NAME in
     # Use `az account show --out json` to find the subscription ID.
     SUBSCRIPTION_ID="8655985a-2f87-44d7-a541-0be9a8c2779d"
```
petenorth: I wouldn't normally  have the subscription id in the repo (basically from a security perspective).
Similarly with the resource group name.
Also the name of the file development.json.txt, curious as to why the extension is *.txt

pezholio: Cool, fair enough. We'll ding those two things. On the naming, I think our reasoning was because it's not the actual parameter file, so saves confusion should someone use it in error.

tekin: perhaps `development.example.json` instead to make it clearer that it's not the real thing?

robbiepaul: I've renamed it to `development.template.json` since the file is used but with substitutions.

robbiepaul: I've removed subscription ID and resource group name variables from the script. We'll also need to rotate the values or purge the git history if they're sensitive.

pezholio: Yeah, I think the only way is to purge the history. I'd say those values are pretty much set in stone. That said, I don't think they're highly sensitive, but makes sense to get rid

bin/azure-deploy
```diff
@@ -2,16 +2,29 @@
...
+	echo "Enter your subscription ID:"
+	read SUBSCRIPTION_ID
+	[[ "${SUBSCRIPTION_ID:?}" ]]
+fi
+
+if [[ -z "$RESOURCE_GROUP_NAME" ]]; then
+	echo "Enter your resource group name:"
+	read RESOURCE_GROUP_NAME
+	[[ "${RESOURCE_GROUP_NAME:?}" ]]
+fi
```
erbridge: What motivated this change? I imagined we'd hardcode the values for the different environments into the `case` statement below, rather than requiring extra setup from users.

erbridge: Ok, I now see what motivated it. I don't really see how this improves security, as if you have access to the subscription or the resource group, you can easily query them with the CLI. We're not gaining anything useful, and we're adding (unnecessary IMO) complication to new developer setup.

erbridge: We're also hardcoding the KeyVault resource group.

erbridge: I think we need a better solution that environment variables, as these will be different for each environment, and requiring the user to manually update them when switching environments is error prone. What are we actually gaining from taking them out of the script?

erbridge: For context: https://stackoverflow.com/a/45663968

pezholio: Yep, I think we can safely leave these in. It will make deployment much easier and I don't think there's any real risk in having these values open.

PR #109, opened 2019-06-20
Re organise feature specs

tekin:
The main claim feature spec file was getting rather large and difficult to work with. This PR splits out the main groups of feature specs into their own files, making it easier to find the specs relevant to specific features or behaviour.


tekin:
Just realised the commit that supposedly removed redundant expectations (1d08244) was bogus - the expectations were there to prove the button text changed. I've removed that commit and force-pushed.

PR #110, opened 2019-06-20
Amend consent page to reflect writing in MPS

titlescreen:
To provide a more consistent experience we have replicated the writing
style to reflect the maths and phys consent page. This is still subject
to change when our content designer does a formal review but its a step
forwards.

### DIFF
--- a/app/views/claims/new.html.erb
+++ b/app/views/claims/new.html.erb
@@ -9,7 +9,7 @@
     </p>
     <ul class="govuk-list govuk-list--bullet">
       <li>
-        that you were empolyed there
+        that you were employed there
       </li>
       <li>
         that you taught 11 to 16 year olds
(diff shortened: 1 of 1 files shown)


app/views/claims/new.html.erb
```diff
@@ -5,35 +5,25 @@
...
 
     <p class="govuk-body">
-      To claim your student loan payments back we will contact the school you tell
-      us about in this claim to confirm:
+      By continuing, you consent to the Department for Education contacting your school to confirm:
     </p>
     <ul class="govuk-list govuk-list--bullet">
       <li>
-        that you were employed to teach there
+        that you were empolyed there
```
pezholio: Typo on `employed`

titlescreen: Ah cool thanks for the heads up.

PR #113, opened 2019-06-24
Change the TRN question bring it inline with MPS

titlescreen:
- Amend help text to provide users a way to seek assistance if they
  don't know their TRN.

- Change width of input to match number of expected characters

### DIFF
--- a/app/views/claims/teacher_reference_number.html.erb
+++ b/app/views/claims/teacher_reference_number.html.erb
@@ -8,11 +8,11 @@
           <%= form.label :teacher_reference_number, t("tslr.questions.teacher_reference_number"), class: "govuk-label govuk-label--xl" %>
         </h1>
 
-        <span class="govuk-hint">This is on the certificate you got when you qualified as a teacher, or your school can tell you.</span>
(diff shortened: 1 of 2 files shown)


app/views/claims/teacher_reference_number.html.erb
```diff
@@ -8,11 +8,11 @@
           <%= form.label :teacher_reference_number, t("tslr.questions.teacher_reference_number"), class: "govuk-label govuk-label--xl" %>
         </h1>
 
-        <span class="govuk-hint">This is on the certificate you got when you qualified as a teacher, or your school can tell you.</span>
+        <span class="govuk-hint">You can get this from your school, the certificate you got when you qualified as a teacher or from the <a href="https://www.gov.uk/guidance/individual-teacher-records-information-for-teachers#contact">teacher qualifications helpdesk</a>.</span>
```
pezholio: Should this be a translation?

titlescreen: Perhaps we should so we can edit questions and hints in the same location.

PR #114, opened 2019-06-24
Amend NINO page to match MPS

titlescreen:
- Change the question name based off maths and physics journey for
  clarity.

- Change the width of the input to match the number of expected
  characters

### DIFF
--- a/app/views/claims/national_insurance_number.html.erb
+++ b/app/views/claims/national_insurance_number.html.erb
@@ -12,7 +12,7 @@
 
         <%= errors_tag current_claim, :national_insurance_number %>
 
-        <%= form.text_field :national_insurance_number, spellcheck: "false", class: css_classes_for_input(current_claim, :national_insurance_number) %>
+        <%= form.text_field :national_insurance_number, spellcheck: "false", class: css_classes_for_input(current_claim, :national_insurance_number, 'govuk-input--width-10') %>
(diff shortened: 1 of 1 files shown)


app/views/claims/national_insurance_number.html.erb
```diff
@@ -12,7 +12,7 @@
 
         <%= errors_tag current_claim, :national_insurance_number %>
 
-        <%= form.text_field :national_insurance_number, spellcheck: "false", class: css_classes_for_input(current_claim, :national_insurance_number) %>
+        <%= form.text_field :national_insurance_number, spellcheck: "false", class: css_classes_for_input(current_claim, :national_insurance_number, 'govuk-input--width-10') %>
```
pezholio: We could probably set a `maxlength` for this field too (so 9 characters)

titlescreen: Hey I was just looking into this and if I set `maxlength => 9` then we stop users entering their national insurance number in the hinted format "QQ 12 34 56 C". I could set it to accept a max length of 13 but then if they hit space twice by accident it will throw an error. 

I think we are better off accepting `minlength =>9`.

pezholio: Yeah, that sounds sensible. I didn't think to account for spaces etc

PR #118, opened 2019-06-24
Set up CI with Azure Pipelines

robbiepaul:
## Potential issues

- Azure Pipelines doesn't have any caching so containers are built from scratch each time. The whole CI process takes ~6 minutes, which I think is fine for now.
- Removing `webdrivers` gem may impact devs slightly as they will need to ensure they have `chromedriver` installed when running feature specs.
- We will need to ensure that when we build the production container that we pass in `--target web` so that none of the test container actions are run.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,6 +40,8 @@ RUN gem install bundler
 
 RUN if [ ${RAILS_ENV} = "production" ]; then \
   bundle install --frozen --retry 3 --without development test; \
+  elif [ ${RAILS_ENV} = "test" ]; then \
+  bundle install --frozen --retry 3; \
   else \
   bundle install --frozen --retry 3 --without test; \
   fi
@@ -49,7 +51,11 @@ RUN if [ ${RAILS_ENV} = "production" ]; then \
 COPY package.json ${DEPS_HOME}/package.json
 COPY yarn.lock ${DEPS_HOME}/yarn.lock
 
-RUN yarn install --frozen-lockfile --production
+RUN if [ ${RAILS_ENV} = "production" ]; then \
+  yarn install --frozen-lockfile --production; \
+  else \
+  yarn install --frozen-lockfile; \
+  fi
 # End
 
 # ------------------------------------------------------------------------------
@@ -97,3 +103,13 @@ EXPOSE 3000
 
 ENTRYPOINT [ "bin/docker-entrypoint" ]
 CMD [ "rails", "server" ]
+
+# ------------------------------------------------------------------------------
+# test
+# ------------------------------------------------------------------------------
+
+FROM web AS test
+
+RUN echo "Building with RAILS_ENV=${RAILS_ENV}, NODE_ENV=${NODE_ENV}"
+RUN apk add chromium chromium-chromedriver
+COPY spec ${APP_HOME}/spec
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,7 +41,7 @@ RUN gem install bundler
 RUN if [ ${RAILS_ENV} = "production" ]; then \
   bundle install --frozen --retry 3 --without development test; \
   elif [ ${RAILS_ENV} = "test" ]; then \
-  bundle install --frozen --retry 3; \
+  bundle install --frozen --retry 3 --without development; \
   else \
   bundle install --frozen --retry 3 --without test; \
   fi
--- a/Gemfile
+++ b/Gemfile
@@ -73,11 +73,11 @@ group :development do
   gem "spring"
   gem "spring-watcher-listen", "~> 2.0.0"
   gem "foreman"
+  gem "webdrivers"
 end
 
 group :test do
   gem "selenium-webdriver"
-  gem "webdrivers"
   gem "launchy"
 end
 
--- a/README.md
+++ b/README.md
@@ -68,6 +68,8 @@ NOTIFY_TEMPLATE_ID=d72e2ff9-b228-4f16-9099-fd9d411c0334
 bundle exec rake
 ```
 
+To run the feature specs you will need Chrome installed.
+
 ### Code linting rules
 
 Code linting is performed using:
--- a/README.md
+++ b/README.md
@@ -82,6 +82,17 @@ Code linting is performed using:
 [Bullet](https://github.com/flyerhzm/bullet) runs around each spec. If it detects an N+1 query it will raise an
 exception and the tests will fail.
 
+## Production
+
+### Building the Docker container
+
+Make sure you build the image with the target of `web` to ensure that only
+production assets are added.
+
+```
+docker build --target web .
+```
+
 ## Service architecture
 
 The service architecture is currently defined and [on confluence](https://dfedigital.atlassian.net/wiki/spaces/TP/pages/1049559041/Service+Architecture).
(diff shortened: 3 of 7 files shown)


.env.development
```diff
@@ -0,0 +1,4 @@
+DFE_SIGN_IN_ISSUER=
+DFE_SIGN_IN_REDIRECT_BASE_URL=
+DFE_SIGN_IN_IDENTIFIER=
+DFE_SIGN_IN_SECRET=
```
erbridge: Are these actually needed?

Dockerfile
```diff
@@ -41,15 +41,19 @@ RUN gem install bundler
 RUN if [ ${RAILS_ENV} = "production" ]; then \
   bundle install --frozen --retry 3 --without development test; \
   else \
-  bundle install --frozen --retry 3 --without test; \
+  bundle install --frozen --retry 3; \
```
erbridge: Let's only install `test` if `RAILS_ENV`is `test`?

azure-pipelines.yml
```diff
@@ -0,0 +1,20 @@
...
+trigger:
+- master
+
+pool:
+  vmImage: 'ubuntu-latest'
+
+variables:
+  POSTGRES_IMAGE: postgres
+  POSTGRESS_USER: postgres
+  POSTGRESS_PASSWORD: secret
```
erbridge: Typos here? Too many `S`es?

docker-compose.test.yml
```diff
@@ -0,0 +1,24 @@
...
+    build:
+      context: .
+      target: test
+      args:
+        RAILS_ENV: test
+    image: local/dfe-teachers-payment-service:test
+    depends_on:
+      - db
+    env_file:
+      - .env.development
```
erbridge: Let's include `.env.test` instead?

docker-compose.test.yml
```diff
@@ -0,0 +1,24 @@
...
+      DFE_TEACHERS_PAYMENT_SERVICE_DATABASE_USERNAME: postgres
+      DFE_TEACHERS_PAYMENT_SERVICE_DATABASE_HOST: db
+    tty: true
+    stdin_open: true
+    command: bundle exec spring server
+
+  db:
+    image: postgres
+    volumes:
+      - ./tmp/db:/var/lib/postgresql/data
```
erbridge: This means it's sharing a database with the `development` Docker compose setup. Was that intentional?

docker-compose.test.yml
erbridge: Actually, let's just remove the `env_file` import, since we don't need it.

docker-compose.test.yml
robbiepaul: Good catch, I'll change it. Azure automatically opened this PR when I went to enable pipelines so it's very much WIP, I'll fix it up

robbiepaul:
[Automatic builds before deploy (i.e. Continuous Integration)](https://trello.com/c/zTQvrHBU/249-automatic-builds-before-deploy-ie-continuous-integration)

Dockerfile
```diff
@@ -97,3 +103,13 @@ EXPOSE 3000
 
 ENTRYPOINT [ "bin/docker-entrypoint" ]
 CMD [ "rails", "server" ]
+
+# ------------------------------------------------------------------------------
+# test
+# ------------------------------------------------------------------------------
+
+FROM web AS test
```
pezholio: As mentioned in the PR description, can you add something to the Readme so we don't forget? Also, might be a good idea to add a script to `bin` that builds the docker container with the relevant flags

robbiepaul:
Ready for another look

app.json
```diff
@@ -5,6 +5,7 @@
       "buildpacks": [
         { "url": "heroku/nodejs" },
         { "url": "https://github.com/heroku/heroku-buildpack-google-chrome" },
+        { "url": "https://github.com/heroku/heroku-buildpack-chromedriver" },
```
pezholio: Do we need the app.json now? We probably don't need to build in Heroku as well as Azure.

Gemfile
```diff
@@ -77,7 +77,6 @@ end
 
 group :test do
   gem "selenium-webdriver"
-  gem "webdrivers"
```
pezholio: Could we not add "webdrivers" to the development group, so we can still install it on dev machines? Webdriver is always a pain to set up locally

app.json
robbiepaul: I figured I'd leave both there for the time being until we decide to switch then remove traces of the Heroku one separately

Gemfile
robbiepaul: Good idea, done 👍

README.md
```diff
@@ -68,6 +68,8 @@ NOTIFY_TEMPLATE_ID=d72e2ff9-b228-4f16-9099-fd9d411c0334
 bundle exec rake
 ```
 
+To run the feature specs you will need Chrome and Chromedriver installed.
+
```
erbridge: Is this still true if we have `webdrivers` in development?

Dockerfile
robbiepaul: I imagine we'll cover this as part of the CD story, I'll mention it in the readme though

README.md
robbiepaul: I've amended the text to just say Chrome

azure-pipelines.yml
```diff
@@ -0,0 +1,16 @@
...
+  - master
+
+pool:
+  vmImage: ubuntu-latest
+
+steps:
+  - script: docker-compose -f docker-compose.test.yml build
+    displayName: 'Building Docker container'
+  - script: docker-compose -f docker-compose.test.yml run --rm web bundle exec rake
+    displayName: 'Running rake default'
```
erbridge: Nitpick: double quotes or no quotes (my preference).

Gemfile
robbiepaul: This is working in CI now, but I need add `require "webdrivers"` for the webdrivers to work on dev. Not sure how best to do this? We could set a `CI` environment variable and require `webdrivers` unless it exists?

robbiepaul: Ideally we'd fix this upstream with webdrivers but I didn't want to spend too much time on it

pezholio: Could you do `require "webrivers" rescue nil` or similar?

robbiepaul: I tried `require "webdrivers" rescue LoadError` because LoadError isn't a StandardError but still getting: 

```
LoadError:
  cannot load such file -- webdrivers
```

erbridge: How about `require "webdrivers" if Rails.env.development?`?

robbiepaul: `RAILS_ENV` is set to `test` in `rails_helper`

erbridge: Oh right. Of course. Ignore me.

robbiepaul: I've added a commit that requires it if an environment variable doesn't exist. It feels a bit icky. The alternative is to require devs to run `brew cask install chromedriver`

pezholio: Could you not rescue from the LoadError?

robbiepaul: I could but not inline, it worked with the long form

```
begin 
  require "webdrivers"
rescue LoadError
  nil
end
```

pezholio: Sweet 👍

robbiepaul: Should I do that instead then @pezholio?

pezholio: I reckon that's the easiest thing to do, and avoids mucking around with env vars etc

robbiepaul: Cool, I've pushed up that change

PR #120, opened 2019-06-25
Ask applicants what subjects they have taught

pezholio:
This PR makes some changes to the flow so we ask applicants what subjects they've taught, followed by asking if they've taught any of the subjects for more than 50% of their time:

Successful path:

image

Unsuccessful path:

image

I've also added some JS so if a user chooses a subject, and then 'not applicable', the unapplicable checkboxes are unchecked, and vice versa:

image

### DIFF
--- a/app/assets/javascripts/components/subjects_taught.js
+++ b/app/assets/javascripts/components/subjects_taught.js
@@ -2,39 +2,33 @@
 
 document.addEventListener("DOMContentLoaded", function() {
   var fieldset = document.querySelector("#tslr_claim_eligible_subjects");
-  var el;
-  var eligibleSubjectsSelector;
-  var notTeachingSubjectsSelector;
 
   if (!fieldset) {
     return;
   }
 
-  fieldset.addEventListener(
-    "click",
-    function(e) {
-      el = e.target;
+  function toggleSubjectValues(event) {
+    var el = event.target;
+    var eligibleSubjectsName = "tslr_claim[eligible_subjects][]";
+    var notTeachingSubjectsName = "tslr_claim[mostly_teaching_eligible_subjects]";
 
-      if (!el) {
-        return;
-      }
-      if (el.checked === false) {
-        return;
-      }
+    if (!el) {
+      return;
+    }
+    if (el.checked === false) {
+      return;
+    }
 
-      eligibleSubjectsSelector = "tslr_claim[eligible_subjects][]";
-      notTeachingSubjectsSelector = "tslr_claim[mostly_teaching_eligible_subjects]";
+    if (el.name == eligibleSubjectsName) {
+      var radioButton = document.querySelector("input[type='radio'][name='" + notTeachingSubjectsName + "']");
+      radioButton.checked = false;
+    } else if (el.name == notTeachingSubjectsName) {
+      var checkboxes = document.querySelectorAll("input[name='" + eligibleSubjectsName + "']");
+      checkboxes.forEach(function(item, index) {
+        item.checked = false;
+      });
+    }
+  }
 
-      if (el.name == eligibleSubjectsSelector) {
-        var radioButton = document.querySelector("input[type='radio'][name='" + notTeachingSubjectsSelector + "']");
-        radioButton.checked = false;
-      } else if (el.name == notTeachingSubjectsSelector) {
-        var checkboxes = document.querySelectorAll("input[name='" + eligibleSubjectsSelector + "']");
-        checkboxes.forEach(function(item, index) {
-          item.checked = false;
-        });
-      }
-    },
-    false
-  );
+  fieldset.addEventListener("click", toggleSubjectValues, false);
 });
--- a/app/assets/javascripts/components/subjects_taught.js
+++ b/app/assets/javascripts/components/subjects_taught.js
@@ -1,0 +1,40 @@
+"use strict";
+
+document.addEventListener("DOMContentLoaded", function() {
+  var fieldset = document.querySelector("#tslr_claim_eligible_subjects");
+  var el;
+  var eligibleSubjectsSelector;
+  var notTeachingSubjectsSelector;
+
+  if (!fieldset) {
+    return;
+  }
+
+  fieldset.addEventListener(
+    "click",
+    function(e) {
+      el = e.target;
+
+      if (!el) {
+        return;
+      }
+      if (el.checked === false) {
+        return;
+      }
+
+      eligibleSubjectsSelector = "tslr_claim[eligible_subjects][]";
+      notTeachingSubjectsSelector = "tslr_claim[mostly_teaching_eligible_subjects]";
+
+      if (el.name == eligibleSubjectsSelector) {
+        var checkbox = document.querySelector("input[type='checkbox'][name='" + notTeachingSubjectsSelector + "']");
+        checkbox.checked = false;
+      } else if (el.name == notTeachingSubjectsSelector) {
+        var checkboxes = document.querySelectorAll("input[name='" + eligibleSubjectsSelector + "']");
+        checkboxes.forEach(function(item, index) {
+          item.checked = false;
+        });
+      }
+    },
+    false
+  );
+});
--- a/app/assets/javascripts/components/subjects_taught.js
+++ b/app/assets/javascripts/components/subjects_taught.js
@@ -26,8 +26,8 @@ document.addEventListener("DOMContentLoaded", function() {
       notTeachingSubjectsSelector = "tslr_claim[mostly_teaching_eligible_subjects]";
 
       if (el.name == eligibleSubjectsSelector) {
-        var checkbox = document.querySelector("input[type='checkbox'][name='" + notTeachingSubjectsSelector + "']");
-        checkbox.checked = false;
+        var radioButton = document.querySelector("input[type='radio'][name='" + notTeachingSubjectsSelector + "']");
+        radioButton.checked = false;
       } else if (el.name == notTeachingSubjectsSelector) {
         var checkboxes = document.querySelectorAll("input[name='" + eligibleSubjectsSelector + "']");
         checkboxes.forEach(function(item, index) {
--- a/app/helpers/claims_helper.rb
+++ b/app/helpers/claims_helper.rb
@@ -38,6 +38,12 @@ module ClaimsHelper
     ]
   end
 
+  def subject_list(subjects)
+    connector = " or "
+    subjects.map! { |subject| I18n.t("tslr.questions.eligible_subjects.#{subject}") }
+    subjects.to_sentence(last_word_connector: connector, two_words_connector: connector)
+  end
+
   private
 
   def academic_years(year_range)
--- a/app/helpers/claims_helper.rb
+++ b/app/helpers/claims_helper.rb
@@ -20,8 +20,8 @@ module ClaimsHelper
       [t("tslr.questions.qts_award_year"), academic_years(claim.qts_award_year), "qts-year"],
       [t("tslr.questions.claim_school"), claim.claim_school_name, "claim-school"],
       [t("tslr.questions.current_school"), claim.current_school_name, "current-school"],
-      [t("tslr.questions.subjects_taught"), subject_list(claim.eligible_subjects, "and"), "subjects-taught"],
-      [t("tslr.questions.mostly_teaching_eligible_subjects", subjects: subject_list(claim.eligible_subjects, "or")), (claim.mostly_teaching_eligible_subjects? ? "Yes" : "No"), "mostly-teaching-eligible-subjects"],
+      [t("tslr.questions.subjects_taught"), subject_list(claim.eligible_subjects), "subjects-taught"],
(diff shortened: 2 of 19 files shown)


app/views/claims/subjects_taught.html.erb
```diff
@@ -1,39 +1,24 @@
...
-              <%= form.radio_button(:mostly_teaching_eligible_subjects, false, class: "govuk-radios__input") %>
-              <%= form.label :mostly_teaching_eligible_subjects_false, "No", class: "govuk-label govuk-radios__label" %>
+          <div class="govuk-checkboxes">
+            <% TslrClaim.eligible_subjects.each do |subject, value| -%>
+              <div class="govuk-checkboxes__item">
+                <%= check_box_tag "tslr_claim[eligible_subjects][]", subject, current_claim.eligible_subjects.include?(subject), class: "govuk-checkboxes__input", id: "eligible_subjects_#{subject}"  %>
+                <%= label_tag "tslr_claim[eligible_subjects][]", t("tslr.questions.eligible_subjects.#{subject}"), class: "govuk-label govuk-checkboxes__label", for: "eligible_subjects_#{subject}"  %>
+              </div>
+            <% end %>
+            <div class="govuk-checkboxes__item">
```
robbiepaul: Although it's not part of the guidance on checkboxes we may want to include a text divider like on the pattern for radios: https://design-system.service.gov.uk/components/radios/#radio-items-with-a-text-divider

It feels a bit strange having 'Not applicable' next to the eligible subjects, but feel free to ignore. @titlescreen what do you think?

app/helpers/claims_helper.rb
```diff
@@ -38,6 +38,12 @@ def identity_answers(claim)
     ]
   end
 
+  def subject_list(subjects)
+    connector = " or "
+    subjects.map! { |subject| I18n.t("tslr.questions.eligible_subjects.#{subject}") }
```
robbiepaul: Is this mutating the original value passed in to the helper? I know this is a smell in other languages, but not sure how Ruby handles it

pezholio: Good call. That's a bad habit from a while back. Usually Rubocop picks that up

app/views/claims/subjects_taught.html.erb
pezholio: I've changed it to this now, which I think is nicer, but has caused a whole load of other fires:

image

app/assets/javascripts/components/subjects_taught.js
```diff
@@ -0,0 +1,40 @@
...
+  var eligibleSubjectsSelector;
+  var notTeachingSubjectsSelector;
+
+  if (!fieldset) {
+    return;
+  }
+
+  fieldset.addEventListener(
+    "click",
+    function(e) {
```
erbridge: Can we pull this out into a named function, so it's clear what it's doing?

app/assets/javascripts/components/subjects_taught.js
```diff
@@ -0,0 +1,40 @@
+"use strict";
+
+document.addEventListener("DOMContentLoaded", function() {
+  var fieldset = document.querySelector("#tslr_claim_eligible_subjects");
+  var el;
+  var eligibleSubjectsSelector;
+  var notTeachingSubjectsSelector;
```
erbridge: No need to declare these outside of the function scope since we're not reusing their values.

app/views/claims/subjects_taught.html.erb
```diff
@@ -1,39 +1,24 @@
...
+              <%= form.check_box(:mostly_teaching_eligible_subjects, {class: "govuk-checkboxes__input"}, false, '')  %>
+              <%= form.label :mostly_teaching_eligible_subjects_false, "Not applicable, I didn't teach any of these subjects", class: "govuk-label govuk-checkboxes__label" %>
```
erbridge: Is "Not applicable" actually an answer to this question? Isn't it "No, I didn't teach any of these subjects", or just "I didn't teach any of these subjects"?

db/migrate/20190624101725_add_eligible_subjects_to_tslr_claims.rb
```diff
@@ -0,0 +1,5 @@
+class AddEligibleSubjectsToTslrClaims < ActiveRecord::Migration[5.2]
+  def change
+    add_column :tslr_claims, :eligible_subjects, :integer, array: true, null: false, default: []
```
erbridge: :+1: on the default!

app/assets/javascripts/components/subjects_taught.js
```diff
@@ -0,0 +1,40 @@
...
+
+      if (!el) {
+        return;
+      }
+      if (el.checked === false) {
+        return;
+      }
+
+      eligibleSubjectsSelector = "tslr_claim[eligible_subjects][]";
+      notTeachingSubjectsSelector = "tslr_claim[mostly_teaching_eligible_subjects]";
```
erbridge: These are specifically `name`s not selectors, so let's name the variables to match that.

app/models/tslr_claim.rb
```diff
@@ -46,7 +57,9 @@ class TslrClaim < ApplicationRecord
 
   validates :employment_status,         on: [:"still-teaching", :submit], presence: {message: "Choose the option that describes your current employment status"}
 
-  validates :mostly_teaching_eligible_subjects, on: [:"subjects-taught", :submit], inclusion: {in: [true, false], message: "Select either Yes or No"}
+  validates :eligible_subjects,         on: [:"subjects-taught", :submit], presence: {message: "Choose a subject, or select \"not applicable\""}, unless: ->(c) { c.mostly_teaching_eligible_subjects == false }
```
erbridge: If changing the "no" answer to the question, the error message will also need updating.

app/views/claims/mostly_teaching_eligible_subjects.html.erb
```diff
@@ -0,0 +1,34 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render("shared/error_summary", instance: current_claim, errored_field_id_overrides: { mostly_teaching_eligible_subjects: "tslr_claim_mostly_teaching_eligible_subjects_true"}) if current_claim.errors.any? %>
+      <%= form_for current_claim, url: claim_path do |form| %>
+        <%= form.hidden_field :mostly_teaching_eligible_subjects %>
+        <%= form_group_tag current_claim do %>
+          <fieldset class="govuk-fieldset">
+            <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
+              <h1 class="govuk-fieldset__heading">
+                <%= t("tslr.questions.mostly_teaching_eligible_subjects", subjects: subject_list(current_claim.eligible_subjects)) %>
```
erbridge: `Did you teach %{subjects} for more than 50% of your teaching time between 6 April XXXX and 5 April XXXX?` with `or` (the default) reads like an exclusive or, not an inclusive or. Is that intentional?

app/views/claims/subjects_taught.html.erb
erbridge: Does this at all work without JS? This is mixing metaphors quite a lot. I think it makes more sense to be a checkbox rather than a radio button.

spec/models/tslr_claim_spec.rb
```diff
@@ -127,9 +127,21 @@
   end
 
   context "when saving in the “subjects-taught” validation context" do
+    it "validates a subject has been provided" do
+      expect(TslrClaim.new).not_to be_valid(:"subjects-taught")
+      expect(TslrClaim.new(eligible_subjects: [:computer_science, :physics])).to be_valid(:"subjects-taught")
+    end
+
+    it "does not validates a subject has been provided when mostly_teaching_eligible_subjects is false" do
```
erbridge: Typo: `validate` singular.

app/views/claims/mostly_teaching_eligible_subjects.html.erb
pezholio: I'd say so, as long as they teach one of those subjects for 50% of their time or more, then they'll be eligible

app/assets/javascripts/components/subjects_taught.js
```diff
@@ -0,0 +1,34 @@
+"use strict";
+
+var toggleSubjectValues = function(event) {
```
erbridge: This will pollute the global namespace. Better to either wrap this in an immediately invoked function or define it inside the `DOMContentLoaded` callback (my preference). Also, it's more readable to use the `function name(args)` form, rather than variable assignment, IMO.

app/views/claims/subjects_taught.html.erb
pezholio: Good catch, it didn't. I've fixed that now, which has also simplified the code a lot and helped with something I'd been butting up against all morning! On the checkbox vs radio button front, after doing both, it feels more natural clicking a radio button to reset the other values - @titlescreen what do you think?

pezholio:
[Ask applicants what subjects they have taught](https://trello.com/c/eW71LjbT/324-ask-applicants-what-subjects-they-have-taught)

pezholio:
[Ask applicants what subjects they have taught](https://trello.com/c/eW71LjbT/324-ask-applicants-what-subjects-they-have-taught)

app/views/claims/mostly_teaching_eligible_subjects.html.erb
erbridge: What if they, e.g. teach Chemistry 25% and Biology 25%? Are they eligible?

app/views/claims/subjects_taught.html.erb
erbridge: If you don't have JS does the radio button still make sense? Something to consider.

app/assets/javascripts/components/subjects_taught.js
```diff
@@ -0,0 +1,34 @@
...
+      var checkboxes = document.querySelectorAll("input[name='" + eligibleSubjectsName + "']");
+      checkboxes.forEach(function(item, index) {
+        item.checked = false;
+      });
+    }
+  }
+
+  if (!fieldset) {
+    return;
+  }
```
erbridge: Probably want to do this guard at the top, so we can be confident that the rest of the code only runs on the expected pages.

app/views/claims/mostly_teaching_eligible_subjects.html.erb
pezholio: Dunno. I've popped the question into Slack.

app/helpers/claims_helper.rb
```diff
@@ -20,7 +20,8 @@ def claim_answers(claim)
       [t("tslr.questions.qts_award_year"), academic_years(claim.qts_award_year), "qts-year"],
       [t("tslr.questions.claim_school"), claim.claim_school_name, "claim-school"],
       [t("tslr.questions.current_school"), claim.current_school_name, "current-school"],
-      [t("tslr.questions.mostly_teaching_eligible_subjects"), (claim.mostly_teaching_eligible_subjects? ? "Yes" : "No"), "subjects-taught"],
+      [t("tslr.questions.subjects_taught"), subject_list(claim.eligible_subjects), "subjects-taught"],
```
erbridge: This is a list of subjects, is it correct to list them with "or"? I think it was correct as it was with "and"?

config/locales/en.yml
```diff
@@ -44,7 +44,8 @@ en:
       claim_school: "Which school were you employed at between 6 April XXXX and 5 April XXXX?"
       employment_status: "Are you still employed to teach at a school in England?"
       current_school: "Which school are you currently employed at?"
-      mostly_teaching_eligible_subjects: "Did you teach eligible subjects for more than 50% of your teaching time between 6 April XXXX and 5 April XXXX?"
+      subjects_taught: "Did you teach any of the following subjects between 6 April XXXX and 5 April XXXX?"
+      mostly_teaching_eligible_subjects: "Did you teach %{subjects} for more than 50% of your teaching time between 6 April XXXX and 5 April XXXX?"
```
erbridge: This might be out of scope for this change, but I'm not sure this wording is clear.

> Did you teach Biology or Chemistry for more than 50% of your teaching time between 6 April XXXX and 5 April XXXX?

This could be read as:

> Did you teach (Biology or Chemistry) for more than 50% of your teaching time between 6 April XXXX and 5 April XXXX?

or:

> Did you teach Biology for more than 50% of your teaching time or Chemistry for more than 50% of your teaching time between 6 April XXXX and 5 April XXXX?

This might be clearer if we rewrote it as something like:

> Did teaching Biology and Chemistry cover 50% of your teaching time between 6 April XXXX and 5 April XXXX?

robbiepaul:
Because there's a small number of subjects that isn't likely to change we could use booleans instead of the complication of having an array enum?

pezholio:
> Because there's a small number of subjects that isn't likely to change we could use booleans instead of the complication of having an array enum?

Yeah, I'm looking at this now 👍

config/locales/en.yml
pezholio: Yeah, I like that 👍

PR #121, opened 2019-06-25
Make questions and check answers consistent

titlescreen:
Make sure the questions asked on page and the summary of answers use the
same questions. This should make it more consistent and easier to
understand as you will now click to change a question and see that same
question shown on the question page.

### DIFF
--- a/app/helpers/claims_helper.rb
+++ b/app/helpers/claims_helper.rb
@@ -28,12 +28,12 @@ module ClaimsHelper
 
   def identity_answers(claim)
     [
-      ["Full name", claim.full_name, "full-name"],
-      ["What is your address?", claim.address, "address"],
-      ["What is your date of birth?", l(claim.date_of_birth), "date-of-birth"],
+      ["What's your full name?", claim.full_name, "full-name"],
+      ["What's your address?", claim.address, "address"],
+      ["What's your date of birth?", l(claim.date_of_birth), "date-of-birth"],
       ["What's your teacher reference number?", claim.teacher_reference_number, "teacher-reference-number"],
-      ["National Insurance number", claim.national_insurance_number, "national-insurance-number"],
-      ["Email address", claim.email_address, "email-address"],
+      ["What's your National Insurance number?", claim.national_insurance_number, "national-insurance-number"],
+      ["What's your email address?", claim.email_address, "email-address"],
       ["Account number", claim.bank_account_number, "bank-details"],
       ["Sort code", claim.bank_sort_code, "bank-details"],
     ]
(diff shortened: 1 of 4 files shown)


app/helpers/claims_helper.rb
```diff
@@ -28,9 +28,9 @@ def claim_answers(claim)
   def identity_answers(claim)
     [
       ["Full name", claim.full_name, "full-name"],
-      ["Address", claim.address, "address"],
-      ["Date of birth", l(claim.date_of_birth), "date-of-birth"],
-      ["Teacher reference number", claim.teacher_reference_number, "teacher-reference-number"],
+      ["What is your address?", claim.address, "address"],
```
robbiepaul: We should probably be consistent with using statements or questions:

https://www.gov.uk/service-manual/design/form-structure#structure-your-form-to-help-users

titlescreen: That's a good shout Robbie I'll run through and phrase everything as a question. 👍

app/helpers/claims_helper.rb
```diff
@@ -27,12 +27,12 @@ def claim_answers(claim)
...
-      ["Date of birth", l(claim.date_of_birth), "date-of-birth"],
-      ["Teacher reference number", claim.teacher_reference_number, "teacher-reference-number"],
-      ["National Insurance number", claim.national_insurance_number, "national-insurance-number"],
-      ["Email address", claim.email_address, "email-address"],
+      ["What's your full name?", claim.full_name, "full-name"],
+      ["What's your address?", claim.address, "address"],
+      ["What's your date of birth?", l(claim.date_of_birth), "date-of-birth"],
+      ["What's your teacher reference number?", claim.teacher_reference_number, "teacher-reference-number"],
+      ["What's your National Insurance number?", claim.national_insurance_number, "national-insurance-number"],
+      ["What's your email address?", claim.email_address, "email-address"],
```
erbridge: Let's use the i18n keys here instead of hardcoding?

app/helpers/claims_helper.rb
```diff
@@ -27,12 +27,12 @@ def claim_answers(claim)
...
       ["Account number", claim.bank_account_number, "bank-details"],
       ["Sort code", claim.bank_sort_code, "bank-details"],
```
erbridge: Should we stick "Bank" on the front of these for context? i.e. "Bank account number" and "Bank sort code".

erbridge: Do these belong in a different section for "Payment details" or similar? They're not really about identity.

titlescreen: Yeah I get what you are saying, if we already have sections for Claim and Identify then I think this would be a good idea to have its own section too.

titlescreen: Yeah that would make it clearer, I will get on with that 😄

robbiepaul:
@titlescreen do you need a hand resolving these conflicts?

PR #125, opened 2019-06-27
Ask applicants what subjects they have taught

pezholio:
Reup of #120 

This PR makes some changes to the flow so we ask applicants what subjects they've taught, followed by asking if they've taught any of the subjects for more than 50% of their time


pezholio:
[Ask applicants what subjects they have taught](https://trello.com/c/eW71LjbT/324-ask-applicants-what-subjects-they-have-taught)

app/views/claims/subjects_taught.html.erb
```diff
@@ -1,39 +1,27 @@
...
             <div class="govuk-radios__item">
-              <%= form.radio_button(:mostly_teaching_eligible_subjects, true, class: "govuk-radios__input") %>
-              <%= form.label :mostly_teaching_eligible_subjects_true, "Yes", class: "govuk-label govuk-radios__label" %>
-            </div>
-            <div class="govuk-radios__item">
-              <%= form.radio_button(:mostly_teaching_eligible_subjects, false, class: "govuk-radios__input") %>
-              <%= form.label :mostly_teaching_eligible_subjects_false, "No", class: "govuk-label govuk-radios__label" %>
+              <%= form.hidden_field :mostly_teaching_eligible_subjects, value: '' %>
+              <%= form.radio_button(:mostly_teaching_eligible_subjects, false, class: "govuk-radios__input")  %>
+              <%= form.label :mostly_teaching_eligible_subjects_false, "Not applicable, I didn't teach any of these subjects", class: "govuk-label govuk-radios__label" %>
```
erbridge: This is in the i18n.

config/locales/en.yml
```diff
@@ -53,6 +54,13 @@ en:
       student_loan_amount: "How much student loan did you repay while you were at %{claim_school_name} between 6 April XXXX and 5 April XXXX?"
       email_address: "What is your email address?"
       bank_details: "Enter bank account details"
+      eligible_subjects:
+        biology_taught: "Biology"
+        chemistry_taught: "Chemistry"
+        physics_taught: "Physics"
+        computer_science_taught: "Computer Science"
+        languages_taught: "Languages"
+        not_applicable: "Not applicable, I didn't teach any of these subjects"
```
erbridge: As mentioned on the last PR, "not applicable" doesn't feel like a valid answer to the question (it would imply that the user thinks the question isn't relevant to them). Shouldn't this be "No, I didn't teach any of these subjects", or just "I didn't teach any of these subjects"?

spec/features/ineligible_tslr_claims_spec.rb
```diff
@@ -43,13 +43,29 @@
...
-    expect(page).to have_text(I18n.t("tslr.questions.mostly_teaching_eligible_subjects"))
+    choose I18n.t("tslr.questions.eligible_subjects.not_applicable")
+    click_on "Continue"
+
+    expect(claim.reload.mostly_teaching_eligible_subjects).to eq(false)
+    expect(page).to have_text("You’re not eligible")
+    expect(page).to have_text("You must have spent at least half your time teaching an eligible subject.")
+  end
+
+  scenario "does not teach an eligible subject" do
```
erbridge: These two scenarios have the same name. Is that right?

config/locales/en.yml
pezholio: I'm going on what was agreed in the card. I reckon we should stick with what's been agreed and see what comes out of testing 👍

app/views/claims/subjects_taught.html.erb
pezholio: Whoops, good catch

spec/features/ineligible_tslr_claims_spec.rb
pezholio: I've tweaked this to be a bit more descriptive

config/locales/en.yml
erbridge: Fair enough 👍

PR #132, opened 2019-07-02
Get GOVUK Verify working locally

tekin:
This is the first step towards integrating with GOV.UK Verify. Things to note:

This only implements the very basics of the successful verification journey, there is more work to follow. Specifically:

- Handling the [Verify failure scenarios](https://www.docs.verify.service.gov.uk/get-started/handle-failure-scenarios/#handle-failure-scenarios)
- More sophisticated handling of the Verify responses (currently makes big assumptions about the attributes returned, e.g. they first one is the correct one and that it is actually verified)
- Preventing users from changing/editing the information returned by Verify
- Capturing gender (we'll want to do this so it can be passed to HMRC for Payroll)
- Making it easier to run the VSP locally by including it in `Profile.dev` with appropriate configuration

We'll tackle these as separate stories next sprint.

The integration is feature-switched and can be turned on locally in development mode by setting the `GOVUK_VERIFY_ENABLED=1` environment variable. We won't be turning it for staging/production environments until we have filled in more of the gaps


app/controllers/application_controller.rb
```diff
@@ -5,9 +5,23 @@ class ApplicationController < ActionController::Base
     if: -> { ENV.key?("BASIC_AUTH_USERNAME") },
   )
 
-  helper_method :signed_in?
+  helper_method :signed_in?, :govuk_verify_enabled?, :current_claim
+
+  private
+
+  def send_unstarted_claiments_to_the_start
```
pezholio: This is still in the ClaimsController. Does it need to be deleted now?

app/controllers/verify/authentications_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+    #   https://www.docs.verify.service.gov.uk/get-started/set-up-successful-verification-journey/#handle-a-response
+    def create
+      verify_authentication_response = Verify::ServiceProvider.new.translate_response(params["SAMLResponse"], session[:verify_request_id], "LEVEL_2")
+      current_claim.update!(claim_parameters_from_response(verify_authentication_response))
+      redirect_to claim_path("teacher-reference-number")
+    end
+
+    private
+
+    def claim_parameters_from_response(response)
```
pezholio: Should this be in the controller? Might be better in the model, or better still, a small service object

app/controllers/verify/authentications_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+      {
+        full_name: full_name_from_response(response),
+        address_line_1: response.fetch("attributes").fetch("addresses").first.fetch("value").fetch("lines")[0],
```
pezholio: Might be more elegant do use `dig` here, e.g:

```ruby
response.dig("attributes", "addresses").first.dig("value", "lines")[0]
```

config/application.rb
```diff
@@ -12,6 +12,8 @@
 require "sprockets/railtie"
 # require "rails/test_unit/railtie"
 
+require_relative "../lib/verify"
```
pezholio: Will Rails not autorequire everything in `lib`?

config/initializers/verify.rb
```diff
@@ -0,0 +1,3 @@
+Verify.configure do |config|
```
pezholio: 👍 

Feels like (once it's all done and dusted) this could be something we could break out into a gem

config/routes.rb
```diff
@@ -16,6 +16,15 @@
     end
   end
 
+  namespace :verify do
+    resources :authentications, only: [:new, :create]
+
+    if Rails.env.test?
+      require "verify/fake_sso"
+      mount Verify::FakeSso.new("/verify/authentications"), at: "/fake_sso"
```
pezholio: 👍

app/controllers/application_controller.rb
tekin: good spot!

app/controllers/verify/authentications_controller.rb
tekin: using `dig` would mean that `nil` would get returned if those keys were missing, whereas using `fetch` means that this will fail noisily if any keys aren't there, which at this stage is what I think we should be doing.

config/application.rb
tekin: nope, not since Rails 5, it's discouraged for thread-safety issues - https://blog.bigbinary.com/2016/08/29/rails-5-disables-autoloading-after-booting-the-app-in-production.html

app/controllers/verify/authentications_controller.rb
tekin: Yep, I definitely think we should break out the processing of the verify responses into their own object(s), but I think we'll be in a better place to extract that object when we are handling the failure cases - the shape/name/responsibility of the abstraction(s) will be clearer at that point.

tekin: Also, remind me to schedule a dev discussion about "service objects" at some point...

pezholio: Ha, yeah, we've had a brief chat about that. Just force of habit!

pezholio: But yeah, I think it's fine where it is for now

PR #136, opened 2019-07-03
Handle Verify failure scenarios

pezholio:
This adds a class to handle translated parameters from the Verify Service Provider, work out what type of response we have, and either update the claim with the verified details, or show an error.

Outstanding questions:

- [ ] What wording do we want to use when a user hits a failure scenario?
- [ ] Do we want to let them try again, or do we exit from the flow entirely? (delete the claim and clear the session etc).
- [ ] If it's an `AUTHENTICATION_FAILURE` scenario, do we want to guard against the use trying again? (or maybe limit the number of tries to guard against potential fraud)


tekin:
closing in favour of https://github.com/DFE-Digital/dfe-teachers-payment-service/pull/156

PR #141, opened 2019-07-04
Add scheduled tasks and background worker to CIP

erbridge:
Eject the app service from the building blocks so we can have more control over its parameters, including dropping the "staging" slot set up with different assumptions over how the app service will be used.

Switch to Docker Compose mode for the app service. This allows running the app service as a multi-container environment, supporting a worker container.

Switch to a deployment for setting up the KeyVault, so we can easily set access policies.

Set up `delayed_cron_job` to manage scheduled tasks through a `CronJob` superclass. The rake task that enqueues the jobs runs after each `db:migrate` and `db:schema:load`, automatically (unless running in development), meaning it runs on every deploy.


erbridge:

erbridge:
[Import schools on first deploy](https://trello.com/c/MB5FRnzf/314-import-schools-on-first-deploy)

lib/tasks/jobs.rake
```diff
@@ -0,0 +1,17 @@
+namespace :db do
+  desc "Schedule all cron jobs"
+  task schedule_jobs: :environment do
+    glob = Rails.root.join("app", "jobs", "**", "*_job.rb")
+
+    Dir.glob(glob).each { |f| require f }
+    CronJob.subclasses.each { |job| job.schedule }
```
pezholio: With other cron extensions I've seen (like [Resque Scheduler](https://github.com/resque/resque-scheduler)), they've used a YAML file to specify scheduled jobs. I like this subclass approach, but I'm not sure if it's a bit too opaque.

spec/jobs/cron_job_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+      it "leaves the job enqueued" do
+        existing_job = TestCronJob.send(:enqueued_job)
+
+        TestCronJob.schedule
+
+        expect(TestCronJob.send(:enqueued_job)).to eq(existing_job)
+      end
+
+      it "doesn't enqueue another instance of the job" do
+        expect { TestCronJob.schedule }.not_to(change { TestCronJob.send(:jobs).where(cron: nil).count })
```
pezholio: I think it might be clearer to just check the count here, so:

```ruby
schedule_count = TestCronJob.send(:jobs).where(cron: nil).count

TestCronJob.schedule

expect(TestCronJob.send(:jobs).where(cron: nil).count).to eq(schedule_count)
```

spec/jobs/cron_job_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+      it "leaves the job enqueued" do
+        existing_job = TestWithPerformOnScheduleCronJob.send(:enqueued_job)
+
+        TestWithPerformOnScheduleCronJob.schedule
+
+        expect(TestWithPerformOnScheduleCronJob.send(:enqueued_job)).to eq(existing_job)
+      end
+
+      it "doesn't enqueue another instance of the job" do
+        expect { TestWithPerformOnScheduleCronJob.schedule }.not_to(change { TestWithPerformOnScheduleCronJob.send(:jobs).where(cron: nil).count })
```
pezholio: As before, I reckon just testing the count is clearer

spec/lib/tasks/jobs_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  before do
+    Rake::Task.define_task(:environment)
+    Rake::Task.define_task(:"db:migrate")
+    Rake::Task.define_task(:"db:schema:load")
+    Rake.application.rake_require "tasks/jobs"
+  end
+
+  it "schedules and enqueues SchoolDataImporterJob" do
```
pezholio: This seems to rely on some state outside the test (namely `SchoolDataImporterJob`), so if we change / add jobs, this test may fail. Can we set up some stubs this so the state remains inside the tests?

PR #142, opened 2019-07-04
Setup staging and production infrastructure

pezholio:
This allows us to deploy to different environments, Test (or Edge), Development (or Staging) and Production. We've also documented how to deploy to prouction and get the relevant permissions, as there's some weirdness around this.


robbiepaul:
I've added the changes we made earlier, I just need to update the docs with the changes to deployment process

bin/azure-deploy
```diff
@@ -57,19 +67,13 @@ sed \
   -e "s/\${keyVaultName}/$KEYVAULT_NAME/g" \
   "$SECRETS_PARAMETERS_TEMPLATE_FILE_PATH" > "$SECRETS_PARAMETERS_FILE_PATH"
 
-echo
-echo "Are you ready to deploy secrets?"
-echo "  Hit return to continue, or CTRL+C to stop."
-read -r
```
erbridge: Can we leave these in and skip them via a flag or env var?

bin/azure-deploy
```diff
@@ -57,19 +67,13 @@ sed \
...
-
 echo "Starting secrets deployment..."
 az group deployment create \
   --name "$SECRETS_DEPLOYMENT_NAME" \
   --resource-group "$SECRETS_RESOURCE_GROUP_NAME" \
   --template-file "$SECRETS_TEMPLATE_FILE_PATH" \
   --parameters "@$SECRETS_PARAMETERS_FILE_PATH" \
   --mode Complete \
-  --rollback-on-error \
```
erbridge: This should only ever be an issue on first deploy. I had considered using a flag on `bin/azure-deploy` to skip this flag on first deployment. Maybe now is the time to do that?

robbiepaul:
Yeah, I agree. Although the impression I get is that we're supposed to deploy through the central Service Principle Name. I could be wrong though. There's a reply from David Goodyear on Slack: https://ukgovernmentdfe.slack.com/archives/CKYNU0VG9/p1562767543024000

robbiepaul:
Actually looking at it we might be able to sign in to the CLI as the service principle? https://docs.microsoft.com/en-us/cli/azure/authenticate-azure-cli?view=azure-cli-latest#sign-in-with-a-service-principal

erbridge:
In that case, rather than adding infrastructure deployment to CD, let's change the script to log in as the appropriate service principle.

erbridge:
I think for now, if we rollback to before Robbie's commits, this is good to go? We can then do the work to use the service principle in another PR.

pezholio:
That does leave the question, how do we actually deploy using the release pipeline? The default azure Web App deployment task in Pipelines stomps all over our docker-compose.yml, so won't work.

The way I envisaged it was that we continuously deploy to the development envrionment from the `master` branch, and  to the test environment from the `edge` branch. Deploying to production will still requires a human to manually kick off the deploy from the pipeline, and we can do manual checks of the config before deploying, and also have confidence that it works because we saw an automatic deploy go to development.

robbiepaul:
I've removed the two commits from yesterday. Stu's right though it does leave the question of how we're going to deploy code changes via the pipeline though.

PR #145, opened 2019-07-08
Add workers to process background jobs on Azure Cloud Infrastructure Platform

pezholio:
This does the following:

* Ejects the app service from the building blocks so we can have more control over its parameters, including dropping the "staging" slot set up with different assumptions over how the app service will be used.
* Switches to Docker Compose mode for the app service. This allows running the app service as a multi-container environment, supporting a worker container.
* Switch to a deployment for setting up the KeyVault, so we can easily set access policies.


pezholio:

EwanNoble:
Would advise against Docker Compose for two reasons:

- Firstly, as we talked about on Slack you will probably want to separate the underlying resources so that your background worker process has no negative effects on your web service, using Docker Compose would mean you are running both apps on the same resources. 
- It's also still a Preview feature and unless there is a specific benefit it provides I wouldn't recommend committing to using it.

I would suggest creating adding an additional App Service Plan and App Service to achieve this separation.

I haven't reviewed the deployment script, if you are deploying your environments to Azure from Azure Pipelines I would suggest using the native tools available to do so, there is examples of this in other TWD projects already.

pezholio:
@EwanNoble Is there any prior art on getting a worker set up? I'm guessing I need to send a different command to the Docker container in the worker App Service, but the Azure documentation is pretty opaque to say the least.

EwanNoble:
> @EwanNoble Is there any prior art on getting a worker set up? I'm guessing I need to send a different command to the Docker container in the worker App Service, but the Azure documentation is pretty opaque to say the least.

I would imagine that you'd have a build pipeline that creates an image for your web service and an image for your background worker. You will then pass both of these into your template and assign them to their respective app services.

EwanNoble:
Or are you referring to App Service startup commands?

pezholio:
Yeah, the startup commands really. Currently we have one Dockerfile with a startup `CMD` of `rails server`, which we tweak in the docker-compose file for the worker, but I'm not sure how we can override this if we're not using docker-compose

EwanNoble:
There is a property on the app services called [appCommandLine](https://docs.microsoft.com/en-us/azure/templates/microsoft.web/2018-11-01/sites#siteconfig-object) which you can specify if you need to run a startup command. You can put this into your template for now whilst you're not using the building blocks and perhaps PR a change to parameterise this.

pezholio:
Ah, cool. That's really useful, ta 👍

PR #146, opened 2019-07-08
Import schools on first deploy

pezholio:
Sets up [`delayed_cron_job`](https://github.com/codez/delayed_cron_job) to manage scheduled tasks through a `CronJob` superclass. The rake task that enqueues the jobs runs after each `db:migrate` and `db:schema:load`, automatically (unless running in development), meaning it runs on every deploy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,12 +77,10 @@ including how to run the verify dependencies locally.
 
 ### Running `CronJob`s
 
-Recurring jobs are scheduled by the `db:schedule_jobs` rake task. This task
-normally runs as part of `db:migrate` or `db:schema:load`, but is disabled in
-development environments. To schedule them during development, run:
+To schedule recurring jobs, run the following:
 
 ```
-rake db:schedule_jobs
+rake jobs:schedule
 ```
 
 ## Running specs, brakeman, and code linting
--- a/README.md
+++ b/README.md
@@ -75,6 +75,16 @@ The service uses GOV.UK Verify to verify the identity of teachers that are
 claiming. See [docs/govuk-verify](/docs/govuk-verify.md) for details on this,
 including how to run the verify dependencies locally.
 
+### Running `CronJob`s
+
+Recurring jobs are scheduled by the `db:schedule_jobs` rake task. This task
+normally runs as part of `db:migrate` or `db:schema:load`, but is disabled in
+development environments. To schedule them during development, run:
+
+```
+rake db:schedule_jobs
+```
+
 ## Running specs, brakeman, and code linting
 
 ```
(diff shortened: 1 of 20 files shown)


pezholio:
[Import schools on first deploy](https://trello.com/c/MB5FRnzf/314-import-schools-on-first-deploy)

app/jobs/cron_job.rb
```diff
@@ -0,0 +1,39 @@
+class CronJob < ApplicationJob
+  class_attribute :cron_expression
+  class_attribute :perform_on_schedule
+
+  class << self
+    def schedule
```
tekin: This method schedule two jobs if `perform_on_schedule` is set to true: one with a `cron` value of `nil`; one with a `cron` value set to the expression defined on the class. Is that intended? I found this confusing. From my reading of the code, the impact of this is that a job would get scheduled for immediate execution every time we deployed the code, which makes me uncomfortable, especially given how long the school import takes to run. I don't believe we want or need to update the schools data on every deploy.

Could we use the [simple example base class](https://github.com/codez/delayed_cron_job#scheduling) given by the gem documentation instead? I'm not convinced we'd ever want to auto-run jobs on every deploy, and I'd feel more comfortable with the maintenance burden and potential for bugs if we weren't diverging too far from the gem's recommended base class.

lib/tasks/jobs.rake
```diff
@@ -0,0 +1,17 @@
+namespace :db do
```
tekin: I know the example on the gem puts this task under the `db` namespace but I think it would make more sense to have this job namespaced under the existing `jobs` namespace and call the job `schedule`. e.g. 

```ruby
namespace :jobs do
  task schedule: :environment do
  ...
  end
end
```

tekin:
Perhaps this story for this was badly named... In reality, we don't actually need to automate the school import so it happens on first deploy. Code to do that is not going to have much value long term - in theory it would literally get executed twice ever! Much easier just to let the schedule kick in and do it on the next run. The rest of the work to have it execute on a daily schedule is good though one the auto-execute stuff is stripped out.

README.md
```diff
@@ -75,6 +75,14 @@ The service uses GOV.UK Verify to verify the identity of teachers that are
 claiming. See [docs/govuk-verify](/docs/govuk-verify.md) for details on this,
 including how to run the verify dependencies locally.
 
+### Running `CronJob`s
+
+To schedule recurring jobs, run the following:
+
+```
+rake db:schedule_jobs
```
tekin: This needs updating to reflect the new name for the rake task.

PR #150, opened 2019-07-09
Refactor page sequence code

tekin:
We had logic for determining the sequence of pages in both the model and the controller. With more sequence-related logic coming down the pipe (i.e. the multi-page student loan plan questions), moving all the sequence-related logic into a single place is going to make our lives a lot easier.

### DIFF
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -51,13 +51,13 @@ class ClaimsController < ApplicationController
 
   def next_claim_path
     return ineligible_claim_path if current_claim.ineligible?
-    return claim_path("check-your-answers") if edited_answer?
+    return claim_path("check-your-answers") if edited_answer?  && !on_school_flow?
 
     claim_path(next_slug)
   end
 
   def edited_answer?
-    current_claim.can_be_submitted? && !current_claim.submitted? && !on_school_flow?
+    current_claim.can_be_submitted? && !current_claim.submitted?
   end
 
   def on_school_flow?
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -50,8 +50,7 @@ class ClaimsController < ApplicationController
   end
 
   def next_claim_path
-    return ineligible_claim_path if current_claim.ineligible?
-    return claim_path("check-your-answers") if edited_answer?  && !on_school_flow?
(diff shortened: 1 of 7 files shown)


app/models/page_sequence.rb
```diff
@@ -0,0 +1,65 @@
+# frozen_string_literal: true
+
+# Represents the pages in the claim process.
+#
+# Depending on the state of `claim`,  the sequence of slugs will change,
+# filtering out pages that are not relevant. For example, the sequence for a
+# claim that states it is still employed at the same school as the
+# `claim_school` will skip the `current-school` page as `current_school` is
+# inferred to be the same as `claim_school.
+class PageSequence
```
pezholio: Just wondering if this should be called TslrPageSequence, as there will no doubt be a sequence for Maths and Physics too? There may be some logic we can share too

tekin: The answer to that one is: I don't think we know enough to decide yet. It may be that passing in the claim will give the generic `PageSequence` enough to create a policy-appropriate sequence. There will almost certainly be logic shared between the two, but how that shakes out will be easier to decide at the point we are introducing a second sequence.

tekin: For example, maybe we end up with a generic `PageSequence` with policy-specific rules/slugs that it is composed of. We'll be better equipped to decide when we're working on the work to introduce Maths and Physics support

pezholio: Yep, cool. Fair enough. I think this can easily be changed further down the line 👍

app/controllers/claims_controller.rb
```diff
@@ -67,8 +67,7 @@ def on_school_flow?
   end
 
   def next_slug
-    current_slug_index = current_claim.page_sequence.index(params[:slug])
-    current_claim.page_sequence[current_slug_index + 1]
+    PageSequence.new(current_claim, params[:slug]).next_slug
```
mec: This feels like we crate a new PageSequence everytime we call `next_slug`, I think I would've created a single PageSequence instance when the claims controller is inatilized? What's the thinking behind this approach? Does Rails create a new controller for each request so `next_slug` is only ever called once?

tekin: Yep, each request instantiates a new instance of the controller, and so because we only call `next_slug` once, this only gets instantiated once. We may want to memoize this and/or extract it to a helper method at some point, but there is no reason to do so now.

PR #152, opened 2019-07-10
Amend the student loan amount questions

titlescreen:
UR was showing that people were likely to just guess rather than enter
the precise amount with the expectation that DfE or HMRC would pick up
the discrepancy and correct it.

The text was amended to encourage users to provide some more accurate
information and will be tested in a future sprint.


config/locales/en.yml
```diff
@@ -53,8 +53,8 @@ en:
       teacher_reference_number: "What's your teacher reference number?"
       national_insurance_number: "What's your National Insurance number?"
       student_loan_amount:
-        "How much student loan did you repay while you were at %{claim_school_name} between 6 April XXXX and 5 April
-        XXXX?"
+        "Exactly how much student loan did you repay while you worked at Penistone Grammar School between 6 April XXXX
```
pezholio: Should be `Exactly how much student loan did you repay while you worked at %{claim_school_name}`

titlescreen: Changing now.

db/schema.rb
```diff
@@ -82,9 +82,9 @@
     t.string "national_insurance_number", limit: 9
     t.string "email_address", limit: 256
     t.boolean "mostly_teaching_eligible_subjects"
+    t.datetime "submitted_at"
     t.string "bank_sort_code", limit: 6
     t.string "bank_account_number", limit: 8
-    t.datetime "submitted_at"
```
tekin: We should avoid these unexpected column order changes in `schema.rb`, otherwise they will dance back and forth as others run migrations. They happen when locally migrations have been run out of order (normally when switching between branches and running migrations before they've been merged into master). Something to keep an eye out in the future. I'll push up a PR to fix this.

PR #154, opened 2019-07-10
Make sure cron jobs can be scheduled independently

pezholio:
As part of the initialisation process, we were running the `rake schools_data:import` task, but running `rake jobs:schedule` immediately afterwards stomps all over our previously scheduled job. This makes sure our CronJob superclass `jobs` method only cares about jobs with a cron attached to them


robbiepaul:

PR #155, opened 2019-07-10
Amend QTS question to use ITT

titlescreen:
Users were struggling to answer the QTS question, they were unsure if
they answered Jun YEAR when they got QTS or Sept YEAR when they started
at the school (Newly Qualified Teacher).

We changed it to use Initial Teacher Training as there were no issues
with this in the maths and physics journey.

Amended CI to use the alias for the question so we don't have to keep
updating it.

### DIFF
--- a/spec/requests/claims_spec.rb
+++ b/spec/requests/claims_spec.rb
@@ -24,7 +24,7 @@ RSpec.describe "Claims", type: :request do
 
       it "renders the requested page in the sequence" do
         get claim_path("qts-year")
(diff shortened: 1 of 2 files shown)


spec/requests/claims_spec.rb
```diff
@@ -24,7 +24,7 @@
 
       it "renders the requested page in the sequence" do
         get claim_path("qts-year")
-        expect(response.body).to include("Which academic year were you awarded qualified teacher status")
+        expect(response.body).to include(I18n.t("tslr.questions.qts_award_year"))
```
pezholio: 👍

PR #156, opened 2019-07-10
Handle GOV.UK Verify failure cases

tekin:
This builds on the work started by @pezholio in https://github.com/DFE-Digital/dfe-teachers-payment-service/pull/136.

Adds handling for the Verify failure scenarios, as outlined on [Handle failure scenarios docs](https://www.docs.verify.service.gov.uk/get-started/handle-failure-scenarios/#handle-failure-scenarios).

This also adds a verify confirmation page on successful verification so that we don't unceremoniously drop the user straight onto the TRN question page.

Still work to do, particularly on the `Verify::Response` class to make it more generic (i.e. the `claim_parameters` method is specific to our application), but that work would be better done as part of the [capture most recent verified data](https://trello.com/c/ptDe751W/355-only-capture-the-most-recent-verified-version-of-each-name-or-address-from-verify) story.

## TODO:

- [ ] run the new content pages past Patrick and update as appropriate

### Verify confirmation
The user was able to successfully complete verification
Screenshot 2019-07-10 at 15 48 25

### Verification failed
The user completed the verify process but failed to pass verification
Screenshot 2019-07-10 at 15 48 34

### Verification not completed
The user didn't get to the end of the verification process and quit early
Screenshot 2019-07-10 at 15 48 45


lib/verify/service_provider.rb
```diff
@@ -55,14 +56,23 @@ def generate_request
...
 
-      JSON.parse(response.body)
+      handle_response response.body
+    end
+
+    private
+
+    def handle_response(json_response)
+      JSON.parse(json_response).tap do |parameters|
+        raise ResponseError.new(parameters["message"], parameters["code"]) if parameters.key?("code")
```
pezholio: 👌 I like the idea of passing the exception down the pipe so we know about it

tekin:
Merging this for now as Patrick won't have time to give feedback on the content until next week. We can update the pages accordingly then.

PR #157, opened 2019-07-10
Update cron job to run at 4am

pezholio:
We had a Rollbar error this morning (https://rollbar.com/dxw/teachers-payment-service/items/8/) when importing schools because the CSV for that day didn't yet exist. I've checked with the GIAS team, and they've told us these are generated at around 3am, so this queues the job at 4am when it's more likely that the CSV will be around. If it's not yet imported, the job will retry.

### DIFF
--- a/app/jobs/school_data_importer_job.rb
+++ b/app/jobs/school_data_importer_job.rb
@@ -1,5 +1,5 @@
 class SchoolDataImporterJob < CronJob
-  self.cron_expression = "0 0 * * *"
+  self.cron_expression = "0 0 4 * * * "
 
   queue_as :school_data
 


mec:
Should we close the Rollbar entry?

pezholio:
Yeah, I'll do it 👍

app/jobs/school_data_importer_job.rb
```diff
@@ -1,5 +1,5 @@
 class SchoolDataImporterJob < CronJob
-  self.cron_expression = "0 0 * * *"
+  self.cron_expression = "0 0 4 * * * "
```
erbridge: Does this work? Cron doesn't normally do second resolution. The format I'm familiar with is `minute hour day_of_month month day_of_week`.

pezholio: Yeah, I tested it. The Cron implementation the gem uses is `second minute hour day_of_month month day_of_week` https://github.com/codez/delayed_cron_job/blob/master/lib/delayed_cron_job/cronline.rb

erbridge: That's helpful of them, given they link to a definition that doesn't do seconds in their docs...

pezholio: Ha! I didn't even see that. Might do an upstream PR to fix that.

PR #163, opened 2019-07-11
Run school import in seeds.rb

tekin:
This makes the first run experience better for a developer trying to run the application (they don't need to know to run the import) and also removes the app-specific knowledge from `docker-entrypoint`.

### DIFF
--- a/bin/docker-entrypoint
+++ b/bin/docker-entrypoint
@@ -3,12 +3,12 @@ set -e
 
 if [ "$RAILS_ENV" != "production" ]; then
   echo "In Dev / Test - Initializing database"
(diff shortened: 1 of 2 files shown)


tekin:
@titlescreen when this gets merged, you can run `rake db:drop db:setup` to get your local database schema back in sync and avoid unnecessary `schema.rb` changes like https://github.com/DFE-Digital/dfe-teachers-payment-service/commit/d7fe6325c9f9a90c76ded65162ff18121078b8d2#diff-1acd2e7e27a227829d5d14a91c863bb6R85 creeping in to your PRs

PR #164, opened 2019-07-11
Submittiable behaviour tweaks

mec:
Whilst working on the Student loan plan journey, we wanted to make the page sequence easier to manage which this refatoring enables.

After this we'll have a more robust way to mange the page sequence, which _should_ make the student loan plan journey much simpler.

### DIFF
--- a/app/models/tslr_claim.rb
+++ b/app/models/tslr_claim.rb
@@ -106,6 +106,10 @@ class TslrClaim < ApplicationRecord
     submitted_at.present?
   end
 
+  def submittable?
+    valid?(:submit) && !submitted?
+  end
+
   def ineligible?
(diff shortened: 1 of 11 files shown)


app/models/tslr_claim.rb
```diff
@@ -91,24 +93,21 @@ class TslrClaim < ApplicationRecord
   scope :submitted, -> { where.not(submitted_at: nil) }
 
   def submit!
-    if ineligible?
-      errors.add(:base, full_ineligibility_reason)
-      return false
-    end
-
-    if can_be_submitted?
+    if valid?(:submit)
```
erbridge: Is there a reason this isn't using `submittable?`?

mec: No reason at all. Not using it was allowing multiple submissions which we had no coverage for. 

Great spot, thanks @erbridge!

PR #168, opened 2019-07-15
Update Verify content based on Patrick’s feedback

tekin:


app/views/verify/authentications/failed.html.erb
```diff
@@ -2,22 +2,35 @@
...
     </p>
 
     <ul class="govuk-list govuk-list--bullet">
       <li>
-        your answer to a question doesn't match the records held by the Verify provider
+        the company you chose does not have enough information about you
       </li>
       <li>
-        the Verify provider that you chose does not have access to enough information about you
+        the company’s records are not up to date
```
pezholio: Is it clear that `the company` referred to here refers to `the company you chose`, or should we repeat this?

app/views/verify/authentications/failed.html.erb
```diff
@@ -2,22 +2,35 @@
...
+    <h2 class="govuk-heading-m">
+      Next steps
+    </h2>
+
+    <ul class="govuk-list govuk-list--bullet">
+      <li>
+        check your answers are correct
+      </li>
+      <li>
+        try using a different company
```
pezholio: `try using a different company for verification` maybe?

app/views/verify/authentications/failed.html.erb
PatrickTaylorDfE: We call it the company you chose up above this, so though I considered it, repeating the 'you chose' seemed a bit excessive to me.

app/views/verify/authentications/failed.html.erb
PatrickTaylorDfE: If we were going to do this then I would probably go for 'try choosing a different company in GOV.UK Verify'

app/views/verify/authentications/failed.html.erb
pezholio: Cool fair enough 👍

app/views/verify/authentications/failed.html.erb
tekin: @PatrickTaylorDfE do you have a preference either way?

PatrickTaylorDfE: My preference would be to leave as is.

pezholio: Cool, let's leave it then. I'm probably overthinking it 😄

PR #169, opened 2019-07-15
Only capture the most recent verified name and address

robbiepaul:
It is possible to have values returned from Verify that aren't verified. We don't want to capture these values as it's important that the data we have is current, correct and valid.

This PR introduces the concept of `most_recent_verfied_value` that will filter out non-verified attributes, then sort them by `to` date. This makes an assumption that this value will always be present.

### DIFF
--- a/lib/verify/response.rb
+++ b/lib/verify/response.rb
@@ -42,7 +42,7 @@ module Verify
     private
 
     def address
-      @address ||= parameters.fetch("attributes").fetch("addresses").first.fetch("value")
+      @address ||= most_recent_verified_value(parameters.fetch("attributes").fetch("addresses"))
     end
 
     def address_lines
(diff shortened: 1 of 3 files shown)


robbiepaul:
[Only capture the most recent "verified" version of each name or address from Verify](https://trello.com/c/ptDe751W/355-only-capture-the-most-recent-verified-version-of-each-name-or-address-from-verify)

lib/verify/response.rb
```diff
@@ -42,19 +42,25 @@ def scenario
...
+      first_name = most_recent_verified_value(parameters.fetch("attributes").fetch("firstNames"))
+      middle_name = most_recent_verified_value(parameters.fetch("attributes").fetch("middleNames"))
+      surname = most_recent_verified_value(parameters.fetch("attributes").fetch("surnames"))
 
       [first_name, middle_name, surname].join(" ")
     end
+
+    def most_recent_verified_value(attributes)
+      attributes.sort_by { |attribute| Date.strptime(attribute["to"], "%Y-%m-%d") }
+        .select { |attribute| attribute["verified"] }
```
pezholio: Would it be better to use `find` here, then you can drop the call to `.last` on the next line

lib/verify/response.rb
```diff
@@ -42,19 +42,25 @@ def scenario
...
+        .last.fetch("value")
```
pezholio: What happens if there are no verified attributes?

lib/verify/response.rb
robbiepaul: I've changed this to use `find`. I had to add a `reverse` in there too to make sure we were getting the most recent value.

lib/verify/response.rb
robbiepaul: From the quote on the card it looks like we can always expect to retrieve one verified value.

> "unfortunately the rules are a bit weird on verified or not […] for names you should be ables to expect just one current verified value, but you may also get a current non-verified value. It may also be possible to get a current unverified name and only a verified previous name (possibly due to you recently changing your name)

PR #174, opened 2019-07-16
375 tslr student loan amount

titlescreen:
Amend question text to include the words exactly

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -53,8 +53,8 @@ en:
       teacher_reference_number: "What's your teacher reference number?"
       national_insurance_number: "What's your National Insurance number?"
       student_loan_amount:
-        "How much student loan did you repay while you were at %{claim_school_name} between 6 April 2018 and 5 April
-        2019?"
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
```diff
@@ -53,8 +53,8 @@ en:
       teacher_reference_number: "What's your teacher reference number?"
       national_insurance_number: "What's your National Insurance number?"
       student_loan_amount:
-        "How much student loan did you repay while you were at %{claim_school_name} between 6 April 2018 and 5 April
-        2019?"
+        "Exactly how much student loan did you repay while you worked at %{claim_school_name} between 6 April XXXX and 5
```
pezholio: Do we need to replace the `XXXX`s with real years?

db/schema.rb
```diff
@@ -82,9 +82,9 @@
     t.string "national_insurance_number", limit: 9
     t.string "email_address", limit: 256
     t.boolean "mostly_teaching_eligible_subjects"
+    t.datetime "submitted_at"
```
pezholio: This has happened again, can you edit your commit to remove this file?

titlescreen: I think I've sorted it now

PR #175, opened 2019-07-16
Tag as `latest` when deploying a Docker image

erbridge:
This only happens on `master`, so means `latest` will always be the most recent build.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -39,9 +39,13 @@ steps:
       pass: $(dockerPassword)
     displayName: Login to DockerHub
     condition: eq(variables['Build.SourceBranch'], 'refs/heads/master')
-  - script: docker build --file=Dockerfile --tag=$(dockerRegistry)/$(dockerImageName):$(Build.BuildNumber) --target=web .
+  - script:
+      docker build --file=Dockerfile --tag=$(dockerRegistry)/$(dockerImageName):$(Build.BuildNumber) --target=web .
     displayName: Build Docker image for deployment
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -39,9 +39,13 @@ steps:
...
-  - script: docker build --file=Dockerfile --tag=$(dockerRegistry)/$(dockerImageName):$(Build.BuildNumber) --target=web .
+  - script:
+      docker build --file=Dockerfile --tag=$(dockerRegistry)/$(dockerImageName):$(Build.BuildNumber) --target=web .
     displayName: Build Docker image for deployment
     condition: eq(variables['Build.SourceBranch'], 'refs/heads/master')
-  - script: docker push $(dockerRegistry)/$(dockerImageName):$(Build.BuildNumber)
+  - script: |
+      docker tag $(dockerRegistry)/$(dockerImageName):$(Build.BuildNumber) $(dockerRegistry)/$(dockerImageName):latest
+      docker push $(dockerRegistry)/$(dockerImageName):$(Build.BuildNumber)
+      docker push $(dockerRegistry)/$(dockerImageName):latest
```
pezholio: Can we get rid of the docker-related stuff in the pipeline now then? So all we need to do to deploy is just restart the app right?

erbridge: I think it's still better to be explicit about setting the `development` tag, so we can e.g. rollback. Also keeps development deployments closer to production ones.

pezholio: 👌

PR #176, opened 2019-07-16
Admin sessions time out after inactivity

robbiepaul:
> Sessions that never expire extend the time-frame for attacks such as cross-site request forgery (CSRF), session hijacking and session fixation.

We should ensure that admin sign-in sessions do not last longer than 30 minutes of inactivity.

There's a bit of duplication of logic between the code here and the code in the claims controller. We think it's best to keep this separate and not mix admin session logic with our normal user session logic.

### DIFF
--- a/app/controllers/admin/base_admin_controller.rb
+++ b/app/controllers/admin/base_admin_controller.rb
@@ -1,11 +1,25 @@
 module Admin
   class BaseAdminController < ApplicationController
-    before_action :ensure_authenticated_user
+    TIMEOUT_LENGTH_IN_MINUTES = 30
+
+    before_action :ensure_authenticated_user, :end_expired_sessions, :update_last_seen_at
 
     private
 
     def ensure_authenticated_user
       redirect_to admin_sign_in_path unless signed_in?
+    end
+
+    def end_expired_sessions
+      if admin_session_timed_out?
+        session.destroy
+        redirect_to admin_sign_in_path, notice: "Your session has timed out due to inactivity, please sign-in again"
+      end
+    end
+
+    def admin_session_timed_out?
+      signed_in? && session.key?(:last_seen_at) &&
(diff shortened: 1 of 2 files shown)


robbiepaul:
[admin sign-in sessions should time out after 30 minutes](https://trello.com/c/YIaPE0xf/336-admin-sign-in-sessions-should-time-out-after-30-minutes)

app/controllers/admin/base_admin_controller.rb
```diff
@@ -1,11 +1,25 @@
...
     private
 
     def ensure_authenticated_user
       redirect_to admin_sign_in_path unless signed_in?
     end
+
+    def end_expired_sessions
+      if admin_session_timed_out?
+        session.destroy
+        redirect_to admin_sign_in_path, notice: "Your session has timed out due to inactivity, sign-in again"
```
pezholio: I'd probably add a 'please' to the 'sign-in again' instruction, just to be nicer 😄

app/controllers/admin/base_admin_controller.rb
```diff
@@ -1,11 +1,25 @@
 module Admin
   class BaseAdminController < ApplicationController
-    before_action :ensure_authenticated_user
+    TIMEOUT_LENGTH_IN_MINUTES = 30
+
+    before_action :ensure_authenticated_user, :end_expired_sessions, :update_last_seen_at
```
pezholio: Is `update_last_seen_at` in the `ApplicationController` now?

robbiepaul: Yeah, it was moved there as part of the Verify work

app/controllers/admin/base_admin_controller.rb
robbiepaul: 👍

PR #179, opened 2019-07-16
Prevent editing of verified name, address and date of birth

robbiepaul:
Now that the Verify integration is in place, we can remove the ability to manually set/change the personal information that is returned by Verify, i.e. name, address and date of birth.

Setting as draft as we're not quite ready to integrate with Verify on staging just yet.


robbiepaul:
[Prevent editing/changing of Verified name/address/dob by users](https://trello.com/c/8Q9zh0de/376-prevent-editing-changing-of-verified-name-address-dob-by-users)

README.md
```diff
@@ -74,11 +74,10 @@ NOTIFY_TEMPLATE_ID=d72e2ff9-b228-4f16-9099-fd9d411c0334
...
 claiming.
 
-If you want to test Verify in development, add the following to your `.env`
-file.
+In order to integrate with Verify in development, add the following to your
+`.env` file.
 
 ```
-GOVUK_VERIFY_ENABLED=1
 GOVUK_VERIFY_VSP_HOST=http://localhost:50300
```
erbridge: Given this isn't really configurable or optional, maybe we should create a `.env.development` again instead of getting developers to maintain their own config?

app/models/tslr_claim.rb
```diff
@@ -39,21 +39,21 @@ class TslrClaim < ApplicationRecord
 
   validates :mostly_teaching_eligible_subjects, on: [:"mostly-teaching-eligible-subjects", :submit], inclusion: {in: [true, false], message: "Select either Yes or No"}
 
-  validates :full_name,                         on: [:"full-name", :submit], presence: {message: "Enter your full name"}
+  validates :full_name,                         on: :submit, presence: {message: "Enter your full name"}
```
erbridge: These messages should probably be updated to ask the user to re-Verify, or something. Asking a user to enter their name, but not giving them any mechanism to do so is a road to frustration. Though what would cause these fields to be missing that the user could do anything about anyway?

robbiepaul: Good point, I'll raise this at stand-up tomorrow

robbiepaul: I totally forgot about this at stand-up. I'd be interested to know your thoughts @titlescreen @tekin 

I think we need to keep the validation on `submit` for these values as it's what makes a claim "submittable". I don't think a user should be able to get in a state where they'll see this error message because we get their details from Verify (are we confident that we always get all the fields?). 

I'm not sure about the error message, we could go with something more generic like "Full name is required" for the cases where we might manually create/update claim records.

tekin: Technically a user could jump to /check-your-answers before going through verify, or at least they can at the moment, but yes, in theory a user going through the normal flow shouldn't ever see this message. It might be worth adding something to the backlog so that we prevent users moving beyond the verify step until they've actually got through verify.

erbridge: So perhaps it makes sense to rephrase the error message to ask the user to (re-)Verify, and make the "change" links take them to the Verify step again (though they'd need some way to decide _not_ to re-Verify other than their browser back button)?

robbiepaul: I had a chat with Rich about this and we feel the re-verification step needs some design thinking. Maybe it needs to happen as soon as the user is returned from Verify?

We know that we'll get back name, address and date of birth from Verify so it feels like an edge case and if it is an issue it should surface during private beta.

I agree that we should prevent users from skipping ahead to the check-your-answers page.

erbridge:
Maybe this belongs in a separate body of work, but maybe, rather than removing asking for address entirely, we ask it if verify returns blank.

tekin:
Closing this in favour of https://github.com/DFE-Digital/dfe-teachers-payment-service/pull/245

PR #182, opened 2019-07-17
Format currency to 2 decimal places in CSV rows

erbridge:


tekin:
What's the rationale for this one?

erbridge:
I was taking a quick glance at the bug about the number input not always showing 2 decimal places and noticed that the CSV was also only showing 1 decimal place. I assert that it's easier to parse currencies by eye if they have 2 decimal places. The comma was a bonus, but also easier to parse.

If you don't think this change is appropriate, fine to close. Just felt like an easy improvement.

tekin:
Yeah, I think it's probably not worth messing with the formatting too much now, at least until we understand more about the rest of the flow for claim processing and handing off for payment/payroll, etc. For example we may find that the commas cause issues for systems we pass the data onto.

erbridge:
Fair enough.

PR #183, opened 2019-07-17
Document how to release a production build

erbridge:
We normally deploy from existing development releases so we use the deployment configuration at the time of first release, so we call scripts with the appropriate arguments, for example.

We decided that we don't need to mess with Git tags or similar because Azure DevOps keeps a record of releases and supports filtering by the most current release to each stage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,33 @@ Code linting is performed using:
 [Bullet](https://github.com/flyerhzm/bullet) runs around each spec. If it
 detects an N+1 query it will raise an exception and the tests will fail.
 
+## Deployment
+
+### Development
+
+Development is automatically built and deployed when commits are pushed to
+`master`.
+
+### Production
+
+To do a production release:
+
+1. Log in to this project on [Azure DevOps](azure-devops).
+2. Navigate to _Pipelines_ > _Builds_.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -112,6 +112,32 @@ Code linting is performed using:
...
+`master`.
+
+### Production
+
+To do a production release:
+
+1. Log in to this project on [Azure DevOps](azure-devops).
+2. Navigate to _Pipelines_ > _Builds_.
+3. Find the build you want to release and note its _Build #_ (e.g.
+   `20190717.2`). You should only be releasing builds of `master`.
```
tekin: is there an obvious way to identify the specific build for a particular revision?

erbridge: You can filter by branch, but otherwise you're looking to match commit messages by eye in the UI. Once you're on a build, you can see the commit hash and click through to GitHub.

tekin: ok, cool. I wonder if the "You should only be releasing builds of `master`" sentence is entirely true. For example, what if we have a feature that requires a multi-stage release, say with the first release adding a new column before the second makes use of it, we may legitimately want to release a release that isn't master.

erbridge: Sure. I'll remove.

tekin:
Looking good! 👍🏽

PR #184, opened 2019-07-17
Restrict admin access based on DfE sign in roles

pezholio:
This uses the [DfE Signin API](https://github.com/DFE-Digital/login.dfe.public-api) to check a user has access to our service after a successful login, as they may be a DfE Signin user, but not have the relevant permissions to access our service. 

I've added a module called `DfeSignin` that we can potentially extend, should be need to add more endpoints. It should be easy enough to break out into a gem if we have the time/inclination.

The relevant secrets have been added to the development keyvault, but not the production one yet, as we need to get different secrets for production. This will also mean we'll need to do a production deploy once this is merged 👍


pezholio:
[Restrict admin access based on DfE Sign-in roles](https://trello.com/c/I22q4ZfC/384-restrict-admin-access-based-on-dfe-sign-in-roles)

azure/app/parameters/production.template.json
```diff
@@ -88,6 +88,38 @@
...
+      }
+    },
+    "DFE_SIGN_IN_ROLE": {
+      "reference": {
+        "keyVault": {
+          "id": "/subscriptions/${subscriptionId}/resourceGroups/${keyVaultResourceGroupName}/providers/Microsoft.KeyVault/vaults/${keyVaultName}"
+        },
+        "secretName": "DfeSignInApiRole"
+      }
+    },
```
erbridge: Let's keep `test` up to date, too, even if we're not using it right now.

Dockerfile
```diff
@@ -118,6 +118,7 @@ RUN apk add chromium chromium-chromedriver
 COPY .prettierignore ${APP_HOME}/.prettierignore
 COPY .prettierrc ${APP_HOME}/.prettierrc
 COPY .rspec ${APP_HOME}/.rspec
+COPY .env.test ${APP_HOME}/.env.test
```
erbridge: This shouldn't be necessary. `docker-compose.test.yml` should already hook up the environment variables from that file.

app/controllers/admin/auth_controller.rb
```diff
@@ -11,17 +11,36 @@ def sign_out
...
     def failure
     end
 
     private
 
+    def authorised?
+      DfeSignin::Authorisation.new(
+        user_id: user_id,
+        organisation_id: organisation_id
+      ).has_role?(ENV["DFE_SIGN_IN_ROLE"])
```
erbridge: Can we pull the role out somewhere and verify that it's set at startup?

pezholio: I've tweaked it to do an `ENV.fetch`, that was it'll error on initialization if it's not set

app/models/dfe_signin/authorisation.rb
```diff
@@ -0,0 +1,26 @@
+module DfeSignin
```
tekin: What do you think about getting rid of this class altogether and having the `has_role?` method directly on the `DfeSignin::UserAccess` class instead? From what I can see, the `DfeSignin::UserAccess` object has everything it needs to be able to say whether it has a role or not.

docker-compose.test.yml
```diff
@@ -12,6 +12,10 @@ services:
     environment:
       DFE_TEACHERS_PAYMENT_SERVICE_DATABASE_USERNAME: postgres
       DFE_TEACHERS_PAYMENT_SERVICE_DATABASE_HOST: db
+      DFE_SIGN_IN_API_CLIENT_ID: teacherpayments
+      DFE_SIGN_IN_API_SECRET: secret
+      DFE_SIGN_IN_API_ENDPOINT: https://example.com
+      DFE_SIGN_IN_ROLE: Access to Teacher Payments
```
tekin: There's also an `.env.test` file where these environment variables can be added. That way the build won't fail in non-docker environments.

README.md
```diff
@@ -53,6 +53,10 @@ DFE_SIGN_IN_ISSUER=https://pp-oidc.signin.education.gov.uk:443
 DFE_SIGN_IN_REDIRECT_BASE_URL=https://localhost:3000
 DFE_SIGN_IN_IDENTIFIER=<paste identifier>
 DFE_SIGN_IN_SECRET=<paste secret>
+DFE_SIGN_IN_API_CLIENT_ID=teacherpayments
+DFE_SIGN_IN_API_SECRET=<paste secret>
+DFE_SIGN_IN_API_ENDPOINT=https://pp-api.signin.education.gov.uk
+DFE_SIGN_IN_ROLE=<paste role>
```
tekin: Given that we know we are dealing with a single "Access to Teacher Payments " role for admin access to the app, how do you feel about simply defining it as a constant somewhere appropriate in the application rather than having it as something that is configured with an environment variable?  Perhaps in the auth controller, e.g. `Admin::AuthController::DFE_SIGN_IN_ADMIN_ROLE_CODE`. Also, We should use the `"code"` as DfE Sign-in team have said the code stays constant across all environments. In our case, the code is `"teacher_payments_access"`.

config/initializers/dfe_signin.rb
```diff
@@ -0,0 +1,7 @@
+DFE_SIGN_IN_ROLE = ENV.fetch("DFE_SIGN_IN_ROLE")
```
tekin: As mentioned above, let's define this constant closer to where it is used, rather than as a global constant.

lib/dfe_signin.rb
```diff
@@ -0,0 +1,21 @@
+module DfeSignin
```
tekin: How about if the filename and case of this namespace matched the underscores used elsewhere, i.e. `dfe_sign_in.rb` and `DfeSignIn`?

lib/dfe_signin/user_access.rb
```diff
@@ -0,0 +1,40 @@
...
+    def call
+      request = Net::HTTP::Get.new(uri)
+      request["Authorization"] = "bearer #{DfeSignin::Utils.generate_jwt_token}"
+      request["Content-Type"] = "application/json"
+
+      response = Net::HTTP.start(uri.hostname, uri.port, use_ssl: true) { |http|
+        http.request(request)
+      }
+
+      raise ExternalServerError if response.code.eql?("500")
```
tekin: Might be helpful if we passed on the message returned in the response into the exception to give us more context about what went wrong. Assuming we get some body back with the response from sign-in.

lib/dfe_signin/utils.rb
```diff
@@ -0,0 +1,14 @@
+module DfeSignin
+  class Utils
```
tekin: How about if this was `module Utils`, rather than a class? That way there wouldn't be any need for the class method stuff.

spec/lib/dfe_signin/dfe_signin_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    before do
+      DfeSignin.configure do |config|
+        config.client_id = 123
+        config.secret = "sekrit"
+        config.base_url = "http://example.com"
+      end
+    end
+
+    after do
+      require Rails.root.join("config", "initializers", "dfe_signin")
```
tekin: I don't think this will do what you expect - `require` won't re-evaluate the config as it tracks the files that have been required before and won't require the same file twice. If you want to re-execute the code in a file you'll want to use `load` instead.

How about instead of overriding/resetting the config in this spec, we test that the config matches the values from the environment variables the initializer file sets them to?

spec/lib/dfe_signin/utils_spec.rb
```diff
@@ -0,0 +1,33 @@
+require "rails_helper"
+
+RSpec.describe DfeSignin::Utils do
+  before do
+    DfeSignin.configure do |config|
```
tekin: Again, can we rely on the config being already set and test against those values rather than re-configuring during the spec?

spec/models/dfe_signin/authorisation_spec.rb
```diff
@@ -0,0 +1,23 @@
+require "rails_helper"
+
+RSpec.describe DfeSignin::Authorisation do
+  subject { described_class.new(organisation_id: 123, user_id: 456) }
+
+  describe "authorised?" do
+    context "when user has access to the service" do
+      before { allow(subject).to receive(:roles).and_return(["Role 1", "Role 2"]) }
```
tekin: Stubbing this internal method makes this spec brittle to changes to the implementation. It also means the code that parses the response body to pull out the roles does not have unit test coverage. Can these specs be re-written without any stubbing?

spec/features/admin_sessions_spec.rb
```diff
@@ -1,6 +1,10 @@
 require "rails_helper"
 
 RSpec.feature "Admin sessions" do
+  before do
+    allow_any_instance_of(DfeSignin::Authorisation).to receive(:has_role?) { true }
```
tekin: By stubbing out the response of this class we lose integration coverage that would give us confidence the whole system works end-to-end. Ideally our feature specs would never stub our own code. If we need to stub anything, then it should be the underlying responses we get back from the auth calls. Have a look at `spec/support/omniauth.rb` to see where we're already doing that. It should be possible to modify the mock auth response there so that we get back an auth hash that looks like what we'd expect.

spec/requests/admin_spec.rb
```diff
@@ -17,23 +17,43 @@
...
         )
-        post admin_dfe_sign_in_path
-        follow_redirect!
       end
 
-      it "renders the admin page and sets a session" do
-        get admin_path
+      context "when the user is authorised to access the service" do
+        before do
+          allow_any_instance_of(DfeSignin::Authorisation).to receive(:has_role?) { true }
```
tekin: As in the feature-spec, we shouldn't be stubbing our own internal code. It should be possible to set the auth response using `OmniAuth.config.mock_auth[:dfe]=` so that it returns a response with the appropriate role in it.

spec/requests/admin_spec.rb
```diff
@@ -17,23 +17,43 @@
...
+          end
+        end
       end
 
-      context "and they sign out" do
-        it "unsets the session" do
-          delete admin_sign_out_path
+      context "when the user is not authorised to access the service" do
+        before do
+          allow_any_instance_of(DfeSignin::Authorisation).to receive(:has_role?) { false }
```
tekin: As above, let's use `OmniAuth.config.mock_auth[:dfe]=` instead to set a auth hash that doesn't have the expected role instead.

spec/requests/admin_timeout_spec.rb
```diff
@@ -8,6 +8,7 @@
       "provider" => "dfe",
       "info" => {"email" => "test-dfe-sign-in@host.tld"}
     )
+    allow_any_instance_of(DfeSignin::Authorisation).to receive(:has_role?) { true }
```
tekin: Update `OmniAuth.config.mock_auth[:dfe]` above to contain the role rather than stubbing this code.

app/models/dfe_signin/authorisation.rb
pezholio: Yeah, that's probably a good call actually. I only had this model when we had the old `authorised?` method, which was application-specific, but given `has_role?` can be used for any role, it's probably better to sit in the lib code

lib/dfe_signin.rb
pezholio: Yeah, that makes sense, I was treating Signin as one word, but DfE Sign-In is hyphonated, so probably makes sense to camel case them as seperate words

lib/dfe_signin/utils.rb
pezholio: Good plan - done! 👍

spec/features/admin_sessions_spec.rb
pezholio: I've fixed this in the sessions spec by using Webmock to stub the response from the DfE Signin API, but it seemed like overkill to repeat this work for every single login (I could have used hooks and done all the Webmock stub there, but I felt this was sacrificing readability/findability), so have kept the stubs in the other feature and request steps, as we're not _really_ testing the API response here, we just care if we're logged in or not

lib/dfe_sign_in/user_access.rb
```diff
@@ -0,0 +1,47 @@
...
+        request["Authorization"] = "bearer #{generate_jwt_token}"
+        request["Content-Type"] = "application/json"
+
+        response = Net::HTTP.start(uri.hostname, uri.port, use_ssl: true) { |http|
+          http.request(request)
+        }
+
+        raise ExternalServerError if response.code.eql?("500")
+
+        JSON.parse(response.body) if response.code.eql?("200")
```
tekin: If the response code isn't "500" or "200" this method will return/assign `nil`, which will be make it hard to debug and track down issues as we will have no trace of the what the response was from the API. Instead, perhaps we could raise in _all_ cases, except when the response is 200. Also, instead of just `raise ExternalServerError`, can we include the code and body as a message for the exception to aid debugging? For example, something like `raise ExternalServerError, "#{response.code}: #{response.body}"`

config/initializers/dfe_sign_in.rb
```diff
@@ -0,0 +1,5 @@
+DfeSignIn.configure do |config|
+  config.client_id = ENV.fetch("DFE_SIGN_IN_API_CLIENT_ID")
+  config.secret = ENV.fetch("DFE_SIGN_IN_API_SECRET")
+  config.base_url = ENV.fetch("DFE_SIGN_IN_API_ENDPOINT")
```
robbiepaul: Do you think we should be using `fetch` here? At the moment it's possible to run development without the need for DfE environment vars. This change would require all devs to at least put empty variables in their `.env`. Or perhaps we could reintroduce the shared `.env.development` file? (It will probably be reintroduced as part of #179 anyway)

spec/support/dfe_sign_in_helpers.rb
```diff
@@ -0,0 +1,26 @@
+module DfeSignInHelpers
+  # Stubs the DfE Sign-in OpenID response and the call to the DfE Sign-in API
+  # that we use to determine the roles that the user is authorised with.
+  def stub_dfe_sign_in_with_role(role_code)
```
robbiepaul: 👍

azure/app/parameters/development.template.json
```diff
@@ -94,6 +94,30 @@
...
+    },
+    "DFE_SIGN_IN_API_SECRET": {
+      "reference": {
+        "keyVault": {
+          "id": "/subscriptions/${subscriptionId}/resourceGroups/${keyVaultResourceGroupName}/providers/Microsoft.KeyVault/vaults/${keyVaultName}"
+        },
+        "secretName": "DfeSignInApiSecret"
+      }
+    },
+    "DFE_SIGN_IN_API_ENDPOINT": {
```
robbiepaul: I think we need to add these to the `appSettings` in `azure/app/template.json` as well. I'm happy to add them as a commit if you like

tekin: yep, please do!

config/initializers/dfe_sign_in.rb
tekin: A shared `.env.development` that covered everything needed to run the app locally will be hard because we need secrets in it, for example `DFE_SIGN_IN_API_SECRET` and we wouldn't want to include those in the repo.

I think it's reasonable to expect developers to manage/maintain the environment variables needed to run the application. Using `fetch` here makes it super-explicit what is missing as well, making it easier to know what needs adding.

robbiepaul: Yeah I was thinking a shared `.env.development` with empty values so you could at least run the app. I think we may need to amend the README as it makes these variables out to be optional

tekin: I'm not convinced a `.env.development` file in the repo with incomplete variables is a good idea.  Firstly it'll be impossible to update it without Git picking up the changes. What does a developer do in that situation? Ignore the changes when they create their commits? Add it to their local ignore file, but then remember to remove it when they need to make legit changes to it?

Also, yes the application will run, but it will be broken – when a developer attempts to authenticate with DfE Sign-in they will get a confusing error:

```DfeSignIn::ExternalServerError  – 403: {"success":false,"message":"invalid signature"}```

A developer would have to figure out that it's because of the empty environment variable, and then also figure out how to set the value without committing changes to `.env.development`.

I much prefer the developer seeing an early an explicit `key not found: "DFE_SIGN_IN_API_SECRET"` error on launch. OK, it means developers will need to set the environment variables after this is merged, but I think that one-time cost outweighs the cost of an incomplete developer env file in the repo.

If we do want to sign-post the environment variables beyond what is already in the `README`, one thing we could do is add an `.env.example` file for developers to copy and fill in the gaps when they first clone the repo.

Comment (robbiepaul) [reply config/initializers/dfe_sign_in.rb]: I think my contention mainly stems from this code: https://github.com/DFE-Digital/dfe-teachers-payment-service/blob/master/config/environments/development.rb#L69-L79

Previously it had been possible to run the app in development with OmniAuth in test/mock mode which I've just realised will no longer be possible after this change is in. We should probably remove that code at some point and update the docs, but happy to address that separately.

Comment (tekin) [reply config/initializers/dfe_sign_in.rb]: Ah I wasn't aware of that code, good spot. Yep, we should definitely remove that as we can no longer rely on just mocking the OpenID bit, we'd also have to mock out the API call that checks for the role.

**Comment by tekin** [comment]
@erbridge @mec @pezholio @titlescreen once this is merged you will need to set some extra environment variables to get the app running:

```
DFE_SIGN_IN_API_CLIENT_ID=teacherpayments
DFE_SIGN_IN_API_SECRET=<secret>
DFE_SIGN_IN_API_ENDPOINT=https://pp-api.signin.education.gov.uk
```

You can use the secret from the Heroku config (or give me a shout if you need help getting it)

PR #185, opened 2019-07-18
Add Rollbar access token and environment to ARM templates

robbiepaul:


robbiepaul:
[Set up Rollbar on Azure "development" and "production" environments](https://trello.com/c/TdYbo1Jz/434-set-up-rollbar-on-azure-development-and-production-environments)

erbridge:
This looks good, but I think we also need to add the reference to the `appSettings` for the app service.

robbiepaul:
Ah yes, thanks!

robbiepaul:
I've added those two references in `appSettings`

PR #188, opened 2019-07-18
Extract ClaimUpdate object for managing changes to an in-progress claim

tekin:
Some refactoring work that extracts a new class for handling updates to an in-progress claim, combining logic that was previously spread across the controller, `PageSequence`, and `TslrClaim` model itself.

It does also modify the behaviour slightly for users that are changing their answers:

- if a user changes the answer to their claim school, the subsequent employment_status and current_school answers are reset as those answers are coupled to claim_school and the question has effectively changed
- a user clicking "change" on the current school answer gets sent to the "still teaching" question, as that answer is coupled to the current school and should also be updated

More details in the individual commit messages.

### DIFF
--- a/app/helpers/claims_helper.rb
+++ b/app/helpers/claims_helper.rb
@@ -19,7 +19,7 @@ module ClaimsHelper
     [
       [t("tslr.questions.qts_award_year"), academic_years(claim.qts_award_year), "qts-year"],
       [t("tslr.questions.claim_school"), claim.claim_school_name, "claim-school"],
-      [t("tslr.questions.current_school"), claim.current_school_name, "current-school"],
+      [t("tslr.questions.current_school"), claim.current_school_name, "still-teaching"],
       [t("tslr.questions.subjects_taught"), subject_list(claim.subjects_taught), "subjects-taught"],
       [t("tslr.questions.mostly_teaching_eligible_subjects", subjects: subject_list(claim.subjects_taught)), (claim.mostly_teaching_eligible_subjects? ? "Yes" : "No"), "mostly-teaching-eligible-subjects"],
       [t("tslr.questions.student_loan_amount", claim_school_name: claim.claim_school_name), number_to_currency(claim.student_loan_repayment_amount), "student-loan-amount"],
--- a/app/models/claim_update.rb
+++ b/app/models/claim_update.rb
@@ -20,6 +20,7 @@ class ClaimUpdate
       claim.submit! && send_confirmation_email
     else
       claim.attributes = params
+      update_claim_school_dependent_attributes
       update_employment_status_dependent_attributes
       claim.save(context: context)
     end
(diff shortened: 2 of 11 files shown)


app/models/claim_update.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+# Performs an update on a claim in the context of a user performing an action
+# through the user interface. The `context` tells the object the page on which
+# the user is performing the action, enabling it to carry out actions and
+# validations on the claim that are appropriate to that context. For example,
+# Performing an update in the `"check-your-answers" context will attempt to
```
erbridge: Stray backtick.

app/helpers/claims_helper.rb
```diff
@@ -19,7 +19,7 @@ def claim_answers(claim)
     [
       [t("tslr.questions.qts_award_year"), academic_years(claim.qts_award_year), "qts-year"],
       [t("tslr.questions.claim_school"), claim.claim_school_name, "claim-school"],
-      [t("tslr.questions.current_school"), claim.current_school_name, "current-school"],
+      [t("tslr.questions.current_school"), claim.current_school_name, "still-teaching"],
```
erbridge: :+1:

spec/models/claim_update_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+    it "updates the claim and returns a truthy value" do
+      expect(claim_update.perform).to be_truthy
+      expect(claim.reload.full_name).to eq "Ada Lovelace"
+    end
+  end
+
+  context "with parameters that are not valid for the context" do
+    let(:claim) { create(:tslr_claim) }
+    let(:context) { "full-name" }
+    let(:params) { {full_name: nil} }
```
erbridge: I would expect this to be testing unrelated parameter changes, e.g. `{postcode: "TE5 T42"}`. Perhaps it could be better named, and we could reject changes to unrelated fields? See https://trello.com/c/EvX3BYxI/219-only-allow-the-expected-values-for-a-given-page

tekin: Ah, perhaps the spec isn't clear then. It's meant to be verifying that validation that is specific to the the context gets fired, i.e. if they're updating on the "full-name" page, we'd expect the validation that requires the full-name be present in that context fires. Is there a way I could make that clearer? To be fair I probably shouldn't be using the full-name context anyway as it will disappear shortly when Verify is enabled...

Regarding that other Trello card, I think we should tackle that as a separate piece of work if we decide to do it, as it'll require some more major changes to define and honour the parameters that are editable in each context.

app/models/claim_update.rb
tekin: ah thanks, fixed!

spec/models/claim_update_spec.rb
erbridge: > Ah, perhaps the spec isn't clear then...

I think it's just that word "valid" again. Maybe this would be clearer rephrased to something like: "without required parameters for the context"?

> Regarding that other Trello card...

Yes agreed. I just thought I'd bring it up as some context on what my line of thinking was.

tekin: OK, I reworded the spec as suggested. I've also switched to use a context that isn't about to be removed so we won't have to update this spec later.

PR #189, opened 2019-07-18
Change the school eligibility checkers to allow special schools with unspecified phases

erbridge:
Most special schools have a "Not Applicable" education phase, so this adds support for those schools.

We use the school type, rather than the school type group for determining if a school is a special school because e.g. free special schools are grouped under free schools, and so would be left out. There maye be other examples.

### DIFF
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -82,6 +82,17 @@ class School < ApplicationRecord
     institution_funded_by_other_government_department: 56,
   }.freeze
 
+  SPECIAL_SCHOOL_TYPES = %w[
+    community_special_school
+    non_maintained_special_school
+    other_independent_special_school
+    foundation_special_school
+    special_post_16_institutions
+    academy_special_sponsor_led
+    free_school_special
+    academy_special_converter
+  ].freeze
+
   enum phase: PHASES
   enum school_type_group: SCHOOL_TYPE_GROUPS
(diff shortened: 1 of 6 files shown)


erbridge:
[Change the school eligibility checkers to allow SEN schools with any phase](https://trello.com/c/gbUiELuD/446-change-the-school-eligibility-checkers-to-allow-sen-schools-with-any-phase)

erbridge:
We have not. How do I get the dump of eligible schools and who do I send it to? Perhaps someone could help me do that?

tekin:
I'd use a crude script in the Rails console. Something like:

```ruby

**Comment by tekin** [comment]
I'd use a crude script in the Rails console. Something like:

```ruby
Rails.logger.level = Logger::WARN
School.all.each do |school|
  puts [school.urn, school.name].join(" - ") if Tslr::SchoolEligibility.new(school).check
end;nil
```

As to the person, probably start with Chris Cuff or George Poulton.

robbiepaul:
This is what I used previously when exporting eligible schools:
```
require "csv"
CSV.open("tmp/schools.csv", "wb") do |csv|
  csv << School.attribute_names
  School.includes(:local_authority).select(&:eligible_for_tslr?).each do |school|
    csv << school.attributes.values
  end
end
```

PR #190, opened 2019-07-18
Turn QTS academic year question into radio buttons

robbiepaul:
# Screenshot

Teachers-claim-back-your-student-loan-repayments-2019-07-18-15-07-48

- Change QTS award year to radio buttons - radio buttons are considered more accessible than selects
- Add an ineligible option to give ineligible claimants a way to fail at this step


app/models/tslr_claim.rb
```diff
@@ -31,7 +31,7 @@ class TslrClaim < ApplicationRecord
   validates :claim_school,                      on: [:"claim-school", :submit], presence: {message: "Select a school from the list"}
   validates :current_school,                    on: [:"current-school", :submit], presence: {message: "Select a school from the list"}
 
-  validates :qts_award_year,                    on: [:"qts-year", :submit], inclusion: {in: VALID_QTS_YEARS, message: "Select the academic year you were awarded qualified teacher status"}
+  validates :qts_award_year,                    on: [:"qts-year", :submit], inclusion: {in: ClaimsController.helpers.options_for_qts_award_year.map(&:last), message: "Select the academic year you were awarded qualified teacher status"}
```
erbridge: Is there a way to do this without calling controller code from the model?

app/views/claims/qts_year.html.erb
```diff
@@ -12,8 +12,14 @@
...
-        <%= form.select :qts_award_year, options_for_qts_award_year, { include_blank: true }, { class: css_classes_for_select(current_claim, :qts_award_year) } %>
-
+        <div class="govuk-radios">
+          <%= form.collection_radio_buttons :qts_award_year, options_for_qts_award_year, :last, :first do |b| %>
+            <div class="govuk-radios__item">
+              <%= b.radio_button class: "govuk-radios__input" %>
+              <%= b.label class: "govuk-label govuk-radios__label" %>
+            </div>
+          <% end %>
+        </div>
```
erbridge: Does this need the form helper magic that `form.select` had for handling errors?

robbiepaul: I don't think so because the radios/checkboxes don't get any extra styling in an error state. https://design-system.service.gov.uk/components/error-message/

robbiepaul: Currently looks like: 
errors

erbridge: Ok, nice!

app/models/tslr_claim.rb
robbiepaul: I've added a commit that extracts this code out to a shared `QtsYears` model. I'm not sure if it's a conventional thing to do in Rails, I'm open to suggestions

erbridge: I also don't know if it's conventional, but I like it 🤷‍♂.

app/models/qts_years.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+# Represents the year a teacher can qualify in
+
+class QtsYears
```
tekin: Given that we're not going to be instantiating a `QtsYear` instance and this is a collection of class methods, it probably makes more sense as a module and not a class. Also given that it is specific to the student loans policy, how do you feel about namespacing it under the `TslrClaim` model?

app/views/claims/qts_year.html.erb
```diff
@@ -12,8 +12,14 @@
         </span>
 
         <%= errors_tag current_claim, :qts_award_year %>
-        <%= form.select :qts_award_year, options_for_qts_award_year, { include_blank: true }, { class: css_classes_for_select(current_claim, :qts_award_year) } %>
-
+        <div class="govuk-radios">
```
tekin: The `shared/error_summary` render call above will want `errored_field_id_overrides` set so that the error summary can correctly link the error message to the first radio in the list. See the [`mostly_teaching_eligible_subjects.html.erb`](https://github.com/DFE-Digital/dfe-teachers-payment-service/blob/6e519a866117902775efee57a6245a53c152ec26/app/views/claims/mostly_teaching_eligible_subjects.html.erb#L3) partial for an example of how to do this

spec/models/qts_years_spec.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe QtsYears do
+  describe ".first_eligible_year" do
+    it "returns the first year of the eligible QTS years" do
+      stub_const("QtsYears::ELIGIBLE_YEARS", ["2018-2019", "2019-2020"])
```
tekin: On balance I don't think the stubbing in these specs is very helpful as it locks the spec to the implementation of the method and the underlying data structures, which makes the code harder to refactor. In reality, the QTS years for this policy are set in stone, so I think we can be explicit about that and simply say `expect(QtsYears.first_eligible_year).to eq "2013"`.

app/models/tslr_claim.rb
tekin: The pattern we've followed elsewhere has been to use an `enum` column, e.g. `employment_status` and the [forthcoming student loan plan questions](https://github.com/DFE-Digital/dfe-teachers-payment-service/compare/335-capture-student-loan-repayment-plan). If we wanted to keep things really simple, we might look to turn this column into an enum and then hard-code the form fields based on the available options. The actual options aren't going to change so there is less of a need for the options, etc to be dynamically generated using code.

robbiepaul: I've changed the field to be an enum. The migration may not work and will probably need a `db:reset` for it to work properly, not sure how we're going to do that on staging (maybe just manually destroy the database and start fresh)?

app/models/tslr_claim.rb
```diff
@@ -25,13 +15,24 @@ class TslrClaim < ApplicationRecord
     no_school: 2,
   }, _prefix: :employed_at
 
+  enum qts_award_year: {
+    "before_2013": 0,
+    "2013-2014": 1,
```
tekin: How do you feel about using underscores instead of dashes so they can be symbols and the magic enum boolean methods are available?

db/migrate/20190722101055_change_qts_year_to_integer.rb
```diff
@@ -0,0 +1,9 @@
+class ChangeQtsYearToInteger < ActiveRecord::Migration[5.2]
+  def up
+    change_column :tslr_claims, :qts_award_year, "integer USING CAST(qts_award_year AS integer)"
```
tekin: PostgreSQL will complain that it can't cast a string to an integer. We can use a `CASE` to convert the string values to their corresponding enum values though:

```ruby
class ChangeQtsYearToInteger < ActiveRecord::Migration[5.2]
  def up
    change_column :tslr_claims, :qts_award_year, "integer USING CASE qts_award_year
        WHEN '2013-2014' THEN 1
        WHEN '2014-2015' THEN 2
        WHEN '2015-2016' THEN 3
        WHEN '2016-2017' THEN 4
        WHEN '2017-2018' THEN 5
        WHEN '2018-2019' THEN 6
        WHEN '2019-2020' THEN 7
      END"
  end

  def down
    change_column :tslr_claims, :qts_award_year, "varchar USING CASE qts_award_year
        WHEN 1 THEN '2013-2014'
        WHEN 2 THEN '2014-2015'
        WHEN 3 THEN '2015-2016'
        WHEN 4 THEN '2016-2017'
        WHEN 5 THEN '2017-2018'
        WHEN 6 THEN '2018-2019'
        WHEN 7 THEN '2019-2020'
      END"
  end
end
```

tekin:
Just checking, have we considered how this change impacts the CSV export?

db/migrate/20190722101055_change_qts_year_to_integer.rb
robbiepaul: Thanks for this 👍

app/models/tslr_claim.rb
robbiepaul: I don't have a strong opinion about it. My thinking was that we're probably not going to use those magic enum methods for eligible years and less code would need updating if they remained in this format. Happy to change it though

tekin: My gut is telling me that underscores are probably safer if there isn't an explicit reason to prefer a dash.

robbiepaul:
I just checked the CSV import, it looks like it's using the string value:

Screenshot 2019-07-22 at 14 17 51

app/models/tslr_claim.rb
robbiepaul: I've added a commit that replaces the dashes with underscores

db/migrate/20190722101055_change_qts_year_to_integer.rb
pezholio: This is really cool, I had no idea it was A Thing in Postgres

tekin: yeah, it's pretty neat, although wouldn't recommend it on a large table!

PR #192, opened 2019-07-22
Set up more alerts on CIP

erbridge:
Also, use resource IDs rather than names for `dependsOn`, for added clarity and reduced chance of clashes.


erbridge:
[Set up Monitor on CIP](https://trello.com/c/r9k06lDD/252-set-up-monitor-on-cip)

PR #193, opened 2019-07-23
Claimant answers questions to determine their student loan plan

tekin:
A claimant can now answer a series of questions that determines whether they are still paying a student loan, and if so, the student loan repayment plan they are on.

Things still outstanding to do:

- Play back the plan on the check-your-answers screen
- The flow that will allow a user to go back "change" their answers to the student loan question

Both those things will be tackled as a separate PR building on this work.

 https://trello.com/c/ioo1e1L1/335-capture-the-student-loan-repayment-plan

### DIFF
--- a/app/models/claim_update.rb
+++ b/app/models/claim_update.rb
@@ -22,6 +22,7 @@ class ClaimUpdate
       claim.attributes = params
       update_claim_school_dependent_attributes
       update_employment_status_dependent_attributes
+      determine_student_loan_plan
       claim.save(context: context)
     end
   end
@@ -46,5 +47,9 @@ class ClaimUpdate
     if claim.employment_status_changed?
       claim.current_school = claim.employed_at_claim_school? ? claim.claim_school : nil
     end
+  end
+
+  def determine_student_loan_plan
+    claim.student_loan_plan = StudentLoans.determine_plan(claim.student_loan_country, claim.student_loan_start_date)
   end
 end
--- a/app/models/claim_update.rb
+++ b/app/models/claim_update.rb
@@ -50,6 +50,10 @@ class ClaimUpdate
   end
 
   def determine_student_loan_plan
-    claim.student_loan_plan = StudentLoans.determine_plan(claim.student_loan_country, claim.student_loan_start_date)
+    claim.student_loan_plan = if claim.has_student_loan?
+      StudentLoans.determine_plan(claim.student_loan_country, claim.student_loan_start_date)
+    else
+      TslrClaim::NO_STUDENT_LOAN
+    end
   end
 end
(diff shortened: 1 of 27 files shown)


app/presenters/tslr_claim_csv_row.rb
```diff
@@ -28,6 +28,10 @@ def student_loan_repayment_amount
     "£#{model.student_loan_repayment_amount}"
   end
 
+  def student_loan_repayment_pan
```
robbiepaul: Typo: `plan`

tekin: ta, fixed.

app/models/tslr_claim.rb
```diff
@@ -62,6 +67,11 @@ class TslrClaim < ApplicationRecord
   validates :national_insurance_number, on: [:"national-insurance-number", :submit], presence: {message: "Enter your National Insurance number"}
   validate  :ni_number_is_correct_format
 
+  validates :has_student_loan,                  on: [:"student-loan", :submit], inclusion: {in: [true, false], message: "Select yes if you have a student loan"}
+  validates :student_loan_country,              on: [:"student-loan-country", :submit], presence: {message: "Select the country in which you first applied for your student loan"}
+  validates :student_loan_courses,              on: [:"student-loan-how-many-courses", :submit], presence: {message: "Select the number of higher education courses you have studied"}
```
robbiepaul: I think there needs to be some conditional validation here. When I select 'Scotland' then try to submit the form I get the following errors:

Screenshot 2019-07-23 at 11 57 57

lib/student_loans.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module StudentLoans
+  COUNTRIES = [
+    ENGLAND = "england",
```
robbiepaul: Nice, I didn't know this syntax was possible 👍

app/models/tslr_claim.rb
tekin: Yep, noticed that too, just pushed up a commit that resolves the issue

app/models/claim_update.rb
```diff
@@ -50,6 +50,10 @@ def update_employment_status_dependent_attributes
   end
 
   def determine_student_loan_plan
-    claim.student_loan_plan = StudentLoans.determine_plan(claim.student_loan_country, claim.student_loan_start_date)
+    claim.student_loan_plan = if claim.has_student_loan?
+      StudentLoans.determine_plan(claim.student_loan_country, claim.student_loan_start_date)
+    else
+      TslrClaim::NO_STUDENT_LOAN
```
robbiepaul: 👍

PR #198, opened 2019-07-24
Set up HTTPS for custom domains on Azure

erbridge:
Certificates need to be added to the key vault where they will be imported and assigned to the app service. The key vault needs additional permissions to allow that process.

### DIFF
--- a/azure/secrets/template.json
+++ b/azure/secrets/template.json
@@ -11,6 +11,8 @@
     }
   },
   "variables": {
+    "principal_Microsoft_Azure_App_Service": "a6621090-e704-45ec-b65f-50257f9d4dcd",
+
     "keyVaultName": "[concat(parameters('resourceNamePrefix'), '-kv')]"
   },
   "resources": [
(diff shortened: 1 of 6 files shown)


erbridge:
[Set up domains for Production and Development](https://trello.com/c/tJXItPli/341-set-up-domains-for-production-and-development)

azure/secrets/template.json
```diff
@@ -11,6 +11,8 @@
     }
   },
   "variables": {
+    "principal_Microsoft_Azure_App_Service": "a6621090-e704-45ec-b65f-50257f9d4dcd",
```
robbiepaul: Just wanted to double check you intended to hardcode this UUID

erbridge: I did. The resource principals are constant for all environments inside a specific tenant. That is, every DfE subscription shares the same ID.

I've used `_`s as a delimeter and in place of spaces in the name of the principal, since the easiest way to look up these IDs is by manually adding them to a Key Vault's access policies and looking at the template export.

PR #199, opened 2019-07-24
Student loan answers can be changed

tekin:
The answers to the student loan questions now appear on the check-your-answers page and can now be "changed". Changes to certain questions result in dependent questions being reset such that the user has to re-answer them. More details in the individual commits.

### DIFF
--- a/app/models/claim_update.rb
+++ b/app/models/claim_update.rb
@@ -20,11 +20,11 @@ class ClaimUpdate
       claim.submit! && send_confirmation_email
     else
       claim.attributes = params
-      update_claim_school_dependent_attributes
-      update_employment_status_dependent_attributes
-      update_has_student_loan_dependent_attribute
-      update_student_loan_country_dependent_attribute
-      update_student_loan_courses_dependent_attribute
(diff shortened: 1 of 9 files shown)


app/models/claim_update.rb
```diff
@@ -37,13 +54,15 @@ def send_confirmation_email
...
+  def reset_dependent_answers
+    DEPENDENT_ANSWERS.each do |attribute_name, dependent_attribute_name|
+      if claim.changed.include?(attribute_name)
+        claim.attributes = {dependent_attribute_name => nil}
+      end
     end
   end
 
-  def update_employment_status_dependent_attributes
+  def determine_inferred_current_school
```
erbridge: Can this just be `infer_current_school`?

tekin: 👍

tekin:
Regarding reporting the plan. there is a separate card in the design backlog to look at "hidden" information. Mark pointed out in a conversation yesterday that the design system states that check-your-answers page should only ever play back questions and answers that the user has been asked/answered. Anything else should be exposed to the user elsewhere in the journey.

PR #207, opened 2019-07-29
Make school search JS more generic

tekin:
In preparation for adding support for the Maths and Physics policy, we're slowly making existing code less coupled to TSLR specifically.

### DIFF
--- a/app/views/claims/_school_search.html.erb
+++ b/app/views/claims/_school_search.html.erb
@@ -1,8 +1,8 @@
 <% if @schools.blank? %>
 
-  <%= render("shared/error_summary", instance: current_claim) if current_claim.errors.any? %>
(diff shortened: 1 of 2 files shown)


app/views/claims/_school_search.html.erb
```diff
@@ -26,7 +26,7 @@
       <div id="school-search-container">
 
         <%= errors_tag current_claim, :school_search %>
-        <%= text_field_tag :school_search, params[:school_search], id: :tslr_claim_school_search, class: css_classes_for_input(current_claim, :school_search), value: school_search_value %>
+        <%= text_field_tag :school_search, params[:school_search], id: :school_search, class: css_classes_for_input(current_claim, :school_search), value: school_search_value %>
```
mec: Could we use `errored_field_id_overrides on`render("shared/error_summary")` to set the error message anchor to `school_serach` to make the error message link to the input field?

tekin: Good spot!

PR #211, opened 2019-07-30
Create a script to create and store VSP certificate requests

erbridge:
We put them in the secrets key vaults. There is a manual step to do the enrollment with Verify using their web interface.

There is no "test" environment for VSP certificates here, omitted until we have a use case for Verify that wouldn't allow reusing "development" certificates.

This script was already used to generate the certificate requests that we submitted for dev and prod environments.

### DIFF
--- a/bin/request-vsp-certs
+++ b/bin/request-vsp-certs
@@ -103,11 +103,36 @@ function store_key_and_csr_and_challenge_phrase {
     --value "$CHALLENGE_PHRASE"
 }
 
-SIGNING_FILE_PREFIX="TeacherPayments${ENVIRONMENT_SHORT_NAME}VspSamlSigning${VERSION_NUMBER}"
-SIGNING_CSR_COMMON_NAME="Teacher Payments Service $ENVIRONMENT_SHORT_NAME VSP SAML Signing $VERSION_NUMBER"
+function request_cert {
+  local TYPE=$1
+  local FILE_PREFIX=$2
+  local CSR_COMMON_NAME=$3
 
-ENCRYPTION_FILE_PREFIX="TeacherPayments${ENVIRONMENT_SHORT_NAME}VspSamlEncryption${VERSION_NUMBER}"
-ENCRYPTION_CSR_COMMON_NAME="Teacher Payments Service $ENVIRONMENT_SHORT_NAME VSP SAML Encryption $VERSION_NUMBER"
+  echo "Fetching or generating $TYPE key and CSR..."
+  fetch_or_generate_key_and_csr "$FILE_PREFIX" "$CSR_COMMON_NAME"
+
+  if [ "$KEY_EXISTED" ]; then
+    echo "A $TYPE key already existed for $ENVIRONMENT_NAME version $VERSION_NUMBER."
+    read -rp "  Hit return to continue with the pre-existing key, or CTRL+C to stop."
+  fi
+
+  if [ "$CSR_EXISTED" ]; then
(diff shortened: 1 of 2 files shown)


erbridge:
[Get "Integration" (i.e. staging) Verify certificates for running a VSP](https://trello.com/c/FYRKfGar/345-get-integration-ie-staging-verify-certificates-for-running-a-vsp)

tekin:
It makes me happy to see this put together, one less manual, multi-stag, error-prone process to worry about! 👏🏽

My bash isn't strong enough to give this a detailed review, but a couple of general things:

- Can we include some documentation in the repo? There's already a `docs/govuk-verify.md` file.
- Will this script overwrite an existing key (matching version number) without warning? I guess there isn't a risk of losing the data as Azure retain versions of key values, but it could result in braking things in hard to trace ways.

erbridge:
> It makes me happy to see this put together, one less manual, multi-stag, error-prone process to worry about! 👏🏽

My thoughts exactly.

> Can we include some documentation in the repo?

I have written something up on https://dfedigital.atlassian.net/wiki/spaces/TP/pages/1116110992/Verify+Service+Provider. A step of the process is to "enrol" the certificate, but the URLs to do that aren't publicly advertized, so it didn't feel right to include them here. I'll see about adding something to the in-repo docs, though.

> Will this script overwrite an existing key (matching version number) without warning?

Local files: yes. On the Key Vault, we keep a version history. Are you suggesting that we query the vault to check if we have a version with that number already and ask for confirmation before adding?

tekin:
> I have written something up on https://dfedigital.atlassian.net/wiki/spaces/TP/pages/1116110992/Verify+Service+Provider. A step of the process is to "enrol" the certificate, but the URLs to do that aren't publicly advertized, so it didn't feel right to include them here. I'll see about adding something to the in-repo docs, though.

Ah great. Weird that this isn't public. Is there a reason? We'll want to evolve this documentation so it forms a single to include the rest of the steps, i.e. what you actually do with the requests once you have them and also what you do once we get the certs back, but that can be done separately to this PR.

> Local files: yes. On the Key Vault, we keep a version history. Are you suggesting that we query the vault to check if we have a version with that number already and ask for confirmation before adding?

The scenario I think we should try and protect against is someone generating a new set of keys/certs and accidentally overwriting an existing version without realising and it not coming to light until later after a deploy causes the new values to be picked up and potentially breaking the system. One way to solve it might be to update this script to check and warn. Another might be to update the document so that it has big angry text warning the person doing the generating that they should be careful and ensure they use the next version number as they could accidentally overwrite an existing key. Happy for you to decide which is most appropriate.

erbridge:
I have added some mostly duplicated docs to the repo. Probably we can delete that Confluence page now. I have left documenting what we do with them when we get the certificates back to documenting as we do them. I imagine another script coming into existence to go alonside that.

I don't know if there's a reason for the enrollment URLs being private, but they sent us a private Google Doc when we gained access, which is the only place it's documented.

I'll think about how best to prevent accidental overwriting.

erbridge:
I've updated the script to reuse keys and CSRs for that version if they exist, and prompt the user to confirm that that's acceptable.

erbridge:
@robbiepaul Should be able to ask for access if you're approved by Verify.

erbridge:
@tekin I did consider cleaning them up, but I wanted to leave them in case something went wrong with storing them in the Key Vault, so deletion is a considered manual step.

PR #212, opened 2019-07-31
Introduce student loans eligibility model and port the QTS award year attribute

mec:
This is the first in as series of changes to move the eligibility of the student loans claim into it's own class. This work will invlove a lot of big changes so we want to keep the PRs as small and concise as we can to lessen the load on the reviewer.

This first PR introduces the `StudentLoans` namespace and the `Eligibility` class.

We then move on to port the `qts_award_year` from `TslrClaim` to the new eligitibility class. We will have to used some temporary hacks to keep this work green during the move, using some delegates, which will be removed once this work completes.

Along the way there have been some tightening up of tests and fixes as we encountered them.

https://trello.com/c/yt1gsap9

### DIFF
--- a/app/models/student_loans/eligibility.rb
+++ b/app/models/student_loans/eligibility.rb
@@ -16,5 +16,15 @@ module StudentLoans
     }, _prefix: :awarded_qualified_status
 
     validates :qts_award_year, on: [:"qts-year", :submit], presence: {message: "Select the academic year you were awarded qualified teacher status"}
+
+    def ineligible?
+      ineligible_qts_award_year?
+    end
+
+    private
+
+    def ineligible_qts_award_year?
+      awarded_qualified_status_before_2013?
+    end
   end
 end
--- a/app/models/student_loans/eligibility.rb
+++ b/app/models/student_loans/eligibility.rb
@@ -21,6 +21,12 @@ module StudentLoans
       ineligible_qts_award_year?
     end
 
+    def ineligibility_reason
+      [
+        :ineligible_qts_award_year,
+      ].find { |eligibility_check| send("#{eligibility_check}?") }
+    end
+
     private
 
     def ineligible_qts_award_year?
--- a/app/models/student_loans/eligibility.rb
+++ b/app/models/student_loans/eligibility.rb
@@ -27,8 +27,6 @@ module StudentLoans
       ].find { |eligibility_check| send("#{eligibility_check}?") }
     end
 
-    private
-
     def ineligible_qts_award_year?
       awarded_qualified_status_before_2013?
     end
--- a/app/models/student_loans/eligibility.rb
+++ b/app/models/student_loans/eligibility.rb
@@ -3,5 +3,18 @@
 module StudentLoans
   class Eligibility < ApplicationRecord
     self.table_name = "student_loans_eligibilities"
+
+    enum qts_award_year: {
+      "before_2013": 0,
+      "2013_2014": 1,
+      "2014_2015": 2,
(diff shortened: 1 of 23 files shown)


app/models/student_loans/eligibility.rb
```diff
@@ -0,0 +1,34 @@
...
+      "2013_2014": 1,
+      "2014_2015": 2,
+      "2015_2016": 3,
+      "2016_2017": 4,
+      "2017_2018": 5,
+      "2018_2019": 6,
+      "2019_2020": 7,
+    }, _prefix: :awarded_qualified_status
+
+    validates :qts_award_year, on: [:"qts-year"], presence: {message: "Select the academic year you were awarded qualified teacher status"}
```
erbridge: Do we no longer need to validate on `:submit`?

erbridge: Possibly obvious Rails question, but when does this validator get called with a context?

app/views/claims/qts_year.html.erb
```diff
@@ -1,26 +1,28 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <%= render("shared/error_summary", instance: current_claim, errored_field_id_overrides: {qts_award_year: "tslr_claim_qts_award_year_before_2013"}) if current_claim.errors.any? %>
+    <%= render("shared/error_summary", instance: current_claim, errored_field_id_overrides: {:"eligibility.qts_award_year" => "tslr_claim_eligibility_attributes_qts_award_year_before_2013"}) if current_claim.errors.any? %>
 
     <%= form_for current_claim, url: claim_path do |form| %>
       <%= form_group_tag current_claim do %>
+        <%= form.fields_for :eligibility, include_id: false do |fields| %>
         <h1 class="govuk-label-wrapper">
```
erbridge: Indentation nitpick: shouldn't this all be indented another level now?

app/models/student_loans/eligibility.rb
tekin: Ah, good spot, we should definitely be validating the presence on `:submit`!

tekin: The `ClaimUpdate` class [updates the claim in current context](https://github.com/DFE-Digital/dfe-teachers-payment-service/blob/master/app/models/claim_update.rb#L43), which gets called by the [controller](https://github.com/DFE-Digital/dfe-teachers-payment-service/blob/master/app/controllers/claims_controller.rb#L44)

app/views/claims/qts_year.html.erb
tekin: 👍🏽

mec: 👍

app/models/student_loans/eligibility.rb
erbridge: Ah ok, and that context gets passed through to child models?

tekin: Yep, that's right.

mec: To the best of my knowledge custom validators have to be called explicitally with either `valid?(:context)`, 'invalid?(:context)` or `save(:context)`.

That makes our calls on line 43 in `claim_update.rb` and line 121 in `tslr_claim.rb` models.

https://guides.rubyonrails.org/active_record_validations.html#on

Looking at line 121 in the tslr claim model makes me think you are right about needing to validtate on `:submit`, feels like something we should have test coverage for but I am struggling to find any, I'll pick this up with @tekin.

PR #213, opened 2019-07-31
Create an App Service Key Vaults and enable soft delete

erbridge:
This will be used for SSL certificates, and is separate from the main vault to avoid exposing secrets that don't need to be exposed.

Turning on soft delete means that we have a 90 day grace window to recover unpurged secrets or key vaults that have been deleted.


erbridge:
After talking to Craig, we don't need to do this.

PR #215, opened 2019-07-31
Set up a Verify Service Provider (VSP) on staging

robbiepaul:
This PR adds the config to set up a GOV.UK Verify VSP app service on our `development` infrastructure. The VSP Docker image is built, tagged and pushed when changes are made to the master branch on its repo: https://github.com/DFE-Digital/teacher-payments-service-verify

## Private keys

The private keys used by the VSP need to be in PKCS8 format and base64 encoded. We add a step to the script that requests certs to also save a base64 encoded version.

## Testing

We can use this VSP to test the Verify integration environment: https://www.docs.verify.service.gov.uk/test-in-integration/testing/set-up-tests

### DIFF
--- a/azure/app/parameters/development.template.json
+++ b/azure/app/parameters/development.template.json
@@ -162,7 +162,7 @@
         "keyVault": {
           "id": "${keyVaultId}"
         },
-        "secretName": "VspSamlSigningKey"
+        "secretName": "TeacherPaymentsDevVspSamlSigning2KeyBase64"
       }
     },
     "VSP.SAML_PRIMARY_ENCRYPTION_KEY": {
@@ -170,7 +170,7 @@
         "keyVault": {
           "id": "${keyVaultId}"
         },
-        "secretName": "VspSamlPrimaryEncryptionKey"
+        "secretName": "TeacherPaymentsDevVspSamlEncryption2KeyBase64"
       }
     }
   }
--- a/azure/app/parameters/development.template.json
+++ b/azure/app/parameters/development.template.json
@@ -151,6 +151,9 @@
     "ROLLBAR_ENV": {
       "value": "development"
     },
+    "GOVUK_VERIFY_ENABLED": {
+      "value": "1"
+    },
     "VSP.VERIFY_ENVIRONMENT": {
       "value": "INTEGRATION"
     },
--- a/azure/app/parameters/development.template.json
+++ b/azure/app/parameters/development.template.json
@@ -11,6 +11,9 @@
     "appServiceCertificateSecretName": {
       "value": "sslCertificate-wildcard-additional-teaching-payment-education-gov-uk"
     },
+    "vspAppServiceDockerImage": {
+      "value": "dfedigital/teacher-payments-service-verify:${vspDockerImageTag}"
+    },
     "databaseName": {
       "value": "development"
     },
@@ -147,6 +150,25 @@
     },
     "ROLLBAR_ENV": {
       "value": "development"
+    },
+    "VSP.VERIFY_ENVIRONMENT": {
+      "value": "INTEGRATION"
+    },
+    "VSP.SAML_SIGNING_KEY": {
+      "reference": {
+        "keyVault": {
+          "id": "${keyVaultId}"
+        },
+        "secretName": "VspSamlSigningKey"
+      }
+    },
+    "VSP.SAML_PRIMARY_ENCRYPTION_KEY": {
+      "reference": {
+        "keyVault": {
+          "id": "${keyVaultId}"
+        },
+        "secretName": "VspSamlPrimaryEncryptionKey"
+      }
     }
   }
 }
--- a/azure/app/parameters/production.template.json
+++ b/azure/app/parameters/production.template.json
@@ -145,6 +145,9 @@
     "ROLLBAR_ENV": {
       "value": "production"
     },
+    "GOVUK_VERIFY_ENABLED": {
+      "value": ""
+    },
     "VSP.VERIFY_ENVIRONMENT": {
       "value": "PRODUCTION"
     },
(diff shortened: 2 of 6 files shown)


robbiepaul:

robbiepaul:
I just tried testing this on the temp app and got a 500 error: https://rollbar.com/dxw/teachers-payment-service/items/27

The VSP is up and responding to the healthcheck URL. Will investigate more tomorrow

azure/app/parameters/production.template.json
```diff
@@ -141,6 +141,28 @@
...
+        "secretName": "VspSamlSigningKey"
+      }
+    },
+    "VSP.SAML_PRIMARY_ENCRYPTION_KEY": {
+      "reference": {
+        "keyVault": {
+          "id": "${keyVaultId}"
+        },
+        "secretName": "VspSamlPrimaryEncryptionKey"
+      }
```
erbridge: If these secrets are missing from the Key Vault, the deployment will fail. I suspect, if we want to be able to push this to `production` after merge, we need to set these to an empty string for now.

azure/app/parameters/development.template.json
```diff
@@ -8,6 +8,9 @@
     "appServiceHostName": {
       "value": "development.additional-teaching-payment.education.gov.uk"
     },
+    "vspAppServiceDockerImage": {
+      "value": "dfedigital/teacher-payments-service-verify:${vspDockerImageTag}"
+    },
```
erbridge: This needs to be specified for `production` and `test` as well, or they will be undeployable.

azure/app/parameters/production.template.json
erbridge: Same applies to `test`.

azure/app/template.json
```diff
@@ -183,6 +205,59 @@
...
+          "httpLoggingEnabled": true,
+          "linuxFxVersion": "[variables('vspAppServiceRuntimeStack')]",
+          "appSettings": [
+            {
+              "name": "VERIFY_ENVIRONMENT",
+              "value": "[parameters('VSP.VERIFY_ENVIRONMENT')]"
+            },
+            {
+              "name": "SERVICE_ENTITY_IDS",
+              "value": "[string(createArray(parameters('appServiceHostName')))]"
```
erbridge: It's worth pointing out that this means that the `tmp` resource groups won't actually work with the VSP because they share `appServiceHostName` with their non-`tmp` pair. Their utility is mostly in testing the infrastructure deployments, but if you want to check it actually works, you'll probably need to deploy to a real environment.

bin/request-vsp-certs
```diff
@@ -90,6 +117,13 @@ function store_key_and_csr_and_challenge_phrase {
       --file "${FILE_PREFIX}.key"
   fi
 
+  if ! [ "$KEY_EXISTED" ] || ! [ "$BASE64_KEY_EXISTED" ]; then
```
erbridge: Probably worth writing down why we made this decision somewhere (in the commit message?). "If we generated the key, we always want a new version of the base64 encoded version, regardless of whether we already had a base64 encoded key or not."

azure/app/parameters/test.template.json
```diff
@@ -144,6 +144,28 @@
...
+        "secretName": "VspSamlSigningKey"
+      }
+    },
+    "VSP.SAML_PRIMARY_ENCRYPTION_KEY": {
+      "reference": {
+        "keyVault": {
+          "id": "${keyVaultId}"
+        },
+        "secretName": "VspSamlPrimaryEncryptionKey"
+      }
```
erbridge: Let's make sure we record that we need to change these keys over when the new certs are approved (a Trello card?). Does updating the versions also go in the docs somewhere?

azure/app/template.json
```diff
@@ -183,6 +205,59 @@
...
+      "apiVersion": "2016-08-01",
+      "name": "[variables('vspAppServiceName')]",
+      "kind": "app,linux,container",
+      "location": "[resourceGroup().location]",
+      "dependsOn": ["[variables('vspAppServicePlanDeploymentName')]"],
+      "properties": {
+        "httpsOnly": true,
+        "serverFarmId": "[variables('vspAppServicePlanId')]",
+        "siteConfig": {
+          "alwaysOn": "[parameters('appServiceAlwaysOn')]",
```
erbridge: If this is set to `false`, we might have pretty slow responses on the first request after a deploy. Just a heads up.

azure/app/parameters/production.template.json
robbiepaul: I've added them to the `production` KeyVault. We don't have a `test` KeyVault yet

azure/app/parameters/test.template.json
robbiepaul: I've created a new Trello card for it: https://trello.com/c/YD86WbIp

I'll check about the normal process for rotating certs

azure/app/parameters/production.template.json
erbridge: Ah, I meant changing the parameters file to be `"value": ""` for now. I think that's better than adding placeholder values to a production Key Vault.

azure/app/template.json
robbiepaul: Happy to change this to `true` if you think it'd be better

azure/app/parameters/production.template.json
robbiepaul: Ahh ok, yeah that's better. I'll change it

azure/app/template.json
erbridge: No, I think it's fine. Just thought it was worth pointing out.

robbiepaul:
I've just got confirmation that the new certs are active, so I'll put change the variables to point at the new values

PR #216, opened 2019-07-31
Add a script for importing VSP certificates into the Key Vault

erbridge:
This has already been run on the certificates we just received back from Verify.

### DIFF
--- a/bin/import-vsp-certs
+++ b/bin/import-vsp-certs
@@ -1,0 +1,129 @@
+#!/bin/bash
+set -e
+
+if [ $# -lt 2 ]; then
+  echo "Usage: $0 ENVIRONMENT_NAME VERSION_NUMBER"
+  exit 1
+fi
+
+ENVIRONMENT_NAME=$1
+VERSION_NUMBER=$2
+
+case $ENVIRONMENT_NAME in
+  "development")
+    ENVIRONMENT_SHORT_NAME="Dev"
+    AZURE_SUBSCRIPTION_ID="8655985a-2f87-44d7-a541-0be9a8c2779d"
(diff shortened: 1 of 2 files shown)


erbridge:
[Get "Integration" (i.e. staging) Verify certificates for running a VSP](https://trello.com/c/FYRKfGar/345-get-integration-ie-staging-verify-certificates-for-running-a-vsp)

bin/import-vsp-certs
```diff
@@ -0,0 +1,129 @@
...
+
+function store_certificate {
+  local FILE_PREFIX=$1
+  local CERTIFICATE=$2
+
+  if ! [ "$CERT_EXISTED" ]; then
+    az keyvault secret set \
+      --vault-name $KEY_VAULT_NAME \
+      --name "${FILE_PREFIX}Certificate" \
+      --value "$CERTIFICATE"
```
pezholio: Does this work OK with keyvault? I had to use the `--file` option as the linebreaks got mangled if I set a value

erbridge: Yes, it does. Just confirmed by looking at what's in the Key Vault, imported by this script.

pezholio: 🆒

pezholio:
We're using a lot of shell scripts now, do you think it's worth adding `shellcheck` to our CI?

erbridge:
Possibly. I have it integrated into my editor, but yes. Add a card to the backlog?

pezholio:
Yeah, it might be better as a seperate story

erbridge:
`shellcheck` card added to the backlog.

PR #221, opened 2019-08-01
Capture a value for gender

pezholio:
This adds two methods for capturing a "gender" attribute. The first method is via Verify, if Verify does not return either `Male` or `Female` for a gender (which is what HMRC expect), we ask a user later in the flow using the following form:

image

I've also added a new `verified_fields` element to the claim model, which allows us to skip questions that have already been answered by Verify. This also means we can remove the edit links on the confirmation page, as well as prevent nefarious users from editing verified information. This may also make #179 moot, as we may still want to ask users for information we could not glean from Verify (we know, for example, that EU residents will still need to enter their address manually)


pezholio:
[Capture a value for "gender"](https://trello.com/c/vPAaVCve/347-capture-a-value-for-gender)

app/views/claims/gender.html.erb
```diff
@@ -0,0 +1,47 @@
...
+      <%= form.submit "Continue", class: "govuk-button" %>
+
+      <details class="govuk-details" data-module="govuk-details">
+        <summary class="govuk-details__summary">
+          <span class="govuk-details__summary-text">
+            If you don't know
+          </span>
+        </summary>
+        <div class="govuk-details__text">
+          If you don't know which gender is held on record we will contact HMRC or the Teachers Pension Scheme. We don't need to record this information but it is required to make tax contributions on your behalf to HMRC.
```
erbridge: Suggest we use "what" instead of "which" when talking about gender. Feels like that calls out the fact that we understand gender is more than a binary choice.

config/locales/en.yml
```diff
@@ -70,6 +70,7 @@ en:
         April 2019?"
       email_address: "Email address"
       bank_details: "Enter bank account details"
+      gender: "Which gender does your school's payroll system associate with you?"
```
erbridge: "What" vs "which" as above.

config/locales/en.yml
```diff
@@ -95,6 +96,10 @@ en:
           before_first_september_2012: "All my degree courses started before 1 September 2012"
           some_before_some_after_first_september_2012: "Some of my degree courses started before 1 September 2012 and some started after 1 September 2012"
           on_or_after_first_september_2012: "All of my degree courses started after 1 September 2012"
+      gender:
+        male: Male
+        female: Female
+        dont_know: Don't Know
```
erbridge: Shouldn't this be "I don't know"?

app/controllers/verify/authentications_controller.rb
```diff
@@ -35,6 +35,12 @@ def no_auth
 
     private
 
+    def claim_parameters
```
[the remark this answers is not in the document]
pezholio: I guess what I was going for here was to keep the parameters coming from the `Verify::Response` super generic and only return params that related directly to the user's identity.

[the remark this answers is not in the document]
tekin: Yeah, it's tricky because although that's the intent, `Verify::Response#claim_parameters` isn't really that generic. For a start the name `claim_parameters` is bound to our domain, and also we've got logic in there that is deciding which of the fields that come back are ones we care about, and also we map them to a structure that matches our domain model's attribute names, i.e. joining the names, only returning one of the names/addresses. At least by putting everything together in the same method, we own the fact that it isn't generic, and maybe at a later date we can split out the response parsing from the mapping to our model. That's something that @robbiepaul is almost certainly going to touch on with https://trello.com/c/E7ngqpON/530-fix-the-most-recent-verified-attributes.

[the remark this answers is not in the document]
pezholio: Ah, course. It's called `claim_parameters` after all isn't it?

app/views/claims/_check_your_answers_section.html.erb
```diff
@@ -4,18 +4,20 @@
 
 <dl class="govuk-summary-list govuk-!-margin-bottom-9">
   <%- answers.each do |(label, answer, slug)| %>
-    <div class="govuk-summary-list__row">
-      <dt class="govuk-summary-list__key">
-        <%= label %>
-      </dt>
+    <% unless current_claim.verified_fields.include?(slug) %>
```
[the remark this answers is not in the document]
pezholio: Cool, good shout. I've had a go at rectifying this. I've kept the feature tests in as I think it's still important to test the full journey here

pezholio:
Ready for another 👀 now

spec/features/changing_answers_spec.rb
```diff
@@ -242,4 +242,40 @@
     expect(claim.student_loan_start_date).to eq StudentLoans::BEFORE_1_SEPT_2012
     expect(claim.student_loan_plan).to eq StudentLoans::PLAN_1
   end
+
+  context "with a field that can be verified" do
```
tekin: More general points about specs and nested contexts: ideally, nesting of `describe` and `context` should be kept to a minimum as deeply nested specs can be difficult to maintain, read and follow. I think we should avoid going more than a single level deep where possible. In these cases below, we only have one actual scenario per two nested contexts, and so I'd suggest they could be re-written much more simply without any contexts at all, e.g.

`scenario "user cannot change the value of an identity field that was acquired from Verify"` 

and

`scenario "user can change the answer to an identity question that wasn't acquired from Verify"`

spec/models/student_loans/permitted_parameters_spec.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe StudentLoans::PermittedParameters do
+  let(:claim) { create(:tslr_claim) }
```
tekin: Does this actually need to be a persisted claim? If not we should avoid creating and just pass in a `TslrClaim.new` to keep the specs snappy.

spec/models/student_loans/permitted_parameters_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+  let(:permitted_parameters) { described_class.new(claim) }
+
+  describe "params" do
+    context "when no responses have come from verify" do
+      it "includes all parameters" do
+        expect(permitted_parameters.params).to eq([
```
tekin: Rather than listing out all the keys explicitly, this test will be less brittle (i.e. need updating every time we change the attributes) if we used the `StudentLoans::PermittedParameters::PARAMETERS ` constant here.

spec/models/student_loans/permitted_parameters_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+      end
+    end
+
+    context "when a field has come from verify" do
+      before do
+        claim.verified_fields = ["gender"]
+      end
+
+      it "skips that field" do
+        expect(permitted_parameters.params).to eq([
```
tekin: As above, it would be good if we could make this spec more robust. e.g. `StudentLoans::PermittedParameters::PARAMETERS - [:gender]`

lib/verify/response.rb
```diff
@@ -26,34 +26,45 @@ def verified?
...
       @address_lines ||= address.fetch("lines")
     end
 
     def full_name
-      first_name = most_recent_verified_value(parameters.fetch("attributes").fetch("firstNames"))
-      middle_name = most_recent_verified_value(parameters.fetch("attributes").fetch("middleNames"))
-      surname = most_recent_verified_value(parameters.fetch("attributes").fetch("surnames"))
+      first_name = most_recent_verified_value(attributes.dig("firstNames"))
+      middle_name = most_recent_verified_value(attributes.dig("middleNames"))
+      surname = most_recent_verified_value(attributes.dig("surnames"))
```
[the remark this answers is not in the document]
erbridge: Is it true that we always get more than one name for a user? My name, for instance, is a single word, though Verify may not support this.

config/locales/en.yml
```diff
@@ -110,6 +111,7 @@ en:
       address_line_4: "Address 4"
       postcode: "Postcode"
       date_of_birth: "Date of Birth"
+      gender: "Gender"
```
erbridge: Can this be "Payroll Gender" instead. I'm wary of support agents, or claims checkers using this value to assume pronouns for users.

tekin: that's a good shout 👍

lib/verify/response.rb
```diff
@@ -26,34 +26,45 @@ def verified?
...
     def address_lines
       @address_lines ||= address.fetch("lines")
     end
 
     def full_name
-      first_name = most_recent_verified_value(parameters.fetch("attributes").fetch("firstNames"))
-      middle_name = most_recent_verified_value(parameters.fetch("attributes").fetch("middleNames"))
-      surname = most_recent_verified_value(parameters.fetch("attributes").fetch("surnames"))
+      first_name = most_recent_verified_value(attributes.dig("firstNames"))
+      middle_name = most_recent_verified_value(attributes.dig("middleNames"))
```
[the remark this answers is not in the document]
mec: We've got a card to look at EU citizens, let's cover that issue there and use fetch here for this iteration.

lib/verify/response.rb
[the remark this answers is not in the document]
tekin: As far as the Verify documentation goes, we should always get back a surnames value. This might be a question we can only answer by actually hitting Verify for real. I'd suggest we'll be in a better position to tackle this as a separate iteration once we've got the basics covered and can see how Verify behaves in this case.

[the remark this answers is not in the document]
erbridge: I've added a card to the backlog to look at unusual data in Verify.

app/models/page_sequence.rb
```diff
@@ -19,6 +19,7 @@ class PageSequence
     "full-name",
     "address",
     "date-of-birth",
+    "gender",
```
erbridge: Should this also be `payroll-gender`?

app/models/student_loans/permitted_parameters.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    attr_reader :claim
+
+    def initialize(claim)
+      @claim = claim
+    end
+
+    def keys
+      PARAMETERS.dup.tap do |parameters|
+        parameters.delete(:payroll_gender) if claim.verified_fields.include?("gender")
```
erbridge: Shouldn't this be `payroll_gender` now? Are we missing test coverage of this?

app/models/tslr_claim.rb
```diff
@@ -46,6 +52,8 @@ class TslrClaim < ApplicationRecord
 
   validates :mostly_teaching_eligible_subjects, on: [:"mostly-teaching-eligible-subjects", :submit], inclusion: {in: [true, false], message: "Select either Yes or No"}
 
+  validates :payroll_gender,                    on: [:gender, :submit], presence: {message: "Choose the option for the gender your school’s payroll system associates with you"}
```
erbridge: Change the context to `:"payroll-gender"` if renaming the page.

spec/lib/verify/response_spec.rb
```diff
@@ -34,6 +34,52 @@
...
+    context "when the gender is other" do
+      let(:response_filename) { "identity-verified-other-gender.json" }
+
+      it "returns nil for gender" do
+        expect(subject.claim_parameters[:payroll_gender]).to eq(nil)
+      end
+
+      it "does not return gender in the verified fields" do
+        expect(subject.claim_parameters[:verified_fields]).to_not include(:payroll_gender)
+      end
```
erbridge: `s/gender/payroll gender/` in spec names?

spec/requests/claims_spec.rb
```diff
@@ -166,6 +166,20 @@
         end
       end
 
+      context "when a field has come from Verify" do
+        before do
+          in_progress_claim.update!(verified_fields: ["gender"])
```
erbridge: Should be `payroll_gender`, I think.

spec/models/student_loans/permitted_parameters_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  describe "keys" do
+    context "when no fields have come from verify" do
+      it "returns all permitted parameters" do
+        expect(permitted_parameters.keys).to eq(StudentLoans::PermittedParameters::PARAMETERS)
+      end
+    end
+
+    context "when fields have come from verify" do
+      before do
+        claim.verified_fields = ["gender"]
```
erbridge: Should be `payroll_gender`, which explains why the test was passing.

tekin: hmmm, I thought all these had been updated... Will have another look.

app/models/page_sequence.rb
tekin: We kept the slug for the page the same. I suggest we look to change it separately in the interest of getting this merged and unblocking the dependent work.

app/models/tslr_claim.rb
tekin: Reckon we keep the page slug for now.

PR #222, opened 2019-08-01
Verify requests use SSL for https URLs

robbiepaul:
I'm not sure how to test this, WebMock doesn't give many options for making assertions about requests.

This was causing requests to our integration VSP to be rejected as it's running on https. We still want to be able to use the development VSP so we simply check the port inferred from the URI. Any URIs with `https` will have the port 443 inferred.

### DIFF
--- a/lib/verify/service_provider.rb
+++ b/lib/verify/service_provider.rb
@@ -34,7 +34,7 @@ module Verify
       uri = URI(self.class.generate_request_url)
       request = Net::HTTP::Post.new(uri, {})
       request.content_type = "application/json"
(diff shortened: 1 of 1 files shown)


lib/verify/service_provider.rb
```diff
@@ -74,5 +74,9 @@ def handle_response(json_response)
         raise ResponseError.new(parameters["message"], parameters["code"]) if parameters.key?("code")
       end
     end
+
+    def uri_uses_ssl?(uri)
+      uri.port == 443
```
erbridge: This should be `default_port` not `port` to allow for HTTPS on custom ports (e.g. for local SSL testing).

robbiepaul:

PR #225, opened 2019-08-05
The VSP Entity ID needs https to match the value on the Verify Hub

robbiepaul:
If the value doesn't match the value on the hub we see a 500 error when making a POST to it.

Also changed the `uri_uses_ssl?` method to check for `default_port` instead of port.


robbiepaul:

PR #226, opened 2019-08-05
Restrict access to VSP to the app service and Verify Hub

erbridge:
Due to restrictions in the way ARM templates work, this requires pulling the VSP out into a child template, as we can't use `reference` in `copy` (I think is expands it at "compile" time, so before the referenced resource exists).

The upside of this is that it makes the templates much more readable. If we gave the rest of the project the same treatment, it would help with readability of interdependency of components (and with finding them).

The downside is that doing test deployments now requires pushing the current commit up to the remote, so that `az` can fetch it. It looks it up by commit hash, but doesn't complain if there are uncommitted changes on HEAD, so only changes to child templates need to be pushed.

This isn't actually tested to be working, due to constraints making that hard to do without pushing to `s118d01`.


erbridge:
[Restrict access to VSP](https://trello.com/c/Dr35AdfA/491-restrict-access-to-vsp)

docs/infrastructure-deployment.md
```diff
@@ -66,6 +66,18 @@ Where `<environment>` is the environment you want to deploy to.
...
 has finished running the changes will be applied to your environment!
 
+## Testing deployment
+
+If you want to test changes to infrastructure with minimal impact to existing
+environments, `bin/azure-deploy` has `--tmp-*` arguments for each resource group
+(e.g. `--tmp-app`), which deploys a new resource group with a `-tmp` suffix.
+
+Deployments rely on having web access to the current version of the Git
+repository (they attempt to fetch the current commit hash from GitHub). To test
```
robbiepaul: 👍

PR #228, opened 2019-08-05
Set up alerts for VSP

erbridge:
This shares common commits with #226 to pull the VSP out into its own template and do the set up to support child templates.


erbridge:
[Set up alerts for VSP](https://trello.com/c/x1hE28gU/527-set-up-alerts-for-vsp)

PR #229, opened 2019-08-05
Temporarily log the Verify response in Rollbar

robbiepaul:
Whilst we are debugging the Verify integration environment it would be helpful to know what the response from Verify looks like.

We should remove this logging before we integrate with the production VSP.


robbiepaul:
I added an item to the checklist on this card to make sure this gets removed: https://trello.com/c/8VhzTdt1/356-govuk-verify-4-get-verify-running-on-production

PR #231, opened 2019-08-05
Set up Application Insights

erbridge:
We should see request logging show up once it's deployed. This doesn't do any more than that on the AI front.

This also fixes the deployment of the tmp resource groups. You aren't allowed to have more than one app service with a given host name, so we disable deployment of the hostname parts when deploying a tmp app. The dependency graph was also missing some explicit dependencies that blocked a fresh deployment.

### DIFF
--- a/config/initializers/application_insights.rb
+++ b/config/initializers/application_insights.rb
@@ -1,0 +1,8 @@
+instrumentation_key = ENV["APPINSIGHTS_INSTRUMENTATIONKEY"]
+
+if instrumentation_key.present?
+  Rails.application.configure do
+    buffer_size = 1
+    config.middleware.use ApplicationInsights::Rack::TrackRequest, instrumentation_key, buffer_size
+  end
+end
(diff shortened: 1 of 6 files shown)


erbridge:
[EPIC : Set up Application Insights on CIP](https://trello.com/c/oiJ1sy1k/251-epic-set-up-application-insights-on-cip)

config/initializers/application_insights.rb
```diff
@@ -0,0 +1,11 @@
+unless Rails.env.test?
```
tekin: Would this be better as `if Rails.env.production?` given that the Azure environments will run in that environment? Or is there is a specific reason we'd want to try and load application insights in someone's local development environment?

erbridge: @pezholio removed the check entirely. I assume that was discussed.

tekin: Yep, I'm happy from my end. might want to check in with @pezholio who's working with Craig today.

PR #232, opened 2019-08-06
Don't raise on bullet warnings in test suite

tekin:
Instead we will warn and log in development mode.

This was causing us pain whilst working on the refactoring work to extract a separate model for eligibility criteria because it was raising circular warnings – if we didn't preload `eligibility: [:claim_school]`, etc when loading the claim in `current_claim`, it would complain that we
weren't eager-loading the associations, and if we did, it would complain that they were being eager-loaded but not used. This is because not all the pages where the `current_claim` is accessed also access the eligibility and/or the claim_school.

We could continue to add to the whitelist for these sorts of situations, but then bullet starts to lose its value as we'll essentially end up having to whitelist both n_plus_one_query and unused_eager_loading for pretty much every association on claim and eligibility.

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -52,7 +52,6 @@ Rails.application.configure do
   config.after_initialize do
     Bullet.enable = true
     Bullet.bullet_logger = true
-    Bullet.raise = true
     Bullet.add_whitelist type: :unused_eager_loading, class_name: "TslrClaim", association: :current_school
     Bullet.add_whitelist type: :n_plus_one_query, class_name: "School", association: :local_authority
   end


config/environments/development.rb
```diff
@@ -77,4 +77,12 @@
       },
     }
   )
+
+  config.after_initialize do
```
pezholio: Could we maybe have `Bullet.add_footer` here? Might make it a bit clearer that something needs fixing when devs are testing in development

tekin: I'm wary of gem code that inserts markup or JS into the page as it can result in weird behaviour. Also because of the circular false positives it will literally appear on half the claim pages, complaining about missing eager-loading of `eligibility`, `claim_school`, and/or `current_school`. Short of doing something where we conditionally eager load the appropriate association on each page depending on what is used, I don't think there is a way to make it so we don't end up ignoring these anyway because of all the false positives. Also, that eager loading would be mostly pointless, because when you call `TslrClaim.includes(:eligibility).find(ID)`, Rails query optimiser decides it's not worth the join and just makes two SQL queries anyway!

pezholio: Yeah, I've felt that pain before

pezholio: Been thinking about this more and maybe it's better to have this stuff in test (maybe with `rails_logger` left out?), so if there are any issues with performance we can look at the bullet logs? As our test coverage is pretty good, we can be confident that a run of the whole test suite is covering pretty much all of the queries, and by scanning the Bullet logs we can see any new N+1 queries.

tekin: Yep, ok PR updated.

PR #234, opened 2019-08-07
Fix email field validation error messages

mec:
If the user submits the email field with no value both the presence and
the format validation is triggered.

Switching the format validation to allow_blank stops this behaviour and
shows the appropriate message.

https://trello.com/c/YBl0fxWQ

### DIFF
--- a/spec/models/tslr_claim_spec.rb
+++ b/spec/models/tslr_claim_spec.rb
@@ -16,6 +16,7 @@ RSpec.describe TslrClaim, type: :model do
     it "validates that the value is in the correct format" do
       expect(build(:tslr_claim, email_address: "notan email@address.com")).not_to be_valid
       expect(build(:tslr_claim, email_address: "name@example.com")).to be_valid
+      expect(build(:tslr_claim, email_address: "")).to be_valid
     end
 
     it "checks that the email address in not longer than 256 characters" do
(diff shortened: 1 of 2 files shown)


spec/models/tslr_claim_spec.rb
```diff
@@ -16,6 +16,7 @@
     it "validates that the value is in the correct format" do
       expect(build(:tslr_claim, email_address: "notan email@address.com")).not_to be_valid
       expect(build(:tslr_claim, email_address: "name@example.com")).to be_valid
+      expect(build(:tslr_claim, email_address: "")).to be_valid
```
pezholio: Actually, do we need another test to check `(build(:tslr_claim, email_address: nil)` is valid? (i.e. the default) I'm sure it is, but we probably need to check this assumption as if it isn't, the whole flow will break.

mec: I felt like we had enough coverage for that case in the validation contexts?

https://github.com/DFE-Digital/dfe-teachers-payment-service/blob/801f43202962a1a42dbc9a176f74898ad93c4866/spec/models/tslr_claim_spec.rb#L252

pezholio: Ah, true dat. Carry on then :)

PR #237, opened 2019-08-07
Sanitize dangerous fields in CSV output

erbridge:
"-", "+", "=", and "@" all have special meaning in Excel. Wrapping strings that contain them in `="<field>"` tells Excel to treat the field as a string, and to not execute any formulae.

When exported to a CSV, we end up with field that look like: `"=""=SUM(A1, A2)"""`, which works as expected.

### DIFF
--- a/app/presenters/tslr_claim_csv_row.rb
+++ b/app/presenters/tslr_claim_csv_row.rb
@@ -8,8 +8,18 @@ class TslrClaimCsvRow < SimpleDelegator
 
   private
 
+  DANGEROUS_STRINGS = [
+    "-",
+    "+",
+    "=",
+    "@",
+  ].freeze
+
   def data
-    TslrClaimsCsv::FIELDS.map { |f| send(f) }
+    TslrClaimsCsv::FIELDS.map do |f|
+      field = send(f)
+      sanitize_field(field)
+    end
   end
 
   def employment_status
@@ -34,6 +44,13 @@ class TslrClaimCsvRow < SimpleDelegator
 
   def submitted_at
     model.submitted_at.strftime("%d/%m/%Y %H:%M")
+  end
+
+  def sanitize_field(field)
+    return field if field.blank?
+
+    field = "=\"#{field}\"" if DANGEROUS_STRINGS.any? { |string| field.include?(string) }
+    field
   end
 
   def model
--- a/spec/presenters/tslr_claim_csv_row_spec.rb
+++ b/spec/presenters/tslr_claim_csv_row_spec.rb
@@ -2,7 +2,7 @@ require "rails_helper"
 
 RSpec.describe TslrClaimCsvRow do
   subject { described_class.new(claim) }
-  let(:claim) { create(:tslr_claim) }
+  let(:claim) { build(:tslr_claim) }
 
   describe "to_s" do
     let(:date_of_birth) { "01/12/1980" }
@@ -12,7 +12,7 @@ RSpec.describe TslrClaimCsvRow do
     let(:row) { CSV.parse(subject.to_s).first }
 
     let(:claim) do
-      create(:tslr_claim, :submittable,
+      build(:tslr_claim, :submittable,
         date_of_birth: Date.parse(date_of_birth),
         mostly_teaching_eligible_subjects: mostly_teaching_eligible_subjects == "Yes",
         student_loan_repayment_amount: student_loan_repayment_amount.delete("£").to_i,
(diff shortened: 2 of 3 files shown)


erbridge:
[Protect against executable Excel-formulas in CSV export](https://trello.com/c/sR3dSFw1/470-protect-against-executable-excel-formulas-in-csv-export)

app/presenters/tslr_claim_csv_row.rb
```diff
@@ -36,6 +46,13 @@ def submitted_at
     model.submitted_at.strftime("%d/%m/%Y %H:%M")
   end
 
+  def sanitize_field(field)
+    return field unless field.present?
+
+    field = '="' + field + '"' if DANGEROUS_STRINGS.any? { |string| field.include?(string) }
```
pezholio: I think I'd prefer `"=\"#{field}\""` here. I know it's a bit ugly with the escaped quotes, but it's more performant as concatenating strings like that creates three objects, whereas interpolating strings creates just the one object, which might become an issue with a big CSV.

spec/presenters/tslr_claim_csv_row_spec.rb
```diff
@@ -43,13 +43,35 @@
         date_of_birth,
         claim.teacher_reference_number,
         claim.national_insurance_number,
-        "Plan 2",
-        claim.email_address,
+        StudentLoans::PLAN_2.humanize,
+        '="' + claim.email_address + '"',
```
pezholio: I'm a bit worried that checking for the presence of an `@` results in a false positive like this. This means that if the CSV is opened in anything other than Excel, the email will look a bit odd. Can we not treat the email field differently? My instinct is that we _could_ trust this field anyway, given there are validations in place to only accept a valid email address (unless there's a way to craft a valid email that also acts as an Excel formula)

app/presenters/tslr_claim_csv_row.rb
erbridge: I was trying to avoid the ugly (and hard to read) escaping, but I didn't know about the performance implications of the way Ruby handles this.

spec/presenters/tslr_claim_csv_row_spec.rb
erbridge: I don't think this _is_ a false positive. "@" has meaning in Excel, and you might be able to embed a function into a valid email address, for instance. My understanding is that other software that consumes CSVs will also parse the extra equals sign away, so the only place this should be visible is in the raw CSV, which hopefully is only being read by developers. This is also a stop-gap that only lives as long as this CSV export does, so I think it's ok. I'd rather deal with it when there's an issue and be more complete in our sanitation.

app/presenters/tslr_claim_csv_row.rb
robbiepaul: Could you use `sprintf`? eg. `sprintf('="%s"', field)`

robbiepaul: Actually standard might complain at that

erbridge: I'm going with the interpolation. I think it's more familiar to read, even with escapes.

spec/presenters/tslr_claim_csv_row_spec.rb
```diff
@@ -43,13 +43,35 @@
...
+        StudentLoans::PLAN_2.humanize,
+        "=\"#{claim.email_address}\"",
         mostly_teaching_eligible_subjects,
         claim.bank_sort_code,
         claim.bank_account_number,
         student_loan_repayment_amount,
       ])
     end
+
+    it "escapes fields with dangerous strings" do
```
tekin: Minor thing, but can we be more specific in this spec line and maybe mention we want to escape characters that could be dangerous when interpreted by Excel? I know that information is in the commit message, but capturing it here in the spec description makes it easier for a developer to understand why these strings are dangerous.

erbridge: Done.

app/presenters/tslr_claim_csv_row.rb
erbridge: Done.

app/presenters/tslr_claim_csv_row.rb
```diff
@@ -36,6 +46,13 @@ def submitted_at
     model.submitted_at.strftime("%d/%m/%Y %H:%M")
   end
 
+  def sanitize_field(field)
+    return field unless field.present?
```
tekin: Assuming `return if field.nil?`is what this is meant to be doing? i.e. guard against it not being a string?

app/presenters/tslr_claim_csv_row.rb
```diff
@@ -36,6 +46,13 @@ def submitted_at
...
+
+    field = "=\"#{field}\"" if DANGEROUS_STRINGS.any? { |string| field.include?(string) }
```
tekin: no need for this assignment, we can rely on Ruby's implicit return.

app/presenters/tslr_claim_csv_row.rb
erbridge: I intentionally return what came in. If it's an empty string, there's also nothing to do here, but I didn't want to replace that with `nil`. Do you think that's wrong?

tekin: That's fair enough, in which case we can switch it from an `unless` to an `if field.blank?`

erbridge: Sure.

erbridge: Done.

spec/presenters/tslr_claim_csv_row_spec.rb
pezholio: Yeah, good point on the stop gap bit. I'm happy in that case 👍

PR #238, opened 2019-08-07
 Allow the minimum set of verified data from Verify

robbiepaul:
We now know more about the data attributes that are returned from Verify. This PR should allow us to have a "minimum viable Verify user" complete the journey with us.

- If the user's details haven't changed then we don't get any `to` or `from` dates
- If the user's details have changed they get a `to` and `from` date except the current value which doesn't have a `to` date
- If a user has no middle name it will be returned as an empty array

### DIFF
--- a/lib/verify/response.rb
+++ b/lib/verify/response.rb
@@ -66,11 +66,12 @@ module Verify
       middle_name = most_recent_verified_value(attributes.fetch("middleNames"))
       surname = most_recent_verified_value(attributes.fetch("surnames"))
 
-      [first_name, middle_name, surname].join(" ")
+      [first_name, middle_name, surname].compact.join(" ")
     end
 
     def most_recent_verified_value(attributes)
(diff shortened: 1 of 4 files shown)


robbiepaul:
[Fix the most recent verified attributes](https://trello.com/c/E7ngqpON/530-fix-the-most-recent-verified-attributes)

robbiepaul:
My understanding is that when there's more than one attribute for a field we get the `from` date on each, so our fixture for the first spec should cover the most recent verified value if there's more than one: 

Fixture: https://github.com/DFE-Digital/dfe-teachers-payment-service/blob/530-fix-the-most-recent-verified-attributes/spec/fixtures/verify/identity-verified.json#L6-L56

Spec: https://github.com/DFE-Digital/dfe-teachers-payment-service/blob/530-fix-the-most-recent-verified-attributes/spec/lib/verify/response_spec.rb#L31-L36

tekin:
👍🏽

PR #239, opened 2019-08-07
Output formatted payroll_gender

mec:
When shown, the payroll_gender attribute should be formatted and not an
internal representation.

image


app/helpers/claims_helper.rb
```diff
@@ -65,4 +65,8 @@ def academic_years(year_range)
 
     "September 1 #{start_year} - August 31 #{end_year}"
   end
+
+  def formatted_payroll_gender(payroll_gender)
+    payroll_gender == "dont_know" ? "Don’t know" : payroll_gender.humanize
```
pezholio: We can probably use translations for this

mec: Why do I never see these things the first time around! Thanks @pezholio! How's this?

PR #241, opened 2019-08-08
Explicitly deny all non-allowed web traffic to the VSP

erbridge:


erbridge:
[Restrict access to VSP](https://trello.com/c/Dr35AdfA/491-restrict-access-to-vsp)

PR #242, opened 2019-08-08
Migrate subjects and teaching time-related attributes to the Eligibility model

tekin:
This moves the attributes related to subjects and teaching time to the Eligibility model. These were the last attributes that determine eligibility, which means we can also now stop delegating the eligibility-related methods. There are still accessor methods being delegated that we'll want to remove, but that can be done as a separate step later.

Part of https://trello.com/c/9gpS69mI/536-separate-eligibility-data-modelling-from-claims-in-prep-for-mp

### DIFF
--- a/app/models/tslr_claim.rb
+++ b/app/models/tslr_claim.rb
@@ -1,14 +1,6 @@
 # frozen_string_literal: true
 
 class TslrClaim < ApplicationRecord
-  SUBJECT_FIELDS = [
-    :biology_taught,
-    :chemistry_taught,
-    :physics_taught,
-    :computer_science_taught,
-    :languages_taught,
-  ].freeze
-
   TRN_LENGTH = 7
 
   NO_STUDENT_LOAN = "not_applicable"
@@ -24,6 +16,8 @@ class TslrClaim < ApplicationRecord
            :claim_school,
            :employment_status,
            :current_school,
+           :subjects_taught,
+           :mostly_teaching_eligible_subjects?,
            to: :eligibility
 
   # NOTE: Temporary delegation of eligibility methods
@@ -31,6 +25,7 @@ class TslrClaim < ApplicationRecord
            :ineligible_claim_school?,
            :employed_at_claim_school?,
            :employed_at_no_school?,
+           :not_taught_eligible_subjects_enough?,
            to: :eligibility, allow_nil: nil
 
   # NOTE: Temporary delegation of left over methods
@@ -45,10 +40,6 @@ class TslrClaim < ApplicationRecord
     female: 1,
     male: 2,
   }
-
-  validate :at_least_one_subject_chosen,        on: [:"subjects-taught", :submit], unless: ->(c) { c.mostly_teaching_eligible_subjects == false }
-
-  validates :mostly_teaching_eligible_subjects, on: [:"mostly-teaching-eligible-subjects", :submit], inclusion: {in: [true, false], message: "Select either Yes or No"}
 
   validates :payroll_gender,                    on: [:gender, :submit], presence: {message: "Choose the option for the gender your school’s payroll system associates with you"}
 
@@ -145,10 +136,6 @@ class TslrClaim < ApplicationRecord
     super(value.to_s.gsub(/[£,\s]/, ""))
   end
 
-  def subjects_taught
-    SUBJECT_FIELDS.select { |s| send(s) == true }
-  end
-
   def no_student_loan?
     !has_student_loan?
   end
@@ -158,10 +145,6 @@ class TslrClaim < ApplicationRecord
   end
 
   private
-
-  def not_taught_eligible_subjects_enough?
-    mostly_teaching_eligible_subjects == false
-  end
 
   def normalise_trn
     self.teacher_reference_number = normalised_trn
@@ -215,14 +198,6 @@ class TslrClaim < ApplicationRecord
       ref = Reference.new.to_s
       break ref unless self.class.exists?(reference: ref)
     }
-  end
-
-  def at_least_one_subject_chosen
-    errors.add(:subjects_taught, "Choose a subject, or select \"not applicable\"") unless at_least_one_subject_chosen?
-  end
-
-  def at_least_one_subject_chosen?
-    SUBJECT_FIELDS.find { |s| send(s) == true }.present?
   end
 
   def claim_must_not_be_ineligible
--- a/app/models/tslr_claim.rb
+++ b/app/models/tslr_claim.rb
@@ -20,14 +20,6 @@ class TslrClaim < ApplicationRecord
            :mostly_teaching_eligible_subjects?,
            to: :eligibility
 
-  # NOTE: Temporary delegation of eligibility methods
-  delegate :ineligible_qts_award_year?,
-           :ineligible_claim_school?,
-           :employed_at_claim_school?,
-           :employed_at_no_school?,
-           :not_taught_eligible_subjects_enough?,
-           to: :eligibility, allow_nil: nil
-
   # NOTE: Temporary delegation of left over methods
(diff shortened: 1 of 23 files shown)


app/models/tslr_claim.rb
```diff
@@ -24,15 +16,10 @@ class TslrClaim < ApplicationRecord
            :claim_school,
            :employment_status,
            :current_school,
+           :subjects_taught,
+           :mostly_teaching_eligible_subjects?,
            to: :eligibility
```
erbridge: Is the intention to remove these later?

db/schema.rb
```diff
@@ -71,6 +71,12 @@
     t.uuid "claim_school_id"
     t.uuid "current_school_id"
     t.integer "employment_status"
+    t.boolean "biology_taught"
+    t.boolean "chemistry_taught"
+    t.boolean "computer_science_taught"
+    t.boolean "languages_taught"
+    t.boolean "physics_taught"
```
erbridge: Was losing the default values intentional?

spec/factories/student_loans/eligibilities.rb
```diff
@@ -2,11 +2,18 @@
 
 FactoryBot.define do
   factory :student_loans_eligibility, class: "StudentLoans::Eligibility" do
-    trait :submittable do
+    trait :eligible do
```
erbridge: 👍

spec/models/student_loans/eligibility_spec.rb
```diff
@@ -116,29 +129,64 @@
...
   end
 
+  context "when saving in the “subjects-taught” context" do
+    it "is not valid if none of the subjects-taught attributes are true" do
+      expect(StudentLoans::Eligibility.new).not_to be_valid(:"subjects-taught")
+      expect(StudentLoans::Eligibility.new(biology_taught: false)).not_to be_valid(:"subjects-taught")
+      expect(StudentLoans::Eligibility.new(biology_taught: false, physics_taught: false)).not_to be_valid(:"subjects-taught")
+    end
+
+    it "validates the when one or more of the subjects-taught attributes are true" do
```
erbridge: Typo? I'm not sure what this is supposed to say.

spec/models/student_loans/eligibility_spec.rb
```diff
@@ -116,29 +129,64 @@
...
+      expect(StudentLoans::Eligibility.new(biology_taught: true)).to be_valid(:"subjects-taught")
+      expect(StudentLoans::Eligibility.new(biology_taught: true, computer_science_taught: false)).to be_valid(:"subjects-taught")
+      expect(StudentLoans::Eligibility.new(chemistry_taught: true, languages_taught: true)).to be_valid(:"subjects-taught")
+    end
+
+    it "validates with no subjects present if mostly_teaching_eligible_subjects is false" do
```
erbridge: Can we rephase "validates" to "is valid". I think that's easier to parse ("validates" refers to the process as well as the result).

spec/models/tslr_claim_spec.rb
```diff
@@ -242,7 +223,8 @@
     end
 
     it "validates the claim is not ineligible" do
```
erbridge: This could also do with rewording, I think. Maybe "validates the claim's eligibility"?

spec/models/tslr_claim_spec.rb
```diff
@@ -285,75 +267,19 @@
...
-    context "with an ineligible claim_school" do
-      let(:claim_attributes) { {eligibility: build(:student_loans_eligibility, claim_school: schools(:hampstead_school))} }
-      it { is_expected.to be true }
-    end
-
-    context "when no longer teaching" do
-      let(:claim_attributes) { {eligibility: build(:student_loans_eligibility, employment_status: :no_school)} }
-      it { is_expected.to be true }
-    end
+    it "delegates to the eligibility" do
```
erbridge: We're not actually testing that it delegates here.

spec/models/tslr_claim_spec.rb
```diff
@@ -285,75 +267,19 @@
...
-
-    context "when taught less than half time in eligible subjects" do
-      let(:claim_attributes) { {mostly_teaching_eligible_subjects: false} }
-      it { is_expected.to eql :not_taught_eligible_subjects_enough }
-    end
-
-    context "when not ineligible" do
-      let(:claim_attributes) { {} }
-      it { is_expected.to be_nil }
+    it "delegates to the eligibility" do
```
erbridge: This is also not testing delegation.

spec/presenters/tslr_claim_csv_row_spec.rb
```diff
@@ -14,7 +14,6 @@
     let(:claim) do
       create(:tslr_claim, :submittable,
         date_of_birth: Date.parse(date_of_birth),
-        mostly_teaching_eligible_subjects: mostly_teaching_eligible_subjects == "Yes",
```
erbridge: Should we also remove the `let` associated with this?

spec/presenters/tslr_claim_csv_row_spec.rb
```diff
@@ -40,7 +39,7 @@
         claim.national_insurance_number,
         "Plan 2",
         claim.email_address,
-        mostly_teaching_eligible_subjects,
+        "Yes",
```
erbridge: This is unnecessarily sensitive to changes to the factory, and assumes defaults not set here. Can we derive our expected value from the claim like we do for the CSV spec?

app/models/tslr_claim.rb
tekin: Definitely, we want to get to a generic "Claim" model that has no eligibility-specific knowledge so it can be used for student loans and maths & physics. We've left this delegation in place to keep these PRs focused on migrating the data (which will be harder to do later) and the plan is to move them later.

db/schema.rb
tekin: yeah, I don't believe they are necessary, and leaving it `null` by default means we can distinguish between a claim that hasn't answered the questions and one that has answered with "No".

spec/models/student_loans/eligibility_spec.rb
tekin: 👍🏽

spec/models/student_loans/eligibility_spec.rb
tekin: 👍🏽

spec/models/tslr_claim_spec.rb
tekin: 👍🏽

spec/presenters/tslr_claim_csv_row_spec.rb
tekin: 👍🏽

spec/models/tslr_claim_spec.rb
tekin: Good shout on these, it felt wrong and I've now just removed the delegation altogether.

spec/presenters/tslr_claim_csv_row_spec.rb
tekin: Yeah, this is a tricky one. It's implicit in the `submittable` claim factory that it's also `eligible`, which means the answer to `mostly_teaching_eligible_subjects` attribute will be `true`, which the row code translates to "Yes". Deriving the value would require replicating the code under test, i.e. `claim.eligibility.mostly_teaching_eligible_subjects? ? "Yes" : "No"`, which doesn't really feel better. I suppose one thing we could do is be explicit and set all the attributes on the claim that is being created?

erbridge: Yeah, it's a good point. Maybe we should explicitly set values that would require some processing then.

tekin:
Hopefully that's all comments addressed now.

PR #243, opened 2019-08-08
Add production VSP configuration parameters

robbiepaul:
So that we can get the production VSP up-and-running we need to configure several parameters used in the ARM templates.

We don't want to enable the feature flag just yet until we have a production VSP running and we can access the Verify production environment.


robbiepaul:
[GOV.UK Verify 4. Get Verify running on production](https://trello.com/c/8VhzTdt1/356-govuk-verify-4-get-verify-running-on-production)

PR #245, opened 2019-08-08
Remove temporary non-Verify part of the journey

tekin:
This cleans up the conditional logic that was in place whilst we were waiting for GOV.UK Verify integration to be wrapped up. Technically we're still waiting for our production Verify account to be set up, but we're expecting that imminently. A couple of things:

- full-name and date-of-birth have been dropped from the sequence
- address remains in the sequence as we know we'll need it for EU citizen support, but the page will be excluded from the sequence if the address has come from Verify.

https://trello.com/c/8Q9zh0de/376-prevent-editing-changing-of-verified-name-address-dob-by-users


README.md
```diff
@@ -77,11 +77,10 @@ NOTIFY_TEMPLATE_ID=d72e2ff9-b228-4f16-9099-fd9d411c0334
...
 
-If you want to test Verify in development, add the following to your `.env`
-file.
+Add the following to your `.env` file to point at the local VSP in development
+mode.
 
 ```
-GOVUK_VERIFY_ENABLED=1
 GOVUK_VERIFY_VSP_HOST=http://localhost:50300
 ```
```
erbridge: Can we put this in a `.env.development` instead of getting developers to do this?

app/models/tslr_claim.rb
```diff
@@ -120,6 +120,14 @@ def student_loan_country_with_one_plan?
     StudentLoans::PLAN_1_COUNTRIES.include?(student_loan_country)
   end
 
+  def address_verified?
+    (ADDRESS_ATTRIBUTES & verified_fields).any?
```
erbridge: New syntax to me. What does this do? I'm assuming from context that `&` gets the intersection of the two arrays?

tekin: Yep, binary ampersand performs set intersection in Ruby when used with arrays.

README.md
tekin: Just pushed up an additional commit to rejig the env vars 👍🏽

spec/features/tslr_claim_spec.rb
```diff
@@ -29,34 +29,15 @@
...
-
-    expect(claim.reload.address_line_1).to eql("123 Main Street")
-    expect(claim.address_line_2).to eql("Downtown")
-    expect(claim.address_line_3).to eql("Twin Peaks")
-    expect(claim.address_line_4).to eql("Washington")
-    expect(claim.postcode).to eql("M1 7HL")
-
-    expect(page).to have_text(I18n.t("tslr.questions.date_of_birth"))
-
-    fill_in_date_of_birth
```
robbiepaul: Do we lose coverage for this feature now that this is gone?

app/models/student_loans/permitted_parameters.rb
```diff
@@ -36,9 +36,7 @@ def initialize(claim)
     end
 
     def keys
-      PARAMETERS.dup.tap do |parameters|
-        parameters.delete(:payroll_gender) if claim.verified_fields.include?("payroll_gender")
-      end
+      PARAMETERS.dup - claim.verified_fields.map(&:to_sym)
```
pezholio: Nice 👍

spec/features/tslr_claim_spec.rb
```diff
@@ -29,34 +29,15 @@
...
-
-    expect(page).to have_text(I18n.t("tslr.questions.payroll_gender"))
-    choose "Male"
```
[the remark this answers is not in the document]
tekin: Yeah, I think you may be right...

app/models/tslr_claim.rb
```diff
@@ -120,6 +120,10 @@ def student_loan_country_with_one_plan?
...
```
pezholio: I like this approach, but I do worry - what if we only get back a postcode from Verify? (and would that even be A Thing?). If it's a possible outcome, maybe we should consider the address unverified?

spec/features/tslr_claim_spec.rb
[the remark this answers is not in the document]
tekin: Can we get this merged and tackle that separately? I just started to add a feature spec for completing missing verify fields but it's going to be involved and it would be good to get this merged today if we can so we're ready for Monday.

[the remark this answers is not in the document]
robbiepaul: Sounds good to me 👍

app/models/tslr_claim.rb
tekin: Yeah, I thought about that and decided if we only got back a postcode (which I don't think will happen – Verify should always give us an actual address, right?), then we should count it as "verified" and deal with the consequences then.

PR #246, opened 2019-08-08
Validate the length of address_line_4

erbridge:


erbridge:
Leaving it for now, in the interest of keeping things moving, but yes, that would be good.

PR #248, opened 2019-08-08
Set up redundancy and resource levels on infrastructure

pezholio:
This allows us to explicitly set the number of instances we're running in a service plan. There is the option to do autoscaling, but this gives us redundancy if one instance was to fall over.

### DIFF
--- a/azure/resource_groups/app/template.json
+++ b/azure/resource_groups/app/template.json
@@ -44,12 +44,20 @@
       "type": "string",
       "defaultValue": "1"
     },
+    "vspAppServiceDockerImage": {
+      "type": "string"
+    },
+    "vspAppServicePlanTier": {
+      "type": "string",
+      "defaultValue": "Standard"
+    },
+    "vspAppServicePlanSize": {
+      "type": "string",
+      "defaultValue": "2"
+    },
     "vspAppServicePlanInstances": {
       "type": "int",
       "defaultValue": 2
-    },
-    "vspAppServiceDockerImage": {
-      "type": "string"
     },
     "verifyHubPossibleOutboundIpAddresses": {
       "type": "array"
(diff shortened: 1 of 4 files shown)


azure/resource_groups/app/template.json
```diff
@@ -32,9 +32,33 @@
...
+    },
+    "appServicePlanTier": {
+      "type": "string",
+      "defaultValue": "Standard"
+    },
+    "appServicePlanSize": {
+      "type": "string",
+      "defaultValue": "1"
+    },
+    "vspAppServicePlanInstances": {
```
erbridge: Can this move down to live alongside the other VSP app service plan parameters?

azure/resource_groups/app/template.json
```diff
@@ -32,9 +32,33 @@
...
     "vspAppServiceDockerImage": {
       "type": "string"
     },
+    "vspAppServicePlanTier": {
+      "type": "string",
+      "defaultValue": "Standard"
+    },
+    "vspAppServicePlanSize": {
+      "type": "string",
+      "defaultValue": "2"
```
erbridge: Is there a reason for this to have a different default from the web app service plan?

azure/resource_groups/app/template.json
```diff
@@ -32,9 +32,33 @@
...
+      "type": "string",
+      "defaultValue": "Standard"
+    },
+    "appServicePlanSize": {
+      "type": "string",
+      "defaultValue": "1"
+    },
+    "vspAppServicePlanInstances": {
+      "type": "int",
+      "defaultValue": 1
```
erbridge: Is there a reason for this to have a different default from the web app service plan?

erbridge: This should also be set to 2 on `production` for the same reasons that the web app service will be.

azure/resource_groups/app/template.json
pezholio: Nah, I guess I kinda saw the VSP as an ancilliary service from the main app, but there's a good chance it'll get just as much traffic as the main app, so happy to change it and review usage as we get real traffic

PR #261, opened 2019-08-12
Change the claim received email

titlescreen:
- Update the email with copy from content designer
- Inset the reference number so it will be easily visible against the
  headers below
- Drop in merge fields for dynamic information
- Amend mailer spec so it now epects some of the current text

### DIFF
--- a/app/views/claim_mailer/submitted.text.erb
+++ b/app/views/claim_mailer/submitted.text.erb
@@ -1,16 +1,22 @@
-You application for student loan repayment reimbursement between
-6 April 2018 and 5 April 2019 has been received.
+Dear <%= @claim.full_name %>,
 
-^ Your reference number is <%= @claim.reference %>
+We‘ve received your claim to get back the student loan repayments you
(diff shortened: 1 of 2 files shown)


app/views/claim_mailer/submitted.text.erb
```diff
@@ -1,16 +1,22 @@
...
 
-We have sent your application to the Department for Education.
+#What happens next
 
-They will verify the information you have provided and then send
-you the outcome of your application within X weeks.
+We‘ll contact the school to confirm the details you gave us.
 
-You will receive email updates as your application progresses.
+We‘ll tell you if your claim was successful within XXXX.
```
pezholio: Is this `XXXX` deliberate?

titlescreen: Yeah we don't know the timings of this atm so I put in the standard this needs to be updated later placeholder. :)

pezholio: Ah, cool. I figured that was the case 😄

PR #262, opened 2019-08-12
Adjust escaping of formula-triggering characters

tekin:
Just wrapping the entire value in ="STRING_HERE" is not going to prevent
a formula triggering, for example a value of test"+SUM(A1- A2)+" would
get escaped to ="test"+SUM(A1- A2)+"", which would still attempt to
evaluate the entered info as a formula.

Instead we escape the first character if it is one of the known
formula-triggering characters. This may result in operability issues
further down the line, but these will be easy to deal with by the human
operator that will be working with the files. The excess of backslashes
in the gsub is the only way to add the '\\' escape character and match
substitution character '\0'.

See http://georgemauer.net/2017/10/07/csv-injection.html for more
details of how opening CSV files with Excel can result in formula
injection.

### DIFF
--- a/lib/excel_utils.rb
+++ b/lib/excel_utils.rb
@@ -1,0 +1,13 @@
+module ExcelUtils
+  FORMULA_TRIGGERS = %w[- + = @].freeze
+
+  # Escapes potentially formula-triggering characters from the front of the
+  # string, i.e. +,-,=,@
+  def self.escape_formulas(string)
+    if ExcelUtils::FORMULA_TRIGGERS.include?(string&.chr)
+      "\\" + string
+    else
+      string
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


lib/excel_utils.rb
```diff
@@ -0,0 +1,9 @@
+module ExcelUtils
+  FORMULA_TRIGGERS = %w[- + = @].freeze
+
+  # Escapes potentially formula-triggering characters from the front of the
+  # string, i.e. +,-,=,@
+  def self.escape_formulas(string)
+    string&.gsub(/^[-+=@]/, '\\\\\0')
```
pezholio: Should this not be using the `FORMULA_TRIGGERS` constant? I also wonder whether we should do something like this:

```ruby
def self.escape_formulas(string)
   return string unless FORMULA_TRIGGERS.include?(string[0])
  
   '\' + string
end
```

To save running `gsub` against every single string and to be a bit clearer as to the intent of the code (I usually avoid interpolation like that, but it avoids a rabbit hole of slash escaping)

tekin: 👍🏽I've adjusted the method

PR #266, opened 2019-08-13
Enable strict content security policy

tekin:
We only allow scripts, images and styles served over https and from our own domain. This will need updating as and when we add third-party scripts, for example Google Analytics.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -15,3 +15,4 @@
 //= require a11y-dialog/a11y-dialog.min.js
 //= require accessible-autocomplete/dist/accessible-autocomplete.min.js
 //= require_tree .
+//= stub js_check.js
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/application.js
```diff
@@ -15,3 +15,4 @@
 //= require a11y-dialog/a11y-dialog.min.js
 //= require accessible-autocomplete/dist/accessible-autocomplete.min.js
 //= require_tree .
+//= stub js_check.js
```
pezholio: What does `stub` do here?

tekin: It excludes the file from the asset bundle. We don't want it in there as we're including it inline in the layout, otherwise the JS would get executed twice – see https://github.com/rails/sprockets#stub

tekin: I've updated the commit message with this info!

PR #268, opened 2019-08-13
Add in text about how we use data with DQT and TPS

titlescreen:
As we are accessing the Database of Qualified Teachers and the Teacher
Pension Service, need need to explain to users how this information will
be used and who will access it.

### DIFF
--- a/app/views/static_pages/privacy_policy.html.erb
+++ b/app/views/static_pages/privacy_policy.html.erb
@@ -103,9 +103,7 @@
 
     <p class="govuk-body">
       To process your claim we will compare the information you have provided with records kept on the Teachers
-      Pension Service and the Database of Qualified Teachers. This is so we can validate your claim, we will not allow
-      the Teachers Pension Service or the Database of Qualified Tachers acees to the information you have provided to
-      this service.
+      Pension Service and the Database of Qualified Teachers. This is so we can validate your claim.
     </p>
 
     <h2 class="govuk-heading-l">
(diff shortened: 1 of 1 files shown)


app/views/static_pages/privacy_policy.html.erb
```diff
@@ -97,6 +97,17 @@
       </li>
     </ul>
 
+    <h2 class="govuk-heading-l">
+      How we access your personal data
+    </h2>
+
+    <p class="govuk-body">
+      To process your claim we will compare the information you have provided with records kept on the Teachers
+      Pension Service and the Database of Qualified Teachers. This is so we can validate your claim, we will not allow
```
tekin:
```suggestion
      Pension Service and the Database of Qualified Teachers. This is so we can validate your claim. we will not allow
```

app/views/static_pages/privacy_policy.html.erb
```diff
@@ -97,6 +97,17 @@
...
+      the Teachers Pension Service or the Database of Qualified Tachers acees to the information you have provided to
```
tekin:
```suggestion
      the Teachers Pension Service or the Database of Qualified Tachers access to the information you have provided to
```

app/views/static_pages/privacy_policy.html.erb
titlescreen: Amended this and added in the ’validate your claim. We will not allow’.

PR #271, opened 2019-08-14
Update accessiblity and T&Cs with support email

tekin:
We use the service-wide support email address on these pages as we can't know which particular policy the user will be interested in when viewing these pages.

### DIFF
--- a/app/views/static_pages/accessibility_statement.html.erb
+++ b/app/views/static_pages/accessibility_statement.html.erb
@@ -47,13 +47,9 @@
     </h2>
 
     <p class="govuk-body">
-      If you need information on this website in a different format or cannot use the service:
+      If you need information on this website in a different format or cannot use the service,
+      contact us at <%= mail_to t("support_email_address"), t("support_email_address"), class: "govuk-link" %>.
     </p>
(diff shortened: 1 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -6,4 +6,8 @@ def page_title(title)
   def claim_in_progress?
     session.key?(:claim_id)
   end
+
+  def support_email_address
+    "additionalteachingpayment@digital.education.gov.uk"
+  end
```
erbridge: In here rather than the i18n file?

tekin: I hadn't considered using the locale file for this.. I guess it was partly in anticipation of the method having to return a different support email under different circumstances (i.e. for the two different policies), but perhaps that's a touch of YAGNI. Should I just use the locale file instead?

erbridge: That's my hunch. It's what we're doing for e.g. journey name, so it would be consistent.

tekin: ok, I've updated it to load from the locale file instead

PR #272, opened 2019-08-14
Add a healthcheck endpoint

pezholio:
This adds a simple endpoint that returns 200 and a JSON response, so we can quickly check the service is up


tekin:
Do we want this health check to also establish that there is a connection with the database?

pezholio:
My first instinct was yes, but we'll get errors from Rollbar (or whatever we end up using) if the connection to the database is down

pezholio:
Ha, you're right. It's not actually JSON, it's just plaintext.

PR #273, opened 2019-08-14
Remove the KeyVault activity alerts

robbiepaul:
These are proving too noisy, we get several every time someone deploys something. We can always reinstate these later when the service is more stable.

### DIFF
--- a/azure/resource_groups/alerts/template.json
+++ b/azure/resource_groups/alerts/template.json
@@ -36,9 +36,6 @@
     },
     "databaseServerId": {
       "type": "string"
-    },
-    "keyVaultId": {
-      "type": "string"
     }
   },
   "variables": {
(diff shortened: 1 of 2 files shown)


azure/resource_groups/alerts/template.json
```diff
@@ -36,9 +36,6 @@
     },
     "databaseServerId": {
       "type": "string"
-    },
-    "keyVaultId": {
-      "type": "string"
```
erbridge: You'll need to also remove this from `bin/azure-deploy`.

robbiepaul: Thanks F, I've added a commit that I think removes it from the right place 👍

bin/azure-deploy
```diff
@@ -232,7 +232,7 @@ if [ $DEPLOY_ALERTS ]; then
     --parameters "$(
       filter-azure-outputs \
         "$SECRETS_DEPLOYMENT_RESULT" \
-        "['keyVaultId']" \
+        "[]" \
```
erbridge: 👍

PR #274, opened 2019-08-14
Add in Google Tag Manager

titlescreen:
So we can add analytics and make it easily maintained in the future we
added in Google Tag Manager to the layout.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -18,25 +18,10 @@
 
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
 
-    <!-- Google Tag Manager -->
-    <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
-    new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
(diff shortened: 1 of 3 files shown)


app/views/layouts/application.html.erb
```diff
@@ -18,10 +18,25 @@
 
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
 
+    <!-- Google Tag Manager -->
+    <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+    new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+    j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+    'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+    })(window,document,'script','dataLayer','GTM-P5C5JJ2');</script>
```
robbiepaul: We should probably make this value an environment variable. We won't want it set in development and might want to use a different container on staging so it doesn't pollute our analytics.

Let me know if you need a hand with it

titlescreen: Thanks Robbie, I've roped int Tekin as the CSP doesn't see to be working for this.

PR #275, opened 2019-08-14
GOV.UK Verify support for EU citizens

robbiepaul:
GOV.UK Verify returns a different set of data for EU citizens, which is missing their address. Therefor, for EU citizens to be able to use the service, we need to ask them to provide their address explicitly.

We also need to enable the eIDAS feature on our VSP.

See https://www.docs.verify.service.gov.uk/using-verify-data/data-from-verify/#european-identities-and-eidas

I also took the opportunity to add a feature spec for users being able to provide a payroll gender that got lost as part of some other refactoring work.

### DIFF
--- a/lib/verify/response.rb
+++ b/lib/verify/response.rb
@@ -42,6 +42,7 @@ module Verify
         address_line_1: address_line(1),
         address_line_2: address_line(2),
         address_line_3: address_line(3),
+        address_line_4: address_line(4),
         postcode: postcode,
         date_of_birth: attributes.fetch("datesOfBirth").first.fetch("value"),
         payroll_gender: gender,
--- a/lib/verify/response.rb
+++ b/lib/verify/response.rb
@@ -39,10 +39,10 @@ module Verify
     def identity_paramteters
       @identity_paramteters ||= {
         full_name: full_name,
-        address_line_1: address_lines[0],
-        address_line_2: address_lines[1],
-        address_line_3: address_lines[2],
-        postcode: address.fetch("postCode"),
+        address_line_1: address_line(1),
+        address_line_2: address_line(2),
+        address_line_3: address_line(3),
+        postcode: postcode,
         date_of_birth: attributes.fetch("datesOfBirth").first.fetch("value"),
         payroll_gender: gender,
       }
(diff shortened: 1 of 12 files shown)


robbiepaul:
[GOV.UK Verify 6. Support for EU Citizens (different Verify flow/info)](https://trello.com/c/05yrEEzA/431-govuk-verify-6-support-for-eu-citizens-different-verify-flow-info)

spec/features/govuk_verify_spec.rb
```diff
@@ -53,6 +53,48 @@
       @claim.reload
       expect(@claim.full_name).to eql("Isambard Kingdom Brunel")
     end
+
+    context "with unspecified gender parameter" do
```
tekin: I think we have this coverage already on `missing_verify_information_spec.rb`?

lib/verify/response.rb
```diff
@@ -39,10 +39,10 @@ def scenario
...
       @identity_paramteters ||= {
         full_name: full_name,
-        address_line_1: address_lines[0],
-        address_line_2: address_lines[1],
-        address_line_3: address_lines[2],
-        postcode: address.fetch("postCode"),
+        address_line_1: address_line(1),
+        address_line_2: address_line(2),
+        address_line_3: address_line(3),
+        postcode: postcode,
```
erbridge: No address line 4?

spec/features/govuk_verify_spec.rb
robbiepaul: ah yes, I missed that!

lib/verify/response.rb
```diff
@@ -69,7 +69,7 @@ def full_name
     end
 
     def most_recent_verified_value(attributes)
-      return if attributes.empty?
+      return if attributes.blank?
```
tekin: Sounds like what you're saying is that this guard is only meant to trigger when `attributes` is `nil`. In which case, would it make more sense for this to read `attributes.nil?`?

lib/verify/response.rb
robbiepaul: It's not clear how many address lines are returned from Verify. I'll add it in just in case it is returned

lib/verify/response.rb
robbiepaul: It's possible for `middleNames` to come back as an empty array though

tekin:
> Do we still have the form to collect the address if it's missing? I've lost track.

Yep, we never lost that, it still exists as `app/views/claim/address.html.erb`

robbiepaul:
@tekin Yeah I agree, maybe it's something we could look at as part of https://trello.com/c/oWyInDUE/557-store-the-entire-govuk-verify-response-to-aid-debugging-during-early-beta?


@erbridge I've added a commit to allow for 4 lines of address

PR #276, opened 2019-08-14
Add google analytics to the site

titlescreen:
Also update the cookies page with Google info

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -18,6 +18,10 @@
 
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
 
+
+    <%= javascript_include_tag "https://www.googletagmanager.com/gtag/js?id=UA-145681604-1" %>
+    <%= javascript_include_tag "google_analytics" %>
+
     <%= stylesheet_link_tag    'application', media: 'all' %>
   </head>
 
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -18,9 +18,11 @@
 
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
 
+    <% if ENV["GOOGLE_ANALYTICS_ID"] %>
 
-    <%= javascript_include_tag "https://www.googletagmanager.com/gtag/js?id=UA-145681604-1" %>
-    <%= javascript_include_tag "google_analytics" %>
+      <%= javascript_include_tag "https://www.googletagmanager.com/gtag/js?id=#{ENV["GOOGLE_ANALYTICS_ID"]}" %>
+      <%= javascript_include_tag "google_analytics", data:{"ga-id" => ENV["GOOGLE_ANALYTICS_ID"]} %>
+    <% end %>
 
     <%= stylesheet_link_tag    'application', media: 'all' %>
   </head>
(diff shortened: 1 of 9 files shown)


app/views/static_pages/cookies.html.erb
```diff
@@ -67,21 +67,21 @@
...
+          <td class="govuk-table__cell">_ga</td>
           <td class="govuk-table__cell">This helps us count how many people visit <%= t("service_name") %> by tracking if you’ve visited before</td>
           <td class="govuk-table__cell">After 2 years</td>
         </tr>
         <tr>
           <tr class="govuk-table__row">
-          <td class="govuk-table__cell">XXXX_gid</td>
+          <td class="govuk-table__cell">_gid</td>
           <td class="govuk-table__cell">This helps us count how many people visit <%= t("service_name") %> by tracking if you’ve visited before</td>
           <td class="govuk-table__cell">After 24 hours</td>
```
erbridge: Are we confident that these are the cookies used by the tag manager, too?

erbridge:
This needs to set up the infrastructure side of things to pass the env var in before we'll be able to turn it on.

app/views/static_pages/cookies.html.erb
titlescreen: When we were looking at implementing Tag Manager it gave people too much opportunity to be naughty so we changed it to standard Google Analytics. It is using the Google Analytics Global Tag, not Google Tag Manager.

erbridge: Ah ok. It looks to me like we're still including the tag manager script in the header, though. Or am I mistaken?

app/views/layouts/application.html.erb
```diff
@@ -18,6 +18,12 @@
 
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
 
+    <% if ENV["GOOGLE_ANALYTICS_ID"] %>
+
+      <%= javascript_include_tag "https://www.googletagmanager.com/gtag/js?id=#{ENV["GOOGLE_ANALYTICS_ID"]}" %>
```
erbridge: Does this line belong here if we're not using Tag Manager?

tekin: It's what Google say to "copy and paste" into the site to enable Google Analytics, along with the code below.

erbridge: Ah ok. Misled by the name.

tekin: Yeah, it's confusing, but it looks like even without using full Tag Manager, Google Analytics is still loaded using this JS.

azure/resource_groups/app/parameters/development.template.json
```diff
@@ -139,6 +139,14 @@
         "secretName": "DfeSignInApiEndpoint"
       }
     },
+    "GOOGLE_ANALYTICS_ID": {
+      "reference": {
+        "keyVault": {
+          "id": "${keyVaultId}"
+        },
+        "secretName": "GoogleAnalyticsId"
```
robbiepaul: We could argue that this doesn't need to be stored in KeyVault as it's not secret. We could just provide it as a value?

erbridge: Putting it in the Key Vault gives us some advantages in that we can change it and redeploy without committing changes. Though it does add some indirection, and increase the number of "secrets" we store that aren't secret.

PR #278, opened 2019-08-14
Tighten Content-Security-Policy

tekin:
Only allow whitelisted JavaScript and CSS, which at the moment means anything from our domain, plus JavaScript from the Google Analytics and Tag Manager domains.

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -5,12 +5,12 @@
 # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
 
 Rails.application.config.content_security_policy do |policy|
-  policy.default_src :self, :https
+  policy.default_src :self
   policy.font_src :self, :https, :data
   policy.img_src :self, :https, :data
   policy.object_src :none
-  policy.script_src :self, :https
-  policy.style_src :self, :https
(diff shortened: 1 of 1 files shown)


config/initializers/content_security_policy.rb
```diff
@@ -5,12 +5,12 @@
 # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
 
 Rails.application.config.content_security_policy do |policy|
-  policy.default_src :self, :https
+  policy.default_src :self
   policy.font_src :self, :https, :data
   policy.img_src :self, :https, :data
   policy.object_src :none
-  policy.script_src :self, :https
-  policy.style_src :self, :https
```
erbridge: What effect does dropping `:https` from these have?

tekin: These lines define whitelists. The old version meant anything from `:self` (ie our domain) and anything over `https`. Removing `:https` means we no longer allow external styles or scripts.

tekin: More info: https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP

erbridge: Ah ok. I thought it was restricting to HTTPS only from self.

PR #280, opened 2019-08-14
Set up healthcheck alerts

erbridge:
This GETs /healthcheck from 6 regions and fails if at least 3 of those regions return anything other than a 200 response within 30 seconds.

It lives in `app`, despite being an alert, because it needs to be in the same resource group as the Application Insights resource, and the web app service depends on that.

Inline XML is less than ideal, but we need to modify it at runtime making it difficult to pull out into a file.


erbridge:
[We are alerted when the service is down](https://trello.com/c/jpn1jkpf/493-we-are-alerted-when-the-service-is-down)

PR #281, opened 2019-08-14
Filter out or anonymise personally identifiable information

robbiepaul:
The disadvantage of maintaining a separate list of these fields in `filter_parameter_logging.rb` means we'll have to be vigilant with updating this list when new sensitive parameters get added. I considered extracting them out to a separate constant so they'd be more visible in our codebase but we'd have a similar problem of keeping it up-to-date.

### DIFF
--- a/config/initializers/filter_parameter_logging.rb
+++ b/config/initializers/filter_parameter_logging.rb
@@ -3,18 +3,5 @@
 # Configure sensitive parameters which will be filtered from the log file.
 Rails.application.config.filter_parameters += [:password]
 
-# Personally identifiable information
-Rails.application.config.filter_parameters += [
-  :address_line_1,
-  :address_line_2,
-  :address_line_3,
-  :address_line_4,
-  :postcode,
-  :payroll_gender,
-  :teacher_reference_number,
-  :national_insurance_number,
-  :email_address,
-  :bank_sort_code,
-  :bank_account_number,
-  :SAMLResponse,
-]
+# Personally identifiable information in a claim
+Rails.application.config.filter_parameters += Claim.filtered_params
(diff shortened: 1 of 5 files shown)


robbiepaul:
[Sanitise PII (and SAML payloads) fields/parameters from logs/etc](https://trello.com/c/uNKaz3RE/572-sanitise-pii-and-saml-payloads-fields-parameters-from-logs-etc)

config/initializers/filter_parameter_logging.rb
```diff
@@ -2,3 +2,19 @@
...
+  :address_line_4,
+  :postcode,
+  :payroll_gender,
+  :teacher_reference_number,
+  :national_insurance_number,
+  :email_address,
+  :bank_sort_code,
+  :bank_account_number,
+  :SAMLResponse
+]
```
erbridge: Can we put these in the model? I suspect it will be easier to remember to keep them up to date if they're nearer to the code they refer to.

Also, if we want to test that we've made active decisions about all attributes, we could make it a hash, where the value determines whether or not we filter them, and we check that all the possible fields are a key in that hash.

For example on the `Claim` model:

```ruby
FILTER_PARAMETERS = {
  address_line_1: true,
  address_line_2: true,
  student_loan_plan: false,
}.freeze
```

tekin: I really like this idea. We could have a spec that fails unless all attributes on the Claim model have been explicitly designated as filtered or not. So any future columns added without updating the hash would fail the test suite.  👍🏽

robbiepaul:
@erbridge ready for another look. I've added a `.filtered_params` static method to the claim model

PR #282, opened 2019-08-14
Pin the platform building blocks to a specific commit

erbridge:
They just made their first backwards incompatible change (changing the way database firewall rules are named), and we want to explicitly update rather than grabbing `master` every time.

This also makes rolling back more reliable, as we will roll back our dependency versions, too.

### DIFF
--- a/azure/resource_groups/app/template.json
+++ b/azure/resource_groups/app/template.json
@@ -138,7 +138,7 @@
     }
   },
   "variables": {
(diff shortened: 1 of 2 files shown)


erbridge:
[Pin to a specific building blocks version](https://trello.com/c/SmFqR1LQ/578-pin-to-a-specific-building-blocks-version)

erbridge:
The issue was more around changes in names creating duplicate database firewall rules. I'm putting in a PR today to fix the bug.

PR #288, opened 2019-08-15
Recreate the Heroku start page within our app

pezholio:
This replicates the start page on https://dfe-tslr-start-page.herokuapp.com/, and changes the root path to direct users to there by default. I'm not sure this is what we'll want to do long term (especially when we move to support Maths and Physics claims too), but it's enough for private beta.

### DIFF
--- a/app/views/static_pages/start_page.html.erb
+++ b/app/views/static_pages/start_page.html.erb
@@ -1,0 +1,36 @@
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-xl">
+        <%= t('student_loans.journey_name') %>
+      </h1>
+
+      <p class="govuk-body">
+        Claim student loan repayments if you:
+      </p>
+
+      <ul class="govuk-list govuk-list--bullet">
+        <li>
+          teach biology, chemistry, physics, computer science or languages (not including English)
+          for over half your working hours, including your planning, preparation and assessment (PPA) time
+        </li>
+        <li>have been teaching for 1 to 5 years</li>
+        <li>teach in certain places - <a href="https://www.gov.uk/guidance/teachers-student-loan-reimbursement-guidance-for-teachers-and-schools#eligibility-criteria">check which places are eligible</a></li>
+        <li>received qualified teacher status in or after the 2013 to 2014 academic year</li>
+      </ul>
+
+      <h2 class="govuk-heading-m">Before you apply</h2>
+
+      <p class="govuk-body">You'll need:</p>
+
+      <ul class="govuk-list govuk-list--bullet">
+        <li>the exact amount you have repaid – get this from your annual student loan statement, your P60 or by contacting the <a href="https://www.gov.uk/guidance/contact-slc-repayment-enquiries">student loans company</a></li>
+        <li>your National Insurance Number</li>
+        <li>your bank account details</li>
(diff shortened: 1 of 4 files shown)


app/views/static_pages/start_page.html.erb
```diff
@@ -0,0 +1,36 @@
...
+        Claim student loan repayments if you:
+      </p>
+
+      <ul class="govuk-list govuk-list--bullet">
+        <li>
+          teach biology, chemistry, physics, computer science or languages (not including English)
+          for over half your working hours, including your planning, preparation and assessment (PPA) time
+        </li>
+        <li>have been teaching for 1 to 5 years</li>
+        <li>teach in certain places - <a href="https://www.gov.uk/guidance/teachers-student-loan-reimbursement-guidance-for-teachers-and-schools#eligibility-criteria">check which places are eligible</a>.</li>
```
titlescreen: I think we can drop the full stop at the end of the li so its consistent with the others.

app/views/static_pages/start_page.html.erb
```diff
@@ -0,0 +1,36 @@
...
+        <li>received qualified teacher status in or after the 2013 to 2014 academic year</li>
+      </ul>
+
+      <h2 class="govuk-heading-m">Before you apply</h2>
+
+      <p class="govuk-body">You'll need:</p>
+
+      <ul class="govuk-list govuk-list--bullet">
+        <li>the exact amount you have repaid – get this from your annual student loan statement, your P60 or by contacting the <a href="https://www.gov.uk/guidance/contact-slc-repayment-enquiries">student loans company</a>.</li>
```
titlescreen: As above I think we can remove the full stop, a minor thing but I think worth doing. Also I'm aware that I probably put that in on the original Heroku version :)

app/views/static_pages/start_page.html.erb
```diff
@@ -0,0 +1,36 @@
...
+        <li>
+          teach biology, chemistry, physics, computer science or languages (not including English)
+          for over half your working hours, including your planning, preparation and assessment (PPA) time
+        </li>
+        <li>have been teaching for 1 to 5 years</li>
+        <li>teach in certain places - <a href="https://www.gov.uk/guidance/teachers-student-loan-reimbursement-guidance-for-teachers-and-schools#eligibility-criteria">check which places are eligible</a></li>
+        <li>received qualified teacher status in or after the 2013 to 2014 academic year</li>
+      </ul>
+
+      <h2 class="govuk-heading-m">Before you apply</h2>
```
erbridge: Should this be `govuk-heading-l`? Isn't that the usual class for `h2`s?

pezholio: Not sure. I'll defer to @titlescreen on that one, I nabbed it wholesale from the repo.

pezholio:
[Recreate the Heroku start page within our app](https://trello.com/c/GZOoxPde/576-recreate-the-heroku-start-page-within-our-app)

app/views/static_pages/start_page.html.erb
titlescreen: Had a look on MPS prototype and https://www.gov.uk/check-mot-history to compare and see what others do and they are using the `govuk-heading-m` font size. The use of headings is sequential so we are good form accessibility point of view, I just don't think the before you start information needs to stand out enough to justify the `govuk-heading-l`.

PR #290, opened 2019-08-15
Refactor Azure templates

erbridge:


erbridge:
Yes. It's a bit difficult to be certain, but it deployed and I changed very little other than pulling things out.

PR #291, opened 2019-08-15
Change how we show academic years

titlescreen:
Because screen readers don't read out "-" as a word we should say
e.g. September 1 2013 to August 31 2014

### DIFF
--- a/app/helpers/claims_helper.rb
+++ b/app/helpers/claims_helper.rb
@@ -60,7 +60,8 @@ module ClaimsHelper
     return unless year_range.present?
 
     start_year, end_year = year_range.split("_")
-
-    "September 1 #{start_year} - August 31 #{end_year}"
+    start_date = Date.new(start_year.to_i, 9, 1)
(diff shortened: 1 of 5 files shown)


app/helpers/claims_helper.rb
```diff
@@ -61,6 +61,6 @@ def academic_years(year_range)
 
     start_year, end_year = year_range.split("_")
 
-    "September 1 #{start_year} - August 31 #{end_year}"
+    "September 1 #{start_year} to August 31 #{end_year}"
```
erbridge: While we're here, can we switch to using the localization function: `l(date)`? Dates should be `1 September 2013`, not `September 1 2019` according to the design system, and i18n is set up to do that.

config/locales/en.yml
```diff
@@ -79,13 +79,13 @@ en:
       qts_award_year: "Which academic year did you complete your initial teacher training?"
       qts_award_years:
         before_2013: "Before September 1 2013"
```
erbridge: This should also be "1 September 2013".

PR #294, opened 2019-08-15
Have a support role within the app

pezholio:
This adds a support for a support role in the app. If a user logs in to the admin section with a role of `teacher_payments_support`, then they are logged in, but cannot download the full CSV of claims

### DIFF
--- a/spec/features/admin_csv_download_spec.rb
+++ b/spec/features/admin_csv_download_spec.rb
@@ -1,7 +1,7 @@
 require "rails_helper"
 
 RSpec.feature "Download CSV of claims" do
-  context "User is logged in" do
+  context "User is logged in as an admin user" do
     before do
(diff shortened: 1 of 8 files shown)


spec/features/admin_csv_download_spec.rb
```diff
@@ -21,6 +21,24 @@
...
 
+  context "User is logged in as a support user" do
+    before do
+      stub_dfe_sign_in_with_role(Admin::AuthController::DFE_SIGN_IN_SUPPORT_ROLE_CODE)
+      visit admin_path
+      click_on "Sign in"
+    end
+
+    scenario "User cannot see a link to download claims" do
+      expect(page).to_not have_content("Download claims")
```
erbridge: Can we test this more robustly? If the copy changed at the same time as the logic to turn it on and off, this would still pass. Can we look for the link by the href instead?

pezholio:
[Have a Support role within the app](https://trello.com/c/sXvUHfsF/574-have-a-support-role-within-the-app)

PR #295, opened 2019-08-15
Store the entire GOV.UK Verify response to aid debugging during early beta

pezholio:
This adds an additional JSON column to the Claims table where the entirity of the Verify response is called, regardless of if it's a succcessful verification or not. How we actually access this data (given we don't have SSH access) is a different matter

### DIFF
--- a/app/controllers/verify/authentications_controller.rb
+++ b/app/controllers/verify/authentications_controller.rb
@@ -23,6 +23,7 @@ module Verify
         current_claim.update!(@response.claim_parameters)
         redirect_to claim_url("verified")
       else
+        current_claim.update!(verify_response: @response.parameters)
         redirect_to verify_path_for_response_scenario(@response.scenario)
       end
     end
--- a/db/migrate/20190815151449_add_verify_response_to_claim.rb
+++ b/db/migrate/20190815151449_add_verify_response_to_claim.rb
@@ -1,0 +1,5 @@
+class AddVerifyResponseToClaim < ActiveRecord::Migration[5.2]
+  def change
+    add_column :claims, :verify_response, :json
+  end
+end
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -32,6 +32,7 @@ class Claim < ApplicationRecord
     submitted_at: false,
     updated_at: false,
     verified_fields: false,
+    verify_response: true,
   }.freeze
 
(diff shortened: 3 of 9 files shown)


pezholio:
[Store the entire GOV.UK Verify response to aid debugging during early beta](https://trello.com/c/oWyInDUE/557-store-the-entire-govuk-verify-response-to-aid-debugging-during-early-beta)

db/migrate/20190815151449_add_verify_response_to_claim.rb
```diff
@@ -0,0 +1,5 @@
+class AddVerifyResponseToClaim < ActiveRecord::Migration[5.2]
+  def change
+    add_column :claims, :verify_response, :json, null: false, default: "{}"
```
erbridge: Does it make sense to be setting a default here? Doesn't it make more sense for it to default to null so we can tell the difference between those who have been through Verify, and those who haven't?

pezholio: From what I remember, if you don't set a default in JSON columns, it can cause unexpected behaviour

app/controllers/verify/authentications_controller.rb
```diff
@@ -23,6 +23,7 @@ def create
         current_claim.update!(@response.claim_parameters)
         redirect_to claim_url("verified")
       else
+        current_claim.update!(verify_response: @response.parameters)
```
tekin: This means we would only store unverified responses, which is not actually that interesting; they should all look something like `{ "scenario": "AUTHENTICATION_FAILED" }`. The intent of this story is to store the response for "verified" claims so we can deal with situations where, for example, we find the address we're extracted doesn't match the address on record.

tekin: oh, just noticed that this is in addition to setting the response in `claim_parameters`. ignore!

db/migrate/20190815151449_add_verify_response_to_claim.rb
tekin: @pezholio can you confirm this? In my mind we shoudn't need a default, unless we're planning on running queries against the JSON, which I can't see us needing to do here.

pezholio: Yeah, I've removed it and it seems fine. I guess the only time it would cause problems would be if we were querying the JSON

PR #296, opened 2019-08-15
Reverse the 50% question to ask about leadership duties

erbridge:
This adds a new question into the flow that asks if the user was in a leadership position. If they were, they are then asked if that took up more than half of their time. If they weren't, they are no longer asked about how much time they spent teaching eligible subjects.


erbridge:
[Implement reverse 50% question in application](https://trello.com/c/KFcpP9fN/494-implement-reverse-50-question-in-application)

app/views/claims/subjects_taught.html.erb
```diff
@@ -21,9 +21,9 @@
               <% end %>
               <div class="govuk-radios__divider">or</div>
               <div class="govuk-radios__item">
-                <%= fields.hidden_field :mostly_teaching_eligible_subjects, value: '' %>
-                <%= fields.radio_button(:mostly_teaching_eligible_subjects, false, class: "govuk-radios__input")  %>
-                <%= fields.label :mostly_teaching_eligible_subjects_false, t('student_loans.questions.eligible_subjects.not_applicable'), class: "govuk-label govuk-radios__label" %>
+                <%= fields.hidden_field :mostly_teaching, value: '' %>
+                <%= fields.radio_button(:mostly_teaching, false, class: "govuk-radios__input")  %>
+                <%= fields.label :mostly_teaching_false, t('student_loans.questions.eligible_subjects.not_applicable'), class: "govuk-label govuk-radios__label" %>
```
erbridge: This is technically wrong. Just because the user hasn't taught eligible subjects doesn't mean they weren't teaching enough. I've also just realized that that we currently share an eligibility failure reason between this and the case where a teacher did mostly leadrship stuff, which needs to change. I'd like to do this without requiring another database field just for this, but I'm not sure how to do that reliably with `subjects_taught` as an array. Any thoughts?

tekin: @erbridge can we check with policy? I think they definitely need to be teaching at least one of these subjects to be eligible.

erbridge:
Ah good spot. I think I agree. I'll make that change.

app/views/claims/subjects_taught.html.erb
tekin: oh, just seen what you mean... I'll have a ponder but it might be that this requires an additional database column..

erbridge: Yeah, that's what I'm thinking. Since I'm changing the columns, I'll leave the existing column (with a rename) for now.

app/models/claim_update.rb
```diff
@@ -23,6 +23,8 @@ class ClaimUpdate
 
   DEPENDENT_ELIGIBILITY_ANSWERS = {
     "claim_school_id" => "employment_status",
+    "subjects_taught" => "had_leadership_position",
```
tekin: Are these two actually dependent?

app/models/page_sequence.rb
```diff
@@ -14,7 +14,8 @@ class PageSequence
     "still-teaching",
     "current-school",
     "subjects-taught",
-    "mostly-teaching-eligible-subjects",
+    "leadership-position",
+    "mostly-not-teaching",
```
tekin: The "not" in this slug I think makes it hard to parse. Is there a phrasing that works without a "not"?

erbridge: Changed to talk about the leadership question it asks.

app/models/claim_update.rb
erbridge: Nope. It was a relic of an intermediary step. It's gone now.

app/views/claims/mostly_performed_leadership_duties.html.erb
```diff
@@ -0,0 +1,42 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render("shared/error_summary", instance: current_claim, errored_field_id_overrides: { mostly_performed_leadership_duties: "claim_eligibility_attributes_mostly_performed_leadership_duties_true"}) if current_claim.errors.any? %>
```
robbiepaul: Should this be `"eligibility.mostly_performed_leadership_duties": ` so that it links to the correct ID?

erbridge: Ugh. Yes. I suspect there will be others, since this was edited from the now-deleted `mostly_teaching_eligible_subjects` view.

erbridge: I fixed this and went through and fixes the rest of the eligibility error tags, too.

PR #298, opened 2019-08-16
Add a basic cookie banner

robbiepaul:
The styles are mostly copied from the "govuk_publishing_components" which GOV.UK uses as there's no official pattern in the Design System for it.

It will always show until the user clicks 'accept cookies'. No non-essential cookies should be loaded unless the user has clicked accept.

Google Analytics is considered non-essential so it won't start tracking until the page after they click "Accept cookies" on, this might skew our analytics slightly and we might want to tweak this/add events at a later date.

By default the cookie banner will always show to users without JavaScript, but won't have the accept button.

https://ico.org.uk/about-the-ico/news-and-events/news-and-blogs/2019/07/blog-cookies-what-does-good-look-like/

## Screenshot

Screenshot 2019-08-20 at 14 33 27

### DIFF
--- a/app/assets/javascripts/components/cookie_banner.js
+++ b/app/assets/javascripts/components/cookie_banner.js
@@ -1,28 +1,15 @@
 "use strict";
 
 document.addEventListener("DOMContentLoaded", function() {
-  var cookiePolicyName = "accept_cookie_policy";
-  var acceptCookies = getCookie(cookiePolicyName);
   var cookieBanner = document.querySelector("#global-cookie-message");
   var acceptButton = document.querySelector("#accept-cookies");
 
-  if (!acceptCookies) {
-    cookieBanner.style.display = "block";
+  if (!TeacherPayments.cookies.checkNonEssentialCookiesAccepted()) {
+    cookieBanner.classList.add("govuk-cookie-banner--visible");
   }
 
   acceptButton.onclick = function() {
-    setCookie(cookiePolicyName, 1, 365);
-    cookieBanner.style.display = "none";
+    TeacherPayments.cookies.acceptNonEssentialCookies();
+    cookieBanner.classList.remove("govuk-cookie-banner--visible");
   };
-
-  function getCookie(name) {
-    var v = document.cookie.match("(^|;) ?" + name + "=([^;]*)(;|$)");
-    return v ? v[2] : null;
-  }
-
-  function setCookie(name, value, days) {
-    var date = new Date();
-    date.setTime(date.getTime() + 24 * 60 * 60 * 1000 * days);
-    document.cookie = name + "=" + value + ";path=/;expires=" + date.toGMTString();
-  }
 });
--- a/app/assets/javascripts/components/cookie_banner.js
+++ b/app/assets/javascripts/components/cookie_banner.js
@@ -1,0 +1,28 @@
+"use strict";
+
+document.addEventListener("DOMContentLoaded", function() {
+  var cookiePolicyName = "accept_cookie_policy";
+  var acceptCookies = getCookie(cookiePolicyName);
+  var cookieBanner = document.querySelector("#global-cookie-message");
+  var acceptButton = document.querySelector("#accept-cookies");
+
+  if (!acceptCookies) {
+    cookieBanner.style.display = "block";
+  }
+
+  acceptButton.onclick = function() {
+    setCookie(cookiePolicyName, 1, 365);
+    cookieBanner.style.display = "none";
+  };
+
+  function getCookie(name) {
+    var v = document.cookie.match("(^|;) ?" + name + "=([^;]*)(;|$)");
+    return v ? v[2] : null;
+  }
+
+  function setCookie(name, value, days) {
+    var date = new Date();
+    date.setTime(date.getTime() + 24 * 60 * 60 * 1000 * days);
+    document.cookie = name + "=" + value + ";path=/;expires=" + date.toGMTString();
+  }
+});
--- a/app/assets/javascripts/cookie_functions.js
+++ b/app/assets/javascripts/cookie_functions.js
@@ -1,0 +1,20 @@
+window.TeacherPayments = window.TeacherPayments || {};
+
+window.TeacherPayments.cookies = {
+  acceptCookieName: "accept_cookies",
+  checkNonEssentialCookiesAccepted: function() {
+    return this._get(this.acceptCookieName);
+  },
+  acceptNonEssentialCookies: function() {
+    this._set(this.acceptCookieName, 1, 90);
+  },
+  _set: function(name, value, days) {
+    var date = new Date();
+    date.setTime(date.getTime() + 24 * 60 * 60 * 1000 * days);
+    document.cookie = name + "=" + value + ";path=/;expires=" + date.toGMTString();
+  },
+  _get: function(name) {
+    var v = document.cookie.match("(^|;) ?" + name + "=([^;]*)(;|$)");
+    return v ? v[2] : null;
+  }
+};
--- a/app/assets/javascripts/google_analytics.js
+++ b/app/assets/javascripts/google_analytics.js
@@ -1,7 +1,14 @@
-window.dataLayer = window.dataLayer || [];
-function gtag() {
-  dataLayer.push(arguments);
-}
-gtag("js", new Date());
+var currentScript = document.currentScript;
 
-gtag("config", document.currentScript.getAttribute("data-ga-id"), { anonymize_ip: true });
+document.addEventListener("DOMContentLoaded", function() {
+  if (!window.TeacherPayments.cookies.checkNonEssentialCookiesAccepted()) {
+    return;
+  }
+  window.dataLayer = window.dataLayer || [];
+  function gtag() {
+    dataLayer.push(arguments);
+  }
+  gtag("js", new Date());
+
+  gtag("config", currentScript.getAttribute("data-ga-id"), { anonymize_ip: true });
+});
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -34,6 +34,16 @@
 
     <a href="#main-content" class="govuk-skip-link">Skip to main content</a>
 
+    <div id="global-cookie-message" class="govuk-cookie-banner" role="region" aria-label="cookie banner">
+      <div class="govuk-width-container">
+        <p class="govuk-cookie-banner__message">This service uses cookies to make the site simpler.</p>
+        <div class="govuk-cookie-banner__buttons">
+          <button class="govuk-button govuk-button--secondary" type="submit" id="accept-cookies">Accept cookies</button>
+          <%= link_to('Cookie information', cookies_path, class: "govuk-button govuk-button--secondary ") %>
+        </div>
+      </div>
+    </div>
+
     <header class="govuk-header " role="banner" data-module="header">
       <div class="govuk-header__container govuk-width-container">
 
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -36,7 +36,7 @@
 
     <div id="global-cookie-message" class="govuk-cookie-banner" role="region" aria-label="cookie banner">
       <div class="govuk-width-container">
-        <p class="govuk-cookie-banner__message">This service uses cookies to make the site simpler.</p>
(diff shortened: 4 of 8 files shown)


robbiepaul:
@titlescreen Can you have a look at the styling as there's a weird vertical alignment issue happening that I couldn't fix

robbiepaul:
[Cookie banner](https://trello.com/c/xILDS1N2/523-cookie-banner)

app/views/layouts/application.html.erb
```diff
@@ -34,6 +34,16 @@
...
 
+    <div id="global-cookie-message" class="govuk-cookie-banner" role="region" aria-label="cookie banner">
+      <div class="govuk-width-container">
+        <p class="govuk-cookie-banner__message">This service uses cookies to make the site simpler.</p>
+        <div class="govuk-cookie-banner__buttons">
+          <button class="govuk-button govuk-button--secondary" type="submit" id="accept-cookies">Accept cookies</button>
+          <%= link_to('Cookie information', cookies_path, class: "govuk-button govuk-button--secondary ") %>
+        </div>
+      </div>
+    </div>
```
erbridge: Should this default to hidden for the case of no-JS?

app/views/static_pages/cookies.html.erb
```diff
@@ -66,6 +66,11 @@
         </tr>
       </thead>
       <tbody>
+        <tr class="govuk-table__row">
+          <td class="govuk-table__cell">accept_cookie_policy</td>
+          <td class="govuk-table__cell">This lets us know that you have accepted our cookie policy</td>
+          <td class="govuk-table__cell">After 1 year</td>
+        </tr>
```
erbridge: This is the table for Google Analytics. This should be in its own section, I think.

app/assets/javascripts/components/cookie_banner.js
```diff
@@ -0,0 +1,28 @@
+"use strict";
+
+document.addEventListener("DOMContentLoaded", function() {
+  var cookiePolicyName = "accept_cookie_policy";
+  var acceptCookies = getCookie(cookiePolicyName);
+  var cookieBanner = document.querySelector("#global-cookie-message");
+  var acceptButton = document.querySelector("#accept-cookies");
+
+  if (!acceptCookies) {
+    cookieBanner.style.display = "block";
```
erbridge: Can we do this with classes instead of modifying style properties?

app/views/layouts/application.html.erb
robbiepaul: I'll hide the accept button for non-JS users as per: https://github.com/alphagov/govuk_publishing_components/pull/948

erbridge:
In order for the accept button to be meaningful, we should be checking for the cookie it sets before doing anything else that enables cookies, shouldn't we? Otherwise we've set cookies without the user's concent.

robbiepaul:
I think it needs some thinking about. If we don't set the session cookie the user won't be able to continue their journey at all.

The Service Manual just states we should have a banner that informs the user we're using cookies on their first visit, so maybe we should just remove the accept button and do that?

https://www.gov.uk/service-manual/technology/working-with-cookies-and-similar-technologies

titlescreen:
I think it would be cool to have the accept cookies actually enable the option cookies (cookie banner, GA etc) but I feels like a lot of work to get it in. 

My suggestion would be to remove the accept cookies button because as F said, it's not really doing much at the moment and just notify them about cookies. Perhaps future us can think about adding the accept button back and improving upon what we have?

titlescreen:
@robbiepaul the fix for the vertical alignment is now up.

robbiepaul:
Ready for another look 👍

app/assets/javascripts/components/cookie_banner.js
```diff
@@ -0,0 +1,23 @@
+"use strict";
+
+document.addEventListener("DOMContentLoaded", function() {
+  var seenCookieBannerName = "seen_cookie_banner";
+  var seenCookieBanner = getCookie(seenCookieBannerName);
+  var cookieBanner = document.querySelector("#global-cookie-message");
+
+  if (!seenCookieBanner) {
+    cookieBanner.classList.add("govuk-cookie-banner--visible");
+    setCookie(seenCookieBannerName, 1, 365);
```
erbridge: Is it ok that we only show this to users once? Do we need some sort of "yes, I read this" action to clear it?

app/views/layouts/application.html.erb
erbridge: No JavaScript also means no cookies, but sure :).

app/views/static_pages/cookies.html.erb
```diff
@@ -139,9 +139,9 @@
       </thead>
       <tbody>
         <tr class="govuk-table__row">
-          <td class="govuk-table__cell">XXXX</td>
+          <td class="govuk-table__cell">seen_cookie_banner</td>
           <td class="govuk-table__cell">Saves a message to let us know that you have seen our cookie message</td>
-          <td class="govuk-table__cell">After 1 month</td>
+          <td class="govuk-table__cell">After 1 year</td>
```
erbridge: I don't think this is a "welcome message", either. Maybe the text above just needs a bit of amendment.

app/views/layouts/application.html.erb
robbiepaul: Not always, cookies can be set by response headers too

app/views/static_pages/cookies.html.erb
titlescreen: Amended this section to now refer to cookie banner.

app/assets/javascripts/components/cookie_banner.js
robbiepaul: I'm not sure. I can try to find out.

The now deprecated GOV.UK Template that the Service manual refers to [used a similar method](https://github.com/alphagov/govuk_template/blob/cde91582d0a3bc67d5df1c9484de558548361247/source/assets/javascripts/cookie-bar.js) of just showing the cookie banner on first visit so there will be dozens of services doing the same.

titlescreen: I have changed the banner to show every 90 days, so if you revisit after submitting you shouldn't see it, but its frequent enough that is should catch cookie policy amends.

I think the part about having an accept button, we look at as a separate dev.

app/views/static_pages/cookies.html.erb
erbridge: 90 days now.

robbiepaul:
Ready for another review now. Build should pass, rake passes locally

app/views/layouts/application.html.erb
```diff
@@ -17,6 +17,7 @@
     <%= favicon_link_tag 'govuk-apple-touch-icon.png', rel: 'apple-touch-icon', type: 'image/png' %>
 
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <%= javascript_include_tag "cookie_functions" %>
```
tekin: is there a reason this isn't just included in the asset bundle with the rest of the JS?

app/views/layouts/application.html.erb
```diff
@@ -34,6 +35,16 @@
 
     <a href="#main-content" class="govuk-skip-link">Skip to main content</a>
 
+    <div id="global-cookie-message" class="govuk-cookie-banner" role="region" aria-label="cookie banner">
+      <div class="govuk-width-container">
+        <p class="govuk-cookie-banner__message"><%= t("service_name") %> uses cookies to make the site simpler.</p>
```
tekin: Is this wording correct? Literally the site service can't work without cookies. also "Accept cookies" in this case actually means accept Google Analytics. If they don't accept this we will still set cookies for their claim.

app/views/static_pages/cookies.html.erb
```diff
@@ -77,6 +77,12 @@
           <td class="govuk-table__cell">This helps us count how many people visit <%= t("service_name") %> by tracking if you’ve visited before</td>
           <td class="govuk-table__cell">After 24 hours</td>
         </tr>
+        <tr>
+          <tr class="govuk-table__row">
+          <td class="govuk-table__cell">_gat_gtag_[ua-code]</td>
```
tekin: I think we should put our actual ua-code here.

app/assets/javascripts/google_analytics.js
```diff
@@ -1,7 +1,9 @@
...
+if (window.TeacherPayments.cookies.checkNonEssentialCookiesAccepted()) {
+  window.dataLayer = window.dataLayer || [];
+  function gtag() {
+    dataLayer.push(arguments);
+  }
+  gtag("js", new Date());
 
-gtag("config", document.currentScript.getAttribute("data-ga-id"), { anonymize_ip: true });
+  gtag("config", document.currentScript.getAttribute("data-ga-id"), { anonymize_ip: true });
+}
```
erbridge: I think if you ran this on DOM loaded, you'd be able to rely on the cookie functions without taking them out of the build system. Or put the cookies code before `= require_tree .`.

app/assets/javascripts/cookie_functions.js
```diff
@@ -0,0 +1,26 @@
...
+
+window.TeacherPayments.cookies = {
+  acceptCookieName: "accept_cookies",
+  checkNonEssentialCookiesAccepted: function() {
+    return this._get(this.acceptCookieName);
+  },
+  acceptNonEssentialCookies: function() {
+    this._set(this.acceptCookieName, 1, 90);
+  },
+  setNonEssentialCookie: function(name, value, days) {
```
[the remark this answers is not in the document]
robbiepaul: No, I put a note about it in the commit message. I wanted to make it clear in code that if you want to set a non-essential cookie the user needs to give consent. Unfortunately the only non-essential cookies we have are set by Google Analytics. Happy to remove it though if you think we don't need it

app/views/layouts/application.html.erb
robbiepaul: To ensure that the cookie functions are available to both the `google_analytics.js` and asset bundle. The asset bundle loads in at the end of the page

app/views/static_pages/cookies.html.erb
```diff
@@ -123,12 +129,12 @@
...
     <h2 class="govuk-heading-l">
-      Our introductory message
+      Cookie consent
     </h2>
 
     <p class="govuk-body">
-      You may see a pop-up welcome message when you first visit <%= t("service_name") %>.
-      We’ll store a cookie so that your computer knows you’ve seen it and knows not to show it again.
+      You will see a pop-up cookie consent message when you visit <%= t("service_name") %>.
+      We’ll store a cookie so that your computer knows you’ve consented to non-essential cookies and don't ask you again.
```
erbridge: *If you give us consent to, we'll store a cookie...

app/views/layouts/application.html.erb
robbiepaul: Not sure the wording correct, just used the wording from the GOV.UK cookie banner: https://www.gov.uk 

"Accept cookies" means "Accept non-essential cookies" (but that felt a bit verbose) which is all you need consent for according to: https://ico.org.uk/about-the-ico/news-and-events/news-and-blogs/2019/07/blog-cookies-what-does-good-look-like/

app/assets/javascripts/cookie_functions.js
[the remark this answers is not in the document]
tekin: I think if we want to communicate that there's probably a more explicit and clear way to do it. the chances of a developer that needs to add a non-essential cookie coming across this method and understanding that they need to call it is pretty slim. Perhaps it would be better to simply document it somewhere?

app/views/layouts/application.html.erb
tekin: ah ok. it looks really weird to me, but I guess they've got the prior art, so ¯\_(ツ)_/¯

app/views/layouts/application.html.erb
robbiepaul: I've implemented the solution that @erbridge suggested about loading the Google Analytics code on `DOMContentLoaded` so I've put `cookie_functions.js` back in the asset bundle

titlescreen:
Changed alignment so it will now look like 
Screenshot 2019-08-20 at 15 44 20

app/assets/javascripts/cookie_functions.js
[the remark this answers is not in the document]
robbiepaul: I've removed this unused function and added a note to the readme about storing non-essential cookies

PR #302, opened 2019-08-19
571 make clear expectations about when they will get paid

titlescreen:
Update confirmation page and email sent to set expectations about when they can expect to get paid and use language to give us some wiggle room.

### DIFF
--- a/app/views/claims/confirmation.html.erb
+++ b/app/views/claims/confirmation.html.erb
@@ -20,8 +20,13 @@
     </p>
 
     <p class="govuk-body">
-      They will verify the information you have provided and then send
-      you the outcome of your application within 4 weeks.
+      They will verify the information that you have provided and then
+      send you the outcome of your application in the next 6 weeks.
+    </p>
+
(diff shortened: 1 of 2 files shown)


app/views/claims/confirmation.html.erb
```diff
@@ -20,16 +20,21 @@
     </p>
 
     <p class="govuk-body">
-      They will verify the information you have provided and then send
-      you the outcome of your application within 4 weeks.
+      They will verify the information that you have provided and then
```
tekin: Not for this PR, but I do find it weird that we say use "They" and the the line about sending "your application to the Department for Education" when we generate the email and it comes from "Claim additional payments for teaching" with a reply-to address of studentloanteacherpayment@digital.education.gov.uk. Something to chat to Patrick about when he's back?

titlescreen: I've already sent Patrick an email with screen shots of this because I agree, its a little weird. He should be back tomorrow, so when I see him next I'll pick it up with him.

PR #303, opened 2019-08-19
Update Privacy policy

titlescreen:
DPIA have looked at the privacy policy and recommended some changes.
The 'Why our use of your personal data is lawful' section still mentions
consent which feels off. We are currently awaiting feedback on that
sentence.

### DIFF
--- a/app/views/static_pages/privacy_policy.html.erb
+++ b/app/views/static_pages/privacy_policy.html.erb
@@ -26,8 +26,8 @@
     </p>
 
     <ul class="govuk-list govuk-list--bullet">
+      <li>to claim a payment</li>
       <li>when contacting us for support using the service, to leave feedback or for any other matters</li>
-      <li>as a user of the service</li>
     </ul>
 
     <h2 class="govuk-heading-l">
@@ -35,7 +35,7 @@
     </h2>
 
     <p class="govuk-body">
-      We will collect:
+      We will collect your personal information such as your:
     </p>
 
     <ul class="govuk-list govuk-list--bullet">
@@ -64,7 +64,7 @@
     <p class="govuk-body">
       In order for our use of your personal data to be lawful, we need to meet one (or more) conditions in the data
       protection legislation. For the purpose of this project, your consent forms the basis of this as stated under GDPR
-      Article 6 (1)(a).
+      Article 6 (1)(e).
     </p>
 
     <h2 class="govuk-heading-l">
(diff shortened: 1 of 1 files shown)


app/views/static_pages/privacy_policy.html.erb
```diff
@@ -26,16 +26,16 @@
...
-      <li>as a user of the service</li>
     </ul>
 
     <h2 class="govuk-heading-l">
       What personal information we will collect
     </h2>
 
     <p class="govuk-body">
-      We will collect:
+      If you make a claim we will collect your personal information such as your:
```
erbridge: The second bullet point is unrelated to the user making a claim.

app/views/static_pages/privacy_policy.html.erb
```diff
@@ -112,14 +112,14 @@
 
     <p class="govuk-body">
       Neither DfE nor its delivery partners will keep your personal data longer than we need it for the purpose(s) of
-      research, security and/or delivery of the service, and in any case not longer than XXXX years. Your data will be
-      deleted when DfE and its delivery partners no longer need it for these purposes, or when XXXX years have passed,
+      research, security and/or delivery of the service, and in any case not longer than 7 years. Your data will be
+      deleted when DfE and its delivery partners no longer need it for these purposes, or when 7 years have passed,
```
erbridge: Can we actually say 7 years? Does that mean people will be able to reclaim after 7 years?

app/views/static_pages/privacy_policy.html.erb
```diff
@@ -112,14 +112,14 @@
...
       whichever comes first.
     </p>
 
     <p class="govuk-body">
-      Partial IP addresses will be deleted after XXXX days unless their request caused an error, in which case they remain
-      in our logs for XXXX days.
+      Partial IP addresses will be deleted after 7 days unless their request caused an error, in which case they remain
+      in our logs for 30 days.
```
erbridge: IIRC, TVS did something clever with munging the IPs to they only stored partial ones, but I don't remember the details. We don't store any IPs at all currently, and we're not set up to make this statement at all true if we did start to collect them. Can we just remove it?

app/views/static_pages/privacy_policy.html.erb
titlescreen: Checked and we can remove the reference to the dates.

app/views/static_pages/privacy_policy.html.erb
```diff
@@ -62,9 +62,10 @@
     </h2>
 
     <p class="govuk-body">
-      In order for our use of your personal data to be lawful, we need to meet one (or more) conditions in the data
-      protection legislation. For the purpose of this project, your consent forms the basis of this as stated under GDPR
-      Article 6 (1)(a).
+      In order for our use of your personal data to be lawful, we need to meet one or more conditions in the data
+      protection legislation.  For the purpose of ‘<%= t("service_name") %>‘, the Department processes your
```
erbridge: Second curly quote is backwards. Not sure it matters.

tekin: yep, should be `’`

PR #308, opened 2019-08-19
Play GOV.UK Verify response data back to the user

erbridge:
They need to see what data we're using before they can make an informed decision about whether or not they submit the claim.

image

### DIFF
--- a/app/views/claims/_check_your_answers_section.html.erb
+++ b/app/views/claims/_check_your_answers_section.html.erb
@@ -14,7 +14,7 @@
       </dd>
 
       <dd class="govuk-summary-list__actions">
-        <%= link_to 'Change', claim_path(slug), class: 'govuk-link' %>
+        <%= link_to('Change', claim_path(slug), class: 'govuk-link') unless slug.nil? %>
       </dd>
     </div>
   <% end %>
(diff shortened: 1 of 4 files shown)


app/views/claims/_check_your_answers_section.html.erb
```diff
@@ -14,7 +14,7 @@
       </dd>
 
       <dd class="govuk-summary-list__actions">
-        <%= link_to 'Change', claim_path(slug), class: 'govuk-link' %>
+        <%= link_to('Change', claim_path(slug), class: 'govuk-link') unless slug.nil? %>
```
pezholio: If there's no slug, should we even show the last `<dd>` element?

tekin:
One thing I discussed here with Mark in the office is replaying the information on the `/verified` page. We think that's where it will ultimately happen, plus replaying it to the user sooner rather than later feels like the right thing to do. Plus it doesn't overload the check-your-answers page, which strictly speaking according to the pattern is only for replaying answers to questions the user has been asked, and not meant for other information. Can we work this up to replay it there instead?

app/views/claims/_check_your_answers_section.html.erb
erbridge: It keeps the layout consistent. I was on the fence, though.

erbridge:
[Display fields from Verify as read-only on the check your answers page](https://trello.com/c/HTsgi3Sd/592-display-fields-from-verify-as-read-only-on-the-check-your-answers-page)

PR #312, opened 2019-08-20
Have a support role within the app

pezholio:
This builds on #307 and adds a support for a 'support' role in the app. If a user logs in to the admin section with a role of `teacher_payments_support`, then they are logged in, but cannot download the full CSV of claims

### DIFF
--- a/app/controllers/admin/claims_controller.rb
+++ b/app/controllers/admin/claims_controller.rb
@@ -1,4 +1,6 @@
 class Admin::ClaimsController < Admin::BaseAdminController
+  before_action :ensure_service_operator
+
   def index
     claims = Claim.includes(eligibility: [:claim_school, :current_school]).submitted.order(:submitted_at)
     csv = TslrClaimsCsv.new(claims)
@@ -6,5 +8,11 @@ class Admin::ClaimsController < Admin::BaseAdminController
     respond_to do |format|
       format.csv { send_file csv.file, type: "text/csv", filename: "claims.csv" }
     end
+  end
+
+  private
+
+  def ensure_service_operator
+    render "admin/auth/failure", status: :unauthorized unless service_operator_signed_in?
   end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -52,6 +52,7 @@ class ApplicationController < ActionController::Base
     if admin_session_timed_out?
       session.delete(:user_id)
       session.delete(:organisation_id)
+      session.delete(:role_codes)
       flash[:notice] = "Your session has timed out due to inactivity, please sign-in again"
     end
   end
--- a/app/models/admin_session.rb
+++ b/app/models/admin_session.rb
@@ -2,8 +2,13 @@
 
 class AdminSession < DfeSignIn::AuthenticatedSession
   SERVICE_OPERATOR_DFE_SIGN_IN_ROLE_CODE = "teacher_payments_access"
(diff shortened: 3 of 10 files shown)


pezholio:
[Have a Support role within the app](https://trello.com/c/sXvUHfsF/574-have-a-support-role-within-the-app)

pezholio:
[Have a Support role within the app](https://trello.com/c/sXvUHfsF/574-have-a-support-role-within-the-app)

app/controllers/application_controller.rb
```diff
@@ -24,6 +24,14 @@ def admin_signed_in?
     session.key?(:user_id)
   end
 
+  def admin_session
+    AdminSession.new(session[:user_id], session[:organisation_id], session[:role_codes])
+  end
```
erbridge: Could this do with memoization?

app/models/admin_session.rb
```diff
@@ -2,8 +2,17 @@
 
 class AdminSession < DfeSignIn::AuthenticatedSession
   SERVICE_OPERATOR_DFE_SIGN_IN_ROLE_CODE = "teacher_payments_access"
+  SUPPORT_USER_DFE_SIGN_IN_ROLE_CODE = "teacher_payments_support"
```
erbridge: Can we call them "support agent" instead of "support user" to be clear about their role?

app/controllers/application_controller.rb
```diff
@@ -24,6 +24,14 @@ def admin_signed_in?
...
+
+  def service_operator_signed_in?
```
tekin: I think we should define this in the base admin controller to keep it scoped to that part of the codebase.

app/controllers/application_controller.rb
```diff
@@ -24,6 +24,14 @@ def admin_signed_in?
     session.key?(:user_id)
   end
 
+  def admin_session
```
tekin: As with `service_operator_signed_in?`, perhaps it would be better to put this in the base admin controller.

app/controllers/admin/claims_controller.rb
```diff
@@ -7,4 +9,10 @@ def index
       format.csv { send_file csv.file, type: "text/csv", filename: "claims.csv" }
     end
   end
+
+  private
+
+  def ensure_service_operator
+    redirect_to admin_auth_failure_path unless service_operator_signed_in?
```
tekin:
```suggestion
    render "admin/auth/failure", status: :unauthorized unless service_operator_signed_in?
```

So that the response code is correct?

spec/features/admin_csv_download_spec.rb
```diff
@@ -21,6 +22,23 @@
     end
   end
 
+  context "User is logged in as a support user" do
```
erbridge: *agent

app/controllers/application_controller.rb
```diff
@@ -9,7 +9,7 @@ class ApplicationController < ActionController::Base
     if: -> { ENV.key?("BASIC_AUTH_USERNAME") },
   )
 
-  helper_method :admin_signed_in?, :current_claim
+  helper_method :admin_signed_in?, :current_claim, :service_operator_signed_in?
```
tekin: This change should happen in the base admin controller where the method is defined.

pezholio: Oh, good shout. I thought I'd done that

PR #317, opened 2019-08-20
Consistently use "privacy notice", not "policy"

tekin:
Make the slug, url helpers and link in the footer match the title of the page.


mec:
I think this is the PR that has been replaced with #319

PR #316, opened 2019-08-20
Adjust footer links

tekin:
- remove the dead links to Help and Contact
- add a link to the support email

PR #319, opened 2019-08-20
Privacy notice snags

titlescreen:


mec:
Are we meant to be fixing the footer in this PR as well?

https://github.com/DFE-Digital/dfe-teachers-payment-service/pull/316

mec:
> Are we meant to be fixing the footer in this PR as well?
> 
> #316

Think I see what happend and have sorted it!

PR #322, opened 2019-08-20
Remove parameter on error check

titlescreen:
With the parameter on it appeared that the logic on the form group was
never getting triggered as it had returned a school and did not have an
error on that part. The error was because a user had not selected a
school from the radio.

Was
Screenshot 2019-08-20 at 15 31 18

Now
Screenshot 2019-08-20 at 15 30 26

### DIFF
--- a/app/views/claims/_school_search.html.erb
+++ b/app/views/claims/_school_search.html.erb
@@ -45,7 +45,7 @@
   %>
 
   <%= form_for current_claim, url: claim_path do |form| %>
-    <%= form_group_tag current_claim, school_param do %>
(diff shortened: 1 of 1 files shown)


app/views/claims/_school_search.html.erb
```diff
@@ -45,7 +45,7 @@
   %>
 
   <%= form_for current_claim, url: claim_path do |form| %>
-    <%= form_group_tag current_claim, school_param do %>
+    <%= form_group_tag current_claim do %>
```
erbridge: I suspected this was the issue when I was refactoring. Glad to see my hunch was right :).

PR #323, opened 2019-08-20
Closed schools show close date during search

mec:
When a school changes some of its attributes (name, status, etc) it is closed and then re-opened with the new attributes in the Get information about schools data.

We want to show this close date on the claim school and current school search pages so a user can make an informed desicion about which school to select in cases where there are duplicates or the when the school is actually closed.

This is an iterim solution whilst we work on the right way to approach these 'closed' schools.

Handing this off to @titlescreen to look at the display of the close date in the UI.


mec:
[How do we deal with closed/renamed schools?](https://trello.com/c/syQijn7T/436-how-do-we-deal-with-closed-renamed-schools)

mec:
[How do we deal with closed/renamed schools?](https://trello.com/c/syQijn7T/436-how-do-we-deal-with-closed-renamed-schools)

PR #324, opened 2019-08-20
Add "How we will use the information you provide" page

robbiepaul:
Before we send the user to Verify we want to make it clear how we are going to use any information they provide and any hidden processes we have.

# Screenshot
information_provided_changes


robbiepaul:
[Tell user about hidden information and processes](https://trello.com/c/YEf5SsER/600-tell-user-about-hidden-information-and-processes)

app/views/claims/information_provided.html.erb
```diff
@@ -0,0 +1,48 @@
...
+      To process your claim we will check the information you provide against the information the
+      Department for Education hold.
+    </p>
+
+    <p class="govuk-body">To determine if your claim is valid we will:</p>
+
+    <ul class="govuk-list govuk-list--bullet">
+      <li>
+        check your eligibility against the Database of Qualified Teachers. You may have used with
+        this in the past as the Teacher Self-Service Portal
```
erbridge: Is it weird to have multiple sentences in these? The list items start with lowercase, then have a full stop, then end without one. Just looks weird to me and is something I always struggle to avoid.

app/views/claims/information_provided.html.erb
```diff
@@ -0,0 +1,48 @@
...
+    <%= link_to "Continue to GOV.UK Verify", new_verify_authentications_path, class: "govuk-button" %>
+
+    <details class="govuk-details" data-module="govuk-details">
+      <summary class="govuk-details__summary">
+        <span class="govuk-details__summary-text">
+          What information is sent to HMRC?
+        </span>
+      </summary>
+      <div class="govuk-details__text">
+        We send HMRC your full name, date of birth, payment amount and gender to ensure your tax
```
erbridge: "payment amount"? What does that refer to? Could it be clarified?

app/models/page_sequence.rb
```diff
@@ -17,6 +17,7 @@ class PageSequence
     "leadership-position",
     "mostly-performed-leadership-duties",
     "eligibility-confirmed",
+    "information-provided",
```
erbridge: Seems strange to me to tell people about how we'll use their eligibility data _after_ we've asked them for it.

app/views/claims/information_provided.html.erb
```diff
@@ -0,0 +1,48 @@
...
+    <p class="govuk-body">
+      To process your claim we will check the information you provide against the information the
+      Department for Education hold.
+    </p>
+
+    <p class="govuk-body">To determine if your claim is valid we will:</p>
+
+    <ul class="govuk-list govuk-list--bullet">
+      <li>
+        check your eligibility against the Database of Qualified Teachers. You may have used with
```
erbridge: Extra "with" on the end here.

app/views/claims/information_provided.html.erb
```diff
@@ -0,0 +1,48 @@
...
+      <li>pay you as the Department for Education. You will see this on a future P60</li>
+      <li>send some of your personal information to HMRC</li>
+      <li>pay a tax, student loan and national insurance contribution to HMRC</li>
+    </ul>
+
+    <p class="govuk-body">
+      We will send you details about these payments once your application is complete.
+    </p>
+
+    <%= link_to "Continue to GOV.UK Verify", new_verify_authentications_path, class: "govuk-button" %>
```
erbridge: It's not clear what this next step actually is now. It sounds like we're done at this point from the text above.

app/views/claims/information_provided.html.erb
tekin: Yes, this content has been put together quite hastily and we expect it to change once Patrick has given it the once over. Bulleted lists should ideally be single sentence, no full stop, but because it is two sentences for now, we should end this with a full stop.

app/views/claims/information_provided.html.erb
tekin: On the mock up this button reads "Continue". The page the user links to explains what verify is and how it works. Probably best if we drop the `to GOV.UK Verify` for now.

app/views/claims/information_provided.html.erb
tekin: Yes, it could be. Again, this is hastily put-together content and a stop-gap solution that hasn't had a content designer go over it. We know this page needs work, let's get this out to unblock private beta and revisit this page properly next.

app/views/claims/information_provided.html.erb
robbiepaul: Yeah I thought it was an omission on the mockup. I'll remove it

erbridge: My comment was less about the text on the Continue button, but more about the expectation of what happens next. This page talks about making the claim as if we're now done, so it's going to be jarring for a user to still have 10ish steps to go, including a long interaction with Verify. I think there needs to be some words about how next we'll ask the user some information to verify their identity and collect details about how to contact and pay them.

robbiepaul:
I've updated it with changes from Patrick and Mark:


information_provided_changes

PR #325, opened 2019-08-20
Cache Docker on CI

erbridge:
This involves pushing up 3 new Docker images with `master` builds:

- `cache-web-dependencies`: the `dependencies` image that's dropped from the final image
- `cache-test`: the `test` image built for running tests
- `cache-test-dependencies`: the `dependencies` image for those tests (with `RAILS_ENV=test`)

This is so we can use them as cache sources for builds. This change means all non-`master` builds should now reuse the Docker cache from the latest `master` build. `master` builds build from scratch to ensure we're running fresh images (probably not necessary, but it depends on how much you trust Docker and our use of it).

In order to get this to work locally, I needed to drop the `--no-cache` flag on the package installations. This results in slightly larger Docker images, which I think is ok.

**This isn't fully tested with actual images on Azure.** It was tested by running the commands locally using locally tagged images.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,10 +4,10 @@
 
 FROM ruby:2.6.2-alpine AS base
 
-RUN apk --no-cache add bash
-RUN apk --no-cache add postgresql-dev
-RUN apk --no-cache add tzdata
-RUN apk --no-cache add nodejs
+RUN apk add bash
+RUN apk add postgresql-dev
+RUN apk add tzdata
+RUN apk add nodejs
 
 ENV APP_HOME /app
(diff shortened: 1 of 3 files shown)


erbridge:
[Cache Docker images on CI to speed up builds](https://trello.com/c/5bM6P6Yr/601-cache-docker-images-on-ci-to-speed-up-builds)

robbiepaul:
I had issues testing this locally, I guess that's because the `cache-*` images get pushed on merges to master?

erbridge:
Fixed those typos. Thanks.

PR #328, opened 2019-08-21
Confirm data from GOV.UK Verify

pezholio:
This adds a page which confirms data coming from GOV.UK Verify as per the mockup at https://miro.com/app/board/o9J_kwlloyg=/?moveToWidget=3074457346806977016

image

image


app/views/claims/verified.html.erb
```diff
@@ -5,9 +5,28 @@
...
     <%= button_to "Continue", claim_path, method: :put, class: "govuk-button" %>
+
+    <details class="govuk-details" data-module="govuk-details">
+      <summary class="govuk-details__summary">
+        <span class="govuk-details__summary-text">
+          Some of this information is wrong
+        </span>
+      </summary>
+      <div class="govuk-details__text">
+        If you find that some of this information is incorrect please contact support at <%= mail_to "studentloanteacherpayment@digital.education.gov.uk", "studentloanteacherpayment@digital.education.gov.uk", class: "govuk-link" %> as this could affect your application.
```
titlescreen: Quick question, did we want to use the translation file to save the TSLR specific support address?

app/views/claims/verified.html.erb
```diff
@@ -5,9 +5,28 @@
...
-      You can now continue with your claim
+      This is your <%= current_claim.presented_identity.verified_fields %> from your digital identity:
     </p>
 
+    <dl class="govuk-summary-list govuk-!-margin-bottom-9">
+      <% [:full_name, :date_of_birth, :payroll_gender, :address].each do |k| %>
+        <% if current_claim.presented_identity.send(k) %>
+          <%= render partial: "answer_row", locals: { label: I18n.t("verified_fields.#{k}").capitalize, answer: current_claim.presented_identity.send(k), slug: nil } %>
+        <% end %>
+      <% end %>
```
erbridge: Was there a particular reason to do this in a different way from how we do the check your answers page? We already have all of the data and logic needed there.

pezholio:
[How do we use or display hidden information to the user (such as Verify)](https://trello.com/c/WF4NYZOU/472-how-do-we-use-or-display-hidden-information-to-the-user-such-as-verify)

app/views/claims/verified.html.erb
pezholio: Good shout 👍

app/models/claim.rb
```diff
@@ -127,6 +127,10 @@ def student_loan_country_with_one_plan?
     StudentLoans::PLAN_1_COUNTRIES.include?(student_loan_country)
   end
 
+  def presented_identity
+    @presented_identity ||= VerifiedIdentity.new(self)
+  end
```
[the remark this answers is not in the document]
pezholio: I guess it's because it's expressing the type of data to expect (i.e. it's designed for presentation, not raw data)

app/presenters/verified_identity.rb
```diff
@@ -0,0 +1,36 @@
...
+  def initialize(claim)
+    @claim = claim
+  end
+
+  def verified_fields
+    fields = []
+    fields << I18n.t("verified_fields.full_name") if claim.verified_fields.include?("full_name")
+    fields << I18n.t("verified_fields.date_of_birth") if claim.verified_fields.include?("date_of_birth")
+    fields << I18n.t("verified_fields.payroll_gender") if claim.verified_fields.include?("payroll_gender")
+    fields << I18n.t("verified_fields.address") if claim.verified_fields.include?("address_line_1")
```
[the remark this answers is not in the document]
pezholio: Whoops, good shout

app/presenters/verified_identity.rb
```diff
@@ -0,0 +1,36 @@
...
+  def payroll_gender
+    return nil unless claim.payroll_gender_verified?
+
+    I18n.t("answers.payroll_gender.#{claim.payroll_gender}")
+  end
+
+  def address
+    return nil unless claim.address_verified?
+
+    claim.address("<br>").html_safe
```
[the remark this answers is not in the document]
pezholio: It's pretty low risk, but a good shout. I've moved the call to `.html_safe` to only act on the `<br>` (which is the only HTML we should expect anyway)

app/presenters/verified_identity.rb
```diff
@@ -0,0 +1,36 @@
+class VerifiedIdentity
+  attr_reader :claim
+
+  def initialize(claim)
+    @claim = claim
+  end
+
+  def verified_fields
```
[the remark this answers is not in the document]
pezholio: Yeah, actually, I think it might be better to use the helper rather than the presenter on reflection

app/models/claim.rb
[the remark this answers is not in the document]
erbridge: It's not the only "identity" information we collect, though. Perhaps `presented_verified_identity`?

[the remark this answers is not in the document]
pezholio: I've tweaked all this to use the helpers now, so probably not relevant anymore

app/views/claims/_check_your_answers_section.html.erb
```diff
@@ -1,21 +1,11 @@
...
-
-      <dd class="govuk-summary-list__value">
-        <%= answer %>
-      </dd>
-
-      <dd class="govuk-summary-list__actions">
-        <%= link_to('Change', claim_path(slug), class: 'govuk-link') unless slug.nil? %>
-      </dd>
-    </div>
+    <%= render partial: "answer_row", locals: { label: label, answer: answer, slug: slug } %>
```
erbridge: I don't think we gain anything for having this partial anymore. Can it get rolled back in to this one?

app/views/claims/verified.html.erb
```diff
@@ -5,9 +5,22 @@
     </h1>
 
     <p class="govuk-body">
-      You can now continue with your claim
+      This is your <%= verified_fields(current_claim) %> from your digital identity:
```
erbridge:
```suggestion
      These are your <%= verified_fields(current_claim) %> from your digital identity:
```

app/helpers/claims_helper.rb
```diff
@@ -25,10 +34,10 @@ def eligibility_answers(eligibility)
 
   def verify_answers(claim)
     [].tap do |a|
-      a << ["Full name", claim.full_name]
-      a << ["Address", claim.address] if claim.address_verified?
-      a << ["Date of birth", l(claim.date_of_birth)]
-      a << ["Gender", t("answers.payroll_gender.#{claim.payroll_gender}")] if claim.payroll_gender_verified?
+      a << [I18n.t("verified_fields.full_name").capitalize, claim.full_name]
+      a << [I18n.t("verified_fields.address").capitalize, claim.address("<br>").html_safe] if claim.address_verified?
```
erbridge: `html_safe` issue again.

app/views/claims/verified.html.erb
tekin: we'll want to remember that this will need updating/making dynamic when we introduce Maths & Physics.

app/views/claims/verified.html.erb
pezholio: `These are your full name, address, date of birth, and gender from your digital identity` sounds a bit clumsy to me?

app/helpers/claims_helper.rb
pezholio: Yeah, running `html_safe` on just the `<br>` results in it returning `&lt;br&gt;` in the string. I think the risks of getting dodgy HTML from Verify service providers, especially considering it's run through the VSP first, it extremely low

tekin: we can use `simple_format` to convert newlines to breaks.

app/views/claims/verified.html.erb
erbridge: Maybe. I think it is the "correct" grammar, but it's not a big deal. Feel free to leave.

app/helpers/claims_helper.rb
tekin: actually, scratch that, I've wrapped it in a `sanitize` call.

PR #329, opened 2019-08-21
Subjects taught exported to csv

mec:
Each of the eligible subjects is included in the csv export.


erbridge:
Did you consider putting it in a comma seperated list in a `Subjects Taught` column instead of adding 5 new columns?

mec:
@erbridge yeah I did, this seemed like the most useful and simple way to do it for what is a temporary feature – I guess it depends on how the data is used?

pezholio:
I reckon a comma seperated list is fine. Makes it easier to review at a glance when scanning seperate rows. We chose the approach of having different columns for each subject in the DB to keep the underlying data structure simple, but it makes more sense on the user side to have one column for all subjects I reckon

PR #331, opened 2019-08-21
Remove consent page from journey

tekin:
As part of the work to be clearer about the hidden information and the
things we do with users' data, the new "information-provided" page was
added after "eligibility-confirmed". The introduction of this page, and
the fact that we don't actually need or want to ask for explicit consent
means that the consent page should no longer be there.

I'm not 100% with the solution of making the start page perform a POST
to create the claim, mainly because this can't be the long term solution
because the eventual start page will live on GOV.UK and won't be able to
perform a POST to our service, but to get us out the door for private
beta this afternoon I think this is OK. We can rework the way that the
controller/page sequencing works to make the qts-year question act as
the initial form, but that is a bit more involved.


app/views/static_pages/start_page.html.erb
```diff
@@ -31,6 +31,6 @@
         <li>sign in details for ‘Verify’ – the government’s secure way of proving who you are, if you do not have a Verify account, you’ll need your passport or photocard driving licence to sign up</li>
       </ul>
 
-      <%= link_to "Start now", new_claim_path, class: "govuk-button govuk-!-margin-top-2 govuk-!-margin-bottom-8 govuk-button--start", role: "button" %>
+      <%= link_to "Start now", claims_path, method: :post, class: "govuk-button govuk-!-margin-top-2 govuk-!-margin-bottom-8 govuk-button--start", role: "button" %>
```
robbiepaul: I think this won't POST for non-JavaScript users

tekin: ah good point, I'll covert it to a button.

tekin: done.

PR #338, opened 2019-08-21
Include Markdown files for linting in Dockerfile

tekin:


Dockerfile
```diff
@@ -134,6 +134,7 @@ COPY spec ${APP_HOME}/spec
 
 # Copy files for linting (sorted by vague frequency of change for caching)
 COPY docs ${APP_HOME}/docs
+COPY *.md ${APP_HOME}/
```
pezholio: Should we include everything in /docs too?

tekin: Isn't that covered by the line above?

pezholio: Whoops, yeah. Missed that!

PR #340, opened 2019-08-22
Add diagnostic logs to key vault

pezholio:
This adds a diagnostic setting to our Keyvault, as well as a storage account where we can keep the logs (retained for a year by default)


erbridge:
[Diagnostic logs in Key Vault should be enabled](https://trello.com/c/v0McAhQ4/588-diagnostic-logs-in-key-vault-should-be-enabled)

PR #341, opened 2019-08-22
Report redacted GOV.UK Verify responses using Rollbar

tekin:
This will aid in debugging situations where there is a problem handling the response from GOV.UK Verify.

### DIFF
--- a/lib/verify/redacted_response.rb
+++ b/lib/verify/redacted_response.rb
@@ -1,0 +1,42 @@
+module Verify
+  # Used to report a response from Verify without revealing any Personally
+  # Identifiable Information by replacing any string "values" with the "X"
+  # character. For example:
+  #
+  #   Verify::RedactedResponse.new({"value" => "Secret"}).parameters
+  #   => {"value" => "XXXXXX"}
+  #
+  class RedactedResponse
+    KEYS_TO_REDACT = %w[value postCode].freeze
+
+    attr_reader :parameters
+
+    def initialize(parameters)
+      @parameters = redact_values(parameters.deep_dup)
+    end
+
+    private
+
+    def redact_values(input)
+      if input.is_a?(String)
+        redact_string(input)
+      else
+        input.tap do |hash_to_redact|
+          hash_to_redact.each do |key, value|
+            if value.is_a?(Hash)
+              hash_to_redact[key] = redact_values(value)
+            elsif value.is_a?(Array)
+              hash_to_redact[key].each_with_index { |nested_value, nested_key| hash_to_redact[key][nested_key] = redact_values(nested_value) }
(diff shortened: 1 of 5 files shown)


lib/verify/redacted_response.rb
```diff
@@ -0,0 +1,42 @@
+module Verify
+  # Used to report a response from Verify without revealing any Personally
+  # Identifiable Information by replacing any string "values" with the "X"
+  # character. For example:
+  #
+  #   Verify::RedactedResponse.new({"value" => "Secret"}).parameters
+  #   => {"value" => "XXXXXX"}
+  #
+  class RedactedResponse
+    KEYS_TO_REDACT = %w[value postCode].freeze
```
erbridge: What is `value`? Could this do with a comment explaining the expected format from Verify?

lib/verify/redacted_response.rb
```diff
@@ -0,0 +1,42 @@
...
+            elsif value.is_a?(String) && KEYS_TO_REDACT.include?(key)
+              hash_to_redact[key] = redact_string(value)
+            end
+          end
+        end
+      end
+    end
+
+    def redact_string(string)
+      "X" * string.length
```
erbridge: Are we confident we never leak anything by making the string lengths match?

lib/verify/redacted_response.rb
erbridge: It feels like this should be implemented with a whitelist, not a blacklist, in case the response format changes.

lib/verify/redacted_response.rb
tekin: For the values of things we are redacting (names, addresses, DOBs), I'd like to think it would be virtually impossible to identify an individual from just the length of those strings but happy to change this if we can think of a way it might be.

lib/verify/redacted_response.rb
tekin: We could do. I do think the Verify response format is highly unlikely to change, and if it does specs for this class will fail, which should be enough to signal that the redactor will need updating along with the standard `Verify::Response`

tekin: The two strings in this constant are the hash keys that contain personal information in Verify responses (as seen in the `spec/fixtures/verify` files). What would have made this clearer?

erbridge: I'm more worried about them changing on a live system. It's a system we have no control over, and could change in any way, and it would be good not to be leaking PII if that were to happen.

erbridge: I mean `value` is a bit confusing to read. It took some thought to realize that the Verify response returns hashes for its properties, not the values directly.

lib/verify/redacted_response.rb
erbridge: I agree. Just wanted to raise the question.

PR #342, opened 2019-08-22
Adjust ineligible screens

erbridge:
They are now consitently worded and include the date range where appropriate. The subjects taught ineligibility is now more explicit about what the requirements are.

### DIFF
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -196,6 +196,6 @@ class Claim < ApplicationRecord
   end
 
   def claim_must_not_be_ineligible
-    errors.add(:base, eligibility.ineligibility_reason) if eligibility.ineligible?
+    errors.add(:base, "You’re not eligible for this payment") if eligibility.ineligible?
   end
 end
--- a/app/views/claims/_ineligibility_reason_ineligible_qts_award_year.html.erb
+++ b/app/views/claims/_ineligibility_reason_ineligible_qts_award_year.html.erb
@@ -1,4 +1,4 @@
 <p class="govuk-body">
-  You are only eligible to claim back student loan repayments if you completed
-  your initial teacher training on or after September 1st 2013.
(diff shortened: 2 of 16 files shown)


erbridge:
[Adjust ineligible screen for subjects taught question](https://trello.com/c/W6vWlhmo/372-adjust-ineligible-screen-for-subjects-taught-question)

app/models/claim.rb
```diff
@@ -88,7 +88,7 @@ class Claim < ApplicationRecord
   validate :bank_account_number_must_be_eight_digits
   validate :bank_sort_code_must_be_six_digits
 
-  validate :claim_must_not_be_ineligible, on: :submit
+  validate :claim_must_be_eligible, on: :submit
```
tekin: So technically "eligible" isn't the opposite of "ineligible" – a claim without all the answers to the eligibility questions won't be "ineligible", but that doesn't make it "eligible". The double negative is deliberate so as not to report a incomplete claim as "eligible".

app/views/claims/_ineligibility_reason_ineligible_qts_award_year.html.erb
```diff
@@ -0,0 +1,4 @@
+<p class="govuk-body">
+  You can only get this payment if you completed your initial teacher training
+  on or after September 1st 2013.
```
titlescreen: This looks like an inconsistent date format, I think it should be `1 September 2013`

erbridge: Good spot.

app/models/claim.rb
erbridge: Ah ok. Makes sense. I'll remove it.

erbridge:
These have been OKed by Patrick.

PR #343, opened 2019-08-22
Un-verified middle names are now ignored

robbiepaul:
We saw an error on production where a user returned from Verify with an unverified middle name causing an application error and they couldn't complete their journey.

Middle names are not essential to payroll so we can mark them as optional, if a non-verified value comes back we can ignore it.

https://rollbar.com/dxw/teachers-payment-service/items/69/

### DIFF
--- a/lib/verify/response.rb
+++ b/lib/verify/response.rb
@@ -7,6 +7,7 @@ module Verify
   #   Verify::Response.translate(saml_response: "SOME SAML", request_id: "REQUEST_ID", level_of_assurance: "LEVEL_2")
   #
   class Response
+    class MissingResponseAttribute < StandardError; end
     attr_reader :parameters
 
     def initialize(parameters)
@@ -88,7 +89,7 @@ module Verify
       most_recent_verified_attribute = attributes.sort_by { |attribute| attribute["from"].present? ? Date.strptime(attribute["from"], "%Y-%m-%d") : 0 }
         .reverse
         .find { |attribute| attribute["verified"] }
-
+      raise MissingResponseAttribute, "No verified value found" if required && most_recent_verified_attribute.nil?
       required ? most_recent_verified_attribute.fetch("value") : most_recent_verified_attribute&.fetch("value")
     end
 
--- a/lib/verify/response.rb
+++ b/lib/verify/response.rb
@@ -76,19 +76,20 @@ module Verify
 
     def full_name
       first_name = most_recent_verified_value(attributes.fetch("firstNames"))
-      middle_name = most_recent_verified_value(attributes.fetch("middleNames"))
+      middle_name = most_recent_verified_value(attributes.fetch("middleNames"), required: false)
       surname = most_recent_verified_value(attributes.fetch("surnames"))
 
       [first_name, middle_name, surname].compact.join(" ")
     end
 
-    def most_recent_verified_value(attributes)
+    def most_recent_verified_value(attributes, required: true)
       return if attributes.blank?
 
-      attributes.sort_by { |attribute| attribute["from"].present? ? Date.strptime(attribute["from"], "%Y-%m-%d") : 0 }
+      most_recent_verified_attribute = attributes.sort_by { |attribute| attribute["from"].present? ? Date.strptime(attribute["from"], "%Y-%m-%d") : 0 }
         .reverse
         .find { |attribute| attribute["verified"] }
-        .fetch("value")
+
+      required ? most_recent_verified_attribute.fetch("value") : most_recent_verified_attribute&.fetch("value")
     end
 
     def gender
--- a/lib/verify/response.rb
+++ b/lib/verify/response.rb
@@ -56,7 +56,7 @@ module Verify
     end
 
     def address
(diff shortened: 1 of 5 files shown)


lib/verify/response.rb
```diff
@@ -75,20 +75,21 @@ def address_missing?
     end
 
     def full_name
-      first_name = most_recent_verified_value(attributes.fetch("firstNames"))
-      middle_name = most_recent_verified_value(attributes.fetch("middleNames"))
-      surname = most_recent_verified_value(attributes.fetch("surnames"))
+      first_name = most_recent_verified_value(attributes.fetch("firstNames"), required: true)
```
tekin: if we're defaulting to `true` in the method signature I reckon these `require: true` calls are redundant.

lib/verify/response.rb
```diff
@@ -75,20 +75,21 @@ def address_missing?
...
 
       [first_name, middle_name, surname].compact.join(" ")
     end
 
-    def most_recent_verified_value(attributes)
+    def most_recent_verified_value(attributes, required: true)
       return if attributes.blank?
 
-      attributes.sort_by { |attribute| attribute["from"].present? ? Date.strptime(attribute["from"], "%Y-%m-%d") : 0 }
+      verified_attribute = attributes.sort_by { |attribute| attribute["from"].present? ? Date.strptime(attribute["from"], "%Y-%m-%d") : 0 }
```
tekin:
```suggestion
      most_recent_verified_attribute = attributes.sort_by { |attribute| attribute["from"].present? ? Date.strptime(attribute["from"], "%Y-%m-%d") : 0 }
```

lib/verify/response.rb
```diff
@@ -75,20 +75,21 @@ def address_missing?
...
         .reverse
         .find { |attribute| attribute["verified"] }
-        .fetch("value")
+
+      required ? verified_attribute.fetch("value") : verified_attribute&.fetch("value")
```
tekin: Perhaps we should now be more specific here with our error if we get back a Verify response that has a required field missing? for example:

```ruby

class MissingResponseAttribute < StandardError; end
...
raise MissingResponseAttribute, "No verified value found for #{attributes}" if required && verified_attribute.nil?
verified_attribute.fetch("value")
```

lib/verify/response.rb
```diff
@@ -1,4 +1,5 @@
 module Verify
+  class MissingResponseAttribute < StandardError; end
```
tekin: Define this inside the `Response` class to avoid weirdness with the autoloading and I think this is good to go 👍🏽

robbiepaul: Done 👍

lib/verify/response.rb
```diff
@@ -76,19 +77,20 @@ def address_missing?
...
-    def most_recent_verified_value(attributes)
+    def most_recent_verified_value(attributes, required: true)
       return if attributes.blank?
 
-      attributes.sort_by { |attribute| attribute["from"].present? ? Date.strptime(attribute["from"], "%Y-%m-%d") : 0 }
+      most_recent_verified_attribute = attributes.sort_by { |attribute| attribute["from"].present? ? Date.strptime(attribute["from"], "%Y-%m-%d") : 0 }
         .reverse
         .find { |attribute| attribute["verified"] }
-        .fetch("value")
+      raise MissingResponseAttribute, "No verified value found" if required && most_recent_verified_attribute.nil?
```
tekin: Ah, can we include the attribute name in the message her to aid debugging?

lib/verify/response.rb
```diff
@@ -76,19 +77,20 @@ def address_missing?
...
+      most_recent_verified_attribute&.fetch("value")
```
tekin: the `&` shouldn't be necessary here, right?

tekin: also `&.fetch(key)` is equivalent to `[key]`

robbiepaul: Ah yes I put the `&.` in to allow for `required: false` returning nil there. I'll change it to `[key]`

lib/verify/response.rb
robbiepaul: Yeah, I'll have to pass it as another argument as that method only currently receives the attribute hash 👍

lib/verify/response.rb
robbiepaul: Hmm, `[key]` won't work for optional values where `most_recent_verified_attribute` is nil

tekin: we probably still want `fetch("value")` I think. If the most recent "verified" field doesn't have a "value" we probably want it to noisily fail so we can investigate

robbiepaul: I've put the turnery back in to allow for nil optional values as well

lib/verify/response.rb
tekin: I guess we should be able to figure it out from the backtrace...

robbiepaul: I've added a commit that does it, however it feels a bit messy. What do you think?

tekin: I prefer this version where we pass in the attribute name, rather than the hash!

PR #344, opened 2019-08-22
Improve layout of closed schools on search

titlescreen:
Change the html elements used to display closed schools and apply
appropriate formatting.

### DIFF
--- a/app/assets/javascripts/components/school_search.js.erb
+++ b/app/assets/javascripts/components/school_search.js.erb
@@ -82,15 +82,13 @@ document.addEventListener("DOMContentLoaded", function() {
       suggestion: function(id) {
         var school = findSchool(id);
 
-        var suggestion =
-          '<label class="govuk-label govuk-label--s">' +
-          school.name +
-          "</label>" +
-          '<span class="govuk-hint">' +
-          school.address +
-          "</span>";
+        var suggestion = '<table><tr><td><label class="govuk-label govuk-label--s">' + school.name + "</label></td>";
 
-        if (school.closeDate) suggestion += '<span class="govuk-hint">' + "Closed on " + school.closeDate + "</span>";
+        if (school.closeDate)
+          suggestion += '<td><span class="govuk-hint">' + "Closed on " + school.closeDate + "</span></td>";
+
+        suggestion +=
+          '</tr><tr><td colspan="2"><span class="govuk-hint">' + school.address + "</span></td><tr></table>";
 
         return suggestion;
       }
--- a/app/assets/javascripts/components/school_search.js.erb
+++ b/app/assets/javascripts/components/school_search.js.erb
@@ -82,14 +82,27 @@ document.addEventListener("DOMContentLoaded", function() {
       suggestion: function(id) {
         var school = findSchool(id);
 
-        var suggestion = '<table><tr><td><label class="govuk-label govuk-label--s">' + school.name + "</label></td>";
+        var suggestion =
+          '<div class="school-search__suggestion">' +
+          '<div class="school-search__suggestion-main-section">' +
+          '<label class="govuk-label govuk-label--s">' +
+          school.name +
+          "</label>" +
+          '<span class="govuk-hint">' +
+          school.address +
+          "</span>" +
+          "</div>";
 
-        if (school.closeDate)
-          suggestion += '<td><span class="govuk-hint">' + "Closed on " + school.closeDate + "</span></td>";
+        if (school.closeDate) {
+          suggestion +=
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/components/school_search.js.erb
```diff
@@ -82,15 +82,13 @@ document.addEventListener("DOMContentLoaded", function() {
...
-          "</span>";
-
-        if (school.closeDate) suggestion += '<span class="govuk-hint">' + "Closed on " + school.closeDate + "</span>";
+        var suggestion = '<table><tr><td><label class="govuk-label govuk-label--s">' + school.name + "</label></td>";
+
+        if (school.closeDate)
+          suggestion += '<td><span class="govuk-hint">' + "Closed on " + school.closeDate + "</span></td>";
+
+        suggestion +=
+          '</tr><tr><td colspan="2"><span class="govuk-hint">' + school.address + "</span></td><tr></table>";
```
erbridge: Isn't it bad for accessibility to use tables when you're not actually displaying tabular data? Are they just for layout?

tekin: Ah, that's a very good point...

titlescreen: Ok no worries, will swap to divs and flex or inline it.

app/assets/javascripts/components/school_search.js.erb
```diff
@@ -83,14 +83,15 @@ document.addEventListener("DOMContentLoaded", function() {
...
-          '<span class="govuk-hint">' +
-          school.address +
-          "</span>";
-
-        if (school.closeDate) suggestion += '<span class="govuk-hint">' + "Closed on " + school.closeDate + "</span>";
+          '<div class="flex"><div class="flex-1"><label class="govuk-label govuk-label--s">' + school.name + "</label>";
+        suggestion += '<span class="govuk-hint">' + school.address + "</span></div>";
+
+        if (school.closeDate)
+          suggestion +=
```
erbridge: Let's wrap this in curly braces, so it's clear it's a continuation of that line?

app/assets/javascripts/components/school_search.js.erb
```diff
@@ -83,14 +83,15 @@ document.addEventListener("DOMContentLoaded", function() {
...
-          '<label class="govuk-label govuk-label--s">' +
-          school.name +
-          "</label>" +
-          '<span class="govuk-hint">' +
-          school.address +
-          "</span>";
-
-        if (school.closeDate) suggestion += '<span class="govuk-hint">' + "Closed on " + school.closeDate + "</span>";
+          '<div class="flex"><div class="flex-1"><label class="govuk-label govuk-label--s">' + school.name + "</label>";
+        suggestion += '<span class="govuk-hint">' + school.address + "</span></div>";
```
erbridge: If we go for 1 line per tag (separated by `+`s), it makes it easier to see the structure here. I think we're missing a closing `</div>` if we don't have a close date.

app/assets/stylesheets/components/accessible_autocomplete_tweaks.scss
```diff
@@ -11,3 +11,18 @@
     }
   }
 }
+
+.flex {
```
erbridge: Can they be named more appropriately to indicate what they're for and reduce the chances of naming clashes? e.g. `.autocomplete-tweak__suggestion`, `.autocomplete-tweak__suggestion__main-section`, `.autocomplete-tweak__suggestion__closed-status`.

erbridge: Actually, they should probably be something other than `autocomplete-tweak__*` (`school-search__*`?) since they're used by the school search template itself. Probably we should also pull them into a different file, since they're not strictly about autocompletion.

app/views/claims/_school_search.html.erb
```diff
@@ -67,11 +67,17 @@
...
-                <span class="govuk-hint govuk-radios__hint"><%= b.object.address %></span>
-                <% unless b.object.close_date.blank? %>
-                  <span class="govuk-hint govuk-radios__hint">Closed on <%= l(b.object.close_date) %></span>
-                <% end %>
+                <div class="school-search__suggestion">
+                  <div class"school-search__suggestion-main-section>
+                    <%= b.label class: "govuk-label govuk-radios__label govuk-label--s" %>
+                    <span class="govuk-hint govuk-radios__hint"><%= b.object.address %></span>
+                  </div>
+                  <% unless b.object.close_date.blank? %>
```
erbridge: Could reverse this to `if b.object.close_date.present?`. Don't think it's a blocker, though.

PR #345, opened 2019-08-22
Record the current page in GA when accepting cookies

erbridge:


erbridge:
[Record the current page in GA when accepting cookies](https://trello.com/c/R5YDRSij/612-record-the-current-page-in-ga-when-accepting-cookies)

app/assets/javascripts/cookie_functions.js
```diff
@@ -1,22 +1,35 @@
...
-  _get: function(name) {
+  }
+
+  function getCookie(name) {
     var v = document.cookie.match("(^|;) ?" + name + "=([^;]*)(;|$)");
     return v ? v[2] : null;
   }
-};
+
+  window.TeacherPayments.cookies.checkNonEssentialCookiesAccepted = function checkNonEssentialCookiesAccepted() {
```
robbiepaul: Do we need to repeat the function name here? Could we do `window.TeacherPayments.cookies.checkNonEssentialCookiesAccepted = function() {`

erbridge: It helps with debugging. If there's an error, it will be in the anonymous function, while if the function is named, we get the function name back. This may have improved since the last time I had to think about it, though. It's not as much of a consideration when writing modern JS.

robbiepaul: 👍

PR #350, opened 2019-08-27
Add in contact us page

titlescreen:


erbridge:
It's possible to make them non-static. Perhaps we should consider that.

titlescreen:
Ok no problem, I will update the static pages today then maybe if we have time at some point we can look at making the error pages non-static.

mec:
(sorry, mis-clicked the close button) My feeling is that there are pros and cons to dynamic error pages in Rails to bear in mind as well, it's one of those balance things and for now, the static ones feel like the right balance.

tekin:
Scratch that, just pushed up a change making the footer links use the `link_to` helper!

tekin:
> LGTM 👍 On the static / dynamic error page front - is there not a way to generate the error pages as part of the asset precompilation process?

It's certainly possible to generate a static page from a template but I'm not sure the effort is worth it, especially as we don't expect the page layout to change much.

PR #351, opened 2019-08-27
Create a staging slot to deploy to for zero-downtime

erbridge:
The runtime stack is still declared for the "production" slot so an incomplete deployment followed by a service restart doesn't take the service down. If not specified, the fact that we do a `Complete` deployment would mean that the configuration would be cleared when deploying, making the app fragile to a restart.

The app service doesn't change which version it's running unless it's restarted.

This situation might be improved by setting the main slot to use the currently live version explicitly, so configuration would match what's already live, though this wouldn't stop environment variable changes getting deployed. We could do this by querying the Azure API. I don't believe this is a good idea, however. It's better to deploy a complete configuration of the newest version accidentally than to deploy mixed configuration.

**BEFORE MERGE: This requires changing _both_ Azure DevOps deployment to (re)start the staging slot (possibly unnecessary, but it feels more reliable to be explicit), then swap staging to production.** It looks like you can type into the slot name, so this _can_ be done before merge, even though the slot doesn't exist. I **highly** recommend we set up production at the same time as we set up development, because release configuration gets attached to releases, so you can't easily change it after the development release occurs.

Additional reading:
- https://blogs.msdn.microsoft.com/mvpawardprogram/2017/05/16/deploy-app-azure-app-service/
- https://ruslany.net/2015/09/how-to-warm-up-azure-web-app-during-deployment-slots-swap/

### DIFF
--- a/bin/swap-slots
+++ b/bin/swap-slots
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+ENVIRONMENT_NAME=$1
+
+case $ENVIRONMENT_NAME in
+  "development")
+    SUBSCRIPTION_ID="8655985a-2f87-44d7-a541-0be9a8c2779d"
+    RESOURCE_GROUP="s118d01-app"
+    ;;
+  "production")
+    SUBSCRIPTION_ID="88bd392f-df19-458b-a100-22b4429060ed"
+    RESOURCE_GROUP="s118p01-app"
+    ;;
+  *)
+    echo "Could not find an known environment with the name: $ENVIRONMENT_NAME"
+    exit 1
+    ;;
+esac
+
+while [ "$2" ]; do
+  case $2 in
+    "--tmp-app")
+      RESOURCE_GROUP="$RESOURCE_GROUP-tmp"
+      ;;
+    *)
+      echo "Unexpected argument: $2"
+      exit 1
+      ;;
+  esac
+
+  shift
+done
+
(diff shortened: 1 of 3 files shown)


erbridge:
[Use slots to warm up app service changes without downtime](https://trello.com/c/TGVFnHqB/417-use-slots-to-warm-up-app-service-changes-without-downtime)

bin/swap-slots
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+RESOURCE_GROUP=$1
+APP_NAME="$RESOURCE_GROUP-as"
+
+az webapp deployment slot swap --name "$APP_NAME" --slot staging --resource-group "$RESOURCE_GROUP"
```
erbridge: Should this also restart the app service, then?

pezholio: Yeah, probably makes sense to do that too

bin/swap-slots
```diff
@@ -0,0 +1,35 @@
...
+    exit 1
+    ;;
+esac
+
+while [ "$2" ]; do
+  case $2 in
+    "--tmp-app")
+      RESOURCE_GROUP="$RESOURCE_GROUP-tmp"
+      ;;
+  esac
```
erbridge: This should have a `*)` case to warn about unexpected arguments. See `azure-deploy`.

PR #352, opened 2019-08-27
Fix Verify responses without "from" fields throwing an exception

erbridge:
If it exists, we assume the first attribute without "from" is always the most recent. We also assume that there will only be one, though the code will probably function fine if there are more than one.

### DIFF
--- a/lib/verify/response.rb
+++ b/lib/verify/response.rb
@@ -84,14 +84,31 @@ module Verify
     end
 
     def most_recent_verified_value(attribute_name, required: true)
-      attrs = required ? attributes.fetch(attribute_name) : attributes.dig(attribute_name)
-      return if attrs.blank?
+      if required
+        attrs = attributes.fetch(attribute_name)
 
-      most_recent_verified_attribute = attrs.sort_by { |attribute| attribute["from"].present? ? Date.strptime(attribute["from"], "%Y-%m-%d") : 0 }
-        .reverse
-        .find { |attribute| attribute["verified"] }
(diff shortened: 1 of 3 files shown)


erbridge:
[Verify responses without "from" fields throw an exception](https://trello.com/c/6hoTqFd7/615-verify-responses-without-from-fields-throw-an-exception)

lib/verify/response.rb
```diff
@@ -84,14 +84,31 @@ def full_name
...
     def most_recent_verified_value(attribute_name, required: true)
-      attrs = required ? attributes.fetch(attribute_name) : attributes.dig(attribute_name)
-      return if attrs.blank?
-
-      most_recent_verified_attribute = attrs.sort_by { |attribute| attribute["from"].present? ? Date.strptime(attribute["from"], "%Y-%m-%d") : 0 }
-        .reverse
-        .find { |attribute| attribute["verified"] }
-      raise MissingResponseAttribute, "No verified value found for #{attribute_name}" if required && most_recent_verified_attribute.nil?
-      required ? most_recent_verified_attribute.fetch("value") : most_recent_verified_attribute&.fetch("value")
+      if required
```
tekin: Let's pair on trying to simplify the code in this method. There are now a fair few more explicit `return` and `raise` calls than before, plus more conditionals and I reckon we can get them down.

tekin:
Closing this for now, we're going to refactor and try a slightly different approach to simplify this code.

PR #353, opened 2019-08-27
Fix broken asset url in svg fallback

mec:
The crown logotype svg in the GOVUK Frontend header has a fallback
in-built for older browsers. The hard-coded src value was incorrect
for the Rails asset pipeline.

This will stop us seeing rollbar errors like this one:

https://rollbar.com/dxw/teachers-payment-service/items/56/occurrences/93408935412/


mec:
I must've amended onto a prior commit, closing this and sorting it out.

PR #358, opened 2019-08-28
Stop leaking dates of birth in govuk verify responses inferrable from from fields

mec:
There are two goals here:

1. Redact the from date in the Verify response
2. Ensure the users IP Address is not passed from Rails to Rollbar

Redacting the entire from date was the simplest option at this point, we will continue to monitor the service and if we need to redact only part of the date we can come back to this.

Rails sends a couple of headers and we have expicitly set the Rollbar gem not to send the client IP.

You can test this in Rollbar by removing `|| Rails.env.development?` from the rollbar config and adding the `ROLLBAR_ACCESS_TOKEN` to your `env.development` –  you can get the token from Rollbar, after which there should be no evidence of the IP in Rollbar occurances.

## Important

Once this change is deployed to production we must delete all prioir occorances in Rollbar where the from date could be used to infer a users DOB.

### DIFF
--- a/config/initializers/rollbar.rb
+++ b/config/initializers/rollbar.rb
@@ -7,7 +7,9 @@ Rollbar.configure do |config|
   # Disable Rollbar in "test" and "development" environments
   config.enabled = false if Rails.env.test? || Rails.env.development?
 
-  config.anonymize_user_ip = true
+  # removing personally identifiable information
+  config.collect_user_ip = false
+  config.scrub_headers |= ["X-Client-Ip", "Client-Ip"]
   config.scrub_fields |= Rails.application.config.filter_parameters
 
   # By default, Rollbar will try to call the `current_user` controller method
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- redact all from dates in Verify response
 - fixed a bug where the no JS school contact was not selectable
 - fixed a bug in the GOVUK logotype svg src url
 - Improve layout of closed schools in search results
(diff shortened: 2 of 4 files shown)


config/initializers/rollbar.rb
```diff
@@ -7,7 +7,9 @@
   # Disable Rollbar in "test" and "development" environments
   config.enabled = false if Rails.env.test? || Rails.env.development?
 
-  config.anonymize_user_ip = true
```
pezholio: Does `anonymize_user_ip` not do enough in this case? I was under the impression that is removed some enough octals for it to be considered anonymized? Scrubbing headers probably makes sense though.

mec: > Does `anonymize_user_ip` not do enough in this case? I was under the impression that is removed some enough octals for it to be considered anonymized? Scrubbing headers probably makes sense though.

Yeah, I didn't really understand what was happening, according to this:

https://docs.rollbar.com/docs/ruby#section-managing-sensitive-data

`config.anonymize_user_ip = true` should ' semi-anonymization on the captured IP address by masking out the least significant bits' but that was not what I was seeing in rollbar.

The headers still had what look like the 'real' IP address and the 'masked' address is completely different?! You can take a look in Rollbar if you like – I'd love another opinion to understand what was happening.

mec: @pezholio how wedded to keeping the `anonymize_ip` are you, I love to get this in to stop the potenrial dob leak, as I don't fully understand what Rollbar is doing, my prefered option is to not send the IP at all, but I would be happy either way.

pezholio: Yeah, let's leave it as you've done here. I don't think the IP is useful in Rollbar anyway

mec:
[Stop leaking dates of birth in GOV.UK Verify responses inferrable from "from" fields](https://trello.com/c/T4PmrJW4/616-stop-leaking-dates-of-birth-in-govuk-verify-responses-inferrable-from-from-fields)

PR #361, opened 2019-08-29
Separate parsing Verify responses from the code that gets them

tekin:
This introduces the `Claim::VerifyResponseParametersParser` class for performing the task of translating a response from Verify into the attributes that we mass-assign to claims. As part of the extraction, a bug was fixed where missing `"from"` dates on attributes would cause the code to crash (in commit f2e1fcc). Apologies that this work hasn't been done in a more iterative fashion so it's easier to separate the  bug fix from the extraction, but I found it easier to write the new class from scratch rather than extract and replace it more slowly over several commits. Hopefully following the individual commits along will make this straightforward to review.


app/models/claim/verify_response_parameters_parser.rb
```diff
@@ -0,0 +1,122 @@
...
+    class MissingResponseAttribute < StandardError; end
+
+    def initialize(response_parameters)
+      @response_parameters = response_parameters
+    end
+
+    def attributes
+      identity_attributes.merge(verified_fields: verified_fields, verify_response: @response_parameters)
+    end
+
```
pezholio: Should the methods below here be private too? (given they're never called outside the class)

spec/models/claim/verify_response_parameters_parser_spec.rb
```diff
@@ -0,0 +1,271 @@
...
+          {value: "Barbara", verified: true, from: "1991-12-12", to: "2018-08-08"},
+          {value: "Bob", verified: true, from: "2018-08-09"},
+        ],
+      })
+      parser = Claim::VerifyResponseParametersParser.new(changed_name_parameters)
+
+      expect(parser.first_name).to eq "Bob"
+    end
+
+    it "returns the 'verified' value without a from/to date (assuming this will be the most recent)" do
```
pezholio: Should we assume this? What if the `to` date is not populated? Which one do we trust more?

app/models/claim/verify_response_parameters_parser.rb
tekin: Keeping them public made it easier to write tests for the specific nuances for parsing each of the attributes. If they were made private, the tests would need to be adjusted so that they were all against the `attributes` method. Probably wouldn't make things that much more complicated though to be fair...

spec/models/claim/verify_response_parameters_parser_spec.rb
tekin: I think we can only go off the Verify responses we've seen, which essentially look like the data structure below: one with both a from/to, one with neither.

tekin: The fact that one has a `to` in the past would suggest it isn't current.

pezholio: True. Best not to make assumptions about the state of the data, especially as this is precisely why we have a private beta 👍

tekin: should I update the wording of the spec to remove the word "assuming"?

pezholio: Yeah, please. Might save confusion further down the line

tekin: done.

app/models/claim/verify_response_parameters_parser.rb
tekin: I'm going to suggest we wait until @robbiepaul has done the work to split out the names as otherwise the specs will need significant rewrite to test against `:full_name` and then have to be un-rewritten to test against the individual components later.

tekin:
I've just adjusted the commit that parses addresses. Confirmed with the team at GOV.UK Verify: we will always see an `"addresses"` attribute, even for EU citizens, which simplifies the code a bit.

PR #362, opened 2019-08-29
Reorder check answers to more logical groups

titlescreen:


mec:
For this sort of PR, I would find it super useful to have a bit of information about why this change came about and maybe even some before and after screenshots?

titlescreen:
Ok on worries, will take that on board for next time :)

PR #363, opened 2019-08-29
Introduce schools data export rake tasks

mec:
So we can easily export csv files of eligble schools for both Student loans and Maths and physics.

Decided to add them to the existing `schools_data` namespace to keep things tidy.

Also took the opportunity to rename `eligible_for_tslr?` to `eligible_for_student_loans?` inline with the rest of the app as the rake task calls it.


mec:
[Export Maths and Physics eligible schools list](https://trello.com/c/TL9RuzTQ/636-export-maths-and-physics-eligible-schools-list)

lib/tasks/schools_data.rake
```diff
@@ -0,0 +1,29 @@
...
+    export_school_data(:student_loans)
+  end
+
+  desc "Export schools eligible for Maths and physics to csv"
+  task "export:maths_and_physics": :environment do
+    export_school_data(:maths_and_physics)
+  end
+end
+
+def export_school_data(policy_name)
```
tekin: Can we put this method in the `namespace` so it doesn't get globally defined?

lib/tasks/schools_data.rake
```diff
@@ -0,0 +1,29 @@
...
+  require "csv"
```
tekin: It's more conventional to stick this at the top of the file rather than here.

mec: bah! I moved it down! :(

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- introduced rake tasks to export eligible schools to csv
```
tekin: Capitalise the `I` please.

PR #364, opened 2019-08-29
🚢 Release 003 

pezholio:
- Fixed a bug where missing "from" dates on Verify attributes caused a crash
- Stop user IP address appearing in Rollbar
- Redact all from dates in Verify response
- Fixed a bug where the no JS school contact was not selectable
- Fixed a bug in the GOVUK logotype svg src url
- Improve layout of closed schools in search results
- Add contact us page
- Fix an error in the claim CSV where the current school was displaying
  incorrectly
- Switch to using deployment slots for "zero-downtime" deployment

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,13 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+## [Release 003] - 2019-08-29
+
 - Fixed a bug where missing "from" dates on Verify attributes caused a crash
-- stop user IP address appearing in Rollbar
-- redact all from dates in Verify response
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+## [Release 003] - 2019-08-29
+
 - Fixed a bug where missing "from" dates on Verify attributes caused a crash
 - stop user IP address appearing in Rollbar
```
tekin: Can we correct the case on these lines? They should start with a capital letter.

PR #365, opened 2019-08-29
Extend timeout length when coming back from Verify

pezholio:
As a user can potentially take a long time to authenticate with Verify, we've seen users sessions getting timed out when they return to the app. This extends the timeout period to 90 minutes on Verify-related pages, giving the user a bit more time to complete their actions.

### DIFF
--- a/spec/features/govuk_verify_spec.rb
+++ b/spec/features/govuk_verify_spec.rb
@@ -100,4 +100,20 @@ RSpec.feature "Teacher verifies identity using GOV.UK Verify" do
       expect(@claim.verify_response).to eq(parsed_vsp_translated_response("authentication-failed"))
     end
   end
+
+  context "user takes a particularly long time to get through verify" do
+    before do
+      stub_vsp_translate_response_request
+    end
+
+    scenario "Does not time the user out", js: true do
+      click_on "Continue"
+
+      travel 80.minutes do
+        click_on "Perform identity check"
+        expect(page).to_not have_text("Your session has ended due to inactivity")
+        expect(page).to have_text("This is your full name, address, date of birth, and gender from your digital identity")
+      end
+    end
+  end
 end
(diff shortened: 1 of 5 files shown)


pezholio:
[User gets a timeout after taking more than 30 minutes to complete Verify](https://trello.com/c/PwmbAfJC/619-user-gets-a-timeout-after-taking-more-than-30-minutes-to-complete-verify)

spec/features/govuk_verify_spec.rb
```diff
@@ -100,4 +100,20 @@
       expect(@claim.verify_response).to eq(parsed_vsp_translated_response("authentication-failed"))
     end
   end
+
+  context "user takes a particularly long time to get through verify" do
+    before do
+      stub_vsp_translate_response_request
+    end
+
+    scenario "Does not time the user out", js: true do
```
tekin: can we have another spec (or modify this existing one?) to show that it still times out after the extended timeout period?

spec/features/timeout_spec.rb
```diff
@@ -4,8 +4,8 @@
   let(:one_second_in_minutes) { 1 / 60.to_f }
 
   before do
-    allow_any_instance_of(ClaimsHelper).to receive(:claim_timeout_in_minutes) { one_second_in_minutes }
-    allow_any_instance_of(ClaimsHelper).to receive(:claim_timeout_warning_in_minutes) { one_second_in_minutes }
+    allow_any_instance_of(ApplicationController).to receive(:claim_timeout_in_minutes) { one_second_in_minutes }
```
tekin: It's a shame these feature specs rely so heavily on stubs and setting expectations on method calls as it makes them brittle as demonstrated by the need for these changes. Do you fancy rewriting them to use `travel` to verify the behaviour instead?

robbiepaul: I seem to remember trying `travel` when writing these originally but they're using JavaScript for the timer for the modal popup thing so it wasn't working.

pezholio: Ah, yeah. I remember this now. We could use Sinion or something to stub the time, but this adds a layer of complication

PR #366, opened 2019-08-29
Store claimants names as separate fields

robbiepaul:
HMRC expect employees names to be reported as three separate fields: Forename, Second forename, and Surname. Given this, and given that we receive three separate names from Verify, we should probably store claimants names as three separate fields. 

Added a data migration using the gem [`data_migrate`](https://github.com/ilyakatz/data-migrate) - it seems well supported with active development. The features it provides will allow us to confidently migrate and modify data.

There was one annoyance I found with it, it doesn't create a db schema migration for the `data_migrations` table it generates, so I created one manually. There is an open issue on the gem repo about it: https://github.com/ilyakatz/data-migrate/issues/105


app/views/claim_mailer/submitted.text.erb
```diff
@@ -1,4 +1,4 @@
-Dear <%= @claim.full_name %>,
+Dear <%= @claim.first_name %>,
```
pezholio: Do we want to address them as their full name here? It might be good to have a `full_name` method that pieces the name parts together

spec/presenters/tslr_claims_csv_spec.rb
```diff
@@ -55,7 +57,9 @@
         eligibility.claim_school_name,
         eligibility.employment_status.humanize,
         eligibility.current_school_name,
-        claim.full_name,
+        claim.first_name,
+        nil,
```
pezholio: Might be a good idea to check the middle name is in the CSV, even though it's optional. It's bitten me in the past

db/migrate/20190829145138_create_data_migrations.rb
```diff
@@ -0,0 +1,5 @@
+class CreateDataMigrations < ActiveRecord::Migration[5.2]
+  def change
+    create_table :data_migrations, primary_key: "version", id: :string unless table_exists? :data_migrations
```
pezholio: I know it's not A Thing in migrations usually, but can we wrap the arguments to `create_table` in brackets, given that we have an `unless` call that the end. Or better still, have a `return if table_exists? :data_migrations` guard clause on the first line

README.md
```diff
@@ -118,6 +118,17 @@ Code linting is performed using:
...
+### Creating data migrations
+
+When running a live service sometimes you're required to change existing data in
+some way. Using the gem [data_migrate](https://github.com/ilyakatz/data-migrate)
+we can write data migrations in a similar way we do with schema migrations.
+
+The migrations are stored in the `db/data` folder.
+
+When the service deploys it will run these data migrations at the same time as
+the schema ones.
```
pezholio: Can you add instructions on how to generate a migration, and how to run the data migrations at the same time as the DB migrations? (i.e. run `rake db:migrate:with_data`). I know it's in the gem docs, but makes sense to be here too

db/data/20190829135327_backfill_first_middle_surnames.rb
```diff
@@ -0,0 +1,26 @@
+class BackfillFirstMiddleSurnames < ActiveRecord::Migration[5.2]
+  def up
+    claims_with_full_names.each do |claim|
+      parser = Claim::VerifyResponseParametersParser.new(claim.verify_response)
```
pezholio: Nice, having the parser broken out is super useful here 😃

README.md
robbiepaul: Yeah sure, my worry was that the commands may get out of date as the gem updates but we can cross that bridge when it comes to it 👍

db/data/20190829135327_backfill_first_middle_surnames.rb
robbiepaul: Yeah it made things super simple 👍 thanks @tekin

robbiepaul:
[Store claimants' names as separate fields](https://trello.com/c/0tTLbfZq/614-store-claimants-names-as-separate-fields)

robbiepaul:
Ready for another look

bin/docker-entrypoint
```diff
@@ -11,8 +11,8 @@ else
     echo "No tables exist - setting up the database"
     rails db:setup
   else
-    echo "Running the migrations"
-    rails db:migrate
+    echo "Running the schema & data migrations"
+    rails db:migrate:with_data
```
tekin: Looking at the definition for this rake task makes me wary. It looks like it re-implements `db:migrate`. Unless there's good reason, I'd feel more comfortable running `db:migrate data:migrate` separately.

db/migrate/20190829094432_add_first_middle_last_names_to_claim.rb
```diff
@@ -0,0 +1,7 @@
+class AddFirstMiddleLastNamesToClaim < ActiveRecord::Migration[5.2]
+  def change
+    add_column :claims, :first_name, :string, limit: 100
```
tekin: How were these limits chosen?

robbiepaul: I just halved the full name one, happy to change it though

bin/docker-entrypoint
robbiepaul: 👍

README.md
```diff
@@ -118,6 +118,22 @@ Code linting is performed using:
...
+
+The migrations are stored in the `db/data` folder.
+
+- To generate a migration: `rails g data_migration add_this_to_that`
+- To run the data migration by itself: `rails data:migrate`
+- To run the data migration with the schema migration:
+  `rails db:migrate:with_data`
+
+When the service deploys it will run these data migrations at the same time as
+the schema ones.
```
tekin: can we reword this – if we run `rails db:migrate` followed by `rails data:migrate` sequentially rather than using `rails db:migrate:with_data`, which looks like it intermingles the two. I'd rather we had a workflow that required schema migrations were run first and then data migrations run after. that does some changes will require splitting across more than one release (i.e. a follow up PR to drop the now-unused column), but I think this approach is safer and less likely to result in irreversible accidental data loss.

db/migrate/20190829094432_add_first_middle_last_names_to_claim.rb
tekin: no, It should be fine.

README.md
robbiepaul: Yeah that makes sense, I've made the change

robbiepaul:
@tekin Ready for another review. I added the separation of `db:migrate` and `data:migrate`, updated the readme. 

I also had to add a commit that ignores some warnings from standard for the generated file 55e824cb1754a7c9f5c10a68d934e56473ea815f

PR #371, opened 2019-09-02
A service operator is able to mark a claim as approved

pezholio:
This is an MVP to allow logged in service operator can mark a claim as 'approved'. Users can view all claims awaiting approval, click a link to view the relevant information to allow a claim to be reviewed, and then click a button to approve.

image

image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog]
 
 ## [Release 004] - 2019-09-04
 
+- Allow service operators to approve claims
 - Fix claims export ordering
 - Store claimants' names as separate fields
 - Add ability to run data migrations
--- a/app/controllers/admin/claim_approvals_controller.rb
+++ b/app/controllers/admin/claim_approvals_controller.rb
@@ -1,0 +1,9 @@
+class Admin::ClaimApprovalsController < Admin::BaseAdminController
+  before_action :ensure_service_operator
+
+  def create
+    @claim = Claim.find(params[:claim_id])
+    @claim.approve!(approved_by: admin_session.user_id)
+    redirect_to admin_claims_path, notice: "Claim has been approved successfully"
+  end
+end
--- a/app/controllers/admin/claim_approvals_controller.rb
+++ b/app/controllers/admin/claim_approvals_controller.rb
@@ -4,6 +4,7 @@ class Admin::ClaimApprovalsController < Admin::BaseAdminController
   def create
     @claim = Claim.find(params[:claim_id])
     if @claim.approve!(approved_by: admin_session.user_id)
+      ClaimMailer.approved(@claim).deliver_later
       redirect_to admin_claims_path, notice: "Claim has been approved successfully"
     else
       redirect_to admin_claim_path(@claim), notice: "Claim cannot be approved"
--- a/app/controllers/admin/claim_approvals_controller.rb
+++ b/app/controllers/admin/claim_approvals_controller.rb
@@ -3,7 +3,10 @@ class Admin::ClaimApprovalsController < Admin::BaseAdminController
 
   def create
     @claim = Claim.find(params[:claim_id])
-    @claim.approve!(approved_by: admin_session.user_id)
-    redirect_to admin_claims_path, notice: "Claim has been approved successfully"
+    if @claim.approve!(approved_by: admin_session.user_id)
+      redirect_to admin_claims_path, notice: "Claim has been approved successfully"
+    else
+      redirect_to admin_claim_path(@claim), notice: "Claim cannot be approved"
+    end
   end
 end
--- a/app/controllers/admin/claims_controller.rb
+++ b/app/controllers/admin/claims_controller.rb
@@ -14,6 +14,12 @@ class Admin::ClaimsController < Admin::BaseAdminController
     @claim = Claim.find(params[:id])
   end
 
+  def approve
+    @claim = Claim.find(params[:id])
+    @claim.update(approved_at: DateTime.now)
+    redirect_to admin_claims_path, notice: "Claim has been approved successfully"
+  end
+
   private
 
   def ensure_service_operator
--- a/app/controllers/admin/claims_controller.rb
+++ b/app/controllers/admin/claims_controller.rb
@@ -16,10 +16,7 @@ class Admin::ClaimsController < Admin::BaseAdminController
 
   def approve
     @claim = Claim.find(params[:id])
-    @claim.update(
-      approved_at: Time.zone.now,
-      approved_by: admin_session.user_id
-    )
+    @claim.approve!(approved_by: admin_session.user_id)
     redirect_to admin_claims_path, notice: "Claim has been approved successfully"
   end
 
--- a/app/controllers/admin/claims_controller.rb
+++ b/app/controllers/admin/claims_controller.rb
@@ -16,7 +16,10 @@ class Admin::ClaimsController < Admin::BaseAdminController
 
   def approve
     @claim = Claim.find(params[:id])
-    @claim.update(approved_at: DateTime.now)
+    @claim.update(
+      approved_at: Time.zone.now,
+      approved_by: admin_session.user_id
+    )
     redirect_to admin_claims_path, notice: "Claim has been approved successfully"
   end
 
--- a/app/controllers/admin/claims_controller.rb
+++ b/app/controllers/admin/claims_controller.rb
@@ -10,6 +10,10 @@ class Admin::ClaimsController < Admin::BaseAdminController
     end
   end
 
+  def show
+    @claim = Claim.find(params[:id])
+  end
+
   private
 
   def ensure_service_operator
--- a/app/controllers/admin/claims_controller.rb
+++ b/app/controllers/admin/claims_controller.rb
@@ -2,7 +2,7 @@ class Admin::ClaimsController < Admin::BaseAdminController
   before_action :ensure_service_operator
 
   def index
-    @claims = Claim.includes(eligibility: [:claim_school, :current_school]).submitted.order(:submitted_at)
+    @claims = Claim.includes(eligibility: [:claim_school, :current_school]).awaiting_approval.order(:submitted_at)
 
     respond_to do |format|
       format.csv { send_file TslrClaimsCsv.new(@claims).file, type: "text/csv", filename: "claims.csv" }
--- a/app/controllers/admin/claims_controller.rb
+++ b/app/controllers/admin/claims_controller.rb
@@ -2,11 +2,11 @@ class Admin::ClaimsController < Admin::BaseAdminController
   before_action :ensure_service_operator
 
   def index
-    claims = Claim.includes(eligibility: [:claim_school, :current_school]).submitted.order(:submitted_at)
-    csv = TslrClaimsCsv.new(claims)
+    @claims = Claim.includes(eligibility: [:claim_school, :current_school]).submitted.order(:submitted_at)
 
     respond_to do |format|
-      format.csv { send_file csv.file, type: "text/csv", filename: "claims.csv" }
+      format.csv { send_file TslrClaimsCsv.new(@claims).file, type: "text/csv", filename: "claims.csv" }
+      format.html
     end
   end
 
--- a/app/controllers/admin/claims_controller.rb
+++ b/app/controllers/admin/claims_controller.rb
@@ -13,16 +13,4 @@ class Admin::ClaimsController < Admin::BaseAdminController
   def show
     @claim = Claim.find(params[:id])
   end
-
-  def approve
-    @claim = Claim.find(params[:id])
-    @claim.approve!(approved_by: admin_session.user_id)
-    redirect_to admin_claims_path, notice: "Claim has been approved successfully"
-  end
-
-  private
-
-  def ensure_service_operator
-    render "admin/auth/failure", status: :unauthorized unless service_operator_signed_in?
-  end
 end
--- a/app/helpers/admin/claims_helper.rb
+++ b/app/helpers/admin/claims_helper.rb
@@ -1,0 +1,24 @@
+module Admin
+  module ClaimsHelper
+    include ::ClaimsHelper
+
+    def admin_eligibility_answers(eligibility)
+      [].tap do |a|
+        a << [t("student_loans.questions.admin.qts_award_year"), academic_years(eligibility.qts_award_year)]
+        a << [t("student_loans.questions.admin.claim_school"), eligibility.claim_school_name]
+        a << [t("questions.admin.current_school"), eligibility.current_school_name]
+        a << [t("student_loans.questions.admin.subjects_taught"), subject_list(eligibility.subjects_taught)]
+        a << [t("student_loans.questions.admin.had_leadership_position"), (eligibility.had_leadership_position? ? "Yes" : "No")]
+        a << [t("student_loans.questions.admin.mostly_performed_leadership_duties"), (eligibility.mostly_performed_leadership_duties? ? "Yes" : "No")] if eligibility.had_leadership_position?
+      end
+    end
+
+    def admin_identity_answers(claim)
+      [
+        [t("questions.admin.teacher_reference_number"), claim.teacher_reference_number],
+        [t("questions.admin.national_insurance_number"), claim.national_insurance_number],
+        [t("questions.admin.email_address"), claim.email_address],
+      ]
+    end
+  end
+end
--- a/app/helpers/admin/claims_helper.rb
+++ b/app/helpers/admin/claims_helper.rb
@@ -13,10 +13,13 @@ module Admin
       end
     end
 
-    def admin_identity_answers(claim)
+    def admin_personal_details(claim)
       [
         [t("questions.admin.teacher_reference_number"), claim.teacher_reference_number],
+        [t("verified_fields.full_name").capitalize, claim.full_name],
         [t("questions.admin.national_insurance_number"), claim.national_insurance_number],
+        [t("verified_fields.date_of_birth").capitalize, l(claim.date_of_birth)],
+        [t("verified_fields.address").capitalize, sanitize(claim.address("<br>").html_safe, tags: %w[br])],
         [t("questions.admin.email_address"), claim.email_address],
       ]
     end
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -36,6 +36,7 @@ class Claim < ApplicationRecord
     updated_at: false,
     verified_fields: false,
     verify_response: true,
+    approved_at: false,
   }.freeze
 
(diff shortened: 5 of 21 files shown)


pezholio:
[A Service Operator is able to mark a claim as "approved"](https://trello.com/c/mId5CVFV/618-a-service-operator-is-able-to-mark-a-claim-as-approved)

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ The format is based on [Keep a Changelog]
 - Add ability to run data migrations
 - Add data migration to back fill claimant's full name, middle name and surname
 - Introduced rake tasks to export eligible schools to CSV
+- Introduced rake tasks to export eligible schools to csv
```
erbridge: Duplicate. Also I thought we were top-loading?

app/models/claim.rb
```diff
@@ -103,7 +104,7 @@ class Claim < ApplicationRecord
   before_save :normalise_bank_account_number, if: :bank_account_number_changed?
   before_save :normalise_bank_sort_code, if: :bank_sort_code_changed?
 
-  scope :submitted, -> { where.not(submitted_at: nil) }
+  scope :submitted, -> { where(approved_at: nil).where.not(submitted_at: nil) }
```
erbridge: Can we rename this to be clear that it excludes approved claims?

app/views/admin/claims/_answer_section.html.erb
```diff
@@ -0,0 +1,19 @@
...
+      <dt class="govuk-summary-list__key">
+        <%= label %>
+      </dt>
+
+      <dd class="govuk-summary-list__value">
+        <%= answer %>
+      </dd>
+    </div>
+  <% end %>
+</dl>
```
erbridge: Is this basically the same as the teacher view? Can we share it?

app/views/admin/claims/_answer_section.html.erb
```diff
@@ -0,0 +1,19 @@
+<% unless heading.nil? %>
+  <h2 class="govuk-heading-m">
+    <%= heading %>
+  </h2>
+<% end %>
+
+<dl class="govuk-summary-list govuk-!-margin-bottom-9">
+  <%- answers.each do |(label, answer, slug)| %>
```
erbridge:
```suggestion
  <%- answers.each do |(label, answer)| %>
```

app/views/admin/claims/_claim.html.erb
```diff
@@ -0,0 +1,5 @@
+<tr class="govuk-table__row">
+  <th scope="row" class="govuk-table__header"><%= claim.reference %></th>
+  <td class="govuk-table__cell"><%= claim.first_name %> <%= claim.surname %></td>
```
erbridge: We have logic to create this already for the emails. Can we avoid duplication?

app/views/admin/claims/index.html.erb
```diff
@@ -0,0 +1,28 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <h1 class="govuk-heading-xl">
+      Claims
+    </h1>
+
+    <% if @claims.count > 0 %>
```
erbridge:
```suggestion
    <% unless @claims.empty? %>
```

app/views/admin/claims/index.html.erb
```diff
@@ -0,0 +1,28 @@
...
+      <table class="govuk-table">
+        <thead class="govuk-table__head">
+          <tr class="govuk-table__row">
+            <th scope="col" class="govuk-table__header">Reference</th>
+            <th scope="col" class="govuk-table__header">Applicant Name</th>
+            <th scope="col" class="govuk-table__header"></th>
+          </tr>
+        </thead>
+        <tbody class="govuk-table__body">
+          <%= render partial: "claim", collection: @claims %>
```
erbridge: This is coupled to the headers. What was the motivation for making it a partial?

app/views/admin/claims/index.html.erb
tekin: or
```suggestion
    <% if @claims.any? %>
```

app/views/admin/claims/_claim.html.erb
pezholio: Ah, I thought we'd got rid of that

app/controllers/admin/claims_controller.rb
```diff
@@ -2,14 +2,28 @@ class Admin::ClaimsController < Admin::BaseAdminController
...
     end
   end
 
+  def show
+    @claim = Claim.find(params[:id])
+  end
+
+  def approve
+    @claim = Claim.find(params[:id])
+    @claim.update(
```
tekin: can we wrap the act of approval in a method on the model? That encapsulation will make it easier to add (and test) additional behaviour around the approval process. For example, raising an error if the claim isn't approvable (i.e. it hasn't been submitted, or has already been approved)

app/helpers/admin/claims_helper.rb
```diff
@@ -0,0 +1,24 @@
+module Admin
+  module ClaimsHelper
+    include ::ClaimsHelper
+
+    def eligibility_answers(eligibility)
```
tekin: I'm uneasy with this method having exactly the same name as a method in `ClaimsHelper` as it means one will replace the other if both modules are included in the same place. Same goes for the `identity_answers` method below. This could lead to weird and wonderful bugs. Safer to give these methods a different name.

I'm also not 100% convinced of the value of having these helpers in the first place vs just having plain HTML markup with the values interpolated right there in the template. I guess having this helper does mean we can write a test that verifies the correct labels and values are being generated. The advantage of not having the method is that it gives us much more flexibility to change the view that an admin sees without being coupled to this Ruby code. When we're so early in the stages of defining the UI for admins, that flexibility might be useful as it won't preclude the UI having to display the information as a definition list. What do you think?

app/views/admin/claims/_answer_section.html.erb
pezholio: I'd rather not, becuase the teacher view has slightly different markup (two `<dd>` elements instead of one), and they both have different audiences, so best to keep them apart I reckon

app/controllers/admin/claims_controller.rb
```diff
@@ -2,14 +2,28 @@ class Admin::ClaimsController < Admin::BaseAdminController
...
       format.csv { send_file csv.file, type: "text/csv", filename: "claims.csv" }
+      format.html
     end
   end
 
+  def show
+    @claim = Claim.find(params[:id])
+  end
+
+  def approve
```
tekin: A most RESTful approach would be to implement this as `ClaimApprovalsController#create`.

app/helpers/application_helper.rb
```diff
@@ -1,4 +1,6 @@
 module ApplicationHelper
+  include GovukFormHelper
```
tekin: Why is this being included here?

config/application.rb
```diff
@@ -41,5 +41,7 @@ class Application < Rails::Application
 
     # Set a css_compressor so sassc-rails does not overwrite the compressor
     config.assets.css_compressor = nil
+
+    config.action_controller.include_all_helpers = false
```
tekin: Why are we doing this now?

app/views/admin/claims/index.html.erb
pezholio: To make the markup simpler really. I've never thought about not using a partial for table rows.

spec/features/admin_claim_approve_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  context "User is logged in as a support user" do
+    before do
+      stub_dfe_sign_in_with_role(AdminSession::SUPPORT_AGENT_DFE_SIGN_IN_ROLE_CODE)
+      visit admin_path
+      click_on "Sign in"
+    end
+
+    scenario "User cannot view claims to approve" do
```
tekin: Can we also have some spec coverage to verify calling the approval action as a support user isn't authorised too? That can be done as a request spec calling the action directly.

config/application.rb
pezholio: This was because I needed the helpers to be coupled to the controllers (so I could have two `eligibility_answers` in different helpers depending on context). This may be moot now though if we decide to move the logic to the views

app/helpers/admin/claims_helper.rb
pezholio: I'm all for that, I only really used the helper in this instance because it mirrored the way we did it before. I'm happy to look at moving the logic to the view, especially if we back this up with view tests.

app/controllers/admin/claims_controller.rb
pezholio: Ah, nice. I like that approach 👍

app/helpers/admin/claims_helper.rb
pezholio: I've had a crack at this, and I'm not sure what the benefit of adding all the logic to the view is. It just seems to complicate things more and clog up the views. I do take your point on the naming of the helpers. Better to just prefix the helpers with `admin_` rather than add the complication and possible name collision.

app/helpers/admin/claims_helper.rb
```diff
@@ -0,0 +1,24 @@
...
+        a << [t("student_loans.questions.admin.claim_school"), eligibility.claim_school_name]
+        a << [t("questions.admin.current_school"), eligibility.current_school_name]
+        a << [t("student_loans.questions.admin.subjects_taught"), subject_list(eligibility.subjects_taught)]
+        a << [t("student_loans.questions.admin.had_leadership_position"), (eligibility.had_leadership_position? ? "Yes" : "No")]
+        a << [t("student_loans.questions.admin.mostly_performed_leadership_duties"), (eligibility.mostly_performed_leadership_duties? ? "Yes" : "No")] if eligibility.had_leadership_position?
+      end
+    end
+
+    def admin_identity_answers(claim)
+      [].tap do |a|
```
tekin: Given there are no conditionals in this code, does it make sense to make this simply return an array without the `tap` and append calls?

app/models/claim.rb
```diff
@@ -115,6 +117,17 @@ def submit!
     end
   end
 
+  def approve!(approved_by:)
+    if approvable?
+      update(
+        approved_at: DateTime.now,
```
tekin: Can we make this consistent with the `submit!` method above and use `Time.zone.now`? There is a subtle difference between the two values: one accounts for the configured timezone in the app, the other uses the server's configured timezone, which may not be the same.

app/views/admin/claims/_answer_section.html.erb
```diff
@@ -0,0 +1,19 @@
+<% unless heading.nil? %>
```
tekin: Every call to this partial includes a heading so this conditional is not necessary.

app/views/admin/claims/index.html.erb
tekin: I've always been on the fence regarding using a partial to render a collection. I think if the row markup is simple enough I prefer the convenience of having it all in one template. For more complex pages and markup it can make sense to pull them out to a partial.

Incidentally there is also the Rails convention-over-configuration shorthand for this call:

```suggestion
          <%= render @claims %>
```

app/views/admin/page/index.html.erb
```diff
@@ -4,6 +4,12 @@
       Admin
     </h1>
 
-    <%= link_to 'Download claims', admin_claims_path(format: :csv), class: "govuk-button" if service_operator_signed_in? %>
+    <% if service_operator_signed_in? %>
+      <p>
+        <%= link_to 'Approve claims', admin_claims_path, class: "govuk-button" %>
```
tekin: Should this be styled as a link, not a button?

app/views/admin/claims/_claim.html.erb
```diff
@@ -0,0 +1,5 @@
+<tr class="govuk-table__row">
+  <th scope="row" class="govuk-table__header"><%= claim.reference %></th>
+  <td class="govuk-table__cell"><%= claim.full_name %></td>
+  <td class="govuk-table__cell"><%= link_to 'View claim', admin_claim_path(claim), class: "govuk-button" %></td>
```
tekin: Should this be styled as a link, not a button?

spec/features/admin_claim_approve_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+  context "User is logged in as a service operator" do
+    before do
+      stub_dfe_sign_in_with_role(AdminSession::SERVICE_OPERATOR_DFE_SIGN_IN_ROLE_CODE)
+      visit admin_path
+      click_on "Sign in"
+    end
+
+    scenario "User can approve a claim" do
+      freeze_time do
+        submitted_claims = create_list(:claim, 5, :submitted, submitted_at: DateTime.now, approved_at: nil)
```
tekin: You shouldn't need to set `submitted_at` and `approved_at` if you're using the `submitted` trait:

```suggestion
        submitted_claims = create_list(:claim, 5, :submitted)
```

spec/features/admin_claim_approve_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+        click_on "Approve claims"
+
+        expect(page).to have_content(claim_to_approve.reference)
+
+        find("a[href='#{admin_claim_path(claim_to_approve)}']").click
+        click_on "Approve"
+
+        claim_to_approve.reload
+
+        expect(claim_to_approve.approved_at).to eq(DateTime.now)
```
tekin: Best to stick to `Time.zone.now` to avoid potential timezone-related weirdness?

spec/features/admin_claim_approve_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+        expect(claim_to_approve.approved_by).to eq("123")
```
tekin: If we set an actual `user_id` explicitly in the call to `stub_dfe_sign_in_with_role` above, we can assert that this value matches the value we stub, which will make for a more thorough and less brittle spec.

spec/helpers/admin/claims_helper_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+describe Admin::ClaimsHelper do
+  let(:school) { schools(:penistone_grammar_school) }
+
+  describe "eligibility_answers" do
+    let(:eligibility) do
+      build(
+        :student_loans_eligibility,
+        qts_award_year: "2013_2014",
+        claim_school: school,
+        current_school: school,
```
tekin: To be safe, how about we assign two different schools to make extra sure the values being returned are the correct ones? We already saw a similar bug slip through in the CSV export.

spec/helpers/admin/claims_helper_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+        [I18n.t("student_loans.questions.admin.had_leadership_position"), "Yes"],
+        [I18n.t("student_loans.questions.admin.mostly_performed_leadership_duties"), "No"],
+      ]
+
+      expect(helper.admin_eligibility_answers(eligibility)).to eq expected_answers
+    end
+
+    it "excludes questions skipped from the flow" do
+      eligibility.had_leadership_position = false
+      expect(helper.admin_eligibility_answers(eligibility)).to_not include([I18n.t("student_loans.questions.admin.had_leadership_position"), "Yes"])
```
tekin:
```suggestion
      expect(helper.admin_eligibility_answers(eligibility)).to include([I18n.t("student_loans.questions.admin.had_leadership_position"), "No"])
```

spec/models/claim_spec.rb
```diff
@@ -360,12 +360,13 @@
     end
   end
 
-  describe "submitted" do
+  describe "awaiting_approval" do
     let!(:submitted_claims) { create_list(:claim, 5, :submitted) }
     let!(:unsubmitted_claims) { create_list(:claim, 2, :submittable) }
+    let!(:approved_claims) { create_list(:claim, 5, :submitted, approved_at: Time.zone.now) }
```
tekin:
```suggestion
    let!(:approved_claims) { create_list(:claim, 5, :approved) }
```

spec/models/claim_spec.rb
```diff
@@ -360,12 +360,13 @@
...
 
-    it "returns submitted claims" do
-      expect(subject.class.submitted).to match_array(submitted_claims)
+    it "returns claims awaiting approval" do
```
tekin: I think we talk about "awaiting approval", but also use the "submitted" trait and variable name. Maybe we should included the word submitted in the spec wording to join things up a bit more? Either that or we rename the trait, but that feels a bit more involved and may not make sense elsewhere.

```suggestion
    it "returns submitted claims awaiting approval" do
```

app/controllers/admin/claim_approvals_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Admin::ClaimApprovalsController < Admin::BaseAdminController
+  before_action :ensure_service_operator
+
+  def create
+    @claim = Claim.find(params[:claim_id])
+    @claim.approve!(approved_by: admin_session.user_id)
```
tekin: I feel like this code will potentially mean the application lies. i.e. if you call the action on a claim that cannot be approved, it will redirect you and tell you the "Claim has been approved successfully" even though it hasn't. Should we instead conditionally do the redirect only if the approval is successful and give a suitable message if isn't?

spec/models/claim_spec.rb
```diff
@@ -440,4 +441,41 @@
...
+  end
+
+  describe "approve!" do
+    it "approves a claim" do
+      claim = create(:claim, :submitted)
+
+      freeze_time do
+        claim.approve!(approved_by: "12345")
+
+        expect(claim.approved_at).to eq(DateTime.now)
```
tekin: `Time.zone.now`

spec/requests/admin_claim_approvals_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+        freeze_time do
+          post admin_claim_approvals_path(claim_id: claim.id)
+
+          follow_redirect!
+
+          expect(response.body).to include("Claim has been approved successfully")
+
+          claim.reload
+
+          expect(claim.approved_at).to eq(DateTime.now)
```
tekin:
```suggestion
          expect(claim.approved_at).to eq(Time.zone.now)
```

spec/features/admin_claim_approve_spec.rb
```diff
@@ -27,6 +27,13 @@
 
         expect(page).to have_content("Claim has been approved successfully")
         expect(page).to_not have_content(claim_to_approve.reference)
+
+        expect(ActionMailer::DeliveryJob).to have_been_enqueued.with(
```
tekin: Let's do a more end-to-end check here and check deliveries to the person with the expected subject or something.

PR #375, opened 2019-09-03
Update Slack script to send error if build fails

pezholio:
This adds another option to add to the Slack script that sends an error if the build fails. The release pipeline will need to be changed before this is merged.


pezholio:
[Alert devs in Slack if a release fails](https://trello.com/c/mWzlIm0N/651-alert-devs-in-slack-if-a-release-fails)

PR #376, opened 2019-09-03
Use each instead of find_each in claims export

mec:
Active Record's find_each cannot be ordered and the claims export must
be ordered.

The claim export is a temporary measure for private beta only so we
are not worried about the load on the service that switching to each
will incur as the number of cliams is limited.

Details about find_each in the Rails docs:

https://edgeguides.rubyonrails.org/active_record_querying.html#retrieving-multiple-objects-in-batches

## wot no test!? :O
I decided not to add a test as I wasn't sure creating a whole bunch of claims 
in the DB and slowing the test suite outweighed knowing that this tiny change is right.

Could I have done something smarter for a test? Thoughts/suggestions please?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- Fix claims export ordering
 - Store claimants' names as separate fields
 - Add ability to run data migrations
 - Add data migration to back fill claimant's full name, middle name and surname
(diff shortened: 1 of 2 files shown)


mec:
[CSV export is ordered by submitted_at](https://trello.com/c/FID1Wlle/654-csv-export-is-ordered-by-submittedat)

mec:
> 👍 I think it's fine. You could have maybe created one more claim that was submitted after the first and asserted that it appeared in the next row but I don't think it's needed.

I had two claims in my local db and the `find_each` issue isn't present with so few so I figured that doing that would be kind of redundant. We're seeing the issue in prod so we'd have to add around the number of claims (40ish) as there are there to start seeing real issues.

Thanks @robbiepaul

PR #377, opened 2019-09-03
We can generate a payroll CSV

robbiepaul:
The CSV needs to be in a format that our payroll provider (Cantium) can ingest.

The payroll data should only contain approved claims, but the process to approve claims hasn't been finalised yet, so that will come later. We're just using submitted claims for now.

There are a few fields that don't have values. That's because we either don't capture the data, can't infer it or just not sure what value to put.

Some of the values are bespoke static values that don't make sense (eg. "T" for has a student loan). I decided to put these values in the translations file so we can be a bit more descriptive with the translation key, it also links it with the bespoke headers that Cantium expects.

For now we're just using the first day of the current month for the start date, the end date is one week later. This will need refining later.

One of the acceptance criteria is that the leading zeros on the sort code and account number are preserved. I have confirmed they are preserved in the CSV but it's worth noting that if a user opens the CSV in Excel or similar the software will strip the leading zero unless the user explicitly sets the formatting for that cell. I think this is ok as Cantium will ingest the CSV and it shouldn't be opened and saved by a user.

**Note** I need to confirm my assumption about cumulative vs non-cumulative


robbiepaul:
[We can generate a payroll CSV file for Cantium](https://trello.com/c/Uq4DhbyF/617-we-can-generate-a-payroll-csv-file-for-cantium)

spec/presenters/payroll_data_csv_row_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+        "Scheme B",
+        claim.eligibility.student_loan_repayment_amount.to_s,
+        claim.reference,
+      ])
+    end
+
+    it "escapes fields with strings that could be dangerous in Microsoft Excel and friends" do
+      claim.address_line_1 = "=ActiveCell.Row-1,14"
+
+      expect(row[PayrollDataCsv::FIELDS.index(:address_line_1)]).to eq("\\#{claim.address_line_1}")
```
erbridge: I'm suspiscious of this. Do backslashes actually escape equals signs in Excel?

spec/presenters/payroll_data_csv_row_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    let(:row) { CSV.parse(subject.to_s).first }
+    let(:start_of_month) { Date.today.at_beginning_of_month }
+
+    let(:claim) do
+      build(:claim, :submittable,
+        payroll_gender: :female,
+        date_of_birth: Date.new(1980, 12, 1),
+        student_loan_plan: StudentLoans::PLAN_2,
+        bank_sort_code: "001122",
+        bank_account_number: "01234567",
```
erbridge: I _think_ if you wrap a number in quotes, Excel will treat it as a string. Or maybe it's `="001122"`. This was the solution I originally implemented to (incompletely) deal with formulae in values.

spec/presenters/payroll_data_csv_row_spec.rb
robbiepaul: I'm not sure. When I went through it with Tamas, he seemed happy enough with prepending these with a backslash

spec/presenters/payroll_data_csv_row_spec.rb
robbiepaul: I think the idea is for this CSV to be directly consumed without user editing so the value in Excel shouldn't matter. I think CSV parsers should preserve the leading zeros, I know the Ruby one does: 

```
irb(main):004:0> CSV.parse("001,XYZ,01234567")
=> [["001", "XYZ", "01234567"]]
```

spec/presenters/payroll_data_csv_row_spec.rb
erbridge: Could we test it? When I was researching this, I was put under the impression that escaping was more involved.

spec/presenters/payroll_data_csv_row_spec.rb
erbridge: Yeah agreed. I guess we'll need to rely on guidance to tell people not to open the CSV and resave it.

spec/presenters/payroll_data_csv_row_spec.rb
robbiepaul: Yeah, I'll find a way to test it 👍

tekin: more details on the current approach to escaping in the history – https://github.com/dFE-Digital/dfe-teachers-payment-service/commit/eda1b969

spec/presenters/payroll_data_csv_row_spec.rb
tekin: Yeah I think this is right. If you just "double click" a CSV file and let Excel open it, there is little we can do to prevent the leading zero's being dropped. The way to open the CSV and preserve the leading zeros is to use the File > Import feature and tell Excel how to handle specific columns.

app/models/payroll/claim_csv_row.rb
```diff
@@ -0,0 +1,110 @@
...
+
+require "delegate"
+require "csv"
+require "excel_utils"
+
+module Payroll
+  class ClaimCsvRow < SimpleDelegator
+    DATE_FORMAT = "%m/%d/%Y"
+
+    VALUES = [
```
tekin: Why are these constant values put in this `VALUES` array constant?

robbiepaul: No reason in particular, just felt a bit strange having the them as top level class constants. Felt more readable scoping them under `VALUES` to show they are values in the CSV row. I might be overthinking it :)

app/models/payroll/claim_csv_row.rb
```diff
@@ -0,0 +1,110 @@
...
+      UNITED_KINGDOM = "United Kingdom",
+      BASIC_RATE_TAX_CODE = "BR",
+      CUMULATIVE_TAX_BASIS = "0",
+      NOT_EMPLOYEES_ONLY_JOB = "3",
+      ALL_EMPLOYEES_NI_CATEGORY = "A",
```
tekin: I'm not sure I follow this constant name...

app/models/payroll/claim_csv_row.rb
```diff
@@ -0,0 +1,110 @@
...
+      HAS_STUDENT_LOAN = "T",
+      SCHEME_B_NAME = "Scheme B",
```
tekin: So this isn't something you will have been privy too, but rather than saying "Scheme B" and "Scheme A" I think we're free to just say the policy name so there isn't extra effort required to translate between the policy and a magic "scheme" letter. Can we just say `"Student loans"` for now?

tekin: Actually, can we some how get this from the claim? I'm thinking that'll make it better encapsulated for when Maths & Physics gets added.

app/models/payroll/claims_csv.rb
```diff
@@ -0,0 +1,63 @@
...
+      payroll_gender: "GENDER",
+      start_date: "START_DATE",
+      end_date: "END_DATE",
+      date_of_birth: "BIRTH_DATE",
+      email_address: "EMAIL",
+      address_line_1: "ADDR_LINE_1",
+      address_line_2: "ADDR_LINE_2",
+      address_line_3: "ADDR_LINE_3",
+      address_line_4: "ADDR_LINE_4",
+      county: "ADDR_LINE_5",
```
tekin: Although their sample file has a "County" under `ADDR_LINE_5`, I don't think we should interpret that to mean that the 5 line should be the county and the 6th the postcode. I read the sample as the address should appear in the address lines in the order they appear, i.e. a three line address with the postcode as the third line would have postcode under `ADDR_LINE_3`

app/models/payroll/claim_csv_row.rb
```diff
@@ -0,0 +1,110 @@
...
+    def student_loan_plan
+      if model.student_loan_plan == "plan_1" || model.student_loan_plan == "plan_1_and_2"
+        STUDENT_LOAN_PLAN_1
+      elsif model.student_loan_plan == "plan_2"
+        STUDENT_LOAN_PLAN_2
+      end
+    end
+
+    def bank_name
+      model.full_name
```
tekin: oops, looks like @mec is removing this method in https://github.com/DFE-Digital/dfe-teachers-payment-service/pull/380

spec/presenters/payroll_data_csv_row_spec.rb
erbridge: My question is: does `\` actually escape characters in spreadsheet software? The article linked in that commit suggests the only reliable way of escaping formulae is with a prefixed tab character.

app/models/payroll/claim_csv_row.rb
```diff
@@ -0,0 +1,110 @@
...
+    end
+
+    def scheme_name
+      SCHEME_B_NAME
+    end
+
+    def scheme_amount
+      model.eligibility.student_loan_repayment_amount.to_s
```
tekin: Do we actually need to call `to_s` here? I thought that would happen automatically in `CSV.generate_line`

app/models/payroll/claim_csv_row.rb
```diff
@@ -0,0 +1,110 @@
...
+
+    def student_loan_plan
+      if model.student_loan_plan == "plan_1" || model.student_loan_plan == "plan_1_and_2"
+        STUDENT_LOAN_PLAN_1
+      elsif model.student_loan_plan == "plan_2"
+        STUDENT_LOAN_PLAN_2
+      end
+    end
+
+    def bank_name
```
erbridge: Is this the user's name? I would have expected it to be the name of the bank. Also, the full name isn't necessarily what's actually on the bank account (e.g. a user who wants the payment to go into their partner's account). If we need that information, we should ask.

spec/features/admin_csv_download_spec.rb
```diff
@@ -20,6 +20,19 @@
...
+      submitted_claims = create_list(:claim, 5, :submittable, submitted_at: DateTime.now)
+      create_list(:claim, 2, :submittable)
+
+      expect(page).to have_link(nil, href: payroll_admin_claims_path(format: :csv))
+      click_on "Download payroll data"
+
+      expect(page.response_headers["Content-Type"]).to eq("text/csv")
+
+      csv = CSV.parse(body)
+      expect(csv.count).to eq(submitted_claims.count + 1)
```
tekin: can we add some more specific expectations here that will give us confidence the file being returned is definitely the payroll one?

app/models/payroll/claim_csv_row.rb
tekin: It's the closet we have for now though. The alternative is we add an extra step in the claim process to explicitly ask the user for the name on their bank account, which is definitely outside the scope of this PR. 

I do think it would make sense to rename this method though to something less ambiguous. `banking_name` perhaps?

spec/presenters/payroll_data_csv_row_spec.rb
tekin: hmmm, I seemed pretty confident it would in my commit message. I did test it in Excel, Google Sheets and Numbers, and it prevented the formula being evaluated in all of those. I think the reason the article linked to in in the message suggests a tab character is because the "tab character doesn’t show up visually on Excel, or Google Sheets", and therefor it's a method that doesn't result in the data looking weird. In our case we are being a bit more blunt and visible with the `\`

app/models/payroll/claim_csv_row.rb
mec: I had a feeling in my bones we would use it elsewhere! I can drop that commit from my PR if it is helpful?

spec/presenters/payroll_data_csv_row_spec.rb
erbridge: Ok, cool. Just wanted to make sure. My research suggested that wasn't an option (or rather everyone goes to more extreme lengths to get it done). Is it ok that we have `\`s in the data we send to Cantium?

tekin: I'd say the likelyhood of that is fairly low. For it to happen, a user would have to enter highly unusual data in an address field (whether malicious or otherwise) and the claim would then need to be approved. In that situation some manual intervention may be required (either by them or us) if the presence of a backslash is going to cause a problem. On balance we are probably best crossing that edge-case if/when it arises to remove the security risk.

app/controllers/admin/claims_controller.rb
```diff
@@ -3,13 +3,22 @@ class Admin::ClaimsController < Admin::BaseAdminController
...
     claims = Claim.includes(eligibility: [:claim_school, :current_school]).submitted.order(:submitted_at)
-    csv = TslrClaimsCsv.new(claims)
+    csv = StudentLoans::ClaimsCsv.new(claims)
 
     respond_to do |format|
       format.csv { send_file csv.file, type: "text/csv", filename: "claims.csv" }
     end
   end
 
+  def payroll
```
pezholio: At the risk of being another person throwing my oar in here, this doesn't feel very RESTful, can we maybe control the type of csv being returned via a GET parameter? Especially given there's a lot of repeated code in this action too

app/models/payroll/claim_csv_row.rb
robbiepaul: I'm fine either way. I do think it's a useful thing to have on the claim model.

app/controllers/admin/claims_controller.rb
robbiepaul: I've implemented the solution you suggested. It's worth mentioning the amount of code those two action share will reduce once were using approved only claims in the payroll CSV but I guess we can cross that bridge when we get to it

app/models/payroll/claim_csv_row.rb
tekin: Ah right. I think it's fine to have them as high-level constants. I think it's stranger and potentially more confusing to have a constant with a list of values that doesn't get used anywhere. A comment-as-heading at the top could be used to indicating what these constants represent if you feel it would be useful to give an indication of that.

app/controllers/admin/claims_controller.rb
tekin: I'm less convinced by this change tbh. The repetition was in only two places and was boring, standard rails code, which we've replaced with a helper method containing conditional logic and "magic" string constants. Before, the routing helpers gave us concrete methods that would fail hard if there was a typo in the url helper calls to generate these links. Now they'll happily generate a link to an action that will respond with nothing if the magic string doesn't match the ones in the conditional logic. And as @robbiepaul points out, the two CSV files won't include the same claims: payroll is only meant to include "approved" claims, whilst the other is for "submitted" claims.

Generally, any attempt to DRY up code and introduce abstractions should be done with care as there is a cost in introducing new abstractions and entangling two bits of loguc. I stick to the rule of thumb that unless it is blindingly and obviously the case that the code is 100% identical, then I want to see three instances of duplication before I think about generalising the code to DRY it up.

robbiepaul:
@tekin I've removed the commit that DRY'd up the controller and added a commit to make the class constants top level

app/models/payroll/claim_csv_row.rb
```diff
@@ -0,0 +1,136 @@
...
+    def start_of_month
+      Date.today.at_beginning_of_month
+    end
+
+    def date_of_birth
+      model.date_of_birth.strftime(DATE_FORMAT)
+    end
+
+    def address_lines
+      [
```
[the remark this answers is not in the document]
robbiepaul: Yeah good shout, I'll do this when I rebase against #371

PR #382, opened 2019-09-04
Use self-hosted Google Analytics code

mec:
To make this work, stop using the gtag.js, which is for Google Analytics
and Google Tag Manager and instead use just the analytics.js, see docs
for the differences:

https://developers.google.com/analytics/devguides/collection/gtagjs/
https://developers.google.com/analytics/devguides/collection/analyticsjs/

gtag.js loads https://www.google-analytics.com/analytics.js into the
page so we are back to using a third party script when using gtag.js
locally.

We are not using the extra features from Tag Manager and only require
Google Analytics.

The source of the local analytics.js is:

https://www.google-analytics.com/analytics.js

and was obtained on Wed Sep 4 2019.

Prettier has modified the code style, but it is still uglified.


mec:
[Self-host Google Analytics code](https://trello.com/c/i4gb5nQy/653-self-host-google-analytics-code)

robbiepaul:
I wonder if `analytics.js` should live in `vendor/assets/javascripts/` since it's third-party JS and we shouldn't be modifying it

app/views/layouts/application.html.erb
```diff
@@ -19,8 +19,8 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
 
     <% if ENV["GOOGLE_ANALYTICS_ID"] %>
-      <%= javascript_include_tag "https://www.googletagmanager.com/gtag/js?id=#{ENV["GOOGLE_ANALYTICS_ID"]}" %>
       <%= javascript_include_tag "google_analytics", data: { "ga-id" => ENV["GOOGLE_ANALYTICS_ID"] } %>
+      <%= javascript_include_tag "analytics" %>
```
erbridge: This should come first. The `DOMContentLoaded` event of `google_analytics` expects `ga` to exist and be a function, which won't be true until after `analytics` has loaded, and we don't really have any control over how GA sets itself up.

erbridge: Can we flip the names of these two or make it clearer that `analytics` is GA, not some other analytics library. What was the name of the source file?

erbridge: Ok, saw you included it in the description. In that case, can we namespace it in a folder. I agree with Robbie's comment, that this feels like it belongs in `vendor/`, but perhaps it would be good to put it in `vendor/assets/javascripts/google_analytics/analytics.js` or something like that.

erbridge: Prettier also ignores things in `vendor/`.

vendor/assets/javascripts/google_analytics/analytics.js
```diff
@@ -0,0 +1,2412 @@
+(function() {
```
erbridge: Could we have minified version back again? It makes it easier to confirm that it's changed. Prettier should be ignoring `vendor/`

PR #387, opened 2019-09-04
Support blocking access to the admin routes based on IP

erbridge:
This adds an `ADMIN_ALLOWED_IPS` env var that is a comma seperated list of IP addresses or CIDRs to allow access to the admin routes. The restiction is enforced by a new `RestrictAdminByIpMiddleware` middleware after Rollbar.


erbridge:
[CYBER 6: /admin interface is restricted to DfE IP address range](https://trello.com/c/vgqpGj0m/558-cyber-6-admin-interface-is-restricted-to-dfe-ip-address-range)

config/initializers/admin.rb
```diff
@@ -0,0 +1,7 @@
+require_relative "../../lib/restrict_admin_by_ip_middleware"
+
+Rails.application.configure do
+  allowed_ips = ENV.fetch("ADMIN_ALLOWED_IPS", "").split(",")
```
tekin: Just to be clear, this defaults `allowed_ips` to an empty string if `ADMIN_ALLOWED_IPS` isn't set, which from my reading would mean that no one would be able to access admin (safe by default; good). Is this what we want or would it be better for the system to not boot and fail loudly if it the environment variable doesn't exist? Are there any environments where we expect it wouldn't be set?

config/initializers/admin.rb
```diff
@@ -0,0 +1,7 @@
...
+
+  config.middleware.use RestrictAdminByIpMiddleware, allowed_ips
```
tekin: Should this middleware be inserted lower in the stack to minimise the amount of app/framework code the request will execute?

lib/restrict_admin_by_ip_middleware.rb
```diff
@@ -0,0 +1,32 @@
...
+  def initialize(app, allowed_ips)
+    @app = app
+    @allowed_ips = allowed_ips.map { |ip| IPAddr.new(ip) }
+  end
+
+  def call(env)
+    status, headers, body = @app.call(env)
+
+    if restricted?(env)
+      [403, headers, "Forbidden"]
```
tekin: Rack expects the body to be enumerable so `"Forbidden"` will need to be wrapped in an array:
```suggestion
      [403, headers, ["Forbidden"]]
```

It's really strange that the spec passes though...

erbridge:
Good catch. Yes.

lib/restrict_admin_by_ip_middleware.rb
```diff
@@ -0,0 +1,32 @@
+class RestrictAdminByIpMiddleware
+  def initialize(app, allowed_ips)
+    @app = app
+    @allowed_ips = allowed_ips.map { |ip| IPAddr.new(ip) }
+  end
+
+  def call(env)
+    status, headers, body = @app.call(env)
```
tekin: This should only happen if the request isn't restricted, otherwise the action gets on every request, even restricted ones.

lib/restrict_admin_by_ip_middleware.rb
```diff
@@ -0,0 +1,32 @@
...
+  private
+
+  def restricted?(env)
+    req = Rack::Request.new(env)
+
+    restricted_route?(req) && !allowed_ip?(req)
+  end
+
+  def restricted_route?(req)
+    req.path =~ %r{^/admin}
```
tekin: Ruby gives us [`String#start_with?`](https://apidock.com/ruby/String/start_with%3F), which might be a more expressive way to perform this check.

config/initializers/admin.rb
erbridge: That's a good point. I'll make it fail, then.

config/initializers/admin.rb
erbridge: I'm not totally clear on the stack. Is it ok to put it at position 0?

lib/restrict_admin_by_ip_middleware.rb
erbridge: Good shout.

lib/restrict_admin_by_ip_middleware.rb
erbridge: Any suggestions on what to do about headers?

config/initializers/admin.rb
tekin: `rails middleware` will output the current stack, and the internal Rails stack is documented here: https://guides.rubyonrails.org/rails_on_rack.html#internal-middleware-stack

Perhaps after `Rails::Rack::Logger` so the request will be included in the logs?

lib/restrict_admin_by_ip_middleware.rb
tekin: We can set them explicitly to whatever we feel is appropriate. Anything more than `{"Content-Type" => "text/plain"}` necessary?

azure/resource_groups/app/parameters/production.template.json
```diff
@@ -124,6 +124,9 @@
         "secretName": "DfeSignInApiEndpoint"
       }
     },
+    "ADMIN_ALLOWED_IPS": {
+      "value": "194.72.231.130/31,194.72.231.132/31,213.123.63.150,217.38.175.48/28,217.38.175.64/28,217.38.165.192/28,217.38.165.176/28"
```
pezholio: Would this be better in the keyvault? Would mean we can temporarily add our own IPs in case we need to do some debugging

config/initializers/admin.rb
```diff
@@ -0,0 +1,9 @@
+require "rollbar/middleware/rack"
+
+require_relative "../../lib/restrict_admin_by_ip_middleware"
+
+Rails.application.configure do
+  allowed_ips = ENV.fetch("ADMIN_ALLOWED_IPS").split(",")
+
+  config.middleware.insert_after Rollbar::Middleware::Rails::RollbarMiddleware, RestrictAdminByIpMiddleware, allowed_ips
```
pezholio: Can we skip this middleware completely in dev?

lib/restrict_admin_by_ip_middleware.rb
```diff
@@ -0,0 +1,30 @@
...
+  private
+
+  def restricted?(env)
+    req = Rack::Request.new(env)
+
+    restricted_route?(req) && !allowed_ip?(req)
+  end
+
+  def restricted_route?(req)
+    req.path == "/admin" || req.path.start_with?("/admin/")
```
pezholio: Is it not enough to do `.start_with?("/admin")` here?

azure/resource_groups/app/parameters/production.template.json
erbridge: You'd still need to kick off another deployment. I'd prefer to keep it here, so it's clear what we deployed at a glance. I honestly think we have too much in the Key Vault. It makes us fragile to redeploys, and harder to reason about what settings we're running on for a given version.

lib/restrict_admin_by_ip_middleware.rb
erbridge: What if we have a route `/admin-report` or something (a concieted example, I know) which we don't want to block? Since this is hidden in the middleware, and all environments other than production let anyone in, we might well not notice until it went live.

config/initializers/admin.rb
erbridge: I'd prefer not to. Let's keep dev as close to production as possible. This also makes it easy to test the behaviour locally by removing the env var.

PR #391, opened 2019-09-05
Pluralize student_loan_start_date validation error

mec:
The student_load_start_date is used in two contexts, when a user has
completed one course or has completed two or more courses. This should
be taken into account in the validation error message.

Used the locales file to provide two different validation errors.
Introduced question and validation_error keys to the
student_loan_start_date locale values which meant adjusting a few other
places those string are used.

The current value of student_loan_courses is used switch between them.

Updated an incorrect it block in claim_spec.rb.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -67,6 +67,10 @@ en:
       male: "Male"
       female: "Female"
       dont_know: "Don’t know"
+  validation_errors:
+    student_loan_start_date:
+      one_course: "Select when the first year of your higher education course started"
+      two_or_more_courses: "Select when the first year of each of your higher education courses started"
   verified_fields:
     first_name: "first name"
     middle_name: "middle name"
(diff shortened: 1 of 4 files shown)


mec:
[3.3.3 – Error Suggestion	Suggest fixes when users make errors](https://trello.com/c/m1UdihqS/469-333-error-suggestion-suggest-fixes-when-users-make-errors)

config/locales/en.yml
```diff
@@ -48,8 +48,12 @@ en:
     student_loan_country: "When you applied for your student loan where was your home address?"
     student_loan_how_many_courses: "How many higher education courses have you studied?"
     student_loan_start_date:
-      one_course: "When did the first year of your higher education course start?"
-      two_or_more_courses: "When did the first year of each higher education course start?"
+      one_course:
+        question: "When did the first year of your higher education course start?"
+        validation_error: "Select when the first year of your higher education course started"
```
erbridge: Seems a bit weird to nest this inside `questions`. Do we need another hash at the same level?

mec:
dafadfadsf

PR #392, opened 2019-09-05
Upgrade to govuk-frontend 3.0.0 via dxw_govuk_frontend_rails 3.1.0

erbridge:
This involves moving the govuk-frontend assets into the public root, as the govuk-frontend CSS uses paths to reference assets.

I followed the upgrade instructions, but please check.


titlescreen:
It looks like the highlighting on inputs is still failing I think this update has made it slightly worse, the DAC audit said we needed a ratio of 3.1:1 on UI components. I think we should put something on there to target radio, checkboxes and inputs with a darker border.

I'm sure you are all good to make this change but if you would like any help / me to pick it up then please let me know.

erbridge:
@titlescreen Do you mean the focus state, or something else? There were some comments about changes to focusing in the release notes, but I didn't think they applied.

titlescreen:
We also had some feedback from a dyslexic user who said they couldn't see the hint text, do you think it would be worth changing that to use #505758 or would that be going off brand too much?

erbridge:
I think we should keep this PR to changes related to the upgrade. If there are extra changes we need to make, let's do them separately.

PR #393, opened 2019-09-05
Format student loan amount to two decimal places

robbiepaul:
Introduce a new `currency_value_for_number_field` helper that will take a number, convert it in to a currency value (with two decimal places) but not include a pound sign or commas, suitable for a number_field.

### DIFF
--- a/app/views/claims/student_loan_amount.html.erb
+++ b/app/views/claims/student_loan_amount.html.erb
@@ -16,7 +16,15 @@
 
           <div class="govuk-currency-input">
             <span class="govuk-currency-input__unit">&pound;</span>
-            <%= fields.number_field :student_loan_repayment_amount, class: css_classes_for_input(current_claim, :student_loan_repayment_amount, "govuk-currency-input__input govuk-input--width-5"), step: 'any', min: 0, max: 99999 %>
+            <%= fields.number_field(
+                  :student_loan_repayment_amount,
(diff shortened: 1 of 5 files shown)


robbiepaul:
[Bug edit loan repayment amount decimal places](https://trello.com/c/v9Ux0rtT/318-bug-edit-loan-repayment-amount-decimal-places)

app/views/claims/student_loan_amount.html.erb
```diff
@@ -16,7 +16,15 @@
...
             <span class="govuk-currency-input__unit">&pound;</span>
-            <%= fields.number_field :student_loan_repayment_amount, class: css_classes_for_input(current_claim, :student_loan_repayment_amount, "govuk-currency-input__input govuk-input--width-5"), step: 'any', min: 0, max: 99999 %>
+            <%= fields.number_field(
+                  :student_loan_repayment_amount,
+                  class: css_classes_for_input(current_claim, :student_loan_repayment_amount, "govuk-currency-input__input govuk-input--width-5"),
+                  step: 'any',
+                  min: 0,
+                  max: 99999,
+                  value: currency_value_for_number_field(fields.object.student_loan_repayment_amount),
+                )
```
erbridge: More line wrapping in our views, please! Makes it so much easier to read.

robbiepaul:
@erbridge I've added a spec for testing that part of the user journey, although I'm not sure how much value it adds

erbridge:
I guess the value I see in it is that it confirms that we actually use it.

PR #398, opened 2019-09-09
Use helpful and clear page titles

mec:
Changed the `page_title` helper method to return the page title and the service name ('Teachers: claim back your student loan repayments') followed by 'GOV.UK'. If no page title is set, a sensible default is provided

Added page titles to every step in the user journey using either the question from the locales file or the title of the page in line with the decision we made on page titles.

Page titles are set in the view to couple the page titles to the presentation layer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- Fix add clear and helpful page titles
 - Fix a bug displaying incorrect number of decimal places on monetary number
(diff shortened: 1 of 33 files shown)


mec:
[2.4.2 – Page Titled	Use helpful and clear page titles](https://trello.com/c/wcFZbRTJ/463-242-page-titled-use-helpful-and-clear-page-titles)

tekin:
Does the accessibility statement need updating to reflect this change?

mec:
Update the accessibility statement and date.

PR #399, opened 2019-09-09
Configure OmniAuth path_prefix so that failure path works

robbiepaul:
Passing `path_prefix` as an option doesn't set it correctly, causing the failure path to have the "/admin" part missing. This means users who fail to authenticate themselves on DfE Sign In return to a 404 page.

By configuring OmniAuth's `path_prefix` we can ensure the correct failure path will be generated. We were (incorrectly) overriding this value in our spec helper, setting it globally means we don't have to do that any more.

### DIFF
--- a/config/initializers/omniauth.rb
+++ b/config/initializers/omniauth.rb
@@ -1,4 +1,7 @@
-OmniAuth.config.logger = Rails.logger
+OmniAuth.configure do |config|
+  config.logger = Rails.logger
+  config.path_prefix = "/admin/auth"
+end
 
 dfe_sign_in_issuer = ENV["DFE_SIGN_IN_ISSUER"]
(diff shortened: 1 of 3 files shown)


robbiepaul:
[OmniAuth configured to send failed auth to wrong URL](https://trello.com/c/6erFGBwq/495-omniauth-configured-to-send-failed-auth-to-wrong-url)

config/initializers/omniauth.rb
```diff
@@ -1,4 +1,8 @@
-OmniAuth.config.logger = Rails.logger
+OmniAuth.configure do |config|
+  config.logger = Rails.logger
+  config.path_prefix = "/admin/auth"
+end
+
```
pezholio: I don't think Standard likes this extra line

robbiepaul: Oops! Fixed, thanks

PR #400, opened 2019-09-09
Stop old Azure URLs serving the site

pezholio:
This creates a new environment variable, called `CANONICAL_URL`, which is only set if we're using a hostname for that environment. If that variable is set in the app, then we have a rule in the routes that redirects us to the correct hostname, keeping the path and query string intact.

This should also help with #395

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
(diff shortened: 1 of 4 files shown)


pezholio:
[Stop old Azure URLs serving the site](https://trello.com/c/ZRoAp4WJ/529-stop-old-azure-urls-serving-the-site)

PR #401, opened 2019-09-09
Parsing no major html markup code errors

mec:
For 'Which academic year did you complete your initial teacher training?'

This question originally used a select field that was later replaced with a
set of radio buttons. Radio buttons should be marked-up with `<fieldset>`
and `<legend>` rather than a `<label>` so the html parser error is resolved.

For the address question:

- 'region' is not valid `autocomplete` value, use 'address-level-1' 
see below
- Town or city is the second most specific after County so is
'address-level-2'

See guidance:

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#Values

### DIFF
--- a/app/views/claims/address.html.erb
+++ b/app/views/claims/address.html.erb
@@ -44,8 +44,7 @@
             County
           <% end %>
           <%= errors_tag current_claim, :address_line_4 %>
-          <%= form.text_field :address_line_4, autocomplete: "region",
-            class: "govuk-input govuk-!-width-two-thirds" %>
+          <%= form.text_field :address_line_4, class: "govuk-input govuk-!-width-two-thirds" %>
         <% end %>
 
         <%= form_group_tag current_claim, :postcode do %>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- Fix html validation issues with qts year and address views
+
 ## [Release 005] - 2019-09-10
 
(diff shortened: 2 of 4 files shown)


mec:
[4.1.1 – Parsing	No major code errors](https://trello.com/c/kI5GIjVO/464-411-parsing-no-major-code-errors)

app/views/claims/address.html.erb
```diff
@@ -44,7 +44,7 @@
             County
           <% end %>
           <%= errors_tag current_claim, :address_line_4 %>
-          <%= form.text_field :address_line_4, autocomplete: "region",
+          <%= form.text_field :address_line_4, autocomplete: "address-level1",
```
pezholio: I'm gonna be picky here. I don't think we need this level of address. A postal address only needs a street address, post town and postcode to be valid (a locality (e.g. Village, small town etc) can be added too). A county isn't part of the address specification that the Royal Mail use, so we can probably ding it. (I can bore on about addresses for hours)

pezholio: It also adds added complexity, where the autocomplete value here might be the town for some users

mec: Interesting! You know your addresses! :) Are you saying we loose the field entirely or just the autocomplete attribute?

For me, this story was only about resolving the html parser issues @titlescreen raised, I would be wary of changing anything else.

pezholio: Yeah, I'd get rid of the autocomplete for that one, and probably use `street-address`, `address-line1` etc, rather than the administrative level ones, because they're a bit unclear.

PR #402, opened 2019-09-09
Turn off autocomplete on sensitive inputs

robbiepaul:
It is something that is unique to each person and is often used in identity theft. It looks like people can unknowingly submit 'autocompleted' data on malicious websites/websites that accidentally
use malicious npm packages: https://medium.com/@bramus/stealing-usernames-passwords-and-other-personal-data-via-browsers-and-npm-packages-30938725ebc8

However by turning it off, we'll be breaking this WCAG 2.1 criteria: https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose.html

Some more detail: https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

Also: https://www.owasp.org/index.php/Top_10_2013-A6-Sensitive_Data_Exposure

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- Turned off autocomplete setting on National Insurance number and bank inputs
 - Prefix titles with an error notification if an error is present
 - Fix html validation issues with qts year and address views
 
--- a/app/views/claims/bank_details.html.erb
+++ b/app/views/claims/bank_details.html.erb
@@ -23,6 +23,7 @@
             <%= errors_tag current_claim, :bank_sort_code %>
             <%= form.text_field :bank_sort_code,
               class: css_classes_for_input(current_claim, :bank_sort_code, "govuk-!-width-one-quarter"),
+              autocomplete: "off",
               "aria-describedby" => "sort-code-hint" %>
           <% end %>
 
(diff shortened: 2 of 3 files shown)


robbiepaul:
[Should NI number be autocompletable](https://trello.com/c/c0iamQyJ/544-should-ni-number-be-autocompletable)

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog]
   form fields
 - Restrict access to `/admin` by IP
 - Fix student_loan_start_date validation error message
+- Turned off autocomplete setting on National Insurance number and bank inputs
```
pezholio: Can we fill the changelog from the top to save git conflicts?

robbiepaul:
Yeah I agree, worth a discussion. Was just looking for some low-hanging fruit. I'm in two minds about the change, it's a conflict of user needs. I'll have a look to see what other services are doing in these cases.

titlescreen:
Just to throw in my 2 cents. I think turning off autocomplete for the bank details and national insurance number is the way to go. I get that it will make completing those fields a bit harder, but I think the security requirements outweigh this.

tekin:
Yeah, reckon NI and bank details are safe not to autocomplete. They're not part of the typical "identity" fields that get autocompleted and are sensitive enough that not autocompleting them is probably on balance the right way to go.

PR #403, opened 2019-09-09
Prefix titles with an error when error present

pezholio:
This means a user using a screen reader will be immediately notified that an error is present

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,7 +1,12 @@
 module ApplicationHelper
-  def page_title(title)
+  def page_title(title, show_error: false)
     content_for :page_title do
-      "#{title} – #{t("student_loans.journey_name")} – GOV.UK"
+      [].tap do |a|
+        a << "Error" if show_error
+        a << title
+        a << t("student_loans.journey_name")
+        a << "GOV.UK"
+      end.join(" - ")
     end
   end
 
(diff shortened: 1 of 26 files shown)


pezholio:
[Page titles do not get amended when there is an error on the page](https://trello.com/c/WitlZgSL/668-page-titles-do-not-get-amended-when-there-is-an-error-on-the-page)

app/helpers/application_helper.rb
```diff
@@ -1,7 +1,12 @@
 module ApplicationHelper
-  def page_title(title)
+  def page_title(title, error_present = false)
```
robbiepaul: I always find boolean arguments a little confusing, could we use a keyword/named argument to make it a bit more explicit what the boolean is for?

app/helpers/application_helper.rb
```diff
@@ -1,7 +1,12 @@
...
     content_for :page_title do
-      "#{title} – #{t("student_loans.journey_name")} – GOV.UK"
+      [].tap do |a|
+        a << "Error" if error_present
+        a << title
+        a << t("student_loans.journey_name")
+        a << "GOV.UK"
+      end.join(" - ")
```
robbiepaul: 👍 nice

app/helpers/application_helper.rb
pezholio: Ah, nice. Tweaked that to use a named argument and it makes a lot more sense now 😃

PR #404, opened 2019-09-09
School search form field without descriptive label

mec:
The school search autocomplete had no `<label>` which casued accessibilty issues.

This PR also adds the new GOVUK Frontend v3 focus style to the autocomplete input.

This issue was raised specifically around the JAWS screen reader so it would be great to test it with that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- Fix school search autocomplete missing label tag
+
 ## [Release 005] - 2019-09-10
 
 - Redirect requests from any domain other than the canonical one
--- a/app/assets/stylesheets/components/accessible_autocomplete_tweaks.scss
+++ b/app/assets/stylesheets/components/accessible_autocomplete_tweaks.scss
@@ -1,5 +1,12 @@
 @import "govuk-frontend-rails";
 
+.autocomplete__input:focus {
+  // standard focus state from govuk frontend input component
+  outline: $govuk-focus-width solid $govuk-focus-colour;
+  outline-offset: 0;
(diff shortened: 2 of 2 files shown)


titlescreen:
So I've tested on JAWS and when you click into the school input then it reads the question, but if you use the JAWS shortcut it navigates to a div with a role='combobox' applied rather than the input itself and doesn't read the question.

I think accessible autocomplete is creating this element when it runs and I'm not sure how to suggest fixing this without modifying / waiting for an updated version of accessible auto-complete.

Screenshot 2019-09-10 at 12 48 47

app/views/claims/_school_search.html.erb
```diff
@@ -12,8 +12,11 @@
 
     <%= form_group_tag current_claim, :school_search do %>
 
+  </h1>
       <h1 class="govuk-label-wrapper">
+        <label class="govuk-label govuk-label--xl" for="school_search">
```
titlescreen: I think this needs removing, when I look at the page there are two labels pointing at the same id.

mec: > I think this needs removing, when I look at the page there are two labels pointing at the same id.

Where is the second `<label>` element? 

The missing label is what I thought I was fixing and there was another story with the selection of the autocomplete issue we are waiting on the upstream fix?

https://trello.com/c/MEWC6cq0

Sorry if I am being thick!

mec: Seen the second label! 👍  other question still stands! :)

mec: Hmmm, going back and looking at the story I am wondering if we close this and cover all the autocomplete issues together? All this does now is add the focus state styles?

mec: Yep, let's close this and pick the autocomplete issue up later.

PR #408, opened 2019-09-10
Change focus element on inputs for contrast

titlescreen:
Updating the GDS kit replaced the orange border on the inputs with a
yellow one. This made the contrast worse since the update.

I've had to pick a darker orange colour for the border on focus so we
meet AA standards.

GDS thickens the border when an input has focus, but I'm not sure if that will be sufficient to pass.

### Before
Screenshot 2019-09-10 at 11 20 53

### After
Screenshot 2019-09-10 at 11 20 13

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- Fix accessibility issue where there is insufficient contrast on UI elements
+
 ## [Release 005] - 2019-09-10
 
 - Redirect requests from any domain other than the canonical one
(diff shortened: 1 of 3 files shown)


mec:
I'm amazed at this since the GOVUK team claim it meets AA:

>The focus state of components now meets the new WCAG 2.1 level AA requirements.

https://github.com/alphagov/govuk-frontend/releases/tag/v3.0.0
Does the heavier black border not make a difference?

titlescreen:
Yeah we get a thicker border when the input has focus but as I said before I'm unsure if its sufficient to pass. I will contact DAC to query.

titlescreen:
DAC have said that updated width on Radios should be fine. I will close this.

PR #410, opened 2019-09-10
Update school eligibility to exclude closed schools

robbiepaul:
- Student loans policy: schools that closed before the current claim year are ineligible
- Maths and Physics policy: schools that are closed are ineligible


app/models/student_loans/school_eligibility.rb
```diff
@@ -1,5 +1,6 @@
 module StudentLoans
   class SchoolEligibility
+    START_DATE = Date.new(2018, 4, 6)
```
mec: Could this const have a more explicit name? POLICY_START_DATE, just to be super clear what it represents?

spec/models/student_loans/school_eligibility_spec.rb
```diff
@@ -99,6 +171,16 @@
...
           it { is_expected.to be true }
+
+          context "and it closed after the start of the policy" do
+            let(:school_attributes) { special_free_school_attributes.merge({phase: :not_applicable, close_date: after_start_of_policy}) }
+            it { is_expected.to be true }
+          end
+
+          context "and it closed before the start of the policy" do
+            let(:school_attributes) { special_free_school_attributes.merge({phase: :not_applicable, close_date: before_start_of_policy}) }
+            it { is_expected.to be false }
```
mec: Do all these repeating contexts gives us extra confidence the check method is working correctly, I found the repatition made me lazy reading the tests?

spec/models/maths_and_physics/school_eligibility_spec.rb
```diff
@@ -99,6 +134,11 @@
         context "and it doesn't have an education phase" do
           let(:school_attributes) { special_free_school_attributes.merge({phase: :not_applicable}) }
           it { is_expected.to be true }
+
+          context "and it is closed" do
+            let(:school_attributes) { special_free_school_attributes.merge({phase: :not_applicable, close_date: Date.yesterday}) }
+            it { is_expected.to be false }
+          end
```
mec: Do we need to test the closed context each time? I was wondering if operator precedence in the check method means this could be approached differently?

spec/models/student_loans/school_eligibility_spec.rb
robbiepaul: Yeah I think we wanted to build up a matrix of all the possibilities for eligibility criteria. I think it helps make clear that the school in each example would otherwise be eligible if it weren't for it closing before the start of the policy.

spec/models/maths_and_physics/school_eligibility_spec.rb
robbiepaul: Yeah that's a good point, I'll have a look at re-ordering them

robbiepaul: I've reordered the conditional to filter out closed schools earlier, I've also reduced the number of contexts down

mec:
@robbiepaul just read my own comment back - hope you know I meant I've got to learn not you! :/

robbiepaul:
@mec that's how I read it, I knew what you meant :)

Yeah left to right is correct with short-circuits. Specifically the `Boolean Operators: &&, ||, !, and, or, not` section of that page

PR #411, opened 2019-09-10
Fix RestrictAdminByIpMiddleware allowing IPv6 addresses

robbiepaul:
I noticed this when testing `/admin` locally, my IP address is `::1` which is the IPv6 equivalent for local host.

I'm not sure if IPv6 addresses can come through with ports attached, but this should fix it for the majority of cases anyway.

### DIFF
--- a/spec/lib/restrict_admin_by_ip_middleware_spec.rb
+++ b/spec/lib/restrict_admin_by_ip_middleware_spec.rb
@@ -50,6 +50,18 @@ RSpec.describe RestrictAdminByIpMiddleware do
     expect(body).to eq ["Forbidden"]
   end
 
+  it "handles IPv6 addresses" do
+    unblessed_ipv6 = "2001:0db8:85a3:0000:0000:8a2e:0370:7334"
+
+    code, _env, body = middleware.call(mock_request("/test", unblessed_ipv6))
+    expect(code).to eq 200
+    expect(body).to eq ["OK"]
+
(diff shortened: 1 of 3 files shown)


spec/lib/restrict_admin_by_ip_middleware_spec.rb
```diff
@@ -50,6 +50,18 @@
     expect(body).to eq ["Forbidden"]
   end
 
+  it "handles IPv6 addresses" do
```
pezholio: Does this work if we have a IPv6 address in the list of allowed IPs?

robbiepaul: Yeah I think so, IPv6 is included in the default developer `ALLOWED_IPS`: https://github.com/DFE-Digital/dfe-teachers-payment-service/blob/f7fa8b18f8a3ba911dae89d54de0035ffddea11d/.env.development#L10

robbiepaul: This fix fixes a fix for the an issue where ports were being passed to `client_ip` IPv4 IP addresses :)

pezholio: Ah, course

robbiepaul:
[BUG: IPv6 addresses can visit `/admin`](https://trello.com/c/RABiCFlj/677-bug-ipv6-addresses-can-visit-admin)

PR #414, opened 2019-09-11
Switch to dynamic error pages

erbridge:
Possibly contraversial, but I think they're much easier to maintain and keep consistent with the rest of the application, for very little cost.


tekin:
I think on balance I prefer sticking with serving static pages for errors, mostly because dynamic ones will only work if the application itself is working, whereas static ones can get served by a web server directly, or perhaps even a CDN, even when the application instances have fallen over. We don't expect these static pages to change much at all so the cost of keeping them maintained and consistent is relatively low.

PR #415, opened 2019-09-11
Make azure-deploy exit if parameters are missing

erbridge:
Normally Azure CLI waits for the user to enter any missing parameters.
However, that doesn't work in unattended environments or when using
subshells. Instead, we close stdin and rely on an empty response to tell
us something went wrong.


erbridge:
[Surface missing parameters in `azure-deploy` script](https://trello.com/c/IoI3V3qr/506-surface-missing-parameters-in-azure-deploy-script)

PR #416, opened 2019-09-11
Ensure all radio buttons use fieldsets and legends

erbridge:
Gender was using the wrong treatment. The rest seem to be resolved.


erbridge:
[Question not read out when a radio buttons is selected](https://trello.com/c/HVLD6P1x/661-question-not-read-out-when-a-radio-buttons-is-selected)

PR #424, opened 2019-09-12
Set up deployment slot and availability test for VSP

erbridge:
Before merging:

- [x] Update Azure DevOps to use deployment slots for VSP

After merging:

- [ ] Update [the docs] with the new availability test for the VSP
- [ ] Update [the docs] with the new availability test locations

[the docs]: https://dfedigital.atlassian.net/wiki/spaces/TP/pages/1142423576/Alerting+Monitoring

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- Monitor VSP availability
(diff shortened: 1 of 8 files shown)


erbridge:
Yeah, changing it would break a lot of their services.

erbridge:
@pezholio Could you take a look at the new commits, please? I needed to modify `bin/swap-slots`.

PR #426, opened 2019-09-16
Add empty secondary VSP encryption key parameter

mec:
During key rotation on the VSP the secondary key is the new,
incoming key and once the old cert is removed at the GOV.UK Verify hub
the new key becomes the primary.

Have a empty string as the secondary key in the template files to make
this explicit.

### DIFF
--- a/azure/resource_groups/app/parameters/development.template.json
+++ b/azure/resource_groups/app/parameters/development.template.json
@@ -208,6 +208,9 @@
         "secretName": "TeacherPaymentsDevVspSamlEncryption2KeyBase64"
       }
     },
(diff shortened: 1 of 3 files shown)


mec:
[GOV.UK Verify 5. Maintenance process (keeping the certificates updated/rotated on both staging and production](https://trello.com/c/s4l1g8K6/566-govuk-verify-5-maintenance-process-keeping-the-certificates-updated-rotated-on-both-staging-and-production)

erbridge:
This also needs to explicitly set it up in the template and passing through to the VSP environment variables, too.

mec:
@erbridge I'll catch up with you about the template.

PR #430, opened 2019-09-16
Link schools to Get Information About Schools

pezholio:
If there is any doubt about where a teacher says they have worked and the data Hartlink has about them, this will help approvers quickly check GIAS to see if (for example) the school they work at is part of a Multi-Academy Trust and payroll is handled by a different school within the trust.

# Before

image

# After

image


app/helpers/admin/claims_helper.rb
```diff
@@ -23,5 +23,10 @@ def admin_personal_details(claim)
         [t("questions.admin.email_address"), claim.email_address],
       ]
     end
+
+    def school_with_link(school)
```
robbiepaul: `school_with_link` feels like it's going to return the school, with a link. Perhaps `link_to_school` would be more intention revealing?

PR #436, opened 2019-09-17
Update privacy notice to include IP address

mec:
We are now sending the client IP address to Application Insights so we
state that in the notice.

I've tried to copy the GOV.UK use of IP address in their privacy notice, 
I am no expert in this so we may want other eyes on the content?

Screenshot_2019-09-17 Privacy Notice Claim additional payments for teaching - Teachers claim back your student loan repayme

### DIFF
--- a/app/views/static_pages/privacy_notice.html.erb
+++ b/app/views/static_pages/privacy_notice.html.erb
@@ -48,6 +48,7 @@
         your email address if you include it with feedback to us on the service or agree to be contacted for user research
         purposes
       </li>
+      <li>your Internet Protocol (IP) address</li>
     </ul>
 
     <h2 class="govuk-heading-l">
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- Update privacy notice to include the collection of IP address
+
 ## [Release 009] - 2019-09-17
 
 - Add student loan plan and amount to approver view of a claim


app/views/static_pages/privacy_notice.html.erb
```diff
@@ -48,6 +48,7 @@
         your email address if you include it with feedback to us on the service or agree to be contacted for user research
         purposes
       </li>
+      <li>your Internet Protocol (IP) address</li>
```
robbiepaul: Do we need to say how long we store it for and why we need to store it?

mec: I am not sure, I felt like that was covered in the rest of the text?

robbiepaul: You're right, it does, I just re-read it. I wonder if we should be explicit that we collect IP addresses for security and not anything like analytics etc?

titlescreen: I grabbed this from teaching vacancies but they mention it in the 'Who we will make your personal data available to' section.
`for security purposes: collecting part of the IP addresses of people visiting the Teaching Vacancies website (for example, if we were receiving continuous direct denial of service attacks from an IP address range then we could block it)`

Perhaps we should put something similar in the 'how we will use you information' section

erbridge: Note that we're not doing the same thing as they were. We collect full IP addresses, not partial, and we keep them forever.

mec:
[Update privacy notice](https://trello.com/c/huBcrHEV/703-update-privacy-notice)

app/views/static_pages/privacy_notice.html.erb
mec: We plan to get the right team inside DfE to look at this, I've linked to the Trello card.

mec:
Thanks for all the comments. We are going to talk to the right people at DfE and get an answer. I'll close this now as it's such a small change, we can open or redo once we have an answer rahter than have this PR going stale.

PR #437, opened 2019-09-17
A service operator can easily check employment details in Hartlink

pezholio:
This adds a new column to schools to store the DfE number in the database (which is fetched when the nightly import runs), and show the DfE number in brackets against both the claim and the current school, to make it easier when checking a school against a record in Hartlink. 

image

### DIFF
--- a/app/models/school_data_importer.rb
+++ b/app/models/school_data_importer.rb
@@ -44,6 +44,7 @@ class SchoolDataImporter
     school.local_authority = local_authority
     school.local_authority_district = local_authority_district
     school.close_date = row.fetch("CloseDate")
+    school.establishment_number = row.fetch("EstablishmentNumber")
     school
   end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- Show DfE number against school in approver view of claim
 - Drop redundant full_name column from claims table
 
 ## [Release 010] - 2019-09-17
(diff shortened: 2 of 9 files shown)


pezholio:
[A Service Operator can easily check employment details in Hartlink](https://trello.com/c/ikILqHmn/693-a-service-operator-can-easily-check-employment-details-in-hartlink)

mec:
@pezholio is it worth checking with the team that brackets is the way to go on this, visually?

pezholio:
Checked with Matt and he's happy :)

app/models/school_data_importer.rb
```diff
@@ -54,4 +55,11 @@ def schools_csv_url
   def date_string
     Time.zone.now.strftime("%Y%m%d")
   end
+
+  def generate_dfe_number(row)
+    [
+      row.fetch("LA (code)"),
+      row.fetch("EstablishmentNumber"),
```
erbridge: I think we only need the `EstablishmentNumber`, since we already have the `LA (code)` and this is derived from the two?

pezholio: I thought about that, but generating it on the fly would need us to do a join on the database to fetch the local authority, so may be slighty less performant.

erbridge: True. I guess I just find it a bit odd to duplicate the data. Ignore me if you feel that's appropriate.

tekin: I agree with @erbridge, I think storing the establishment number and deriving the combined "number" in the UI makes more sense. It gives us more flexibility in terms of what we can do with the separate parts.

PR #439, opened 2019-09-17
Release 010

robbiepaul:
- Fix deployment template conflict
- Make closed schools ineligible for current school
- Exclude closed schools from current school search

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+## [Release 010] - 2019-09-17
+
 - Fix deployment template conflict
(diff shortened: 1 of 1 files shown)


robbiepaul:
> Yanked releases are versions that had to be pulled because of a serious bug or security issue

We didn't have to pull it, it just didn't deploy. It does raise a question of what we do in these circumstances though. I think another release with the fix is fine?

erbridge:
Hrmm. Maybe `[UNRELEASED]` instead?

robbiepaul:
Yeah ok, although the changes are released. Shall I duplicate the changes from 009 in to 010?

erbridge:
Ok, nevermind. I see why we wouldn't do this. Leave as it is.

PR #440, opened 2019-09-17
Move student loan repayment amount question

robbiepaul:
**Note this isn't merging in to `master`, we're using a different branch for our epic. We're going to use the test environment to test the feature before deploying to staging**

As part of the work to allow users to add multiple claim schools we need to know how much loan they have repaid at each school.

We need to move the student loan repayment amount to happen just after the other claim school specific questions to facilitate this change.


robbiepaul:
[Move student loan repayment amount question to be after leadership Qs](https://trello.com/c/DeKUDS7J/692-move-student-loan-repayment-amount-question-to-be-after-leadership-qs)

tekin:
Is there a reason why this particular change can't go into the main branch?

In general I think it will help us later if we keep the divergence in any long-lived feature batch to a minimum. e.g. it should be possible to do a PR that makes the relationship a has_many whilst keeping the existing UI/flow and merge that into the master branch.

pezholio:
But yes, what @tekin said. I don't see the harm in merging this into master

erbridge:
I think there's a question about whether or not this change makes sense without the rest. Do we need to validate it against users?

robbiepaul:
I guess it could go in to master branch, that's why we originally split this card out in to it's own thing. After the team catch-up yesterday we felt we wanted to user test the multi-school flow so thought best to deploy the changes to `test` first. But I'm happy to change this one to master.

tekin:
It's definitely worth running past the rest of the team. Personally I feel asking them how much student loan they paid whilst at the school straight after they tell us about the school makes sense regardless of whether they are providing multiple schools or not. The question relates to their employment there so having it near the other employment questions seems logical. But others may feel differently. Run it past other folks, e.g. Mark, Rachael, Rich and see what the consensus is?

tekin:
Also, if we do decide to move this one we should be aware that releasing it could potentially break the journey for an in-progress claim that has past the point we are moving it to but not yet reached the point where we are currently asking it. We may need to set up a way to put the service into "maintenance mode" for this and future breaking-changes.

robbiepaul:
Yeah we started a conversation about this issue on slack https://ukgovernmentdfe.slack.com/archives/CGSMQRB1R/p1568651412030800
It would be good to have a process in place as I can imagine us needing to do this type of breaking change again.

robbiepaul:
After a meeting with Mark and Kristi yesterday we decided the student loan question should appear after the eligibility confirmed screen as it requires the claimant to go away and invest some time finding out their loan repayment amount. If they know they are eligible for the payment they are more likely to invest the time.

We also spoke about wanting to test this flow with users, so with that in mind plus it being a journey-breaking change I think it makes sense to merge this in to the epic branch still.

app/views/claims/eligibility_confirmed.html.erb
```diff
@@ -6,6 +6,6 @@
     <h1 class="govuk-heading-xl">You are eligible to claim back student loan repayments</h1>
     <p class="govuk-body">Based on your answers, you can claim back the student loan repayments you made between 6 April 2018 and 5 April 2019.</p>
 
-    <%= link_to "Continue", claim_path("information-provided"), class: "govuk-button", role: "button" %>
+    <%= link_to "Continue", claim_path("student-loan-amount"), class: "govuk-button", role: "button" %>
```
tekin: could we make this less brittle to changes in order by using the page sequence somehow to determine what this link is rather than hard-coding the slug?

tekin:
ok, makes sense. potentially long-running feature branches do make me nervous but let's see how it goes. one minor comment, otherwise this looks good.

robbiepaul:
Merging this as I've implemented using the `next_slug` as suggested by Tekin

PR #441, opened 2019-09-17
WIP: A service operator is able to mark a claim as rejected

mec:
This is the initial work to reject a claim. There are service implication to this we don't yet fully understand but we know enough to warrant being able to:

- mark a claim as rejected
- record the date and time the rejection occured
- capture the reason for the rejection
- send a notification to the user

For the time being we don't intend to use this feature.

More understanding around the reasons for rejection is needed before we include the reason in the notificaiton.

### DIFF
--- a/app/controllers/admin/claim_rejections_controller.rb
+++ b/app/controllers/admin/claim_rejections_controller.rb
@@ -1,12 +1,18 @@
 class Admin::ClaimRejectionsController < Admin::BaseAdminController
-  before_action :ensure_service_operator, :fetch_claim
+  before_action :ensure_service_operator, :fetch_claim, :refuse_checked_claims
+
+  def new
+    @note = @claim.notes.build
+  end
 
   def create
-    if @claim.reject!(rejected_by: admin_session.user_id)
+    @note = @claim.notes.build(created_by: admin_session.user_id, body: params[:note][:body])
+
+    if @note.valid? && @claim.reject!(rejected_by: admin_session.user_id)
       ClaimMailer.rejected(@claim).deliver_later
       redirect_to admin_claims_path, notice: "Claim has been rejected successfully"
     else
-      redirect_to admin_claim_path(@claim), notice: "Claim cannot be rejected"
+      render :new
     end
   end
 
@@ -14,5 +20,9 @@ class Admin::ClaimRejectionsController < Admin::BaseAdminController
 
   def fetch_claim
     @claim = Claim.find(params[:claim_id])
+  end
+
+  def refuse_checked_claims
+    redirect_to([:admin, @claim], notice: "Claim already checked") unless @claim.needs_checking?
   end
 end
--- a/app/controllers/admin/claim_rejections_controller.rb
+++ b/app/controllers/admin/claim_rejections_controller.rb
@@ -3,6 +3,7 @@ class Admin::ClaimRejectionsController < Admin::BaseAdminController
 
   def create
     if @claim.reject!(rejected_by: admin_session.user_id)
+      ClaimMailer.rejected(@claim).deliver_later
       redirect_to admin_claims_path, notice: "Claim has been rejected successfully"
     else
       redirect_to admin_claim_path(@claim), notice: "Claim cannot be rejected"
--- a/app/controllers/admin/claim_rejections_controller.rb
+++ b/app/controllers/admin/claim_rejections_controller.rb
@@ -1,0 +1,17 @@
+class Admin::ClaimRejectionsController < Admin::BaseAdminController
+  before_action :ensure_service_operator, :fetch_claim
+
+  def create
+    if @claim.reject!(rejected_by: admin_session.user_id)
+      redirect_to admin_claims_path, notice: "Claim has been rejected successfully"
+    else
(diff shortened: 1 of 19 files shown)


app/controllers/admin/claim_rejections_controller.rb
```diff
@@ -0,0 +1,12 @@
+class Admin::ClaimRejectionsController < Admin::BaseAdminController
+  before_action :ensure_service_operator
+
+  def create
```
pezholio: Thinking about this, I'm wondering whether the first step in the rejection process should be a `new` action, which renders a form where a user can add their reasoning, which then submits to the `create` action

mec: Yeah, it's okay to take this is steps though right?

pezholio: Oh yeah, totally. Was just idle musing at this stage 😄

mec: nice! :D

mec: Is there a point where the `reject` and `approve` controllers become a single `actions` controller? Are the advantages/disadvantages to the approaches?

pezholio: Yeah, I think that's the point where we have everything wrapped into a single validation model, but I haven't quite got it straight in my head as to how that'll work

app/models/claim.rb
```diff
@@ -137,8 +152,8 @@ def submittable?
     valid?(:submit) && !submitted?
   end
 
-  def approvable?
-    submitted? && approved_at.nil?
+  def can_be_checked?
```
erbridge: Maybe this is better phrased as `needs_checking?`?

app/views/admin/claims/show.html.erb
```diff
@@ -14,6 +14,12 @@
                   admin_claim_approvals_path(@claim),
                   method: :post,
                   class: "govuk-button",
-                  data: { confirm: "Are you sure you want to approve this claim?" } %>
+                  data: { confirm: "Are you sure you want to approve this claim?" },
+                  "data-module": "govuk-button" %>
+
+    <%= link_to "Reject", new_admin_claim_rejection_path(@claim),
+                class: "govuk-button govuk-button--warning",
+                "data-module": "govuk-button",
```
erbridge: Is this correct to do, given that it's an `<a>` not a `<button>`?

db/schema.rb
```diff
@@ -100,6 +103,7 @@
     t.datetime "updated_at", null: false
     t.uuid "local_authority_district_id"
     t.date "close_date"
+    t.string "dfe_number"
```
erbridge: This snuck in.

robbiepaul:
You've probably already thought of this but it might be good to have a rejection category or some way to categorise rejections that isn't free text. Might help later when providing stats on the service

mec:
@robbiepaul, could we use `rejected_at.present?`or similar to achieve that sort of thing?

robbiepaul:
I mean a category for the rejection reason, for instance: identity not matching, incorrect claim amount etc (sorry I can't remember all the criteria from the claims checking session). Feel free to ignore me, just going from my experience it's good to have a way to collate these types of things

app/views/admin/claims/show.html.erb
mec: My thinking is that as this behaves as a button, whatever JS is used on GOVUK buttons uses this to target buttons, therefore we need it?

I think it's this:

https://github.com/alphagov/govuk-frontend/blob/bf8bedf84f984fa4dabbda1d8f4081820d2e9e2b/src/govuk/all.js#L20

mec:
@robbiepaul Sorry, got you! My bad. It's a great idea, We feel like we need to learn a bit more about why things get rejected before commiting to that – but it's definately a thing!

app/views/admin/claims/show.html.erb
erbridge: Ah, I was under the false impression that that assumed it was a `<button>`, but I have dug further, and it looks like it supports other tags that use `role="button"`: https://github.com/alphagov/govuk-frontend/blob/bf8bedf84f984fa4dabbda1d8f4081820d2e9e2b/src/govuk/components/button/button.js#L25

app/views/admin/claim_rejections/new.html.erb
```diff
@@ -0,0 +1,17 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">
+      Reject claim <%= @claim.reference %>
+    </h1>
+
+    <%= form_for @claim, url: admin_claim_rejections_path(@claim), method: :post do |form| %>
+      <%= form.fields_for @claim.notes.first do |note_fields| %>
+        <%= note_fields.label :body, "Rejection note", class: "govuk-label govuk-radios__label" %>
```
pezholio: This just needs to be a plain `govuk-label`. Sorry, my fault

tekin:
Closing this as it will need re-working once the new claim check model is introduced.

PR #445, opened 2019-09-18
Order tests randomly

pezholio:
We've been running tests in the order they're defined in. This could potentially cause some issues where tests defined further down could rely invisibly on some state that gets defined before that test gets run. This sets the order to random, so every time the test suite gets run, the tests are run in a different order each time.

### DIFF
--- a/spec/features/timeout_spec.rb
+++ b/spec/features/timeout_spec.rb
@@ -2,9 +2,10 @@ require "rails_helper"
 
 RSpec.feature "Teacher Student Loan Repayments claims", js: true do
   let(:one_second_in_minutes) { 1 / 60.to_f }
+  let(:two_seconds_in_minutes) { 2 / 60.to_f }
 
   before do
-    allow_any_instance_of(ApplicationController).to receive(:claim_timeout_in_minutes) { one_second_in_minutes }
+    allow_any_instance_of(ApplicationController).to receive(:claim_timeout_in_minutes) { two_seconds_in_minutes }
     allow_any_instance_of(ApplicationController).to receive(:claim_timeout_warning_in_minutes) { one_second_in_minutes }
     start_claim
   end
@@ -15,6 +16,7 @@ RSpec.feature "Teacher Student Loan Repayments claims", js: true do
   end
 
   scenario "Claimants can refresh their session" do
+    expect(page).to have_content("Your session will expire in #{one_second_in_minutes} minutes")
     expect_any_instance_of(ClaimsController).to receive(:update_last_seen_at)
     click_on "Continue session"
   end
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -14,6 +14,12 @@
 #
 # See http://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
 RSpec.configure do |config|
+  # Run specs in random order to surface order dependencies. If you find an
+  # order dependency and want to debug it, you can fix the order by providing
+  # the seed, which is printed after each run.
+  #     --seed 1234
+  config.order = :random
+
   # rspec-expectations config goes here. You can use an alternate
   # assertion/expectation library such as wrong or the stdlib/minitest
   # assertions if you prefer.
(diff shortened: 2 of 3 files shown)


pezholio:
I have noticed one particular seed that's causing problems for some reason. I've bashed my head against a brick wall for a couple of hours, but can't see what can be causing it:

```
rspec spec/features/timeout_spec.rb --seed 11588
```

It seems to be related to that final `Claimants are automatically redirected to the timeout page` test, which in this case is run first 😕

tekin:
I suspect this is happening because in the spec the timeout warning and the actual timeout are being stubbed to ha in 1 second, which means the dialog will appear and the page will timeout at the same time. I'll see if I can tweak this spec.

spec/features/timeout_spec.rb
```diff
@@ -2,9 +2,10 @@
 
 RSpec.feature "Teacher Student Loan Repayments claims", js: true do
   let(:one_second_in_minutes) { 1 / 60.to_f }
+  let(:two_second_in_minutes) { 2 / 60.to_f }
```
pezholio: Can I be really annoying and ask for proper pluralisation on this? 😄

tekin: done

pezholio:
We've just had a fail related to why I implemented this change in the first place:

https://dfe-ssp.visualstudio.com/S118-Teacher-Payments-Service/_build/results?buildId=21909&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9

There's a [commit here](https://github.com/DFE-Digital/dfe-teachers-payment-service/pull/441/commits/5c50d071ec96557181b28e8dd6f830fa787cc1fe) that addresses it. Basically we need to make sure we clear the mailer deliveries array before each test run, otherwise it just gets filled up

pezholio:
Another failing seed - `30470` makes `govuk_verify_spec` fail with two errrors:

```bash
  1) Teacher verifies identity using GOV.UK Verify successful verification with JavaScript enabled
     Failure/Error: choose year

     Capybara::ElementNotFound:
       Unable to find radio button "1 September 2014 to 31 August 2015" that is not disabled
     # ./spec/support/feature_helpers.rb:39:in `choose_qts_year'
     # ./spec/features/govuk_verify_spec.rb:8:in `block (2 levels) in <main>'
     # ./spec/support/bullet.rb:6:in `block (2 levels) in <main>'

  2) Teacher verifies identity using GOV.UK Verify user takes a particularly long time to get through verify Does not time the user out
     Failure/Error: choose year

     Capybara::ElementNotFound:
       Unable to find radio button "1 September 2014 to 31 August 2015" that is not disabled
     # ./spec/support/feature_helpers.rb:39:in `choose_qts_year'
     # ./spec/features/govuk_verify_spec.rb:8:in `block (2 levels) in <main>'
     # ./spec/support/bullet.rb:6:in `block (2 levels) in <main>'
```

tekin:
Added an additional commit to clear the emails between specs.

pezholio:
Running `30470` with the `headless` argument removed from the webdriver options seems to show the "Start Now" button receive focus (it gets highlighted yellow), but it doesn't look like the click action ever gets sent.

tekin:
This definitely opens a can of worms. Seed `24989` also results in `spec/features/school_search_spec.rb:56` failing. Will investigate some more later.

pezholio:
Arrrrgh

pezholio:
Screenshot of what I was talking about:

capybara-201909191321001909982328

I'm not sure why the seeds would affect the behaviour like this? And so consistently?

pezholio:
I've just run those seeds again locally, and they seem to work now. I've also run the tests 100 times locally and had no fails, so I think this is OK to merge now

spec/spec_helper.rb
```diff
@@ -84,12 +86,6 @@
   #   # particularly slow.
   #   config.profile_examples = 10
   #
-  #   # Run specs in random order to surface order dependencies. If you find an
-  #   # order dependency and want to debug it, you can fix the order by providing
-  #   # the seed, which is printed after each run.
-  #   #     --seed 1234
```
erbridge: Can we move this comment to be alongside the config, too?

PR #446, opened 2019-09-19
Show how many claims are awaiting approval

pezholio:
This shows a service operator at a glance how many claims are awaiting approval, meaning they don't have to download a CSV and count rows in order to keep tabs on how many people have applied

image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- Show how many claims are awaiting approval
 - Show date of birth in a better place and format for claim checking
 - Show DfE number against school in approver view of claim
 - Drop redundant full_name column from claims table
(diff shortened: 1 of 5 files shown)


pezholio:
[A Service Operator is able to see how many claims have been submitted without downloading any files](https://trello.com/c/xURrlF98/711-a-service-operator-is-able-to-see-how-many-claims-have-been-submitted-without-downloading-any-files)

tekin:
So the number the team is currently interested in and reporting is the total number of claims. at the moment whilst we aren't actually approving any claims, this number also equals the number of approved claims, but as soon as we approve one, the number will drift. Can we also show the total claim count so that we always know that number, even after claims start to be marked as "approved"?

pezholio:
Cool, makes sense. I've added a very simple dashboard style view to the admin hompage, as it didn't seem to make much sense to add that number to the Manage Claims page. May need some extra visual flair, but it does the job I think

PR #447, opened 2019-09-19
Changes to the claim data shown to service operators

mec:
During a claim checking session we noticed that we were checking date of birth before national insurance number so it made sense to show that first. We also noticed that the date format in DQT and TPS is DD/MM/YYYY and claim checking could be eased by displaying the DOB in the same format.

# Before

Screenshot_2019-09-19 Teachers claim back your student loan repayments – GOV UK(3)

# After

Screenshot_2019-09-19 Teachers claim back your student loan repayments – GOV UK(2)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- Show date of birth in a better place and format for claim checking
 - Show DfE number against school in approver view of claim
 - Drop redundant full_name column from claims table
 - Updated Accessibility statement to show current issues
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -33,6 +33,7 @@ en:
   date:
     formats:
       default: "%-d %B %Y"
+      day_month_year: "%d/%m/%Y"
   number:
     currency:
       format:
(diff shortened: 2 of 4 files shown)


mec:
[Changes to the claim data shown to service operators](https://trello.com/c/95fXD3fk/709-changes-to-the-claim-data-shown-to-service-operators)

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- Show date of birth in a better place and format for claim checking
```
pezholio: I'd argue that any format other than ISO8601 isn't 'better' :trollface:

config/locales/en.yml
```diff
@@ -33,6 +33,7 @@ en:
   date:
     formats:
       default: "%-d %B %Y"
+      claim_checking: "%d/%m/%Y"
```
pezholio: Actually, can we use a more generic name here? (such as `date_month_year`). We might want to use it elsewhere.

CHANGELOG.md
mec: I'm not getting into your date fetish! 😄

PR #448, opened 2019-09-19
A service operator is able to mark a claim as being investigated

pezholio:
This allows service operators to escalate a claim, in a similar way to how claims can be rejected in #441 

This should not be merged until #441 is finished and merged, as it contains some commits from there too


pezholio:
[A Service Operator is able to mark a claim as "being investigated"](https://trello.com/c/OWwaUbb2/698-a-service-operator-is-able-to-mark-a-claim-as-being-investigated)

PR #451, opened 2019-09-23
Use check model for approvals

pezholio:
This tweaks the Claim model to use a new Checks table, which gives us more flexibility when we want to reject a claim, or flag it as needing more attention


pezholio:
[A Service Operator is able to mark a claim as "rejected"](https://trello.com/c/rtmSzr5H/695-a-service-operator-is-able-to-mark-a-claim-as-rejected)

pezholio:
[A Service Operator is able to mark a claim as "rejected"](https://trello.com/c/rtmSzr5H/695-a-service-operator-is-able-to-mark-a-claim-as-rejected)

app/models/check_create.rb
```diff
@@ -0,0 +1,25 @@
...
+  def initialize(claim_id:, checked_by:, status:)
+    @claim = Claim.find(claim_id)
+    @checked_by = checked_by
+    @status = status
+  end
+
+  def perform
+    return false unless claim.approvable?
+
+    create_check.valid? ? send_confirmation : false
```
robbiepaul: Is it possible that the claim to be marked as approved but `send_confirmation` may return false causing the user to see the message "Claim cannot be approved"? It feels like `perform` should return the result of the action and sending the email should be a side-effect

app/models/check_create.rb
```diff
@@ -0,0 +1,25 @@
+class CheckCreate
```
robbiepaul: We've got the class `CheckCreate` and private method `create_check`, which confused me slightly. I wonder if calling the model something like `CheckClaim` might be more explicit

app/models/claim.rb
```diff
@@ -45,6 +44,7 @@ class Claim < ApplicationRecord
   enum student_loan_plan: STUDENT_LOAN_PLAN_OPTIONS
 
   belongs_to :eligibility, polymorphic: true
+  belongs_to :check, optional: true
   accepts_nested_attributes_for :eligibility, update_only: true
```
erbridge: Can we flip these? It helps cognitively (for me at least), to keep the nested attribute stuff next to the relationship.

pezholio:
So, I've confirmed that there have been no approvals carried out in production, so I think we can safely merge this without having to resort to the added complication of doing a data migration 👍

app/controllers/admin/claims_controller.rb
```diff
@@ -15,7 +15,7 @@ def show
   end
 
   def payroll
-    claims = Claim.includes(eligibility: [:claim_school, :current_school]).approved.order(:approved_at)
+    claims = Claim.preload(eligibility: [:claim_school, :current_school]).joins(:check).approved.order("checks.created_at")
```
erbridge:
```suggestion
    claims = Claim.preload(eligibility: [:claim_school, :current_school]).approved.order("checks.created_at")
```

The join is already in `approved`.

db/schema.rb
```diff
@@ -10,12 +10,19 @@
...
+ActiveRecord::Schema.define(version: 2019_09_23_145104) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
   enable_extension "plpgsql"
 
+  create_table "checks", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.integer "status"
+    t.string "checked_by"
+    t.datetime "created_at", null: false
```
erbridge: I don't know if creation date is a reliable "approval date". Can we be explicit about it? What happens if we reject a claim then decide to amend it and approve it?

spec/factories/check.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory :check do
+    status { :approved }
```
erbridge: Shouldn't this be in an `approved` trait?

spec/models/claim_spec.rb
```diff
@@ -363,7 +363,7 @@
   describe "awaiting_approval" do
     let!(:submitted_claims) { create_list(:claim, 5, :submitted) }
     let!(:unsubmitted_claims) { create_list(:claim, 2, :submittable) }
-    let!(:approved_claims) { create_list(:claim, 5, :submitted, approved_at: Time.zone.now) }
+    let!(:approved_claims) { create_list(:claim, 5, :submitted, check: create(:check, status: :approved)) }
```
erbridge:
```suggestion
    let!(:approved_claims) { create_list(:claim, 5, :approved) }
```

spec/requests/admin_claim_checks_spec.rb
```diff
@@ -47,11 +48,11 @@
       follow_redirect!
     end
 
-    describe "claim_approvals#create" do
+    describe "claim_checks#create" do
       let(:claim) { create(:claim, :submitted) }
 
       it "does not allow a claim to be approved" do
```
erbridge: I don't understand what this spec is testing. Why doesn't it allow a claim to be approved, and this doesn't appear to actually be testing attempting to approve it (it should have the status parameter, shouldn't it?

spec/requests/admin_claim_checks_spec.rb
```diff
@@ -8,29 +8,30 @@
...
+          expect(claim.check.status).to eq("approved")
         end
       end
 
       context "when claim is unapprovable" do
         let(:claim) { create(:claim, :approved) }
 
         it "shows an error" do
-          post admin_claim_approvals_path(claim_id: claim.id)
+          post admin_claim_checks_path(claim_id: claim.id)
```
erbridge:
```suggestion
          post admin_claim_checks_path(claim_id: claim.id, status: "approved")
```

spec/requests/admin_claim_checks_spec.rb
```diff
@@ -47,11 +48,11 @@
       follow_redirect!
     end
 
-    describe "claim_approvals#create" do
+    describe "claim_checks#create" do
```
erbridge: This is the same scope as above, isn't it?

spec/requests/admin_claim_checks_spec.rb
```diff
@@ -47,11 +48,11 @@
...
       let(:claim) { create(:claim, :submitted) }
 
       it "does not allow a claim to be approved" do
-        post admin_claim_approvals_path(claim_id: claim.id)
+        post admin_claim_checks_path(claim_id: claim.id)
```
erbridge:
```suggestion
        post admin_claim_checks_path(claim_id: claim.id, status: "approved")
```

spec/models/check_claim_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      it "queues an email job" do
+        expect { check_create.perform }
+          .to have_enqueued_job(ActionMailer::DeliveryJob)
+          .with("ClaimMailer", "approved", "deliver_now", claim)
+      end
+    end
+
+    context "when check is invalid" do
+      let(:status) { nil }
+      let(:user) { nil }
```
erbridge: Can we split these cases up? We want to make sure we can't create a check without either of them.

spec/models/check_claim_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+        expect { check_create.perform }.to change { Check.count }.by(0)
+      end
+
+      it "does not queue an email job" do
+        expect { check_create.perform }.to_not have_enqueued_job
+      end
+    end
+  end
+
+  context "with a non-approvable claim" do
```
erbridge:
```suggestion
  context "with an already approved claim" do
```

spec/requests/admin_claim_checks_spec.rb
```diff
@@ -8,29 +8,30 @@
...
 
-          expect(claim.approved_at).to eq(Time.zone.now)
-          expect(claim.approved_by).to eq("123")
+          expect(claim.check.created_at).to eq(Time.zone.now)
+          expect(claim.check.checked_by).to eq("123")
+          expect(claim.check.status).to eq("approved")
         end
       end
 
       context "when claim is unapprovable" do
```
erbridge:
```suggestion
       context "when claim is already approved" do
```

spec/factories/claims.rb
```diff
@@ -31,8 +31,7 @@
 
     trait :approved do
       submitted
-      approved_at { Time.zone.now }
-      approved_by { "12345" }
+      association(:check)
```
erbridge: This will need the `approved` trait if you go that way.

spec/factories/check.rb
pezholio: Yeah, but that'll come later. At the moment, the only status we can have is `approved`

spec/models/claim_spec.rb
pezholio: I did think about that, but we thought it was better to be explicit in this case

spec/factories/check.rb
erbridge: If we do it now, it reduces the thrashing on specs when we do add it, making the history cleaner. Adding `rejected` shouldn't men changes to `approved`, imo.

spec/models/claim_spec.rb
erbridge: I think it it makes it harder to make changes to the approval process. You have to come and update this spec, when you've already updated the (identically configured) trait.

db/schema.rb
erbridge: To be clear, I think this should happen now so we don't need to do data migrations later.

pezholio: I am beginning to wonder if we need multiple checks, but that comes with the added complication of which check do we treat as canonical? If something is approved, I don't think we should be able to unapprove it. It might be OK to approve something if it's previously been declined, or approve it after it's been escalated, but I think approved claims need to stay approved.

robbiepaul: Yeah, it's tricky I would argue that we should allow a claim to go from approved to rejected (ie. some new information has come to light) until the point that a payment has been requested? Maybe we'll need a state machine at that point?

erbridge: Oh ok. If checks are read-only, except on creation, can we enforce that somehow?

erbridge: Also, does it imply that `status` is actually like a `result`? `status` implies to me that it's mutable.

spec/requests/admin_claim_checks_spec.rb
erbridge: Ah ha! Ok. This now makes sense.

app/models/claim.rb
```diff
@@ -138,7 +128,7 @@ def submittable?
   end
 
   def approvable?
-    submitted? && approved_at.nil?
+    submitted? && check.nil?
```
erbridge: This should be checking that the `check` is also not approved, right?

app/models/claim.rb
```diff
@@ -35,15 +35,16 @@ class Claim < ApplicationRecord
...
-    approved_by: false,
+    check_id: false,
   }.freeze
 
   enum student_loan_country: StudentLoans::COUNTRIES
   enum student_loan_start_date: StudentLoans::COURSE_START_DATES
   enum student_loan_courses: {one_course: 0, two_or_more_courses: 1}
   enum student_loan_plan: STUDENT_LOAN_PLAN_OPTIONS
 
+  belongs_to :check, optional: true
```
erbridge: I think this should actually be a `has_many` relationship, given the discussion about replacing checks if the status changes, the most recently created check being the "current" one.

pezholio: Managed to sort this. It may need some tweaking/rebasing (and we may need to look at a state machine at some point (but not in the tranche of work)), but the hard work is done.

spec/factories/claims.rb
```diff
@@ -31,8 +31,9 @@
 
     trait :approved do
       submitted
-      approved_at { Time.zone.now }
-      approved_by { "12345" }
+      after(:create) do |claim|
+        create(:check, claim: claim)
+      end
```
erbridge: This can just be `checks { [build(:check)] }` (or `build_list` if your prefer), so you can `build` approved claims.

spec/features/admin_claim_approve_spec.rb
```diff
@@ -23,8 +23,12 @@
 
         claim_to_approve.reload
 
-        expect(claim_to_approve.approved_at).to eq(Time.zone.now)
-        expect(claim_to_approve.approved_by).to eq("12345")
+        expect(claim_to_approve.checks.count).to eq(1)
+
+        check = claim_to_approve.checks.last
```
erbridge: Probably a good idea to pull this out into a method, so we're consistent, and know why we're checking the last. It should really be ordered by creation_date, in case the list is populated differently, too, shouldn't it?

spec/models/check_claim_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  context "with a submittable claim" do
+    let(:claim) { create(:claim, :submitted) }
+
+    context "when a claim is approved" do
+      let(:status) { "approved" }
+
+      it "creates a check" do
+        check_create.perform
+        check = claim.reload.checks.first
```
erbridge: Case in point for my previous comment.

spec/features/admin_claim_approve_spec.rb
mec: Really like this. Do we think it is the `last_check` as a check is only created when performing an action?

db/schema.rb
mec: I think I am with @erbridge on `status` I didn't think we would have multiple checks, but it makes way more sense, so `status` starts to feel wrong.

app/models/claim.rb
mec: This was one of the balance ones from me again. My thinking was that a check can only mean an approval at this point so no check means it is not approved.

spec/factories/check.rb
mec: For me this was about catching the right balance of small steps vs knowing what is coming. As this work was about a check that could only result in an approval, we purposely only took steps to make that possible. 

Funnily enough, we spoke about how to best strike that balance as it's something I struggle with.

spec/features/admin_claim_approve_spec.rb
erbridge: `latest_check`, maybe? Since it will be ordered by date.

mec:
[Introduce a check model for claims](https://trello.com/c/iel91cpn/725-introduce-a-check-model-for-claims)

app/models/claim.rb
```diff
@@ -35,15 +35,15 @@ class Claim < ApplicationRecord
     updated_at: false,
     verified_fields: false,
     verify_response: true,
-    approved_at: false,
-    approved_by: false,
```
erbridge: Should we have some `checks` related fields in here?

app/models/check.rb
```diff
@@ -0,0 +1,18 @@
...
+  enum result: {
+    approved: 0,
+  }
+
+  validates :result, presence: true
+  validates :checked_by, presence: true
+
+  before_create do
+    self.checked_at = Time.zone.now
+  end
```
erbridge: Went for this because the default wasn't working.

mec: Is that short hand for a `before_save` lifecycle hook?

app/models/claim.rb
mec: I think these are only attributes on the `Claim` model, we use this to stop sensitive data from being logged.

app/models/check.rb
erbridge: `before_create` is a different callback from `before_save`, and you can pass a block in instead of the name of a method (not sure which aspect you were asking about).

app/models/claim.rb
erbridge: We have `eligibility_id` and `eligibility_type` in there, too, which makes me think that the thing checking for these isn't working?

mec: It looks like the `accepts_nested_attributes_for` is adding `eligibility_id` and `eligibility_type` to the attributes for the Claim, I am not sure we would access the Check attributes in the same way?

erbridge: Ok, makes sense :).

app/controllers/admin/claim_checks_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  private
+
+  def load_claim
+    @claim = Claim.find(params[:claim_id])
+  end
+
+  def reject_checked_claims
+    if @claim.check.present?
```
erbridge: :+1: to not coupling to approval.

app/controllers/admin/claim_checks_controller.rb
```diff
@@ -0,0 +1,23 @@
+class Admin::ClaimChecksController < Admin::BaseAdminController
+  before_action :ensure_service_operator
+  before_action :load_claim
+  before_action :reject_checked_claims
+
+  def create
+    @claim.create_check!(checked_by: admin_session.user_id, result: params[:result])
```
erbridge: Can we require the `result` param. It's really easy to forget to pass it during specs.

app/models/check.rb
```diff
@@ -0,0 +1,11 @@
...
+  belongs_to :claim
+
+  enum result: {
+    approved: 0,
+  }
+
+  def readonly?
+    persisted?
+  end
+end
```
erbridge: Do we not need to validate the presence of `result` and `checked_by`?

spec/features/admin_claim_approve_spec.rb
```diff
@@ -21,10 +21,7 @@
         find("a[href='#{admin_claim_path(claim_to_approve)}']").click
         perform_enqueued_jobs { click_on "Approve" }
 
-        claim_to_approve.reload
-
-        expect(claim_to_approve.approved_at).to eq(Time.zone.now)
-        expect(claim_to_approve.approved_by).to eq("12345")
+        expect(claim_to_approve.check.checked_by).to eq("12345")
```
erbridge: We should check the creation time, too, since we use that as a check date.

spec/requests/admin_claim_checks_spec.rb
```diff
@@ -8,33 +8,31 @@
...
           follow_redirect!
 
           expect(response.body).to include("Claim has been approved successfully")
 
-          claim.reload
-
-          expect(claim.approved_at).to eq(Time.zone.now)
-          expect(claim.approved_by).to eq("123")
+          expect(claim.check.checked_by).to eq("123")
+          expect(claim.check.result).to eq("approved")
```
erbridge: Shouldn't we be checking the creation date, too?

app/models/check.rb
tekin: yep, we should do that 👍🏽

spec/requests/admin_claim_checks_spec.rb
tekin: yep, will add this.

app/controllers/admin/claim_checks_controller.rb
tekin: I think if we add the validations to the model then we shouldn't need to as this line will fail noisily with an exception if the result is missing.

spec/features/admin_claim_approve_spec.rb
tekin: Actually I've changed my mind! I don't think we need to do that as it's default Rails behaviour that `created_at` gets set when the record is created. Testing that feels like testing the framework.

spec/requests/admin_claim_checks_spec.rb
tekin: As above, I don't think we need to explicitly test this as it's baked-in ActiveRecord behaviour.

spec/features/admin_claim_approve_spec.rb
erbridge: I'm more thinking about the case where someone changes the behaviour to create the empty check at the start of the claim (or something). We rely on the creation date for the time of check, so we should be confident that its creation date _is_ the time of check.

tekin: Yeah, I'm not convinced that sort of scenario is very likely to happen. We've defined a `Check` as a RESTfully created model that gets created when someone performs a check. For it to be created at a time other than when the check is performed would be quite odd.

erbridge: Sure.

app/controllers/admin/claim_checks_controller.rb
erbridge: My point about this one is that we previously had an error case test passing because the page rendered an error code, even though the reason was because of missing parameters, rather than creating another check.

mec: If we were to do this, would it be with strong parameters?

erbridge: I bring it up because it _was_ happening until I did my last rebase. Happy to be overruled.

tekin: Yep, the standard Rails way to do it would probably be `ActionController::Parameters#require`, which incidentally raises an exception if the required parameter isn't present!

spec/requests/admin_claim_checks_spec.rb
```diff
@@ -47,11 +45,11 @@
       follow_redirect!
     end
 
-    describe "claim_approvals#create" do
+    describe "claim_checks#create" do
       let(:claim) { create(:claim, :submitted) }
 
       it "does not allow a claim to be approved" do
-        post admin_claim_approvals_path(claim_id: claim.id)
+        post admin_claim_checks_path(claim_id: claim.id, result: "approved")
```
erbridge: This is the test that was passing despite missing the `result` parameter, btw.

tekin: Ah right, ok. What was the expectation before? It looks right to me now – it verifies specifically an "unauthorized" response, which shouldn't result in a false positive.

app/controllers/admin/claim_checks_controller.rb
tekin: @erbridge do you have a suggestion for what the code to protect against false-positives might look like?

spec/requests/admin_claim_checks_spec.rb
erbridge: The test is the same. Ignore me.

app/controllers/admin/claim_checks_controller.rb
mec: @erbridge that spec feels a bit off to me anyway, is it testing that a support user cannot approve a claim, as in create a check, so would the 'you are not an approver' fail first? 

When I change the behaviour I see what I expect, forgetting the `result` param causes a `ActiveRecord::RecordInvalid` error with the validations.

erbridge: No suggestions for now, sorry.

PR #455, opened 2019-09-25
Split employment status into two questions

erbridge:


erbridge:
[Split employment status into two questions](https://trello.com/c/r0RmbDrs/690-split-employment-status-into-two-questions)

app/views/claims/where_teaching.html.erb
```diff
@@ -10,9 +10,14 @@
           <fieldset class="govuk-fieldset">
 
             <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
-              <h1 class="govuk-fieldset__heading"><%= t("student_loans.questions.employment_status") %></h1>
+              <h1 class="govuk-fieldset__heading"><%= t("student_loans.questions.where_teaching") %></h1>
```
lawrence-dxw: The answers (which start with "Yes, ") need to be changed as a result of this change. There's also a `choose_still_teaching` helper, which uses these answers and which should be renamed.

spec/features/school_search_spec.rb
```diff
@@ -17,12 +18,13 @@
     click_on "Continue"
 
     expect(claim.eligibility.reload.claim_school).to eql schools(:penistone_grammar_school)
-    expect(page).to have_text("Are you still employed to teach at a school in England")
+    expect(page).to have_text(I18n.t("student_loans.questions.where_teaching"))
```
lawrence-dxw: Has the team discussed the preferred approach to testing localised strings? A quick browse shows our current approach to be inconsistent; sometimes the tests call the translation helper, sometimes they include the expected string as a literal. But I don't know if it's because we've never discussed it or because we have now settled on a standard but not yet updated existing code.

spec/models/page_sequence_spec.rb
```diff
@@ -6,57 +6,65 @@
   let(:claim) { build(:claim) }
 
   describe "#slugs" do
-    it "excludes “current-school” when the claimant still works at the school they are claiming against" do
-      claim.eligibility.employment_status = :claim_school
+    it "only excludes “current-school” when the claimant still works at the school they are claiming against" do
```
lawrence-dxw: Minor, and perhaps I'm parsing the sentence wrong but from this description it sounds like where before you were only testing the "if" case, now you're only testing the "only if" case. I'd suggest "excludes “current-school” if and only if the claimant still works at the school they are claiming against". Or two tests.

app/models/page_sequence.rb
```diff
@@ -45,6 +45,7 @@ def initialize(claim, current_slug)
 
   def slugs
     SLUGS.dup.tap do |sequence|
+      sequence.delete("where-teaching") if claim.eligibility.claim_school.present? && claim.eligibility.claim_school.closed?
```
lawrence-dxw: Could this just be `claim.eligibility.claim_school&.closed?` or is `present?` doing something special here?

app/models/page_sequence.rb
```diff
@@ -2,16 +2,17 @@
...
-# `claim_school` will skip the `current-school` page as `current_school` is
-# inferred to be the same as `claim_school.
+# claim school will skip the `current-school` page as the current school is
+# inferred to be the same as claim school.
 class PageSequence
   SLUGS = [
     "qts-year",
+    "currently-teaching",
     "claim-school",
-    "still-teaching",
```
lawrence-dxw: Do we need to think about handling the case of a user who is on the `still-teaching` page and then refreshes after we deploy this?

spec/features/school_search_spec.rb
erbridge: Never discussed. However, these tests are not testing the copy itself, but using copy as a proxy for knowing which page we're on.

app/models/page_sequence.rb
erbridge: Consider the case where `claim_school == nil`. `claim.eligibility.claim_school&.closed?` would be falsy, and we would delete `where-teaching`, which I think is incorrect.

app/models/page_sequence.rb
erbridge: Yes. This is merging into a long-lived feature branch that we'll need to do something about rolling out because it will drastically change the flow.

app/views/claims/where_teaching.html.erb
erbridge: Good catch.

app/models/page_sequence.rb
lawrence-dxw: Does that mean we're planning to handle the consequences as a separate step before we release?

erbridge: Yes.

app/models/page_sequence.rb
lawrence-dxw: Yep, makes sense.

erbridge: Actually I think I'm wrong. I'll check.

app/views/claims/where_teaching.html.erb
erbridge: Done.

spec/models/page_sequence_spec.rb
erbridge: Done.

PR #459, opened 2019-09-26
Allow retrying RSpec failures

robbiepaul:
During the refactor work I found myself reaching for this command several times. It's useful when working through a number of failures in batches.

The command is `rspec --only-failures` and can be re-ran several times while fixing specs until there aren't from the original failures left.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -58,12 +58,12 @@ RSpec.configure do |config|
   #   # aliases for `it`, `describe`, and `context` that include `:focus`
   #   # metadata: `fit`, `fdescribe` and `fcontext`, respectively.
   #   config.filter_run_when_matching :focus
-  #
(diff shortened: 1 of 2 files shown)


spec/spec_helper.rb
```diff
@@ -62,7 +62,7 @@
   #   # Allows RSpec to persist some state between runs in order to support
   #   # the `--only-failures` and `--next-failure` CLI options. We recommend
   #   # you configure your source control system to ignore this file.
-  #   config.example_status_persistence_file_path = "spec/examples.txt"
+  config.example_status_persistence_file_path = "spec/examples.txt"
```
erbridge: Can you uncomment the connected comment and bracketing spaces, too, please?

robbiepaul: 👍

PR #460, opened 2019-09-26
Add the claim start and submission times to the admin view

erbridge:
We needed to know to support a user, which required looking in the database.


spec/helpers/admin/claims_helper_spec.rb
```diff
@@ -4,7 +4,7 @@
   let(:claim_school) { schools(:penistone_grammar_school) }
   let(:current_school) { create(:school, :student_loan_eligible) }
 
-  describe "eligibility_answers" do
+  describe "#eligibility_answers" do
```
lawrence-dxw: Would you mind splitting these renames into a separate commit?

erbridge: Done.

PR #461, opened 2019-09-26
Update privacy notice to include IP address

robbiepaul:
We are now sending the client IP address to Application Insights so we state that in the notice.


robbiepaul:
[Update privacy notice](https://trello.com/c/huBcrHEV/703-update-privacy-notice)

PR #462, opened 2019-09-26
Make all-through schools eligible

erbridge:
This is following a discussion with the policy team.


erbridge:
[Make "All-through" schools eligibile](https://trello.com/c/I43Ho3Be/732-make-all-through-schools-eligibile)

PR #468, opened 2019-09-30
Add Virtual Network to ProjectCore resource group

robbiepaul:
At the moment we're not using the virtual network for anything, however we intend to use it for either protecting resources with a firewall or splitting out worker processes (both require our own virtual network)

Virtual networks are [free of charge](https://azure.microsoft.com/en-us/pricing/details/virtual-network/).

We add the virtual network first as we expect to rely on these in the later resource group deployments.

### DIFF
--- a/azure/resource_groups/ProjectCore/template.json
+++ b/azure/resource_groups/ProjectCore/template.json
@@ -1,0 +1,60 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "resourceNamePrefix": {
+      "type": "string",
+      "defaultValue": "[resourceGroup().name]"
+    },
+    "vnetAddressSpaceCIDR": {
+      "type": "string"
+    }
+  },
+  "variables": {
+    "platformBuildingBlocksDeploymentUrlBase": "https://raw.githubusercontent.com/DFE-Digital/bat-platform-building-blocks/7a4748a0cf366193d31434bd7796d483bd281385/templates/",
+    "vnetDeploymentName": "[concat(parameters('resourceNamePrefix'), '-virtualnetwork')]",
+    "vnetName": "[concat(parameters('resourceNamePrefix'), '-vn')]",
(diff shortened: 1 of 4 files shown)


erbridge:
[CYBER 7: Protect critical infrastructure with a firewall](https://trello.com/c/tJNwUwtt/560-cyber-7-protect-critical-infrastructure-with-a-firewall)

azure/resource_groups/ProjectCore/template.json
```diff
@@ -0,0 +1,47 @@
...
+        },
+        "subnets": [
+          {
+            "name": "default",
+            "properties": {
+              "addressPrefix": "10.211.0.0/27",
+              "serviceEndpoints": [],
+              "delegations": [],
+              "privateEndpointNetworkPolicies": "Enabled",
+              "privateLinkServiceNetworkPolicies": "Enabled"
```
chelnak: I'd recommend using service endpoints and delegations here. Private link has only just come out in preview. I'd probably avoid this until it's GA.

The block template above allows you to pass in a subnet configuration object. You can include service endpoint and delegation config within that object.

For ref: https://docs.microsoft.com/en-us/azure/templates/microsoft.network/2019-04-01/virtualnetworks/subnets#subnetpropertiesformat-object

azure/resource_groups/ProjectCore/template.json
```diff
@@ -0,0 +1,47 @@
...
+      "type": "string",
+      "defaultValue": "[resourceGroup().name]"
+    }
+  },
+  "variables": {
+    "vNetName": "[concat(parameters('resourceNamePrefix'), '-vn')]"
+  },
+  "resources": [
+    {
+      "type": "Microsoft.Network/virtualNetworks",
```
chelnak: Can we use the building block here?

https://github.com/DFE-Digital/bat-platform-building-blocks/blob/master/templates/virtual-network.json

It's already being used for ACI integration elsewhere.

erbridge: Oh, nice call. I assumed there wouldn't be one, yet. Thanks.

azure/resource_groups/ProjectCore/template.json
erbridge: Yeah, we decided to leave those out to be added when we need them for specific uses.

PR #474, opened 2019-10-02
Reword the how many courses question

erbridge:
We only care about the courses they took a loan out for, and we use it calculate the plan they're on, and we have seen that users are confused by this question.


tekin:
@erbridge has this change been run by/confirmed by policy as the right change to make?

PR #475, opened 2019-10-02
Payroll CSV file tweaks

pezholio:
This makes the following changes to the Cantium CSV:

* Changes the date format to `YYYYMMDD`
* Hardcodes the `TITLE` field as `Captain`
* Adds a `ROLL_NUMBER` field (which is currently blank)

I've also simplified the CSV tests as we were testing the output of the CSV row in two different places, which tripped me up a bit


pezholio:
[Payroll CSV file tweaks](https://trello.com/c/GtaqNC6f/741-payroll-csv-file-tweaks)

PR #476, opened 2019-10-02
Make the page sequence handle changes in user journeys

erbridge:
The intention is for users on a specific sequence to continue on that sequence, and rely on data migrations to backfill questions they didn't answer.

If any pages are new (and so missing from old sequences) and backfilling isn't possible, we ask them of the user at the end before checking their answers.

Open question: should the check your answers page, therefore, be fixed to refer to the latest version, causing a bit of jank for the users in that it might report questions they didn't answer or in a different order? Extra work would be required to make that not the case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- Add support for multiple page sequences
 - Update Cantium CSV fields
 
 ## [Release 013] - 2019-10-01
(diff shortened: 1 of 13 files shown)


erbridge:
[Make the page sequence handle changes in user journeys](https://trello.com/c/k1QZ6xVn/758-make-the-page-sequence-handle-changes-in-user-journeys)

tekin:
Going to close this for now until we have time to consider the impact of these changes and get the work prioritised, particularly with Maths & Physics on the horizon.

PR #477, opened 2019-10-03
Prepare for an external start page

pezholio:
This updates the first page of the claim to be a `new` action, rather than a `create` action, which then starts the claim process. This means that the start of the journey will be able to be linked to from a GOV.UK start page. 

The bulk of the lines of code in this PR are test changes, I've endeavoured to tidy up the tests a bit by adding a Request helper that means we're not always repeating ourselves when we're POSTing to create a Claim. This will also make things easier to change if / when we change the journey in any way.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -17,7 +17,7 @@ class ApplicationController < ActionController::Base
   private
 
   def send_unstarted_claiments_to_the_start
-    redirect_to root_url unless current_claim.present?
+    redirect_to root_url unless current_claim.persisted?
   end
 
   def admin_signed_in?
@@ -25,7 +25,7 @@ class ApplicationController < ActionController::Base
   end
 
   def current_claim
-    @current_claim ||= current_claim_from_session
+    @current_claim ||= current_claim_from_session || Claim.new(eligibility: StudentLoans::Eligibility.new)
   end
 
   def current_claim_from_session
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -25,7 +25,11 @@ class ApplicationController < ActionController::Base
   end
 
   def current_claim
-    @current_claim ||= Claim.find(session[:claim_id]) if session.key?(:claim_id)
+    @current_claim ||= current_claim_from_session
+  end
+
+  def current_claim_from_session
+    Claim.find(session[:claim_id]) if session.key?(:claim_id)
   end
 
   def claim_timeout_in_minutes
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -4,6 +4,11 @@ class ClaimsController < ApplicationController
 
   after_action :clear_claim_session, if: :submission_complete?
 
+  def new
+    @current_claim = Claim.new(eligibility: StudentLoans::Eligibility.new)
+    render claim_page_template
+  end
+
   def create
     claim = Claim.create!(eligibility: StudentLoans::Eligibility.new)
     session[:claim_id] = claim.to_param
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -5,13 +5,10 @@ class ClaimsController < ApplicationController
   after_action :clear_claim_session, if: :submission_complete?
 
   def new
-    @current_claim = Claim.new(eligibility: StudentLoans::Eligibility.new)
     render claim_page_template
   end
 
   def create
-    @current_claim = Claim.new(eligibility: StudentLoans::Eligibility.new)
-
     if update_current_claim!
       session[:claim_id] = @current_claim.to_param
       redirect_to claim_path(next_slug)
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -10,10 +10,14 @@ class ClaimsController < ApplicationController
   end
 
   def create
-    claim = Claim.create!(eligibility: StudentLoans::Eligibility.new)
-    session[:claim_id] = claim.to_param
+    @current_claim = Claim.new(eligibility: StudentLoans::Eligibility.new)
 
-    redirect_to claim_path("qts-year")
+    if update_current_claim!
+      session[:claim_id] = @current_claim.to_param
+      redirect_to claim_path(next_slug)
+    else
+      render claim_page_template
+    end
   end
 
   def show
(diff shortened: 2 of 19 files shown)


pezholio:
[Prepare for an external start page](https://trello.com/c/BC1M2w1J/749-prepare-for-an-external-start-page)

app/controllers/claims_controller.rb
```diff
@@ -4,6 +4,11 @@ class ClaimsController < ApplicationController
 
   after_action :clear_claim_session, if: :submission_complete?
 
+  def new
+    @current_claim = Claim.new(eligibility: StudentLoans::Eligibility.new)
+    render "qts_year"
```
tekin: Is there a way we can use the page sequence rather than hard-coding this to a specific view?

pezholio: I did consider that, but seeing that it hadn't been done previously (when redirecting etc), I kinda assumed there was a reason for not doing that. Makes sense to give it a go though 👍

spec/requests/claims_spec.rb
```diff
@@ -1,6 +1,14 @@
 require "rails_helper"
 
 RSpec.describe "Claims", type: :request do
+  describe "claims#new request" do
+    it "renders the correct template with `current_claim` set correctly" do
+      claim = Claim.new
+      expect(Claim).to receive(:new) { claim }
```
tekin: Can we test this new action without having to mock out the call to `Claim.new`? It makes the spec test an internal implementation detail, rather than the outcome that we expect. Is the underlying outcome that we want that the form on the page will `POST` to the `create` action?

app/controllers/claims_controller.rb
tekin: By delegating to the page sequence, we decouple the controller from specific pages, which means future policies just have to define their own sequence and the controller code doesn't have to change or have conditions.

spec/requests/claims_spec.rb
tekin: Can we test this new action without having to mock out the call to `Claim.new`? It makes the spec test an internal implementation detail, rather than the outcome that we expect. Is the underlying outcome that we want that the form on the page will `POST` to the `create` action?

app/controllers/claims_controller.rb
```diff
@@ -4,11 +4,20 @@ class ClaimsController < ApplicationController
...
-    claim = Claim.create!(eligibility: StudentLoans::Eligibility.new)
-    session[:claim_id] = claim.to_param
+    @current_claim = Claim.new(eligibility: StudentLoans::Eligibility.new)
 
-    redirect_to claim_path("qts-year")
+    if update_current_claim!
+      session[:claim_id] = @current_claim.to_param
+      redirect_to claim_path(next_slug)
+    else
+      render "qts_year"
```
tekin: as above with `new`, if we can abstract this away so that the controller action isn't coupled to the string for the first page in the sequence it would be better.

config/routes.rb
```diff
@@ -22,9 +22,11 @@
   get "/accessibility_statement", to: "static_pages#accessibility_statement", as: :accessibility_statement
 
   constraints slug: %r{#{PageSequence::SLUGS.join("|")}} do
-    resources :claims, only: [:new, :create, :show, :update], param: :slug, path: "/claim"
+    resources :claims, only: [:new, :show, :update], param: :slug, path: "/claim"
   end
 
+  resources :claims, only: [:create], params: {slug: "qts-year"}, path: "/claim"
```
tekin: Can we write this as singular route definition, rather than a resource that is only a single action? I think it would be easier to follow written like that. Also can we parameterise the first slug via `PageSequence` rather than hard coding?

spec/requests/verify_authentications_spec.rb
```diff
@@ -129,4 +129,14 @@
       expect(response).to redirect_to(root_path)
     end
   end
+
+  def generate_claim!
```
tekin: is this a duplicate of `create_claim` that's defined in `RequestHelpers`?

spec/support/request_helpers.rb
```diff
@@ -0,0 +1,11 @@
+module RequestHelpers
+  def create_claim!
```
tekin: Maybe rename this `start_claim?`

spec/features/changing_answers_spec.rb
```diff
@@ -23,6 +23,31 @@
     expect(page).to have_content("Check your answers before sending your application")
   end
 
+  context "when changing the year they qualified" do
```
tekin: This feature spec file takes over 40 seconds to run on my machine, mostly because of the way features are being tested with multiple scenarios: each scenario will execute `answer_all_student_loans_claim_questions`, which clicks through every single page of the journey. As a separate issue we should definitely to rework these specs so that there are fewer scenarios, but for now, can we try and write new specs with as few scenarios as possible? In this case, I think this feature can be tested with a single `"changes their QTS year"` scenario containing multiple expectations.

app/controllers/application_controller.rb
```diff
@@ -25,7 +25,11 @@ def admin_signed_in?
   end
 
   def current_claim
-    @current_claim ||= Claim.find(session[:claim_id]) if session.key?(:claim_id)
+    @current_claim ||= current_claim_from_session
```
tekin: Instead of assigning directly to `@current_claim` in the `new` and `create` controller actions, we could do this:

```suggestion
    @current_claim ||= current_claim_from_session || Claim.new(eligibility: StudentLoans::Eligibility.new)
```

spec/requests/claims_spec.rb
tekin: Maybe it's enough to say it renders the first page in the sequence when there isn't a persisted current claim, and it redirects to the actual question slug if there is?

app/controllers/application_controller.rb
pezholio: I wondered about that, but will that not always give us a new claim if there's not a claim_id present in the session? It'll also mean we have less control when we come to support Maths and Physics too.

spec/requests/verify_authentications_spec.rb
pezholio: Whoops, yeah, I forgot to take that out

config/routes.rb
pezholio: Good call 👍

app/controllers/claims_controller.rb
pezholio: Yep, now we have access to `current_claim`, we can use `claim_page_template`

app/controllers/application_controller.rb
tekin: ah that's fair, I guess it would complicate things a little when Maths and Physics is introduced...

PR #478, opened 2019-10-03
Rework URL structure in preparation for Maths & Physics policy

robbiepaul:
In preparation for adding another policy (Maths & Physics) we need to have different URLs for each one. 

- Include the policy name as a parameter available the controller
- Only allow valid policies in the URL
- Include URL helpers in the asset pipeline as calling them statically from assets doesn't set the correct defaults

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -21,12 +21,14 @@ Rails.application.routes.draw do
   get "/cookies", to: "static_pages#cookies", as: :cookies
   get "/accessibility_statement", to: "static_pages#accessibility_statement", as: :accessibility_statement
 
-  constraints slug: %r{#{PageSequence::SLUGS.join("|")}} do
-    resources :claims, only: [:new, :create, :show, :update], param: :slug, path: "/claim"
-  end
+  scope path: ":policy", defaults: {policy: "student-loans"}, constraints: {policy: %r{student-loans}} do
+    constraints slug: %r{#{PageSequence::SLUGS.join("|")}} do
+      resources :claims, only: [:new, :create, :show, :update], param: :slug, path: "/"
+    end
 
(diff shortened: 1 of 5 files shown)


robbiepaul:
[Rework URL structure in preparation for Maths & Physics](https://trello.com/c/RVCRHSHo/750-rework-url-structure-in-preparation-for-maths-physics)

config/routes.rb
```diff
@@ -21,12 +21,14 @@
   get "/cookies", to: "static_pages#cookies", as: :cookies
   get "/accessibility_statement", to: "static_pages#accessibility_statement", as: :accessibility_statement
 
-  constraints slug: %r{#{PageSequence::SLUGS.join("|")}} do
-    resources :claims, only: [:new, :create, :show, :update], param: :slug, path: "/claim"
-  end
+  scope path: ":policy", defaults: {policy: "student-loans"}, constraints: {policy: %r{student-loans}} do
+    constraints slug: %r{#{PageSequence::SLUGS.join("|")}} do
+      resources :claims, only: [:new, :create, :show, :update], param: :slug, path: "/"
```
pezholio: Is it worth sending the policy to the controller? (even though we might not use it yet), so we can set the eligibility model at a later date?

robbiepaul: Yeah the policy is available in the controller via `params[:policy]` at the moment. Were you thinking of something else?

pezholio: Oh yeah, sorry, I missed that 🤦‍♂

robbiepaul:
It's worth mentioning `/student-loans` is currently a 404, not sure if we need to tackle that now or as part of a story later

pezholio:
That might be worth doing after #477 goes in, so we can redirect `/student-loans` to the `new` action, but that might need a bit more thought

PR #480, opened 2019-10-03
Changes to accessibility statement

titlescreen:
As we have upgraded accessible autocomplete, it now appears to be AA
compliant on latest OS / IE / JAWS versions. Added in text saying we
think we are AA compliant and we haven't been externally tested to
confirm this.


tekin:
🍾

PR #481, opened 2019-10-03
A service operator can reject a claim

mec:
This PR introduces the concept of _rejecting_ a claim. The goal was to do the smallest possible work to achieve that without trying to guess at what the process around claim rejection will look like.

We know claims will have to be rejected so we are confident of that, we do not know what the claims checking process looks like around rejections so want to keep the code as flexible as possible.

There were three steps:

1. change the words on the UI so it is clear we are not just approving claims, rather checking them
2. allow a `Check` to have a result of `rejected`
3. send a notification to the user

Some obvious omissions we know about are:

- the UI has had no thought as this is a seperate piece of work – I wonder if we should show the reject button at all?
- no record of why a claim is rejected is asked for or stored
- the email is missing the reason why the claim was rejected as we do not have it

in-line with the teams focus on only approving claims for the moment, we do not intend to use the reject feature until we understand more.

P.S. the commits are in a strange order, I think it is because I cherry-picked the notification one from older work – if someone can tell me how to sort that, that would be amazing! :D

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
-- A service operator can reject a claim
+- A service operator can reject a claim and an email is sent to the claimant
 - Make sure claimants cannot make a claim of £0
 
 ## [Release 015] - 2019-10-07
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- A service operator can reject a claim
 - Make sure claimants cannot make a claim of £0
 
 ## [Release 015] - 2019-10-07
--- a/app/views/admin/claims/show.html.erb
+++ b/app/views/admin/claims/show.html.erb
@@ -14,9 +14,14 @@
 
     <%= button_to "Approve",
                   admin_claim_checks_path(@claim),
-                  method: :post,
                   params: { result: "approved" },
                   class: "govuk-button",
                   data: { confirm: "Are you sure you want to approve this claim?" } %>
+
+    <%= button_to "Reject",
+                admin_claim_checks_path(@claim),
+                params: { result: "rejected" },
+                class: "govuk-button govuk-button--secondary",
+                data: { confirm: "Are you sure you want to reject this claim?" } %>
   </div>
 </div>
--- a/spec/features/admin_claim_reject_spec.rb
+++ b/spec/features/admin_claim_reject_spec.rb
@@ -1,9 +1,11 @@
 require "rails_helper"
 
 RSpec.feature "Rejecting a claim" do
+  let(:user_id) { "userid-345" }
+
   context "when a user is logged in as a service operator" do
     before do
-      stub_dfe_sign_in_with_role(AdminSession::SERVICE_OPERATOR_DFE_SIGN_IN_ROLE_CODE, "12345")
+      stub_dfe_sign_in_with_role(AdminSession::SERVICE_OPERATOR_DFE_SIGN_IN_ROLE_CODE, user_id)
       visit admin_path
       click_on "Sign in"
     end
@@ -22,7 +24,6 @@ RSpec.feature "Rejecting a claim" do
       perform_enqueued_jobs { click_on "Reject" }
 
       expect(claim_to_reject.check.checked_by).to eq(user_id)
-
       expect(page).to have_content("Claim has been rejected successfully")
       expect(page).to_not have_content(claim_to_reject.reference)
 
--- a/spec/features/admin_claim_reject_spec.rb
+++ b/spec/features/admin_claim_reject_spec.rb
@@ -29,12 +29,9 @@ RSpec.feature "Rejecting a claim" do
 
       mail = ActionMailer::Base.deliveries.last
 
-      expect(mail.subject).to eq(
-        "Your claim to get your student loan repayments back has been rejected, reference number: #{claim_to_reject.reference}"
-      )
-      expect(mail.body.raw_source).to match(
-        "Unfortunately your claim to get your student loan payments back has been denied."
-      )
+      expect(mail.subject).to match("been rejected")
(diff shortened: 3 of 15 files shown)


mec:
[A Service Operator is able to mark a claim as "rejected"](https://trello.com/c/rtmSzr5H/695-a-service-operator-is-able-to-mark-a-claim-as-rejected)

spec/features/admin_claim_reject_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  context "when a user is logged in as a service operator" do
+    before do
+      stub_dfe_sign_in_with_role(AdminSession::SERVICE_OPERATOR_DFE_SIGN_IN_ROLE_CODE, user_id)
+      visit admin_path
+      click_on "Sign in"
+    end
+
+    scenario "they can reject a claim" do
+      freeze_time do
```
tekin: Is this `free_time` actually necessary? I can't see anything in the spec that is checking timestamps.

spec/mailers/claim_mailer_spec.rb
```diff
@@ -32,4 +32,20 @@
...
+
+    it "renders the headers" do
+      expect(mail.subject).to eq("Your claim to get your student loan repayments back has been rejected, reference number: #{claim.reference}")
+      expect(mail.to).to eq([claim.email_address])
+    end
+
+    it "renders the body" do
+      expect(mail.body.encoded).to match("Dear John Kennedy,")
+      expect(mail.body.encoded).to match("Unfortunately your claim to get your student loan payments back has been denied.")
+      expect(mail.body.encoded).to match("If you feel that we have made a mistake in the processing of your application or something is incorrect please email studentloanteacherpayment@digital.education.gov.uk giving your reference number: #{claim.reference} if you have any questions about your claim.")
```
tekin: I have mixed feelings about specs that precisely verify that content matches. What we gain in exactness, we lose in the robustness of the spec as it's much harder to make copy adjustments without also breaking specs. What do you think to reworking these expectations to be a bit looser such that they still give us confidence the template is rendering without error, and that it is the right template, but without coupling them so tightly to the actual content so that content can easily be tweaked without having to change the spec?

spec/requests/admin_claim_checks_spec.rb
```diff
@@ -24,7 +24,20 @@
         end
       end
 
-      context "when claim is already approved" do
+      it "can reject a claim" do
+        freeze_time do
```
tekin: I think these `freeze_time` calls are a hangover from when we were explicitly testing against an `approved_at` column. Because we're now relying on the timestamp of the check, I think we can lose them without impacting the spec.

app/views/admin/claims/show.html.erb
```diff
@@ -14,9 +14,14 @@
...
                   admin_claim_checks_path(@claim),
-                  method: :post,
                   params: { result: "approved" },
                   class: "govuk-button",
                   data: { confirm: "Are you sure you want to approve this claim?" } %>
+
+    <%= button_to "Reject",
+                admin_claim_checks_path(@claim),
+                params: { result: "rejected" },
+                class: "govuk-button",
```
robbiepaul: Should we make this a secondary button? https://design-system.service.gov.uk/components/button/#secondary-buttons

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- a service operator can reject a claim and an email is sent to the claimant
```
robbiepaul:
```suggestion
- A service operator can reject a claim and an email is sent to the claimant
```

app/views/admin/claims/show.html.erb
mec: I guess right now we are not trying to solve the UI, as long as we can reject and approve a claim we can come back and resolve how that is achieved at a later date, there is work on that happening already.

I'm not sure how I feel about reject being secondary as well, that is a user performing the check doesn't know the outcome going into the check?

Hapyy to change this though if we think it is the right call to make right now.

robbiepaul: Yeah I think it's a quick win to have the button visually different. We aim to follow the design system and it says "Avoid using multiple default buttons on a single page". Not sure which one to use though

robbiepaul: Also, happy to see this addressed later

spec/mailers/claim_mailer_spec.rb
pezholio: I've got round this in the past by using translations in the emails and using the translations in the tests too, but I guess that makes the content of the emails less clear. I'd say the only thing we _really_ need to test here is for the presence of dynamic content (so, the name in this case).

tekin: Checking dynamic content is good. But I'd say we'd also want to perform a minimum match against the body _and substance_ of the email itself, otherwise the code could be changed torender the "approved" email template and the spec would still pass.

spec/mailers/claim_mailer_spec.rb
```diff
@@ -32,4 +32,20 @@
       expect(mail.body.encoded).to match("Email studentloanteacherpayment@digital.education.gov.uk giving your reference number: #{claim.reference} if you have any questions.")
     end
   end
+
+  describe "#rejected" do
+    let(:claim) { create(:claim, :submitted, first_name: "John", middle_name: "Fitzgerald", surname: "Kennedy") }
+    let(:mail) { ClaimMailer.rejected(claim) }
+
+    it "renders the headers" do
+      expect(mail.subject).to eq("Your claim to get your student loan repayments back has been rejected, reference number: #{claim.reference}")
```
tekin: We use the word "rejected" here in the subject, but "denied" in the body. Should we check this inconsistency with Patrick?

spec/mailers/claim_mailer_spec.rb
pezholio: Yeah, it's a shame we can't check what template is rendered in a similar way to controller specs (even though I know controller specs are no longer The Preferred Way Of Doing Things)

PR #494, opened 2019-10-08
Better bank account capture

pezholio:
This PR adds two new fields to the bank details screen - the name on the bank account (which may differ from the name we get from Notify) and the Building Society roll number (which is optional)

image

image

This also plays back the responses to the user and updates the CSV we send to Cantium with the updated details.

I've also added a data migration to backfill the `banking_name` with the `full_name` for historical applications.


pezholio:
[Better bank account capture](https://trello.com/c/6OhviccR/754-better-bank-account-capture)

app/views/claims/bank_details.html.erb
```diff
@@ -36,6 +44,16 @@
               autocomplete: "off",
               "aria-describedby" => "account-number-hint" %>
```
robbiepaul: Interestingly the draft pattern for bank account number says the number can be between 6 to 8 digits, do we need to update our validation rules?

db/data/20191008102825_copy_full_names_to_banking_names.rb
```diff
@@ -0,0 +1,13 @@
+class CopyFullNamesToBankingNames < ActiveRecord::Migration[5.2]
+  def up
+    Claim.where(banking_name: nil).each do |c|
+      c.update!(banking_name: c.full_name)
+    end
+  end
+
+  def down
+    Claim.all.each do |c|
+      c.banking_name = nil
```
robbiepaul: Do we need to commit this change as an update or save? Maybe even `Claim.update_all(banking_name: nil)`?

app/models/claim.rb
```diff
@@ -91,6 +93,7 @@ class Claim < ApplicationRecord
                             length: {maximum: 256, message: "Email address must be 256 characters or less"},
                             allow_blank: true
 
+  validates :banking_name, on: [:"bank-details", :submit], presence: {message: "Enter the name on your bank account"}
```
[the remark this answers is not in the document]
pezholio: I think it might be a bit OTT to have a regex that matches that pattern, but probably worth limiting to 18 characters

db/data/20191008102825_copy_full_names_to_banking_names.rb
pezholio: Oh good catch, I'm not actually doing anything here am I?

app/models/claim.rb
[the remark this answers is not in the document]
robbiepaul: Yeah, I agree. Although it might be worth checking if Cantium run validations against the value, in which case we might need to (I doubt it though)

PR #495, opened 2019-10-08
A service operator can "prepare" a monthly payroll run and download the CSV

lawrence-dxw:
# Overview

This feature allows a service operator to prepare a payroll file for Cantium. It introduces the concept of a "payroll run", which has an associated list of claims.

When creating a payroll run, the service will select all of the approved claims that are not yet associated with an existing payroll run. The service operator can see a summary of the proposed payroll run before they create it. We make sure that the claims in the created payroll run always match the data shown in the summary. That is, if a new claim is approved whilst the service operator is viewing the summary, then it will not be included in the created payroll run.

We remove the existing link that allows the service operator to download a Cantium payroll CSV of _all_ the approved claims; it has been superseded by this new feature.

There are also a couple of commits for DRYing up some common admin functionality (sign-in) in the request and feature specs.

# What we _haven't_ done

We based the implementation on [wireframes](https://miro.com/app/board/o9J_kw4dZzM=). We decided to leave a few things out:

- We were worried about encouraging a service operator to download a CSV of an existing payroll run; in general we don't want them having personally identifying information on their computers. So we did not add a download link (or a link to the `show` page) to the index view. So the service operator has to download the CSV file at the moment they create the payroll run. Thoughts on this would be appreciated.
- Although we _record_ the DfE Sign-In user ID of the service operator who created the payroll run, we have chosen not to display the user's name, since this would require us figuring out how to fetch that information from DfE Sign-In in an efficient fashion. This seemed like it might be time-consuming, and we are keen to get this feature out ASAP. It seemed to us like the important thing was recording the information for audit purposes.
- We do not perform any validation of the `claim_ids` which are submitted to the "create payroll run" endpoint. This means, for example, that it is possible for two users to be sitting on the "new payroll run" screen at the same time, to both press the "create" button and to hence create two payroll runs. Since we have a very small team of service operators, this seemed like a low-risk scenario, and one that we could fix easily if it were to occur. Opinions appreciated.

# Caveats

- The text is not necessarily final.

# Screenshots

New "payroll" link on the service operator admin front page:
Screenshot_2019-10-08 Teachers claim back your student loan repayments – GOV UK

List of payroll runs (none yet) and button to create a new one:
Screenshot_2019-10-08 Teachers claim back your student loan repayments – GOV UK(2)

Preview of new payroll run before creating it:
Screenshot_2019-10-08 Teachers claim back your student loan repayments – GOV UK(3)

Confirmation that payroll run has been created and link to download the CSV:
Screenshot_2019-10-08 Teachers claim back your student loan repayments – GOV UK(4)

List of payroll runs, showing an example of how we present a payroll run:
Screenshot_2019-10-08 Teachers claim back your student loan repayments – GOV UK(5)


spec/models/payroll_run_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      claim_1 = build(:claim, :approved, eligibility: build(:student_loans_eligibility, :eligible, student_loan_repayment_amount: 1500))
+      claim_2 = build(:claim, :approved, eligibility: build(:student_loans_eligibility, :eligible, student_loan_repayment_amount: 2000))
+
+      payroll_run = PayrollRun.new(claims: [claim_1, claim_2])
+
+      expect(payroll_run.total_award_amount).to eq(3500)
+    end
+  end
+
+  describe ".payrollable_claims" do
```
lawrence-dxw: We deliberated between a class method on `PayrollRun` and a scope on `Claim`. The only reason we decided to put it here is because we worried that, since every feature in the app has _something_ to do with claims, it would be quite easy for `Claim` becoming a dumping ground for a lot of stuff.

app/views/admin/payroll_runs/index.html.erb
```diff
@@ -0,0 +1,39 @@
...
+        <tr class="govuk-table__row">
+          <th scope="col" class="govuk-table__header">Date</th>
+          <th scope="col" class="govuk-table__header">Approved claims</th>
+        </tr>
+      </thead>
+      <tbody class="govuk-table__body">
+        <% @payroll_runs.each do |payroll_run| %>
+          <tr class="govuk-table__row">
+            <th scope="row" class="govuk-table__header"><%= l(payroll_run.created_at.to_date) %></th>
+            <td class="govuk-table__cell"><%= number_with_delimiter(payroll_run.claims.size) %></th>
```
lawrence-dxw: This looks like an N+1 query to me, so I was surprised that Bullet did not catch it. Am I expecting too much of it?

app/controllers/admin/payroll_runs_controller.rb
```diff
@@ -0,0 +1,30 @@
+module Admin
+  class PayrollRunsController < BaseAdminController
+    before_action :ensure_service_operator
+
+    def index
+      @payroll_runs = PayrollRun.all.order(created_at: :desc)
```
tekin: The `.all` here is redundant as the `order` scope will return a relation that defaults to including all the records.

app/controllers/admin/payroll_runs_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+    def show
+      @payroll_run = PayrollRun.find(params[:id])
+
+      respond_to do |format|
+        format.html
+        format.csv do
+          csv = Payroll::ClaimsCsv.new(@payroll_run.claims)
+          send_file csv.file, type: "text/csv", filename: "payroll_data_#{@payroll_run.id}.csv"
```
tekin: I wonder if it would be more useful to us and to Cantium if we include the date of the report in the filename, rather than the ID. That way it's easy to see which month the file corresponds to. The ID is not very useful to anyone other than a developer, and even then we'd need to connect to the database to marry a file against a payroll run record.

app/views/admin/payroll_runs/index.html.erb
```diff
@@ -0,0 +1,39 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <h1 class="govuk-heading-xl">
+      Payroll
+    </h1>
+  </div>
+  <div class="govuk-grid-column-two-thirds">
+    <p class="govuk-body">Make a payroll file containing all the relevant information for approved claims to be paid by a payroll provider.</p>
```
tekin: I think this text should have more of a focus on "preparing" the month's payroll. We're not just generating a file, we're actively assigning a set of claims to be processed as part of the month's payroll. I worry that the way it currently reads it makes it sound like a just building a file the same way the previous download link worked. I'll have a think and suggest some alternative words if you like?

app/views/admin/payroll_runs/index.html.erb
tekin: Yep, it will be, but Bullet will only highlight if n+1 queries are actually being performed. If there is only a single record, then only a single query will be performed and Bullet will be happy. Let's not stress this too much for now. We can use counter caching or something down the road.

spec/requests/admin_payroll_runs_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+  context "when signed in as a service operator" do
+    let(:admin_session_id) { "some_user_id" }
+    before do
+      sign_in_to_admin_with_role(AdminSession::SERVICE_OPERATOR_DFE_SIGN_IN_ROLE_CODE, admin_session_id)
+    end
+
+    describe "admin_payroll_runs#new" do
+      it "uses PayrollRun.payrollable_claims to generate the preview" do
+        claims = build_list(:claim, 3, :approved)
+        expect(PayrollRun).to receive(:payrollable_claims).and_return(claims)
```
tekin: This couples the spec to the internal implementation detail of the controller and the model, making the test more brittle and also making it hard to refactor the code. Can we rewrite this spec to not stub our own code and instead verify something on the response instead? For a start, we should at least verify that we get a successful response. Next is probably verify that something that we'd expect to see if the code is working is present on the page.

db/migrate/20191003083950_create_payroll_runs.rb
```diff
@@ -0,0 +1,8 @@
+class CreatePayrollRuns < ActiveRecord::Migration[5.2]
+  def change
+    create_table :payroll_runs, id: :uuid do |t|
+      t.string :created_by, null: false
+      t.timestamps
```
tekin: One thing that has just occurred to me is that we're not adding indexes on our `created_at` timestamps, when we probably should. Can we add one as part of this migration?

spec/factories/claims.rb
```diff
@@ -34,6 +34,11 @@
       association(:check, factory: [:check, :approved], strategy: :build)
     end
 
+    trait :payrolled do
```
tekin: I'm not sure how I feel about this trait. It's presence invites use, but I'm not sure we'd really want to encourage creating claims with this trait as each one will be associated with its own payroll run, which a) wouldn't be very efficient, and b) wouldn't reflect real-world usage. Maybe it's better to not define this trait and just explicitly set the `payroll_run` inline in the one spec where it is being used?

app/views/admin/payroll_runs/index.html.erb
mec: If you can come up with some content so we can get this in that would be great.

app/controllers/admin/payroll_runs_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+    def index
+      @payroll_runs = PayrollRun.all.order(created_at: :desc)
+    end
+
+    def new
+      @payroll_run = PayrollRun.new(claims: PayrollRun.payrollable_claims)
+    end
+
+    def create
+      payroll_run = PayrollRun.create!(claim_ids: params[:claim_ids], created_by: admin_session.user_id)
```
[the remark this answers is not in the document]
lawrence-dxw: In this case, if the record creation fails, then it isn't as a result of the user's actions, and there is nothing they can do to recover from the error. So I thought that the standard behaviour for an uncaught exception (presumably a 500 response) would be fine.

app/views/admin/payroll_runs/index.html.erb
tekin: For the sake of getting this out in time to run a proper payroll run for October, let's get this out as-is. We can come back and adjust the words as a separate piece of work.

spec/requests/admin_payroll_runs_spec.rb
lawrence-dxw: I've pushed an updated version of this spec, which removes the mocking. I'd like to ask some questions about it if you've got a moment today.

lawrence-dxw: I've pushed a simplified version of the spec, after our conversation.

PR #498, opened 2019-10-09
Admin uses its own layout with with updated navigation

titlescreen:
This set the foundation for navigation within the /admin side of the
service. Admin views will now use a bespoke admin layout which keeps
the familiar look and feel of the user facing side with some tweaks

- admin side now has its own layout
- beta banner is removed as its not really relevant header
- parent service name has been used as the admin side will support more
  than student student
- Existing links have been refactored to use the new navigation bar
- claims page has a back link

image

image

### DIFF
--- a/app/assets/stylesheets/components/navigation.scss
+++ b/app/assets/stylesheets/components/navigation.scss
@@ -1,0 +1,103 @@
+// Typography
+$small-font-size: 0.9em;
+$large-font-size: 1.2em;
+$bold-font-weight: 700;
+
+// Colours
+$warning-color: govuk-colour("blue");
+
+// Breakpoints
+$large-screen: 1300px;
+$big-screen: 1024px;
+$medium-screen: 900px;
+$small-screen: 768px;
+$smaller-screen: 520px;
+$tiny-screen: 480px;
+
+// Borders
+$thick-border: 5px;
+
+.app-navigation {
+  $navigation-height: 50px;
+  $navigation-padding: 20px;
+  border-bottom: 1px solid govuk-colour("mid-grey");
+  font-weight: bold;
+  background-color: #f8f8f8;
+  line-height: 25px;
+
+  @include govuk-clearfix;
+
+  &__list {
+    margin: 0;
+    padding: 0;
+    list-style: none;
+    font-size: 0.85em;
+
+    @include govuk-media-query($from: tablet) {
+      font-size: 1em;
+    }
+
+    & > li {
+      box-sizing: border-box;
+      padding: $navigation-padding 0;
+      float: left;
+      -moz-box-sizing: border-box;
+      -webkit-box-sizing: border-box;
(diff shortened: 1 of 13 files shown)


app/views/admin/claims/index.html.erb
```diff
@@ -1,5 +1,6 @@
+<a href="/admin" class="govuk-back-link">Back</a>
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-full">
```
tekin: Looks like an indentation regression here.

app/views/layouts/admin.html.erb
```diff
@@ -0,0 +1,140 @@
+<!DOCTYPE html>
+<html lang="en" class="govuk-template app-html-class">
+  <head>
+    <title>
+      <%= content_for(:page_title) || "#{t("student_loans.journey_name")} – GOV.UK" %>
```
tekin:
```suggestion
      <%= t("service_name") – GOV.UK" %>
```

app/views/layouts/admin.html.erb
```diff
@@ -0,0 +1,140 @@
...
+              </svg>
+              <span class="govuk-header__logotype-text">
+                GOV.UK
+              </span>
+            </span>
+          </a>
+        </div>
+        <div class="govuk-header__content">
+
+          <%= link_to t("service_name"), root_path, class: "govuk-header__link govuk-header__link--service-name" %>
```
tekin: If you change line 57 in the routes.rb file to this:

```ruby
get "/", to: "page#index", as: :root
```
Then you'll be able to link to the root of admin, rather than to the public-facing service:
```suggestion
          <%= link_to t("service_name"), admin_root_path, class: "govuk-header__link govuk-header__link--service-name" %>
```

app/views/admin/claims/index.html.erb
```diff
@@ -1,5 +1,6 @@
+<a href="/admin" class="govuk-back-link">Back</a>
```
tekin: If you update the `routes.rb` file as described in the comment below, you can change this to:
```suggestion
<%= link_to "Back", admin_root_path, class: "govuk-back-link" %>
```

app/assets/stylesheets/components/navigation.scss
```diff
@@ -0,0 +1,103 @@
...
+
+.app-navigation {
+  $navigation-height: 50px;
+  $navigation-padding: 20px;
+  border-bottom: 1px solid govuk-colour("mid-grey");
+  font-weight: bold;
+  background-color: #f8f8f8;
+  line-height: 25px;
+
+  @include clearfix;
```
mec: Could you use `govuk-clearfix` here to avoid creating our own mixin?

titlescreen:
I think I've addressed everything, thanks for the feedback.

PR #499, opened 2019-10-09
Allow service to be put into maintenance mode

pezholio:
This allows the service to be put into maintenance mode by setting an environment variable. Based on [prior art from the Schools Experience team](https://github.com/DFE-Digital/schools-experience/pull/987)

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,5 +199,14 @@ file in `azure/resource_groups/app/parameters/{environment_name}.template.json`
 
 Then commit the changes, open a pull request, and get it merged in.
 
+Alternatively you can run the following script to immediately put the
+application into maintenance mode:
+
+```bash
+bin/set-maintenance-mode ENVIRONMENT "OPTIONAL_AVAILABILITY_MESSAGE"
+```
+
+Bear in mind here, any subsequent deploys will unset the environment variables.
+
 To restore the service, unset the environment variables, and get the changes
 deployed. The service will become operational again on the next deploy.
--- a/README.md
+++ b/README.md
@@ -180,9 +180,24 @@ possible to put ther service into maintenance mode.
 To do this, we set the environment variable `MAINTENANCE_MODE` to a value of
 `1`. If you know when the service will be operational again, you can also set a
 `MAINTENANCE_MODE_AVAILABILITY_MESSAGE` environment variable with a
-human-readable value of when to expect the service to be operational again, for
-example `the service will be available from 2pm today` or
+human-readable value of when to expect the service to be operational again as a
+full sentence, for example `The service will be available from 2pm today.` or
 `Private beta has ended. You will be able to use the service again in November`.
 
-To restore the service, unset the environment variables, and the service will
-become operational again.
+To do this, it's best to update the variables in the relevant Azure parameter
+file in `azure/resource_groups/app/parameters/{environment_name}.template.json`
+, and change the empty strings to real variables, i.e.:
+
+```json
+    "MAINTENANCE_MODE": {
+      "value": "1"
+    },
+    "MAINTENANCE_MODE_AVAILABILITY_MESSAGE": {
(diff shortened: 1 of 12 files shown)


pezholio:
[Maintenance mode](https://trello.com/c/fHren2bL/800-maintenance-mode)

README.md
```diff
@@ -170,3 +170,12 @@ The service architecture is currently defined [on confluence].
 [on confluence]:
   https://dfedigital.atlassian.net/wiki/spaces/TP/pages/1049559041/Service+Architecture
 [openjdk]: https://adoptopenjdk.net/
+
+## Putting the application into maintenance mode
+
+If we need to take the service offline for any reason (for example, we're
+investigating and issue, or we want to deploy a journey-breaking change), it's
```
mec:
```suggestion
investigating an issue, or we want to deploy a journey-breaking change), it's
```

spec/routes/routes_spec.rb
```diff
@@ -14,4 +14,32 @@
...
   end
+
+  describe "service is unavailable" do
+    before do
+      allow(Rails.application.config.x).to receive(:maintenance_mode) { maintenance_mode }
+      Rails.application.reload_routes!
+    end
+
+    after do
+      allow(Rails.application.config.x).to receive(:maintenance_mode).and_call_original
```
lawrence-dxw: RSpec tears down the mocks at the end of each example, so does that make this unnecessary?

tekin:
Are we missing the last item on the acceptance criteria, i.e. being able to set a custom message on the page. In this case, I reckon making the "You will be able to use the service later today." sentence customisable so we can tell users exactly when we think the service will be back up. See the [pattern in the design system](https://design-system.service.gov.uk/patterns/service-unavailable-pages/)

tekin:
I'm not sure what you mean by exact correct format. I was thinking a sentence as a string? I don't think a second env var with a default that we fall back to in the template will add that much extra complication. I would prefer that over telling users we are down until later in the day, when that may be utterly and knowably false.

spec/routes/routes_spec.rb
pezholio: Yeah, you're right. That was a hangover from when I was having trouble stubbing `ENV`. I'll bin it off

pezholio: No, wait, ignore that, I do have remove the stubs here, because the `after` block is still in the context of the tests, so I still need to remove the stub before reloading the routes, as the routes I've loaded in my `before` block will still remain in memory (as routes don't get reloaded at the start of every test)

spec/requests/maintenance_mode_spec.rb
```diff
@@ -0,0 +1,40 @@
+require "rails_helper"
+
+RSpec.describe "Maintenance Mode", type: :request do
+  context "when MAINTENANCE_MODE is set" do
+    before do
+      allow(Rails.application.config.x).to receive(:maintenance_mode) { true }
```
tekin: A common pattern for a temporary overriding a configuration value without stubbing magic is something like this:
```suggestion
  @original_maintenance_mode_value = Rails.application.config.x.maintenance_mode
  Rails.application.config.x.maintenance_mode = true
```
and then you can reset in the `after` call:
```
after do
  Rails.application.config.x.maintenance_mode = @original_maintenance_mode_value
end
```

config/initializers/maintenance_mode.rb
```diff
@@ -0,0 +1,2 @@
+Rails.application.config.x.maintenance_mode = ENV["MAINTENANCE_MODE"].present?
```
tekin: I've just looked up the docs for custom configuration and it looks like we only need to use `config.x` if we plan to nest configuration. In which case can we use the simpler `.config` instead?

robbiepaul:
ARM template changes look good to me 👍

PR #469, opened 2019-10-01
Add email address to request certs script

mec:
The script we use to generate the VSP certificate signing request should
include the group email address.

It would also be helpful if they were put in a directory that is git ignored so 
developers like myself do not accidentally commit them!

PR #510, opened 2019-10-14
Bump Verify signing and encryption keys to 6

mec:
This is the first PR to update the intergrtaion (development) certificates for Verify.

We update the signing key and we run two versions of the encryption key until verify have removed the old one. At that point there will be another PR to make the new key into the primary key and remove the old one.

#469 can also be merged once we have confirmed that these new keys are working correctly in the development environment.

## What happend to keys v3-5?
v3 was the key that we were never notified of due to a typo in the email address used.
v4-5 were not used as I was updating the script (#469) to include the email address and needed a way to test the generation of the CSRs.

### DIFF
--- a/azure/resource_groups/app/parameters/development.template.json
+++ b/azure/resource_groups/app/parameters/development.template.json
@@ -203,7 +203,7 @@
         "keyVault": {
           "id": "${keyVaultId}"
         },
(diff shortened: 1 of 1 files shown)


mec:
[Rotate VSP certificates](https://trello.com/c/z9s8AbbG/685-rotate-vsp-certificates)

PR #512, opened 2019-10-14
Exclude /admin from maintenance mode

tekin:
Whilst we want to be able to put the service into maintenance mode and prevent users from making claims during scheduled maintenance, or when the service is taken offline outside of the claim window, we still need service operators to be able to access the back-office for the service so they can continue to check claims and provide support to existing claimants.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -12,9 +12,11 @@ Rails.application.routes.draw do
   # setup a simple healthcheck endpoint for monitoring purposes
   get "/healthcheck", to: proc { [200, {}, ["OK"]] }
 
+  # Catch-all for when the service has been placed in maintenance mode.
+  # Excludes /admin so Service Operators can continue to check claims.
   if Rails.application.config.maintenance_mode
     root "static_pages#maintenance"
-    match "*path", to: redirect("/"), via: :all
+    match "*path", to: redirect("/"), via: :all, constraints: lambda { |req| !%r{^/admin($|/)}.match?(req.path) }
   else
     root "static_pages#start_page"
   end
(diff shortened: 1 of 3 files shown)


config/routes.rb
```diff
@@ -12,9 +12,11 @@
   # setup a simple healthcheck endpoint for monitoring purposes
   get "/healthcheck", to: proc { [200, {}, ["OK"]] }
 
+  # Catch-all for when the service has been placed in maintenance mode.
+  # Excludes /admin so Service Operators can continue to check claims.
   if Rails.application.config.maintenance_mode
     root "static_pages#maintenance"
-    match "*path", to: redirect("/"), via: :all
+    match "*path", to: redirect("/"), via: :all, constraints: lambda { |req| !req.path.starts_with?("/admin/") }
```
erbridge: Does this include `/admin` itself?

tekin: Good catch...

tekin: Updated.

config/routes.rb
```diff
@@ -12,9 +12,11 @@
   # setup a simple healthcheck endpoint for monitoring purposes
   get "/healthcheck", to: proc { [200, {}, ["OK"]] }
 
+  # Catch-all for when the service has been placed in maintenance mode.
+  # Excludes /admin so Service Operators can continue to check claims.
   if Rails.application.config.maintenance_mode
     root "static_pages#maintenance"
-    match "*path", to: redirect("/"), via: :all
+    match "*path", to: redirect("/"), via: :all, constraints: lambda { |req| !%r{(^/admin$)|(^/admin/)}.match?(req.path) }
```
erbridge:
```suggestion
    match "*path", to: redirect("/"), via: :all, constraints: lambda { |req| !%r{^/admin($|/)}.match?(req.path) }
```

I think this is clearer, but certainly not necessary.

tekin: I like it 👍

PR #513, opened 2019-10-14
Service operators can easily identify claims that are close to their deadline

robbiepaul:
We tell users they will hear back on the outcome of their claim "within 6 weeks". To help Service Operators ensure they check claims within the 6 weeks, they should be able to easily see when a claim is nearing the end of this period, and when it has passed it.

## Screenshots

Screenshot 2019-10-14 at 14 44 29
Screenshot 2019-10-14 at 14 44 38
Screenshot 2019-10-14 at 14 44 48

### DIFF
--- a/app/helpers/admin/claims_helper.rb
+++ b/app/helpers/admin/claims_helper.rb
@@ -35,6 +35,7 @@ module Admin
       [
         [t("admin.started_at"), l(claim.created_at)],
         [t("admin.submitted_at"), l(claim.submitted_at)],
+        [t("admin.check_deadline"), [l(claim.check_deadline_date), check_deadline_warning(claim)].compact.join.html_safe],
       ]
     end
 
--- a/app/helpers/admin/claims_helper.rb
+++ b/app/helpers/admin/claims_helper.rb
@@ -50,5 +50,19 @@ module Admin
       ].join(" ").html_safe
       sanitize(html, tags: %w[span a], attributes: %w[href class])
     end
+
+    def check_deadline_warning(claim)
+      days_until_check_deadline = days_between(Date.today, claim.check_deadline_date)
+      return if days_until_check_deadline.days > Claim::CHECK_DEADLINE_WARNING_POINT
+
+      check_deadline_warning_class = days_until_check_deadline < 0 ? "tag--alert" : "tag--warning"
+      content_tag(:strong, pluralize(days_until_check_deadline, "day"), class: "govuk-tag #{check_deadline_warning_class}")
+    end
+
+    private
+
+    def days_between(first_date, second_date)
+      (second_date - first_date).to_i
+    end
   end
 end
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -144,6 +144,10 @@ class Claim < ApplicationRecord
     %w[male female].exclude?(payroll_gender)
   end
 
+  def check_deadline_date
+    (submitted_at + CHECK_DEADLINE).to_date
+  end
+
   def address(seperator = ", ")
     Claim::ADDRESS_ATTRIBUTES.map { |attr| send(attr) }.reject(&:blank?).join(seperator)
   end
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -39,6 +39,8 @@ class Claim < ApplicationRecord
     building_society_roll_number: true,
     payroll_run_id: false,
   }.freeze
+  CHECK_DEADLINE = 6.weeks
+  CHECK_DEADLINE_WARNING_POINT = 2.weeks
 
   enum student_loan_country: StudentLoans::COUNTRIES
   enum student_loan_start_date: StudentLoans::COURSE_START_DATES
@@ -115,6 +117,8 @@ class Claim < ApplicationRecord
   scope :submitted, -> { where.not(submitted_at: nil) }
   scope :awaiting_checking, -> { submitted.left_outer_joins(:check).where(checks: {claim_id: nil}) }
   scope :approved, -> { joins(:check).where("checks.result" => :approved) }
+  scope :approaching_check_deadline, -> { awaiting_checking.where("submitted_at < ? AND submitted_at > ?", CHECK_DEADLINE.ago + CHECK_DEADLINE_WARNING_POINT, CHECK_DEADLINE.ago) }
+  scope :passed_check_deadline, -> { awaiting_checking.where("submitted_at < ?", CHECK_DEADLINE.ago) }
 
   delegate :award_amount, to: :eligibility
 
--- a/app/views/admin/claims/index.html.erb
+++ b/app/views/admin/claims/index.html.erb
@@ -6,6 +6,7 @@
     </h1>
 
     <% if @claims.any? %>
+      <% claims_with_warning = Claim.approaching_check_deadline.count + Claim.passed_check_deadline.count %>
 
       <h2 class="govuk-heading-m"><%= pluralize(@claims.count, "claim") %> awaiting checking</h2>
       <table class="govuk-table">
@@ -13,7 +14,9 @@
           <tr class="govuk-table__row">
             <th scope="col" class="govuk-table__header">Reference</th>
             <th scope="col" class="govuk-table__header">Applicant Name</th>
-            <th scope="col" class="govuk-table__header">Check warning</th>
+            <% if claims_with_warning.nonzero? %>
+              <th scope="col" class="govuk-table__header">Check warning</th>
+            <% end %>
             <th scope="col" class="govuk-table__header">Check deadline</th>
             <th scope="col" class="govuk-table__header"></th>
           </tr>
@@ -23,7 +26,9 @@
             <tr class="govuk-table__row">
               <th scope="row" class="govuk-table__header"><%= claim.reference %></th>
               <td class="govuk-table__cell"><%= claim.full_name %></td>
-              <td class="govuk-table__cell"><%= check_deadline_warning(claim) %></td>
+              <% if claims_with_warning.nonzero? %>
+                <td class="govuk-table__cell"><%= check_deadline_warning(claim) %></td>
+              <% end %>
               <td class="govuk-table__cell"><%= l(claim.check_deadline_date) %></td>
               <td class="govuk-table__cell"><%= link_to 'View claim', admin_claim_path(claim), class: "govuk-link" %></td>
             </tr>
--- a/app/views/admin/claims/index.html.erb
+++ b/app/views/admin/claims/index.html.erb
@@ -8,12 +8,13 @@
     <% if @claims.any? %>
 
       <h2 class="govuk-heading-m"><%= pluralize(@claims.count, "claim") %> awaiting checking</h2>
-
       <table class="govuk-table">
         <thead class="govuk-table__head">
           <tr class="govuk-table__row">
             <th scope="col" class="govuk-table__header">Reference</th>
             <th scope="col" class="govuk-table__header">Applicant Name</th>
+            <th scope="col" class="govuk-table__header">Check warning</th>
+            <th scope="col" class="govuk-table__header">Check deadline</th>
             <th scope="col" class="govuk-table__header"></th>
           </tr>
         </thead>
@@ -22,6 +23,8 @@
             <tr class="govuk-table__row">
               <th scope="row" class="govuk-table__header"><%= claim.reference %></th>
               <td class="govuk-table__cell"><%= claim.full_name %></td>
+              <td class="govuk-table__cell"><%= check_deadline_warning(claim) %></td>
+              <td class="govuk-table__cell"><%= l(claim.check_deadline_date) %></td>
               <td class="govuk-table__cell"><%= link_to 'View claim', admin_claim_path(claim), class: "govuk-link" %></td>
             </tr>
           <% end %>
--- a/app/views/admin/page/index.html.erb
+++ b/app/views/admin/page/index.html.erb
@@ -1,17 +1,24 @@
 <div class="govuk-grid-row">
 
-  <div class="govuk-grid-column-two-thirds">
-    &nbsp;
-  </div>
-
-  <div class="govuk-grid-column-one-third">
-    <h2 class="govuk-heading-m">Total claims received</h2>
-    <p class="govuk-body govuk-!-font-size-36"><%= Claim.submitted.count %></p>
-
-    <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
-
+  <div class="govuk-grid-column-one-half">
     <h2 class="govuk-heading-m">Claims awaiting checking</h2>
     <p class="govuk-body govuk-!-font-size-36"><%= Claim.awaiting_checking.count %></p>
+    <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
+
+    <h2 class="govuk-heading-m">Total claims received</h2>
+    <p class="govuk-body govuk-!-font-size-36"><%= Claim.submitted.count %></p>
+    <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
   </div>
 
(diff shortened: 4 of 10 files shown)


robbiepaul:
[Service operators can easily identify claims that are close to the SLA for claim checking](https://trello.com/c/OTJs4W1P/813-service-operators-can-easily-identify-claims-that-are-close-to-the-sla-for-claim-checking)

erbridge:
Does the deadline column in the list view need a heading to give it context?

app/models/claim.rb
```diff
@@ -39,6 +39,7 @@ class Claim < ApplicationRecord
     building_society_roll_number: true,
     payroll_run_id: false,
   }.freeze
+  DEADLINE = 6.weeks
```
erbridge: Think we can derive the deadline text we display to claimants from this, too?

spec/helpers/admin/claims_helper_spec.rb
```diff
@@ -94,8 +94,21 @@
       expect(helper.admin_submission_details(claim)).to eq([
         [I18n.t("admin.started_at"), l(claim.created_at)],
         [I18n.t("admin.submitted_at"), l(claim.submitted_at)],
+        [I18n.t("admin.deadline"), l(claim.deadline_date)],
       ])
     end
+
+    context "when the claim is approaching it's deadline" do
```
erbridge:
```suggestion
    context "when the claim is approaching its deadline" do
```

robbiepaul:
Yeah I was on the fence about that, I guess because it doesn't always show I didn't want to have it as `Weekdays until deadline` but perhaps something like `Deadline warning` would work

erbridge:
Or only show the column at all if there are any? Or modify the tect on the button to explain what it refers to?

app/helpers/admin/claims_helper.rb
```diff
@@ -50,5 +51,21 @@ def display_school(school)
...
+      content_tag(
+        :strong,
+        "#{claim.passed_deadline? ? "- " : ""}#{pluralize(weekdays_between(Date.today, claim.deadline_date), "day")}",
+        class: "govuk-tag claim-deadline__#{claim.passed_deadline? ? "passed" : "approaching"}"
+      )
+    end
+
+    private
+
+    def weekdays_between(first_date, second_date)
```
tekin: What if this method was updated to return a negative value when the second date was before the first date? Would that simplify the `deadline_warning` method?

app/models/claim.rb
tekin: `DEADLINE` is a very vague term. Can we give it a more specific name? `CHECKING_SLA_WINDOW` or something?

app/views/admin/claims/index.html.erb
```diff
@@ -15,13 +15,17 @@
             <th scope="col" class="govuk-table__header">Reference</th>
             <th scope="col" class="govuk-table__header">Applicant Name</th>
             <th scope="col" class="govuk-table__header"></th>
+            <th scope="col" class="govuk-table__header">Deadline</th>
```
tekin:
```suggestion
            <th scope="col" class="govuk-table__header">Checking deadline</th>
```

app/models/claim.rb
```diff
@@ -140,6 +141,22 @@ def payroll_gender_missing?
     %w[male female].exclude?(payroll_gender)
   end
 
+  def awaiting_check?
+    submitted? && check.nil?
+  end
+
+  def approaching_deadline?
```
tekin: I can't help but wonder, what if we had a `days_until_checking_deadline` that returns either a positive or negative number and used that to determine where the claim is on the schedule and also to generate warning markup? Would that make things simpler?

app/models/claim.rb
```diff
@@ -140,6 +141,22 @@ def payroll_gender_missing?
...
+    awaiting_check? && Time.zone.now > (deadline_date - 2.weeks) && !passed_deadline?
+  end
+
+  def passed_deadline?
+    awaiting_check? && Time.zone.now > deadline_date
+  end
+
+  def deadline_date
```
tekin: As with the constant above, would be good to give these methods more specific names, e.g. `checking_deadline_date`

app/models/claim.rb
erbridge: Is there not a `to_weeks` method or something?

app/views/admin/page/index.html.erb
```diff
@@ -1,17 +1,24 @@
...
     <p class="govuk-body govuk-!-font-size-36"><%= Claim.submitted.count %></p>
-
     <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
+  </div>
 
-    <h2 class="govuk-heading-m">Claims awaiting checking</h2>
-    <p class="govuk-body govuk-!-font-size-36"><%= Claim.awaiting_checking.count %></p>
+  <div class="govuk-grid-column-one-half">
+    <h2 class="govuk-heading-m">Claims approaching deadline</h2>
+    <p class="govuk-body govuk-!-font-size-36"><%= @claims_awaiting_checking.select(&:approaching_deadline?).count %></p>
```
tekin: This is going to load all the claims into memory to perform the count. Is there a way to calculate this number, and the number below using a query instead and set them as instance variables in the controller action?

app/controllers/admin/page_controller.rb
```diff
@@ -1,6 +1,7 @@
 module Admin
   class PageController < BaseAdminController
     def index
+      @claims_awaiting_checking = Claim.includes(:check).awaiting_checking
```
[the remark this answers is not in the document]
robbiepaul: Yeah I thought that too, but I don't think it does the eager loading.

Before `.includes(:check)`:

```
11:38:18 web.1    |   Claim Load (26.3ms)  SELECT "claims".* FROM "claims" LEFT OUTER JOIN "checks" ON "checks"."claim_id" = "claims"."id" WHERE "claims"."submitted_at" IS NOT NULL AND "checks"."claim_id" IS NULL
11:38:18 web.1    |   ↳ app/views/admin/page/index.html.erb:15
11:38:18 web.1    |   Check Load (24.7ms)  SELECT  "checks".* FROM "checks" WHERE "checks"."claim_id" = $1 LIMIT $2  [["claim_id", "5a2318f7-b77e-48ee-9b92-bf5b95725f9f"], ["LIMIT", 1]]
11:38:18 web.1    |   ↳ app/models/claim.rb:145
11:38:18 web.1    |   Check Load (28.0ms)  SELECT  "checks".* FROM "checks" WHERE "checks"."claim_id" = $1 LIMIT $2  [["claim_id", "80a4a367-16db-4e41-939b-d46da911980d"], ["LIMIT", 1]]
11:38:18 web.1    |   ↳ app/models/claim.rb:145
11:38:18 web.1    |   Check Load (26.5ms)  SELECT  "checks".* FROM "checks" WHERE "checks"."claim_id" = $1 LIMIT $2  [["claim_id", "31041254-edfa-4b07-b024-a87f3f888bee"], ["LIMIT", 1]]
11:38:18 web.1    |   ↳ app/models/claim.rb:145
```

After `.includes(:check)`:
```
11:42:50 web.1    |   ↳ app/views/admin/page/index.html.erb:9
11:42:50 web.1    |   Claim Load (26.4ms)  SELECT "claims".* FROM "claims" LEFT OUTER JOIN "checks" ON "checks"."claim_id" = "claims"."id" WHERE "claims"."submitted_at" IS NOT NULL AND "checks"."claim_id" IS NULL
11:42:50 web.1    |   ↳ app/views/admin/page/index.html.erb:15
11:42:50 web.1    |   Check Load (27.5ms)  SELECT "checks".* FROM "checks" WHERE "checks"."claim_id" IN ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11) .....
```

However saying that, I'm going to be changing this action so it doesn't load all the claims in memory anyway

app/models/claim.rb
robbiepaul: Yeah it would make it a lot simpler. It would mean that I can remove three methods from the claim model and simplify the `deadline_warning` helper at the same time

app/helpers/admin/claims_helper.rb
```diff
@@ -50,5 +51,21 @@ def display_school(school)
       ].join(" ").html_safe
       sanitize(html, tags: %w[span a], attributes: %w[href class])
     end
+
+    def deadline_warning(claim)
```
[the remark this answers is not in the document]
robbiepaul: I think this method is a lot simpler now that I've updated the code to have a `weekdays_until_checking_deadline`, it still has a guard clause to filter out claims still have more than 10 days left to be checked though.

robbiepaul:
@tekin Ready for another look

app/assets/stylesheets/components/claim_deadline.scss
```diff
@@ -0,0 +1,25 @@
+@import "govuk-frontend-rails";
+
+%claim-deadline {
+  text-transform: lowercase;
+  letter-spacing: 0;
+  min-width: 4em;
+  text-align: center;
+}
```
erbridge: Is this a generic style for tags (or whatever we're calling those things)? Does it make sense to make it a generic component that we import?

app/models/claim.rb
```diff
@@ -115,6 +116,8 @@ class Claim < ApplicationRecord
   scope :submitted, -> { where.not(submitted_at: nil) }
   scope :awaiting_checking, -> { submitted.left_outer_joins(:check).where(checks: {claim_id: nil}) }
   scope :approved, -> { joins(:check).where("checks.result" => :approved) }
+  scope :approaching_checking_deadline, -> { awaiting_checking.where("submitted_at < ?", CHECKING_DEADLINE.ago + 2.weeks).where.not("submitted_at < ?", CHECKING_DEADLINE.ago) }
```
erbridge: Could we pull the `+ 2.weeks` out into another constant? What we consider to be "approaching" is a bit hidden.

app/assets/stylesheets/components/claim_deadline.scss
```diff
@@ -0,0 +1,25 @@
+@import "govuk-frontend-rails";
+
+%claim-deadline {
+  text-transform: lowercase;
```
erbridge: Probably better to make sure we're actually using lowercase text so that copying it (and I guess screen readers?) don't have the raw text with uppercase letters.

app/helpers/admin/claims_helper.rb
```diff
@@ -50,5 +51,22 @@ def display_school(school)
       ].join(" ").html_safe
       sanitize(html, tags: %w[span a], attributes: %w[href class])
     end
+
+    def checking_deadline_warning(claim)
+      weekdays_until_checking_deadline = weekdays_between(Date.today, claim.checking_deadline_date)
+      return if weekdays_until_checking_deadline > 10
```
erbridge: Can we pull this `10` out into a constant so it's easier to follow the purpose of it (and see what it is)?

app/helpers/admin/claims_helper.rb
```diff
@@ -50,5 +51,22 @@ def display_school(school)
...
+    end
+
+    private
+
+    def weekdays_between(first_date, second_date)
+      dates_in_order = [first_date, second_date].sort
+      weekdays = (dates_in_order.first..dates_in_order.last).select { |date| [0, 6].exclude?(date.wday) }.count
+      weekdays *= -1 if first_date > second_date
+      weekdays
+    end
```
erbridge: I'm surprised Ruby doesn't have a `(date_a - date_b).to_days`.

spec/helpers/admin/claims_helper_spec.rb
```diff
@@ -114,4 +127,30 @@
       )
     end
   end
+
+  describe "#checking_deadline_warning" do
+    subject { helper.checking_deadline_warning(claim) }
+    before { travel_to Time.zone.local(2019, 10, 11, 7, 0, 0) }
```
erbridge: Is it important that it's a given date and not just frozen?

robbiepaul: Hmm yea, I guess the spec could be a little clearer. It needs to be frozen because it doesn't count weekends. I'll add another context

app/assets/stylesheets/components/claim_deadline.scss
robbiepaul: Yeah the text is actually lowercase but the `govuk-tag` class makes it appear uppercase. I've change our override to be `text-transform: none`

app/helpers/admin/claims_helper.rb
```diff
@@ -50,5 +51,22 @@ def display_school(school)
...
+
+      checking_deadline_warning_class = weekdays_until_checking_deadline < 0 ? "tag--alert" : "tag--warning"
+      content_tag(:strong, pluralize(weekdays_until_checking_deadline, "day"), class: "govuk-tag #{checking_deadline_warning_class}")
+    end
+
+    private
+
+    def weekdays_between(first_date, second_date)
+      dates_in_order = [first_date, second_date].sort
+      weekdays = (dates_in_order.first..dates_in_order.last).select { |date| [0, 6].exclude?(date.wday) }.count
```
tekin: Reckon we can simplify this line with [`Date.#on_weekday?`](https://api.rubyonrails.org/v5.1/classes/DateAndTime/Calculations.html#method-i-on_weekday-3F):

```suggestion
      weekdays = (dates_in_order.first..dates_in_order.last).count { |date| date.on_weekday? }
```

app/assets/stylesheets/components/claim_deadline.scss
erbridge: Aha! Makes sense.

app/helpers/admin/claims_helper.rb
tekin: Yeah, this value and the `2.weeks` in the scope are effectively the same thing, right?

robbiepaul:
- Renamed 'checking deadline' to 'check deadline'
- Renamed `CHECK_DEADLINE_WARNING_POINT`
- Warning shows number of calendar days now, not weekdays

spec/features/admin_stats_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    @submitted_claims = create_list(:claim, 6, :submitted)
+    @approved_claims = create_list(:claim, 4, :approved, submitted_at: 10.weeks.ago)
+    @unfinished_claims = create_list(:claim, 1, :submittable)
+    @claims_approaching_deadline = create_list(:claim, 2, :submitted, submitted_at: 5.weeks.ago)
+    @claims_passed_deadline = create_list(:claim, 1, :submitted, submitted_at: 10.weeks.ago)
+    sign_in_to_admin_with_role(AdminSession::SERVICE_OPERATOR_DFE_SIGN_IN_ROLE_CODE)
+    visit admin_root_path
+  end
+
+  scenario "Service operator is shown how many claims have been submitted" do
```
tekin: Can we combine these four scenarios into a single "Service operator is shown claim stats"? Each scenario has exactly the same expensive setup and is checking the same page. We only need to create the records, log the user in and render the page once to verify the expected behaviour. That'll make reduce the hit on our test suite run time a chunk.

app/models/claim.rb
```diff
@@ -115,6 +117,8 @@ class Claim < ApplicationRecord
   scope :submitted, -> { where.not(submitted_at: nil) }
   scope :awaiting_checking, -> { submitted.left_outer_joins(:check).where(checks: {claim_id: nil}) }
   scope :approved, -> { joins(:check).where("checks.result" => :approved) }
+  scope :approaching_check_deadline, -> { awaiting_checking.where("submitted_at < ?", CHECK_DEADLINE.ago + CHECK_DEADLINE_WARNING_POINT).where.not("submitted_at < ?", CHECK_DEADLINE.ago) }
```
tekin: This could probably be expressed more simply with a "submitted_at < ? AND submitted_at > ?" but not a blocker.

PR #515, opened 2019-10-15
Update brakeman ignore list with payroll filename

mec:
We ignore this brakeman warning, which a previous commit changed when
the date was added to the filename.

This adds the new fingerprint to the ignore file so we are not
warned about it.


tekin:
This has been covered in https://github.com/DFE-Digital/dfe-teachers-payment-service/pull/514

PR #520, opened 2019-10-16
Change the payroll start and end dates

mec:
The start and end dates in the payroll export is the fictional period a
claimant was employed in order for DfE to pay them.

We did not want the start date to be before the 6th of a given month as
the tax year runs from the 6 April until the 5 May the following year.

By setting the start date to the second Monday in the month and the end
date to the following Sunday, we avoid this issue.

I debated with myself about the spec for this for ages and ended up deciding 
I wanted a context to show what is being tested is the day of the 1st of the 
current month as that has the most impact on the code working correctly 
and wasn't really shown without consulting a calendar – appreciate thoughts 
on that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- Payroll export start date is the second Monday of the month and the end date
+  is the following Sunday
+
 ## [Release 018] - 2019-10-16
 
 - Run schema and data migrations at the same time
--- a/app/models/payroll/claim_csv_row.rb
+++ b/app/models/payroll/claim_csv_row.rb
@@ -41,15 +41,17 @@ module Payroll
     end
 
     def start_date
-      start_of_month.strftime(DATE_FORMAT)
+      second_monday_of_month.strftime(DATE_FORMAT)
     end
 
     def end_date
-      (start_of_month + 7.days).strftime(DATE_FORMAT)
+      second_monday_of_month.end_of_week.strftime(DATE_FORMAT)
     end
 
-    def start_of_month
-      Date.today.at_beginning_of_month
+    def second_monday_of_month
+      day = Date.today.at_beginning_of_month
+      day += 1.days until day.monday?
+      day.next_week
     end
 
     def date_of_birth
--- a/spec/models/payroll/claim_csv_row_spec.rb
+++ b/spec/models/payroll/claim_csv_row_spec.rb
@@ -6,7 +6,6 @@ RSpec.describe Payroll::ClaimCsvRow do
 
   describe "to_s" do
     let(:row) { CSV.parse(subject.to_s).first }
-    let(:start_of_month) { Date.today.at_beginning_of_month }
 
     let(:claim) do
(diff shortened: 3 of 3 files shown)


mec:
[Update payroll file based on amended schedule](https://trello.com/c/5I81vfkl/820-update-payroll-file-based-on-amended-schedule)

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- Payroll export start and end dates are the second Monday and Sunday in the
+  month
```
erbridge: Just a note that if the month starts on a Sunday, the second Sunday is before the second Monday.

app/models/payroll/claim_csv_row.rb
```diff
@@ -41,15 +41,17 @@ def payroll_gender
...
     def end_date
-      (start_of_month + 7.days).strftime(DATE_FORMAT)
+      second_monday_of_month.end_of_week.strftime(DATE_FORMAT)
     end
 
-    def start_of_month
-      Date.today.at_beginning_of_month
+    def second_monday_of_month
+      day = Date.today.at_beginning_of_month
+      day += 1.days until day.monday?
```
erbridge: Not sure how `until` works precisely. If the month starts on a Monday, is `day` at this point that Monday or the Monday after?

erbridge: Nevermind. I see the spec that checks this case.

spec/models/payroll/claim_csv_row_spec.rb
```diff
@@ -23,38 +22,62 @@
...
+          claim.building_society_roll_number,
+          "Student Loans",
+          claim.eligibility.student_loan_repayment_amount.to_s,
+          claim.reference,
+        ])
+      end
+    end
+
+    describe "start and end dates" do
+      context "when the first on the month is a Friday" do
```
tekin:
```suggestion
      context "when the first of the month is a Friday" do
```

spec/models/payroll/claim_csv_row_spec.rb
```diff
@@ -23,38 +22,62 @@
...
+        it "returns the date of the second Monday and Sunday" do
+          travel_to Date.parse "20 February 2019" do
+            row = CSV.parse(subject.to_s).first
+            expect(row[6]).to eq "20190211"
+            expect(row[7]).to eq "20190217"
+          end
+        end
+      end
+
+      context "when the first on the month is a Monday" do
```
tekin:
```suggestion
      context "when the first of the month is a Monday" do
```

CHANGELOG.md
mec: Yeah, it's not clear is it! I wasn't even sure something like this goes in the changelog? 

> Payroll export start date is the second Monday in the month with the end date the Sunday of the same week

How does that read?

Funny because I had a spec for a month starting on Sunday and took it out because I was worried about the spec looking so repetetive!

erbridge: Maybe:

> Payroll export start date is the second Monday in the month and the end date is the following Sunday

PR #522, opened 2019-10-16
Enable maintenance mode on production

erbridge:

### DIFF
--- a/azure/resource_groups/app/parameters/production.template.json
+++ b/azure/resource_groups/app/parameters/production.template.json
@@ -142,10 +142,10 @@
       }
     },
     "MAINTENANCE_MODE": {
-      "value": ""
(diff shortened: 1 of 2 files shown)


azure/resource_groups/app/parameters/production.template.json
```diff
@@ -142,10 +142,10 @@
       }
     },
     "MAINTENANCE_MODE": {
-      "value": ""
+      "value": "1"
     },
     "MAINTENANCE_MODE_AVAILABILITY_MESSAGE": {
-      "value": ""
+      "value": "Private beta has ended. The service will be back in November."
```
tekin: Updated content from the team:

```suggestion
      "value": "It's not possible to make a claim at the moment. The service will be back in November."
```

PR #524, opened 2019-10-17
Remove old VSP encryption key

mec:
We have let Verify know that we are running both old and new encryption keys and they have confirmed so the this step is to remove the reference to the old one and set the new one as the primary.


mec:
[Rotate VSP certificates](https://trello.com/c/z9s8AbbG/685-rotate-vsp-certificates)

PR #531, opened 2019-10-21
Service operators can add notes when they approve or reject a claim

robbiepaul:
It was felt that having radio buttons instead of two separate submit buttons would make the act of approving/rejecting a claim more purposeful.

We have a slight discrepancy between the language in our code and the language in the UI. In the UI we're using "claim decision", in the code we're using "claim check/result". We may want to refactor our code to match the UI.

Original mockup: https://miro.com/app/board/o9J_kwHRvTU=/?moveToWidget=3074457346936539234
Most of the content has come from Faye

## Screenshots
Screenshot 2019-10-21 at 15 24 33
Screenshot 2019-10-21 at 15 22 50


app/controllers/admin/claim_checks_controller.rb
```diff
@@ -4,10 +4,14 @@ class Admin::ClaimChecksController < Admin::BaseAdminController
   before_action :reject_checked_claims
 
   def create
-    return redirect_to admin_claim_path(@claim), alert: "Claim cannot be approved" if params[:result] == "approved" && @claim.payroll_gender_missing?
-    @claim.create_check!(checked_by: admin_session.user_id, result: params[:result])
-    send_claim_result_email
-    redirect_to admin_claims_path, notice: "Claim has been #{@claim.check.result} successfully"
+    return redirect_to admin_claim_path(@claim), alert: "Claim cannot be approved" if check_params[:result] == "approved" && @claim.payroll_gender_missing?
```
pezholio: Could have this in a `before_action`?

spec/features/admin_claim_check_spec.rb
```diff
@@ -36,14 +39,51 @@
...
+    end
+
+    scenario "User can see existing check details" do
+      claim_with_check = create(:claim, :submitted, check: build(:check, result: :approved, notes: "Everything matches"))
+      visit admin_claim_path(claim_with_check)
+
+      expect(page).not_to have_button("Submit")
+      expect(page).to have_content("Claim decision")
+      expect(page).to have_content("Result Approved")
+      expect(page).to have_content("Notes\nEverything matches")
```
pezholio: I'm not sure we need to check for the presence of the text in the `dt` here, it tripped me up a bit. I think we should either just check for the presence of the text we added in the note, or do some more complex checks against the HTML (which may be overkill, and likely to make the tests brittle)

robbiepaul:
[Service operators can add "Notes" when they approve or reject a claim](https://trello.com/c/Sq8vcYY2/842-service-operators-can-add-notes-when-they-approve-or-reject-a-claim)

PR #532, opened 2019-10-21
Add claims and payroll runs to seeds

lawrence-dxw:
Making the experience of playing with the app locally a little more pleasant.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -5,5 +5,6 @@
 #
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
-puts "Importing Schools information, this may take a minute..."
-SchoolDataImporter.new.run
+
+ENV["FIXTURES_PATH"] = "spec/fixtures"
+Rake::Task["db:fixtures:load"].invoke
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -6,5 +6,18 @@
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
 
-ENV["FIXTURES_PATH"] = "spec/fixtures"
-Rake::Task["db:fixtures:load"].invoke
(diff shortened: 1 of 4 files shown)


db/seeds.rb
```diff
@@ -5,5 +5,21 @@
 #
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
-puts "Importing Schools information, this may take a minute..."
-SchoolDataImporter.new.run
+
+unless Rails.env.development?
+  raise "This db/seeds.rb file is only intended for use in a local development environment. It should not be used in the #{Rails.env} environment."
```
robbiepaul: I don't think we should raise an error here. The seeds file is called as part of `db:setup` so could be called when setting up a production environment. We could just wrap the creation of the fixtures/factories in a conditional?

db/seeds.rb
```diff
@@ -5,5 +5,21 @@
...
+end
+
+ENV["FIXTURES_PATH"] = "spec/fixtures"
+Rake::Task["db:fixtures:load"].invoke
+
+class Seeds
+  extend FactoryBot::Syntax::Methods
+
+  create(:payroll_run, :confirmation_report_uploaded, claims_count: 32, created_at: 3.months.ago - 10.days)
```
robbiepaul: If you didn't want to extend `FactoryBot::Syntax::Methods` you can just statically call the factory helpers:

```ruby
FactoryBot.create(:payroll_run, :confirmation_report_uploaded, claims_count: 32, created_at: 3.months.ago - 10.days)
```
I think either is fine though

db/seeds.rb
robbiepaul: Or we could change our Docker entrypoint to just run `db:create db:schema:load`

lawrence-dxw: I think you're right; the seeds can in theory be used to seed a production database as well. I think the best thing would be to just wrap the code in a conditional.

db/seeds.rb
lawrence-dxw: Yeah, I don't really have any strong opinions either way; thought this was just a little more readable.

PR #535, opened 2019-10-22
Remove VSP encryption cert v2

mec:
We've had confirmation from the Verify team that the old encryption key has been removed so we can safely remove it from the configuration now.

This completes the VSP key rotation.

Once this PR is merge we _must_ confrim the development enviroment is able to comple the Verify step.

### DIFF
--- a/azure/resource_groups/app/parameters/development.template.json
+++ b/azure/resource_groups/app/parameters/development.template.json
@@ -211,16 +211,11 @@
         "keyVault": {
           "id": "${keyVaultId}"
         },
-        "secretName": "TeacherPaymentsDevVspSamlEncryption2KeyBase64"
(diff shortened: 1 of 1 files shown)


mec:
I had a thought about automated smoke tests for things like this – is that a thing? 

I thought I heard talk about it on another team at DfE, I imagine something like a feature test that runs on development?

mec:
👆 @robbiepaul you beat me! 😄  I had a question out of this…

robbiepaul:
Yeah I think it's a good idea. I wonder how much work it is, I haven't done it before. We should raise it on the devs channel in Slack maybe?

mec:
[VSP Error on development](https://trello.com/c/xWYcsTfG/853-vsp-error-on-development)

PR #539, opened 2019-10-22
Don't ask for specific year for QTS award year

robbiepaul:
> We did originally think we'd need the precise QTS year from the teacher so that we could use it to estimate their claim amount based on their likely salary. But now that we have access to TPS and use it for claim checking, we have more accurate data for their salary. Therefor we no longer need the precise year and can simplify the question to "before" or "after" the eligibility date.

Includes a data migration that will update any existing eligible QTS award years to be the new value expected.

# Screenshot

Screenshot 2019-10-22 at 17 07 41


db/data/20191022155610_convert_qts_award_year.rb
```diff
@@ -0,0 +1,11 @@
+class ConvertQtsAwardYear < ActiveRecord::Migration[5.2]
```
pezholio: I know it's simple, but can we get a test for this? There's some prior art in #530 for how we did this. I was burnt by a data migration without tests before

robbiepaul: Ah yes, I figured it was such a simple query that we'd just be testing ActiveRecord. I'll add one 👍

PR #541, opened 2019-10-23
Allow users to search for a claim by reference

pezholio:
This allows a service operator to search for a claim by its reference. If a claim is found, it redirects to the relevant claim, otherwise the user gets an error. UI is very basic, so may need some @titlescreen ❤️ 

image

image

### DIFF
--- a/app/controllers/admin/claims_controller.rb
+++ b/app/controllers/admin/claims_controller.rb
@@ -9,4 +9,16 @@ class Admin::ClaimsController < Admin::BaseAdminController
     @claim = Claim.find(params[:id])
     @check = @claim.check || Check.new
   end
+
+  def search
+    return unless params[:reference].present?
+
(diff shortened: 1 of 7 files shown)


pezholio:
[Service Operators can view a claim that has been approved/rejected](https://trello.com/c/8wdXY6np/851-service-operators-can-view-a-claim-that-has-been-approved-rejected)

app/controllers/admin/claims_controller.rb
```diff
@@ -9,4 +9,14 @@ def show
     @claim = Claim.find(params[:id])
     @check = @claim.check || Check.new
   end
+
+  def search
+    return unless params[:reference].present?
+
+    if (claim = Claim.find_by(reference: params[:reference]))
```
robbiepaul: Could we separate assignment from the conditional? I find that having assignments in conditionals can be confusing when scan reading the code.

PR #542, opened 2019-10-23
Bump mail-notify to 0.2.3

tekin:
This latest version updates the notifications-ruby-client gem, which resolves an issue when reporting errors to Rollbar.


tekin:
@pezholio I've updated this so we set a random template ID in dev and test environments.

PR #543, opened 2019-10-23
Update student loan question to latest version

robbiepaul:
The content on this page now matches the latest content in the prototype.

Screenshot 2019-10-23 at 16 23 02

### DIFF
--- a/app/views/claims/student_loan_amount.html.erb
+++ b/app/views/claims/student_loan_amount.html.erb
@@ -12,13 +12,14 @@
             <%= fields.label :student_loan_repayment_amount, t("student_loans.questions.student_loan_amount", claim_school_name: current_claim.eligibility.claim_school_name), class: "govuk-label govuk-label--xl" %>
           </h1>
 
-          <span class="govuk-hint">
-            If you do not know the exact amount, check your P60, your annual
-            student loan statement, or contact the
-            <a href="https://www.slc.co.uk/students-and-customers/contact-information-for-customers/loan-repayment-enquiries.aspx" class="govuk-link">
-              student loans company
-            </a>.
-          </span>
+          <p class="govuk-hint">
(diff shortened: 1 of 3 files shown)


robbiepaul:
[Update student loan question to latest version](https://trello.com/c/FivW85dq/874-update-student-loan-question-to-latest-version)

app/views/claims/student_loan_amount.html.erb
```diff
@@ -13,11 +13,9 @@
           </h1>
 
           <span class="govuk-hint">
-            If you do not know the exact amount, check your P60, your annual
-            student loan statement, or contact the
-            <a href="https://www.slc.co.uk/students-and-customers/contact-information-for-customers/loan-repayment-enquiries.aspx" class="govuk-link">
-              student loans company
-            </a>.
+            Check your annual student loan statement, or your P60 if you had only one employer. You can also add up the
```
pezholio: Just wondering if this might be easier to read if it was broken into two paragraphs? It's quite a chunky block of text as it stands

robbiepaul: Ok, I'll double check with Faye 👍 

This is what it will look like:

Screenshot 2019-10-23 at 16 36 40

titlescreen: I think that seems like a winner

pezholio: Sweet 👍

robbiepaul: Faye gave it the thumbs up, updated the commit.

As an aside I noticed we're not using the `aria-describedby` attribute on our inputs so I created a card for it: https://trello.com/c/4vGvgVmZ/882-add-aria-describedby-to-input-elements-with-hints

PR #545, opened 2019-10-24
Claimants recieve payment notifications once a payroll run has been processed

pezholio:
This adds a new mailer to send payment notifications once a payroll run has been processed with a "payslip" containing all their tax contribution information etc. Once a run has been processed successfully, the `PaymentConfirmation` tax queues all the emails to be sent (copy to be confirmed)

Here's how the email looks:

image


app/mailers/previews/claim_mailer_preview.rb
```diff
@@ -0,0 +1,29 @@
...
+  def approved
+    ClaimMailer.approved(claim)
+  end
+
+  def rejected
+    ClaimMailer.rejected(claim)
+  end
+
+  def payment_confirmation
+    payment = FactoryBot.build(:payment, :paid, claim: claim, created_at: DateTime.now)
```
tekin: I hadn't considered using factories here in a mailer! Whilst I like the idea in principle, I think in practice it has some down sides. For example at the minute, every preview results in two new `LocalAuthority` records being created in the database, which isn't ideal. When creating previews, I normally use a scope to grab an appropriate existing record from the database and assume that one will exist. Perhaps instead of using factories, we can use the `seeds.rb` file to make sure the local database will aways contain appropriate records to preview the emails.

app/models/payment_confirmation.rb
```diff
@@ -23,12 +23,22 @@ def ingest
...
+        end
+      else
+        raise ActiveRecord::Rollback
+      end
     end
   end
 
   private
 
+  def payment_date
```
tekin: This method is called `payment_date` but it returns a time object. Maybe things would be better encapsulated if it was rewritten to be called `payment_date_timestamp` instead and included the `to_i` call?

spec/models/payment_confirmation_spec.rb
```diff
@@ -43,6 +43,17 @@
 
       expect(payroll_run.reload.confirmation_report_uploaded_by).to eq(admin_user_id)
     end
+
+    it "queues emails" do
```
tekin: Could this `it` be rewritten to be more specific about the behaviour we expect? e.g. it sends payment confirmation emails with a payment date of the following Friday?

spec/models/payment_confirmation_spec.rb
```diff
@@ -43,6 +43,17 @@
...
+      travel_to Date.parse("2019-01-01") do
```
tekin: Maybe we should name this date, e.g. `a_tuesday`, and then rename the next one to `the_following_friday` to make the intent clearer?

spec/models/payment_confirmation_spec.rb
```diff
@@ -43,6 +43,17 @@
...
+        friday = Time.parse("2019-01-04").to_i
+        expect { payment_confirmation.ingest }.to have_enqueued_job(ActionMailer::DeliveryJob)
+          .with("ClaimMailer", "payment_confirmation", "deliver_now", payroll_run.claims[0], friday)
```
tekin: I worry when I see specs like this that they are quite tightly coupled to the internal implementation details of how ActionMailer/ActiveJob works. If that should change, then these specs would break, even though the code itself would probably still work. See the [idea of structure vs behaviour in tests in this blog post](https://medium.com/@kentbeck_7670/test-desiderata-94150638a4b3).

I think these specs would be more robust if they were written to check the actual content of the delivered emails, ala this rejection feature spec: https://github.com/DFE-Digital/dfe-teachers-payment-service/blob/6d12a3d7655c5118d2b937b844ad5530c47ca1ac/spec/features/admin_claim_check_spec.rb#L62

The downside to this is that we will end up duplicating coverage a bit in that this spec will test the email content in a similar way to the specs for the mailer, but I think a small amount of duplication and overlapping coverage in specs is ok if it makes them less structure-sensitive. Here we only need to verify enough to know that the correct date is present in the email body.

spec/requests/admin_payment_confirmation_report_upload_spec.rb
```diff
@@ -40,6 +40,12 @@
 
           expect(payroll_run.reload.confirmation_report_uploaded_by).to eq(admin_session_id)
         end
+
+        it "queues emails" do
```
tekin: I would be tempted to add this expectation to the existing `it` above and reword the description to mention sending payment confirmation emails. Whilst separating out `it` examples to better explain behaviour is good in general, for expensive feature and request specs, each `it` we add increases the test suite time by a non-trivial amount, which adds up over time. Where we can avoid it, we should try, to help keep the test suite fast.

spec/requests/admin_payment_confirmation_report_upload_spec.rb
```diff
@@ -57,6 +63,23 @@
         end
       end
 
+      context "the CSV has invalid data" do
```
tekin: Could we can test the not queueing part in one of the existing contexts rather than creating a new one?

app/views/claim_mailer/payment_confirmation.text.erb
```diff
@@ -0,0 +1,38 @@
+Dear <%= @display_name %>,
+
+We’re paying your claim to get back the student loan repayments you made while employed as a teacher during the last financial year, reference: <%= @reference %>.
+
+You will receive <%=number_to_currency(@payment.net_pay) %> on or after <%= l(@payment_date) %>
```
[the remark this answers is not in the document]
pezholio: Alas, no. The only things we can do are top level headings, bullets, callouts and horizontal lines. I'll see how it looks with a callout

spec/models/payment_confirmation_spec.rb
pezholio: Yeah, that makes sense. I was a bit iffy about doing it that way at the time, as we're not _strictly_ queuing a job directly when we use `deliver_later`, but I wanted to avoid duplication. I think it's OK as long as we test the bare minimum.

tekin: yep! 👍

.env.development
```diff
@@ -9,4 +9,4 @@ DFE_SIGN_IN_API_ENDPOINT=https://pp-api.signin.education.gov.uk
 
 ADMIN_ALLOWED_IPS=::1,127.0.0.1
 
-NOTIFY_TEMPLATE_ID=XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX
+NOTIFY_TEMPLATE_ID=d72e2ff9-b228-4f16-9099-fd9d411c0334
```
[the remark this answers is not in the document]
pezholio: Yeah, unfortunately. As the previews call the Notify API, I don't think there's a way round this.

app/mailers/previews/claim_mailer_preview.rb
pezholio: Does `build` actually create database records though? I was under the impression that it just created them in memory.

app/mailers/claim_mailer.rb
```diff
@@ -11,6 +11,19 @@ def rejected(claim)
     view_mail_with_claim_and_subject(claim, "Your claim to get your student loan repayments back has been rejected, reference number: #{claim.reference}")
   end
 
+  def payment_confirmation(claim, payment_date_timestamp)
+    @reference = claim.reference
+    @payment = claim.payment
+    @display_name = [claim.first_name, claim.surname].join(" ")
+    @payment_date = Time.at(payment_date_timestamp)
```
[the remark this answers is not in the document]
pezholio: Good catch 👍

app/mailers/previews/claim_mailer_preview.rb
tekin: Yeah, in theory it's equivalent to `new` (vs `create`), but if you are building a record that has a `belongs_to` association on it, I think factory bot ends up saving the parent records.

app/models/payment_confirmation.rb
```diff
@@ -23,12 +23,22 @@ def ingest
...
+      else
+        raise ActiveRecord::Rollback
+      end
     end
   end
 
   private
 
+  def payment_date_timestamp
+    Date.today.next_occurring(:friday).to_time.to_i
```
tekin:
```suggestion
    Date.today.next_occurring(:friday).to_i
```

app/views/claim_mailer/payment_confirmation.text.erb
```diff
@@ -0,0 +1,30 @@
+Dear <%= @display_name %>,
+
+We’re paying your claim to get back the student loan repayments you made while employed as a teacher during the last financial year, reference: <%= @reference %>.
```
tekin:
```suggestion
We’re paying your claim to get back the student loan repayments you made while employed as a teacher during the last financial year, claim reference: <%= @reference %>.
```

app/views/claim_mailer/payment_confirmation.text.erb
```diff
@@ -0,0 +1,30 @@
...
+
+This payment does not change the amount that was credited to your Student Loans Company (SLC) account in the last financial year.
+
+You will receive <%=number_to_currency(@payment.net_pay) %> on or after <%= l(@payment_date) %>
```
tekin:
```suggestion
You will receive <%=number_to_currency(@payment.net_pay) %> on or after <%= l(@payment_date) %>.
```

tekin: where we going to make this a callout in leu of bold?

app/models/payment_confirmation.rb
pezholio: I don't think `.to_i` is available on Date objects.

spec/models/payment_confirmation_spec.rb
```diff
@@ -43,6 +43,27 @@
 
       expect(payroll_run.reload.confirmation_report_uploaded_by).to eq(admin_user_id)
     end
+
+    it "sends payment confirmation emails with a payment date of the following Friday" do
+      a_tuesday = Date.parse("2019-01-01")
+      friday = "4 January 2019"
```
tekin:
```suggestion
      the_following_friday = "4 January 2019"
```

PR #547, opened 2019-10-24
Remove `payroll_run_id` from Claim

pezholio:
I've also had to delete the specs for the data migration, because the column is no longer around. It's served its purpose though now.


pezholio:
[Remove payroll_run_id column on claims table](https://trello.com/c/ceewDHzW/846-remove-payrollrunid-column-on-claims-table)

db/schema.rb
```diff
@@ -108,7 +106,7 @@
     t.decimal "student_loan_repayment", precision: 7, scale: 2
     t.decimal "tax", precision: 7, scale: 2
     t.decimal "net_pay", precision: 7, scale: 2
-    t.index ["claim_id"], name: "index_payments_on_claim_id", unique: true
+    t.index ["claim_id"], name: "index_payments_on_claim_id"
```
lawrence-dxw: Was this change intentional?

pezholio: Nah, all I did was run the migration. I'll do it all again from cold and see if the same thing happens

pezholio: I've recreated the schema from scratch and it hasn't happened again, so I think something must've got out of sync

PR #548, opened 2019-10-24
Prepare PageSequence for Maths & Physics

tekin:
This work makes `PageSequence` policy-agnostic by allowing the policy-specific slug sequence to be passed in, paving the way for the introduction of the Maths & Physics policy.

### DIFF
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -85,6 +85,10 @@ class ClaimsController < ApplicationController
   end
 
   def page_sequence
-    @page_sequence ||= PageSequence.new(current_claim, params[:slug])
+    @page_sequence ||= PageSequence.new(current_claim, claim_slug_sequence, params[:slug])
+  end
+
+  def claim_slug_sequence
+    current_claim.eligibility.class.parent::SlugSequence.new(current_claim)
   end
 end
--- a/spec/models/page_sequence_spec.rb
+++ b/spec/models/page_sequence_spec.rb
@@ -4,22 +4,23 @@ require "rails_helper"
 
 RSpec.describe PageSequence do
   let(:claim) { build(:claim) }
+  let(:slug_sequence) { OpenStruct.new(slugs: ["first-slug", "second-slug", "third-slug"]) }
 
   describe "#next_slug" do
     it "assumes we're at the beginning of the sequence if no current_slug is specified" do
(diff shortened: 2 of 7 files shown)


app/controllers/claims_controller.rb
```diff
@@ -85,6 +85,10 @@ def check_page_is_in_sequence
   end
 
   def page_sequence
-    @page_sequence ||= PageSequence.new(current_claim, params[:slug])
+    @page_sequence ||= PageSequence.new(current_claim, claim_slug_sequence, params[:slug])
+  end
+
+  def claim_slug_sequence
+    current_claim.eligibility.class.parent::SlugSequence.new(current_claim)
```
pezholio: Could we break out the bit where we get the class out into a method? (or have is as a variable) Just to be clearer about the intention here. Something like:

```
def eligibility_context
  current_claim.eligibility.class.parent
end

def claim_slug_sequence
  eligibility_context::SlugSequence.new(current_claim)
end
```

spec/models/page_sequence_spec.rb
```diff
@@ -4,107 +4,46 @@
...
-      expect(page_sequence.slugs).to include("gender")
-    end
-
-    it "excludes the address page if the address was returned by verify" do
-      claim.verified_fields = ["postcode"]
-      page_sequence = PageSequence.new(claim, "student-loan-country")
-      expect(page_sequence.slugs).to_not include("address")
-    end
-  end
+  let(:slug_sequence) { OpenStruct.new(slugs: ["first-slug", "second-slug", "third-slug"]) }
```
pezholio: I like the idea of using a struct for a sequence here 👍

app/controllers/claims_controller.rb
tekin: Yeah, I've toyed around with naming this thing and pulling it out to a method but feel like until it's being used in more places, it might not be the best time to do it. I like 'eligibility context'. I was also thinking of putting it on the claim itself, that way it's available in more places, e.g. `current_claim.policy_namespace`. Can we delay doing this for now? I think it will happen as the rest of the M&P refactor work continues.

pezholio: Ah yeah, cool. I like the idea of putting it in the model too. Happy to leave it for now in that case 👍

PR #551, opened 2019-10-28
Service operators see when details in a claim have been used before

mec:
This is the first step for the work to prevent people being paid more than once. 

We think that checking for matching details in claims and what to do about it falls to the the service operators, therefore we want to get claims with matching attributes in-front of them as soon as we can.

This first step simply checks all other claims for matching details and flags the claims up to the service operator.

It lets them see the other claims and links to them for comparison. We want to do more work around the UI for comparing claims later, with help from the service operators.

As the service does not distiguish between tax years, nor do the checks.

## Screenshot
Screenshot_2019-10-28 Claim additional payments for teaching – GOV UK

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- A clear warning is shown to service operators when a claim contains details
+  that have been used in other claims. The other claims are listed and linked
+
 ## [Release 022] - 2019-10-24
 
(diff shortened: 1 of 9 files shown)


mec:
[We are able to prevent someone from being paid more than once](https://trello.com/c/qzjblBGB/751-we-are-able-to-prevent-someone-from-being-paid-more-than-once)

mec:
Going to have to go back to this one. I will close this for now.

PR #552, opened 2019-10-28
Claim is ineligible if current school is ineligible

lawrence-dxw:
Inform the user that their claim is ineligible, if the school that they are currently employed at is not eligible.

This is the message informing the user that their claim is ineligible:

Screenshot_2019-10-28 You’re not eligible for this payment - Teachers claim back your student loan repayments - GOV UK(1)

### DIFF
--- a/app/models/student_loans/school_eligibility.rb
+++ b/app/models/student_loans/school_eligibility.rb
@@ -42,7 +42,8 @@ module StudentLoans
     end
 
     def eligible_current_school?
-      @school.state_funded? &&
+      @school.open? &&
+        @school.state_funded? &&
         (eligible_phase? || eligible_special_school?)
     end
 
--- a/app/models/student_loans/school_eligibility.rb
+++ b/app/models/student_loans/school_eligibility.rb
@@ -41,6 +41,11 @@ module StudentLoans
         (eligible_phase? || eligible_special_school?)
     end
 
+    def eligible_current_school?
+      @school.state_funded? &&
+        (eligible_phase? || eligible_special_school?)
+    end
+
     private
 
     def eligible_local_authority?
--- a/app/models/student_loans/school_eligibility.rb
+++ b/app/models/student_loans/school_eligibility.rb
@@ -34,7 +34,7 @@ module StudentLoans
       @school = school
     end
 
-    def check
+    def eligible_claim_school?
       !closed_before_policy_start? &&
         eligible_local_authority? &&
         @school.state_funded? &&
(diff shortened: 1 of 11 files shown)


app/models/student_loans/school_eligibility.rb
```diff
@@ -34,13 +34,18 @@ def initialize(school)
...
-    def check
+    def eligible_claim_school?
       !closed_before_policy_start? &&
         eligible_local_authority? &&
         @school.state_funded? &&
         (eligible_phase? || eligible_special_school?)
     end
 
+    def eligible_current_school?
+      @school.state_funded? &&
```
robbiepaul: We could also have a check for if the school is open or not here?

lawrence-dxw: We do already handle the case where the current school is closed, though not via `StudentLoans::SchoolEligibility`. Instead, `StudentLoans::Eligiblity` has a `current_school_closed?` check.

If we wanted to move the check into the school eligibility checker, we would have to make an additional refactor so that the school eligibility checker can return a _reason_ why the school is ineligible, so that we can preserve the current ineligibility message. It sounds like [we are planning to add more detailed ineligiblity reasons](https://trello.com/c/Mjx6L9yH/863-a-user-should-know-why-their-answer-has-made-them-ineligible) in the future so this might be a refactor that's worth doing, but I don't know whether this story is the place to do it.

Do you have any thoughts on this? (cc @tekin)

tekin: Good question! it does feel like moving `current_school_closed?` to the eligibility makes sense from an encapsulation perspective, but as you point out @lawrence-dxw, that would mean a closed school would result in the generic ineligible current school message and we'd lose the specific one for a closed current school. That said, I don't believe it's possible to choose a closed school anymore using the UI as we exclude closed schools from the search, so maybe that's not actually a big deal?

lawrence-dxw: The "closed current school" message was actually introduced [in the same set of changes](https://github.com/DFE-Digital/dfe-teachers-payment-service/pull/431) as excluding closed schools from the current school search. But F's message there suggests that they considered it an unlikely case ("e.g. by manually setting query parameters").

What do you think, @tekin:

1. keep it as it is;
2. refactor, losing the specific error message;
3. more complex refactor, keeping the specific error message?

tekin: I would go with option 2. It definitely sounds like an unlikely scenario, and we will be better placed to tackle the more complex refactor when we know more about it, i.e. when the story to introduce more detailed ineligibility reasons is prioritised.

lawrence-dxw: I've pushed that change.

lawrence-dxw:
Cool, thanks. I'm just going to push some updated copy that I got from Faye, then will merge.

PR #553, opened 2019-10-28
Load GOVUK Frontend from npm package

mec:
Loading the GOVUK Frontend styles, js and assets from a gem worked great until
it didn't when an update to sprockets broke it.

Althought the gem is fixed, as the GOVUK Frontend is distrubuted as a npm
package it makes more sense to load it from there and keep the dependency
tighter.

This commit removes the gem and makes the changes to load the assets from the
npm package in node_modules.

I propose this as a replacement to #549 which I've closed.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,10 +21,6 @@ gem "jbuilder", "~> 2.9"
 # Use ActiveModel has_secure_password
 # gem 'bcrypt', '~> 3.1.7'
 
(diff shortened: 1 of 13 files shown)


mec:
> We could even remove the `public/assets/*` files as they're now redundant but that shouldn't prevent merging

Don't we need these for the static error pages?

PR #554, opened 2019-10-28
Make sure form is submitted locally

pezholio:
At the moment, the user won't see the confirmation page because, by default, `form_with` attaches the `data-remote` attribute submitting the form via an XMLHTTPRequest in the background. To override this, we usuallt need to set the `local: true` attribute, so the correct format is requested. Rather than remembering to do this each time, I've set a global flag, so we don't have to think about it again (for this project anyway).


spec/features/student_loans_claim_spec.rb
```diff
@@ -1,124 +1,130 @@
 require "rails_helper"
 
 RSpec.feature "Teacher Student Loan Repayments claims" do
-  scenario "Teacher claims back student loan repayments" do
-    visit new_claim_path
-    expect(page).to have_text(I18n.t("student_loans.questions.qts_award_year"))
+  [
+    "enabled",
+    "disabled",
```
tekin: Can we just use `true` and `false` here instead and pass the value straight to the `js:` in the scenario? The temporary variable feels surplus to requirements.

pezholio: I use that for the naming of the tests. It might make sense to reverse it though, and use `true` and `false` and then title the test depending on the value. I'll do that

PR #556, opened 2019-10-29
Service operators can see when details in a claim have been used before

mec:
This is the first step for the work to prevent people being paid more than once. 

We think that checking for matching details in claims and what to do about it falls to the the service operators, therefore we want to get claims with matching attributes in-front of them as soon as we can.

This first step simply checks all other submitted claims for matching details and flags the claims up to the service operator.

It lets them see the other claims and links to them for comparison. We want to do more work around the UI for comparing claims later, with help from the service operators.

We also want to check for matches across combination of details, but this will also come later.

As the service does not distiguish between tax years, nor do the checks.

## Postgres Performance
We know this operation is quite expensive and looked at ways to optimise the query. Because the attributes that get compared (TRN, NINO, email address, etc) are always present in the source claim – they are required fields, the query was always planned to run sequentially and so indexes on columns would not get used.  

## Odd commit order
Rebasing has left the commits in an odd order – sorry!

## Screenshot
Screenshot_2019-10-28 Claim additional payments for teaching – GOV UK

### DIFF
--- a/app/helpers/admin/claims_helper.rb
+++ b/app/helpers/admin/claims_helper.rb
@@ -68,6 +68,14 @@ module Admin
       content_tag(:strong, pluralize(days_until_check_deadline, "day"), class: "govuk-tag #{check_deadline_warning_class}")
     end
 
+    def matching_attributes(first_claim, second_claim)
+      first_attributes = first_claim.attributes.slice(*Claim::MatchingAttributeFinder::ATTRIBUTES_TO_MATCH).to_a
+      second_attributes = second_claim.attributes.slice(*Claim::MatchingAttributeFinder::ATTRIBUTES_TO_MATCH).to_a
+
+      matching_attributes = first_attributes & second_attributes
+      matching_attributes.to_h.compact.keys.map(&:humanize).sort
+    end
+
     private
 
     def days_between(first_date, second_date)
--- a/app/models/claim/matching_attribute_finder.rb
+++ b/app/models/claim/matching_attribute_finder.rb
@@ -1,0 +1,31 @@
+class Claim
+  class MatchingAttributeFinder
+    ATTRIBUTES_TO_MATCH = %w[
+      teacher_reference_number
+      email_address
+      national_insurance_number
+      bank_account_number
+      bank_sort_code
+      building_society_roll_number
+    ].freeze
+
+    def initialize(source_claim, claims_to_compare = Claim.submitted)
+      @source_claim = source_claim
+      @claims_to_compare = claims_to_compare
+    end
+
+    def matching_claims
+      predicates = []
+      values = []
+      ATTRIBUTES_TO_MATCH.each do |attribute|
+        value = @source_claim.read_attribute(attribute)
+        next unless value
+
+        predicates << "#{attribute} = ?"
+        values << value
+      end
+
+      @claims_to_compare.where.not(id: @source_claim.id).where(predicates.join(" OR "), *values)
+    end
+  end
+end
--- a/app/views/admin/claims/show.html.erb
+++ b/app/views/admin/claims/show.html.erb
@@ -10,6 +10,12 @@
       </div>
     <% end %>
 
+    <% if @matching_claims.any? %>
+      <div class="govuk-body-l govuk-flash__warning">
+        <a href="#claims-with-matches">Details in this claim match those in other claims</a>
+      </div>
+    <% end %>
+
     <h1 class="govuk-heading-xl">
       Claim <%= @claim.reference %>
     </h1>
@@ -30,6 +36,32 @@
           This claim cannot be approved, the payroll gender is missing and the claim will need to be referred
         </strong>
       </div>
+    <% end %>
+
+    <% if @matching_claims.any? %>
+      <table id="claims-with-matches" class="govuk-table govuk-!-margin-bottom-9">
+        <caption class="govuk-table__caption govuk-heading-l">Claims with matching details</caption>
+        <thead class="govuk-table__head">
+          <tr class="govuk-table__row">
+            <th scope="col" class="govuk-table__header">Claim</th>
+            <th scope="col" class="govuk-table__header">Matching details</th>
+          </tr>
+        </thead>
+        <tbody class="govuk-table__body">
+          <% @matching_claims.each do |matching_claim| %>
+            <tr class="govuk-table__row">
+              <th scope="row" class="govuk-table__header"><%= link_to matching_claim.reference, [:admin, matching_claim], class: "govuk-link" %>
+              <td class="govuk-table__cell">
+                <ul class="govuk-list">
+                <% matching_attributes(@claim, matching_claim).each do |attribute| %>
+                  <li><%= attribute %></li>
+                <% end %>
+                </ul>
+              </td>
+            </tr>
+          <% end %>
+        </tbody>
+      </table>
     <% end %>
 
     <% if @check.persisted? %>
@@ -62,7 +94,6 @@
           </span>
           <%= form.text_area :notes, class: "govuk-textarea", rows: 5 %>
           <%= form.submit "Submit", class: "govuk-button" %>
-
         <% end %>
       <% end %>
     <% end %>
--- a/spec/helpers/admin/claims_helper_spec.rb
+++ b/spec/helpers/admin/claims_helper_spec.rb
@@ -173,4 +173,29 @@ describe Admin::ClaimsHelper do
       it { is_expected.to be_nil }
     end
   end
+
+  describe "#matching_attributes" do
+    let(:first_claim) {
+      build(
+        :claim,
+        teacher_reference_number: "0902344",
+        national_insurance_number: "QQ891011C",
+        email_address: "genghis.khan@mongol-empire.com",
+        bank_account_number: "34682151",
+        bank_sort_code: "972654",
+        building_society_roll_number: "123456789/ABCD",
+      )
+    }
+    let(:second_claim) { build(:claim, :submitted, teacher_reference_number: first_claim.teacher_reference_number, national_insurance_number: first_claim.national_insurance_number, building_society_roll_number: first_claim.building_society_roll_number) }
+    subject { helper.matching_attributes(first_claim, second_claim) }
+
+    it "returns the humanised names of the matching attributes" do
+      expect(subject).to eq(["Building society roll number", "National insurance number", "Teacher reference number"])
+    end
+
+    it "does not consider nil building society roll number to be a match" do
+      [first_claim, second_claim].each { |claim| claim.building_society_roll_number = nil }
+      expect(subject).to eq(["National insurance number", "Teacher reference number"])
+    end
+  end
 end
--- a/spec/models/claim/matching_attribute_finder_spec.rb
+++ b/spec/models/claim/matching_attribute_finder_spec.rb
@@ -1,0 +1,73 @@
+require "rails_helper"
+
+RSpec.describe Claim::MatchingAttributeFinder do
+  describe "#matching_claims" do
+    let(:source_claim) {
+      create(
+        :claim,
+        teacher_reference_number: "0902344",
+        national_insurance_number: "QQ891011C",
+        email_address: "genghis.khan@mongol-empire.com",
+        bank_account_number: "34682151",
+        bank_sort_code: "972654",
+        building_society_roll_number: "123456789/ABCD",
+      )
+    }
+
+    let!(:claim_with_no_matching_attributes) { create(:claim, :submitted) }
+    let!(:unsubmitted_claim_with_matching_teacher_reference_number) { create(:claim, :submittable, teacher_reference_number: source_claim.teacher_reference_number) }
+
+    subject(:matching_claims) { Claim::MatchingAttributeFinder.new(source_claim).matching_claims }
+
+    it "does not include the source claim, or claims that do not match, or claims that are not submitted" do
+      expect(matching_claims).to be_empty
+    end
+
+    it "includes a claim with a matching teacher reference number" do
+      claim_with_matching_attribute = create(:claim, :submitted, teacher_reference_number: source_claim.teacher_reference_number)
+
+      expect(matching_claims).to eq([claim_with_matching_attribute])
+    end
+
+    it "includes a claim with a matching national insurance number" do
+      claim_with_matching_attribute = create(:claim, :submitted, national_insurance_number: source_claim.national_insurance_number)
+
+      expect(matching_claims).to eq([claim_with_matching_attribute])
+    end
+
+    it "includes a claim with a matching email address" do
+      claim_with_matching_attribute = create(:claim, :submitted, email_address: source_claim.email_address)
+
+      expect(matching_claims).to eq([claim_with_matching_attribute])
+    end
+
+    it "includes a claim with a matching bank account number" do
+      claim_with_matching_attribute = create(:claim, :submitted, bank_account_number: source_claim.bank_account_number)
+
+      expect(matching_claims).to eq([claim_with_matching_attribute])
+    end
+
+    it "includes a claim with a matching bank sort code" do
+      claim_with_matching_attribute = create(:claim, :submitted, bank_sort_code: source_claim.bank_sort_code)
+
+      expect(matching_claims).to eq([claim_with_matching_attribute])
+    end
+
+    it "includes a claim with a matching building society roll number" do
+      claim_with_matching_attribute = create(:claim, :submitted, building_society_roll_number: source_claim.building_society_roll_number)
+
+      expect(matching_claims).to eq([claim_with_matching_attribute])
+    end
+
+    context "when the source claim building society roll number is nil" do
+      before do
+        source_claim.update!(building_society_roll_number: nil)
+      end
+
+      it "does not include any claim with a roll number of nil" do
+        create(:claim, :submitted, building_society_roll_number: nil)
+        expect(matching_claims).to be_empty
+      end
+    end
+  end
+end
--- a/spec/requests/admin_claims_spec.rb
+++ b/spec/requests/admin_claims_spec.rb
@@ -20,10 +20,21 @@ RSpec.describe "Admin claims", type: :request do
   describe "claims#show" do
     let(:claim) { create(:claim, :submitted) }
 
-    it "returns a claim when one exists" do
+    it "returns the claim" do
       get admin_claim_path(claim)
 
       expect(response.body).to include(claim.reference)
+    end
+
+    context "when another claim has matching attributes" do
+      let!(:claim_with_matching_attributes) { create(:claim, :submitted) }
+
+      it "returns the claim and the duplicate" do
+        get admin_claim_path(claim)
+
+        expect(response.body).to include(claim.reference)
+        expect(response.body).to include(claim_with_matching_attributes.reference)
+      end
     end
   end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- A clear warning is shown to service operators when a claim contains details
+  that have been used in other claims. The other claims are listed and linked
+
 ## [Release 025] - 2019-10-31
 
 - Adjusted ineligibility screens for users who may have taught at more than one
--- a/app/assets/stylesheets/components/flash.scss
+++ b/app/assets/stylesheets/components/flash.scss
@@ -18,4 +18,12 @@
     color: govuk-colour("red");
     border-color: govuk-colour("red");
   }
+  &__warning {
+    @extend %govuk-flash;
+    color: govuk-colour("orange");
+    border-color: govuk-colour("orange");
+    a {
+      color: govuk-colour("orange");
+    }
+  }
 }
--- a/app/controllers/admin/claim_checks_controller.rb
+++ b/app/controllers/admin/claim_checks_controller.rb
@@ -18,6 +18,7 @@ class Admin::ClaimChecksController < Admin::BaseAdminController
 
   def load_claim
     @claim = Claim.find(params[:claim_id])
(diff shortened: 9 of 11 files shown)


mec:
[We are able to prevent someone from being paid more than once](https://trello.com/c/qzjblBGB/751-we-are-able-to-prevent-someone-from-being-paid-more-than-once)

spec/models/claim/matching_attribute_finder_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+
+    it "does not include the id of a claim without any matcheing attributes" do
+      expect(claims_with_matching_attributes.keys).not_to include(claim_with_no_matching_attributes.id)
+    end
+
+    it "only includes submitted claims" do
+      expect(claims_with_matching_attributes.keys).not_to include(unsubmitted_claim_with_matching_teacher_reference_number.id)
+    end
+
+    it "includes the id of a claim with a mathching teacher reference number" do
```
lawrence-dxw:
```suggestion
    it "includes the id of a claim with a matching teacher reference number" do
```

(and throughout the rest of the file)

spec/models/claim/matching_attribute_finder_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+  describe "#claim_ids_with_matching_attributes" do
+    let(:source_claim) {
+      create(
+        :claim,
+        teacher_reference_number: "0902344",
+        national_insurance_number: "QQ891011C",
+        email_address: "genghis.khan@mongol-empire.com",
+        bank_account_number: "34682151",
+        bank_sort_code: "972654",
+        building_society_roll_number: "123456789/ABCD",
```
lawrence-dxw: Would it be worth checking that two claims which are both missing a building society roll number are not considered "matching"?

spec/models/claim/matching_attribute_finder_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+    let!(:unsubmitted_claim_with_matching_teacher_reference_number) { create(:claim, :submittable, teacher_reference_number: source_claim.teacher_reference_number) }
+
+    let(:claim_with_matching_attribute) { build(:claim, :submitted) }
+    let(:claims_with_matching_attributes) { Claim::MatchingAttributeFinder.new(source_claim).claim_ids_with_matching_attribute_keys }
+
+    it "does not include the id of the source claim" do
+      expect(claims_with_matching_attributes.keys).not_to include(source_claim.id)
+    end
+
+    it "does not include the id of a claim without any matcheing attributes" do
```
lawrence-dxw:
```suggestion
    it "does not include the id of a claim without any matching attributes" do
```

spec/models/claim/matching_attribute_finder_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+    it "includes the id of a claim with a mathching teacher reference number" do
+      claim_with_matching_attribute.teacher_reference_number = source_claim.teacher_reference_number
+      claim_with_matching_attribute.save
+
+      expect(claims_with_matching_attributes.keys).to include(claim_with_matching_attribute.id)
+      expect(claims_with_matching_attributes[claim_with_matching_attribute.id]).to include("Teacher reference number")
+    end
+
+    it "includes the id of a claim with a mathching national insurance number" do
+      claim_with_matching_attribute.national_insurance_number = source_claim.national_insurance_number
```
lawrence-dxw: You could save a line, losing the `save`, with

```suggestion
      claim_with_matching_attribute.update(national_insurance_number: source_claim.national_insurance_number)
```

And it might be worth using `update!` (or `save!`) just to make sure it fails loudly if the test setup fails.

This applies to the other similar `it` blocks in this file, too.

spec/models/claim/matching_attribute_finder_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+      claim_with_matching_attribute.save
+
+      expect(claims_with_matching_attributes.keys).to include(claim_with_matching_attribute.id)
+      expect(claims_with_matching_attributes[claim_with_matching_attribute.id]).to include("National insurance number")
```
lawrence-dxw: This is usually capitalised as "National Insurance number". I wonder whether it would be better if `MatchingAttributeFinder` didn't try to return human-readable attribute names, and instead returned the attribute names as the original symbols, e.g. `:national_insurance_number`. And then at the _view_ layer we could format them for presentation to the user.

spec/models/claim/matching_attribute_finder_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+      )
+    }
+
+    let!(:claim_with_no_matching_attributes) { create(:claim, :submitted) }
+    let!(:unsubmitted_claim_with_matching_teacher_reference_number) { create(:claim, :submittable, teacher_reference_number: source_claim.teacher_reference_number) }
+
+    let(:claim_with_matching_attribute) { build(:claim, :submitted) }
+    let(:claims_with_matching_attributes) { Claim::MatchingAttributeFinder.new(source_claim).claim_ids_with_matching_attribute_keys }
+
+    it "does not include the id of the source claim" do
```
lawrence-dxw: I'd like to get @tekin's opinion on the structure of the first three `it` blocks here, each of which calls the `#claim_ids_with_matching_attribute_keys` method, making various assertions each time but with the same setup data. They could be combined into a single `it` which would speed up the tests, maybe at the cost of readability. We discussed something similar on a PR Tekin reviewed for me last week.

spec/models/claim/matching_attribute_finder_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+        bank_sort_code: "972654",
+        building_society_roll_number: "123456789/ABCD",
+      )
+    }
+
+    let!(:claim_with_no_matching_attributes) { create(:claim, :submitted) }
+    let!(:unsubmitted_claim_with_matching_teacher_reference_number) { create(:claim, :submittable, teacher_reference_number: source_claim.teacher_reference_number) }
+
+    let(:claim_with_matching_attribute) { build(:claim, :submitted) }
+    let(:claims_with_matching_attributes) { Claim::MatchingAttributeFinder.new(source_claim).claim_ids_with_matching_attribute_keys }
```
lawrence-dxw: We could make it clear that this is the subject of the test instead of some piece of setup:

```suggestion
    subject(:claims_with_matching_attributes) { Claim::MatchingAttributeFinder.new(source_claim).claim_ids_with_matching_attribute_keys }
```

spec/models/claim/matching_attribute_finder_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+        bank_account_number: "34682151",
+        bank_sort_code: "972654",
+        building_society_roll_number: "123456789/ABCD",
+      )
+    }
+
+    let!(:claim_with_no_matching_attributes) { create(:claim, :submitted) }
+    let!(:unsubmitted_claim_with_matching_teacher_reference_number) { create(:claim, :submittable, teacher_reference_number: source_claim.teacher_reference_number) }
+
+    let(:claim_with_matching_attribute) { build(:claim, :submitted) }
```
lawrence-dxw: I got a bit confused when I read this `let` because, as it stands, it is _not_ a claim with a matching attribute; it only becomes one when you modify it inside the `it` blocks. Perhaps it would be clearer to create this claim each time you need it, inside the `it`?

spec/models/claim/matching_attribute_finder_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+        national_insurance_number: "QQ891011C",
+        email_address: "genghis.khan@mongol-empire.com",
+        bank_account_number: "34682151",
+        bank_sort_code: "972654",
+        building_society_roll_number: "123456789/ABCD",
+      )
+    }
+
+    let!(:claim_with_no_matching_attributes) { create(:claim, :submitted) }
+    let!(:unsubmitted_claim_with_matching_teacher_reference_number) { create(:claim, :submittable, teacher_reference_number: source_claim.teacher_reference_number) }
```
lawrence-dxw: :+1: for the name!

spec/models/claim/matching_attribute_finder_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+
+    it "only includes submitted claims" do
+      expect(claims_with_matching_attributes.keys).not_to include(unsubmitted_claim_with_matching_teacher_reference_number.id)
+    end
+
+    it "includes the id of a claim with a mathching teacher reference number" do
+      claim_with_matching_attribute.teacher_reference_number = source_claim.teacher_reference_number
+      claim_with_matching_attribute.save
+
+      expect(claims_with_matching_attributes.keys).to include(claim_with_matching_attribute.id)
```
lawrence-dxw: Maybe this reads a bit nicer:

```suggestion
      expect(claims_with_matching_attributes).to have_key(claim_with_matching_attribute.id)
```

spec/models/claim/matching_attribute_finder_spec.rb
```diff
@@ -0,0 +1,83 @@
+require "rails_helper"
+
+RSpec.describe Claim::MatchingAttributeFinder do
+  describe "#claim_ids_with_matching_attributes" do
```
lawrence-dxw: This doesn't match the name of the method.

app/models/claim/matching_attribute_finder.rb
```diff
@@ -0,0 +1,41 @@
...
+      "bank_sort_code",
+      "building_society_roll_number",
+    ].freeze
+
+    def initialize(source_claim, claims_to_compare = Claim.submitted)
+      @source_claim = source_claim
+      @claims_to_compare = claims_to_compare
+    end
+
+    def claim_ids_with_matching_attribute_keys
```
lawrence-dxw: `matching_attributes_keyed_by_claim_id` might make it clearer what are the keys and what are the values.

app/models/claim/matching_attribute_finder.rb
```diff
@@ -0,0 +1,41 @@
+class Claim
+  class MatchingAttributeFinder
```
lawrence-dxw: I like the implementation of this class; before I started reading it I had no idea how I would implement it if it were me; and this is really nice and readable.

app/views/admin/claims/show.html.erb
```diff
@@ -32,6 +38,32 @@
...
+        <thead class="govuk-table__head">
+          <tr class="govuk-table__row">
+            <th scope="col" class="govuk-table__header">Claim</th>
+            <th scope="col" class="govuk-table__header">Matching details</th>
+          </tr>
+        </thead>
+        <tbody class="govuk-table__body">
+          <% @claim_ids_with_matching_attributes.each do |claim_id, matching_attributes| %>
+            <tr class="govuk-table__row">
+              <th scope="row" class="govuk-table__header"><%= link_to Claim.find(claim_id).reference, admin_claim_path(Claim.find(claim_id)), class: "govuk-link" %>
```
lawrence-dxw: For each matching claim, this makes a query to load the claim so that we can display its reference. But, when forming the list of matching claims, inside `Claim::MatchingAttributeFinder`, we already loaded the list of all the claims, and then discarded them, just returning the claims' IDs. Maybe we could change the interface of that class so that it returns the claim objects instead of / in addition to their IDs?

A separate point that will hopefully become moot: this loads each claim _twice_; once for the reference and another time for the URL.

spec/requests/admin_claims_spec.rb
```diff
@@ -20,10 +20,23 @@
   describe "claims#show" do
     let(:claim) { create(:claim, :submitted) }
 
-    it "returns a claim when one exists" do
-      get admin_claim_path(claim)
+    context "when a claim has no matching attributes" do
```
lawrence-dxw:
```suggestion
    context "when no other claim has matching attributes" do
```

spec/requests/admin_claims_spec.rb
```diff
@@ -20,10 +20,23 @@
...
+      it "returns the claim" do
+        get admin_claim_path(claim)
 
-      expect(response.body).to include(claim.reference)
+        expect(response.body).to include(claim.reference)
+      end
+    end
+
+    context "when a claim has matching attributes" do
```
lawrence-dxw:
```suggestion
    context "when another claim has matching attributes" do
```

spec/features/admin_matching_attributes_spec.rb
```diff
@@ -0,0 +1,28 @@
+require "rails_helper"
+
+RSpec.feature "Admin checking a claims" do
+  let(:user_id) { "userid-345" }
```
lawrence-dxw: I think that if we have no intention of varying the user ID or verifying that it is used anywhere, then we can get rid of this let and use the default value provided by `sign_in_to_admin_with_role`.

app/views/admin/claims/show.html.erb
```diff
@@ -10,6 +10,12 @@
       </div>
     <% end %>
 
+    <% if @claim_ids_with_matching_attributes.present? %>
```
lawrence-dxw: @tekin picked up something similar on a PR of mine: given that we know that `@claim_ids_with_matching_attributes` is a collection (something hash-like), I think it would be clearer if we had

```suggestion
    <% if @claim_ids_with_matching_attributes.any? %>
```

(same a few lines down, too)

spec/models/claim/matching_attribute_finder_spec.rb
mec: 🤦‍♂

spec/models/claim/matching_attribute_finder_spec.rb
mec: Much nicer with `update!` thank you.

spec/models/claim/matching_attribute_finder_spec.rb
mec: Yes! And I have a horrible feeling it will fail and match!

mec:
@lawrence-dxw @tekin I am going to have to stop now, I couldn't get all this done and I am on dxw support from tomorrow so I won't be able to pick up the rest of the fixes.

I think I have a problem with empty building society numbers matching as per @lawrence-dxw: https://github.com/DFE-Digital/dfe-teachers-payment-service/pull/556#discussion_r340172255

There is a generally better approach from @tekin that I beleive will resolve @lawrence-dxw other comments as well.

Not sure who will pick this up, but thank you for doing so!

spec/models/claim/matching_attribute_finder_spec.rb
lawrence-dxw: We've added a new spec and fixed this.

spec/models/claim/matching_attribute_finder_spec.rb
pezholio: This probably isn't worth doing for now, especially given the fact that it's an internal screen

spec/features/admin_matching_attributes_spec.rb
```diff
@@ -0,0 +1,31 @@
+require "rails_helper"
+
+RSpec.feature "Admin checking a claims" do
```
tekin: Can this be reworded to better describe the problem we are trying to solve with this feature? e.g. `"Service operator can see potential duplicate claims"`. Same goes for the filename. As a new developer to the project, if I read `admin_matching_attributes_spec.eb`, what does that say to me about the feature? Could the file be named to better express the intent of the feature?

spec/helpers/admin/claims_helper_spec.rb
```diff
@@ -173,4 +173,24 @@
...
+        :claim,
+        teacher_reference_number: "0902344",
+        national_insurance_number: "QQ891011C",
+        email_address: "genghis.khan@mongol-empire.com",
+        bank_account_number: "34682151",
+        bank_sort_code: "972654",
+        building_society_roll_number: "123456789/ABCD",
+      )
+    }
+    let(:second_claim) { build(:claim, :submitted, teacher_reference_number: first_claim.teacher_reference_number, national_insurance_number: first_claim.national_insurance_number) }
```
tekin: good to see `build` and not `create` used here 👍

spec/models/claim/matching_attribute_finder_spec.rb
tekin: This is no longer the name of the method under test.

tekin: I'm wondering if we could increase the confidence that these specs give us if we switched from using `include` to using `eq` instead and match the actual array of claims we expect. What do you think?

spec/requests/admin_claims_spec.rb
```diff
@@ -20,10 +20,23 @@
   describe "claims#show" do
     let(:claim) { create(:claim, :submitted) }
 
-    it "returns a claim when one exists" do
-      get admin_claim_path(claim)
+    context "when no other claim has matching attributes" do
```
tekin: I think it would be ok to leave this `it` block as it is. Wrapping it in the additional context just makes it look a bit strange and not really make sense without knowledge of the context that comes after it. Whereas the context after makes sense on it's own without the need to have this it in an opposite context.

app/models/claim/matching_attribute_finder.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    def initialize(source_claim, claims_to_compare = Claim.submitted)
+      @source_claim = source_claim
+      @claims_to_compare = claims_to_compare
+    end
+
+    def matching_claims
+      predicates = []
+      values = []
+      ATTRIBUTES_TO_MATCH.each do |attribute|
```
pezholio: I think it's OK to put the SQL in directly here (so we don't need the `predicates` and `values`) as the data is already stored in the database, so is already sanitised. So you could do something like:

```ruby
query = ATTRIBUTES_TO_MATCH.each do |attribute|
  value = @source_claim.read_attribute(attribute)
  next unless value

  "#{attribute} ='#{value}'"
end

@claims_to_compare.where.not(id: @source_claim.id).where(query)
```

lawrence-dxw: Hmm, I don't think so. What I mean is that the Ruby values need to be made suitable for insertion into an SQL query, that is, they need to be typecast (e.g. a `Date` needs to be turned into the DB's representation of a date) and stuff that shouldn't be in an SQL query (e.g. quotes) need to be escaped. We can't just drop the values in like that.

tekin: Yes, what @lawrence-dxw said, we need to let ActiveRecord escape any user-provided values that we use in SQL queries, otherwise we open ourselves up to potential SQL-injection attacks.

app/helpers/admin/claims_helper.rb
```diff
@@ -68,6 +68,14 @@ def check_deadline_warning(claim)
       content_tag(:strong, pluralize(days_until_check_deadline, "day"), class: "govuk-tag #{check_deadline_warning_class}")
     end
 
+    def matching_attributes(first_claim, second_claim)
```
tekin: Looks like this helper will includes building society roll numbers that are `nil` as matches. Can we update it to exclude those?

app/models/claim/matching_attribute_finder.rb
pezholio: Ah, I guess. I didn't think about type casting here.

spec/features/admin_matching_attributes_spec.rb
lawrence-dxw: Agreed - I'll change it to something more descriptive.

spec/models/claim/matching_attribute_finder_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+      expect(matching_claims).not_to include(claim_with_no_matching_attributes)
+    end
+
+    it "only includes submitted claims" do
+      expect(matching_claims).not_to include(unsubmitted_claim_with_matching_teacher_reference_number)
+    end
+
+    it "includes a claim with a matching teacher reference number" do
+      claim_with_matching_attribute.teacher_reference_number = source_claim.teacher_reference_number
+      claim_with_matching_attribute.save
```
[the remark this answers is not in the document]
lawrence-dxw: Hmm, I feel like a change has gone missing somewhere - we previously got rid of `claim_with_matching_attribute` as a `let`. I'll see what went wrong.

spec/requests/admin_claims_spec.rb
lawrence-dxw: Agreed; I'll change it.

spec/models/claim/matching_attribute_finder_spec.rb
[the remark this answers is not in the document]
lawrence-dxw: Not sure what happened, but have sorted it out.

app/helpers/admin/claims_helper.rb
lawrence-dxw: Good catch! Updated.

spec/helpers/admin/claims_helper_spec.rb
```diff
@@ -173,4 +173,29 @@
...
+    }
+    let(:second_claim) { build(:claim, :submitted, teacher_reference_number: first_claim.teacher_reference_number, national_insurance_number: first_claim.national_insurance_number, building_society_roll_number: first_claim.building_society_roll_number) }
+    subject { helper.matching_attributes(first_claim, second_claim) }
+
+    it "returns the humanised names of the matching attributes" do
+      expect(subject).to eq(["Building society roll number", "National insurance number", "Teacher reference number"])
+    end
+
+    it "does not consider nil building society roll number to be a match" do
+      [first_claim, second_claim].each { |claim| claim.update!(building_society_roll_number: nil) }
```
tekin: Could we avoid database writes with this?

```suggestion
      [first_claim, second_claim].each { |claim| claim.building_society_roll_number =  ni) }
```

lawrence-dxw: Good idea; updated.

PR #561, opened 2019-10-30
Backfill note on claims that have already been checked

pezholio:
This adds a data migration to backfill notes on claims that have already been checked. To get around the possibility of adding personal/sensitive data to the repo, the notes are stored as an envrionment variable in CSV format in the Azure keyvault. 

**This variable will have to be set before the development/production deploy happens**

Once the data migration has happened, we can remove the environment variable.

### DIFF
--- a/db/data/20191029113201_backfill_note_on_already_checked_claims.rb
+++ b/db/data/20191029113201_backfill_note_on_already_checked_claims.rb
@@ -1,0 +1,20 @@
+class BackfillNoteOnAlreadyCheckedClaims < ActiveRecord::Migration[5.2]
+  def up
+    return unless ENV["NOTES_TO_BACKFILL"]
+
+    rows = ENV["NOTES_TO_BACKFILL"].split("|")
+
+    rows.each do |row|
+      row = row.split(",")
+      claim = Claim.find_by_reference(row[0])
+
+      next if claim.nil? || claim.check.nil?
+
+      claim.check.update_column(:notes, row[1])
+    end
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -89,6 +89,7 @@ end
 group :test do
   gem "selenium-webdriver"
   gem "launchy"
+  gem "climate_control"
 end
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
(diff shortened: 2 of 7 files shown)


db/data/20191029113201_backfill_note_on_already_checked_claims.rb
```diff
@@ -0,0 +1,21 @@
+require "csv"
+
+class BackfillNoteOnAlreadyCheckedClaims < ActiveRecord::Migration[5.2]
+  def up
+    return unless ENV["NOTES_TO_BACKFILL"]
+
+    rows = CSV.parse(ENV["NOTES_TO_BACKFILL"])
```
pezholio: Open question, I'm wondering if this should be JSON, rather than CSV. I'm unsure if we might have problems storing multiline strings as environment variables in the keyvault. At least with JSON it can be one line.

lawrence-dxw: It would be helpful if we could find some documentation on it, or play around somehow to confirm the behaviour. Because if we go with JSON, there's also the possibility that Azure will do some parsing of the newline control string `\n`. Maybe the safest option would be Base64?

lawrence-dxw: Actually I may have misunderstood what your concern is about the newlines. Is it that Azure may do something funny with them, or that they're hard to represent in a CSV?

lawrence-dxw: (Also, if the data given in the Trello is everything, then I wonder if it's moot; none of them seem to contain newlines?)

lawrence-dxw: Okay, I think I should stop talking. I thought you were talking about newlines in an individual note; you mean the newlines that form the structure of the CSV, right?

pezholio: Yeah, that's right. I bear many scars of trying to deal with CSVs cross-platform, and I'm not sure if Azure will like only Windows-style newlines (`\r\n`) or Unix-style newlines (`\n`) or both.

db/data/20191029113201_backfill_note_on_already_checked_claims.rb
```diff
@@ -0,0 +1,22 @@
+require "csv"
```
tekin: This doesn't appear to be using the `CSV` class. Can this be ditched?

pezholio: Oh, good call

PR #562, opened 2019-10-30
Update Verify Service Provider to version 2.1.0

lawrence-dxw:


lawrence-dxw:
Ah, I didn't know about that endpoint. The firewall stuff wouldn't be a concern because I'm only thinking about local development there. But it would mean having to run the existing VSP just to check whether it's the correct version, which seems like more hassle than it's worth.

PR #564, opened 2019-10-30
Prepare service for external start page

tekin:
**NOTE: the start page URL has not been confirmed yet. That should be updated before this is merged**

Once we're live for public beta the start page will live outside of our service domain and the temporary start page we used during private beta will no longer be needed. This removes the temporary start page and sets up a redirect from the root of the domain to the start page as it will exist on GOV.UK once it is published.

As part of these changes a slight tweak was made to the behaviour for maintenance mode: we now return a 503 (service unavailable) instead of redirecting when in maintenance mode, which is better practice.

## Todo before merge

- [x] get the code reviewed
- [x] update `StudentLoans.start_page_url` with the correct start page URL.


robbiepaul:
I also wonder if this will cause a usability issue for developers. If we visit `https://localhost:3000` it will redirect us to GOV.UK. I think it's fine we just need to communicate it to the team that we should now open the service locally at `https://localhost:3000/student-loans/claim`

PR #577, opened 2019-11-04
Remove NOTES_TO_BACKFILL env var from Azure

pezholio:
We've backfilled notes now, so no longer need this variable.


pezholio:
Ah yeah, the spec definitely. I think we need to keep the migrations in place to maintain the history.

PR #579, opened 2019-11-04
Re-organise session handling code to separate claim session code from admin session code

tekin:
This mostly organises the code for admin sessions and public-facing user sessions such that they are entirely separate, the exception being the update_last_seen_at method, which updates the timestamp for the user regardless of the type of user they are.

A test was deleted as part of this mostly because it was a mostly false requirement - we do not expect admin users to be visiting the public-facing site, and if they do, it's not a major issue that their sign-in session will get extended because of it.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,7 +1,6 @@
 class ApplicationController < ActionController::Base
   CLAIM_TIMEOUT_LENGTH_IN_MINUTES = 30
   CLAIM_TIMEOUT_WARNING_LENGTH_IN_MINUTES = 2
-  ADMIN_TIMEOUT_LENGTH_IN_MINUTES = 30
 
   http_basic_authenticate_with(
     name: ENV["BASIC_AUTH_USERNAME"],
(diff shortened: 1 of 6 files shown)


app/controllers/application_controller.rb
```diff
@@ -1,33 +1,23 @@
...
 
-  def admin_signed_in?
-    session.key?(:user_id)
-  end
-
-  def admin_timeout_in_minutes
-    ADMIN_TIMEOUT_LENGTH_IN_MINUTES
-  end
-
   def claim_timeout_in_minutes
```
lawrence-dxw: I'm wondering if it still makes sense for the claim-related timeout code to remain in `ApplicationController`? It would seem more symmetrical for the claim journey controllers to have their own namespace and base class, the same way that the admin controllers do. This means that the admin controller won't have access to a load of methods that it never uses, making it easier to reason about.

Thinking out loud but there seems to be a bit of an asymmetry now; it seems like pages in the claim journey are treated as the default, and that admin pages are then a specialisation of this

tekin: Interesting suggestion. I've created a commit that extracts a separate base controller for the public-facing controllers: https://github.com/DFE-Digital/dfe-teachers-payment-service/commit/919d546d233c974ed1f5e04ed1b75bd414476020

I'm on the fence about which I prefer. In some ways, it's better because, as you say, admin controllers no longer have code in them that are not relevant to the admin interface. On the downside we've introduced more indirection and additional inheritance. what do you think?

tekin:
Merging this to unblock things, we can extract a separate base controller as a separate PR.

PR #580, opened 2019-11-04
Explain how to complete Verify flow in local dev

lawrence-dxw:
F had to explain this to me when I joined the project; hopefully this
will be useful for new developers.

### DIFF
--- a/docs/govuk-verify.md
+++ b/docs/govuk-verify.md
@@ -33,6 +33,23 @@ You can check that the VSP is running ok by hitting the healthcheck URL:
 curl localhost:50300/admin/healthcheck
 ```
 
+### How to complete the Verify user journey in local development
+
+After beginning the Verify flow, you will be redirected to a URL which looks
+something like
+`https://compliance-tool-reference.ida.digital.cabinet-office.gov.uk/SAML2/SSO`,
+and which will display a JSON object in your browser. To complete the Verify
+flow from here:
+
+1. Follow the `responseGeneratorLocation` URL from this JSON object.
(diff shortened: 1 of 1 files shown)


titlescreen:
LGTM, I understood your explanation of how to get through the local verify. I just wanted to check that there wasn't any local VSP setup that also needs mentioning in the docs?

lawrence-dxw:
I think that the local VSP setup is already explained pretty well in the "Running locally in development" part of the same document.

docs/govuk-verify.md
```diff
@@ -33,6 +33,23 @@ You can check that the VSP is running ok by hitting the healthcheck URL:
 curl localhost:50300/admin/healthcheck
 ```
 
+### How to complete the Verify user journey in local development
+
+After beginning the Verify flow, you will be redirected to a URL which looks
+something like
+https://compliance-tool-reference.ida.digital.cabinet-office.gov.uk/SAML2/SSO,
```
tekin: Visiting this URL returns a `405` error. Maybe we need to include the actual JSON instead?

lawrence-dxw: I wasn't intending that anybody follow this link; I just wanted to give an example of what the URL looks like, so that they know which part of the flow I'm talking about.

tekin: Ah, OK, I wonder if there is a way we can mark this so that it looks less like a visitable link and isn't interpreted as such in Markdown. Maybe wrap it in quote marks? Or backticks?

lawrence-dxw: Good idea - I've added backticks.

PR #584, opened 2019-11-06
Implement the new multiple schools journey

pezholio:
This implements the new multiple schools journey. Now if a teacher did not teach at an eligible school, they are given the opportunity to try again with another school that they may have taught at in the same period. We also change the flow so they are asked if they taught eligibile subjects relative to the school they taught in, if they didn't they are given the opportunity to choose another school where they may have taught eligible subjects.

image

image

image

image

image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog]
 ## [Unreleased]
 
 - Add a link to our service's satisfaction survey to the confirmation page
+- Improve journey for users that worked at more than one school during the claim
+  period
 
 ## [Release 028] - 2019-11-06
 
--- a/app/helpers/claims_helper.rb
+++ b/app/helpers/claims_helper.rb
@@ -73,4 +73,8 @@ module ClaimsHelper
       two_words_connector: connector
     )
   end
+
+  def school_search_question(searching_for_additional_school)
+    searching_for_additional_school ? t("student_loans.questions.additional_school") : t("student_loans.questions.claim_school")
+  end
 end
--- a/app/helpers/claims_helper.rb
+++ b/app/helpers/claims_helper.rb
@@ -19,7 +19,7 @@ module ClaimsHelper
       a << [t("student_loans.questions.qts_award_year"), I18n.t("student_loans.questions.qts_award_years.#{eligibility.qts_award_year}"), "qts-year"]
       a << [t("student_loans.questions.claim_school"), eligibility.claim_school_name, "claim-school"]
       a << [t("questions.current_school"), eligibility.current_school_name, "still-teaching"]
-      a << [t("student_loans.questions.subjects_taught"), subject_list(eligibility.subjects_taught), "subjects-taught"]
+      a << [t("student_loans.questions.subjects_taught", school: eligibility.claim_school_name), subject_list(eligibility.subjects_taught), "subjects-taught"]
       a << [t("student_loans.questions.leadership_position"), (eligibility.had_leadership_position? ? "Yes" : "No"), "leadership-position"]
       a << [t("student_loans.questions.mostly_performed_leadership_duties"), (eligibility.mostly_performed_leadership_duties? ? "Yes" : "No"), "mostly-performed-leadership-duties"] if eligibility.had_leadership_position?
     end
--- a/app/models/student_loans/eligibility.rb
+++ b/app/models/student_loans/eligibility.rb
@@ -21,8 +21,8 @@ module StudentLoans
       SUBJECT_ATTRIBUTES,
     ].flatten.freeze
     ATTRIBUTE_DEPENDENCIES = {
-      "claim_school_id" => "employment_status",
-      "had_leadership_position" => "mostly_performed_leadership_duties",
+      "claim_school_id" => ["taught_eligible_subjects", *SUBJECT_ATTRIBUTES, "employment_status"],
+      "had_leadership_position" => ["mostly_performed_leadership_duties"],
     }.freeze
 
     self.table_name = "student_loans_eligibilities"
@@ -87,8 +87,10 @@ module StudentLoans
     end
 
     def reset_dependent_answers
-      ATTRIBUTE_DEPENDENCIES.each do |attribute_name, dependent_attribute_name|
-        write_attribute(dependent_attribute_name, nil) if changed.include?(attribute_name)
+      ATTRIBUTE_DEPENDENCIES.each do |attribute_name, dependent_attribute_names|
+        dependent_attribute_names.each do |dependent_attribute_name|
+          write_attribute(dependent_attribute_name, nil) if changed.include?(attribute_name)
+        end
       end
       self.current_school = inferred_current_school if employment_status_changed?
     end
--- a/app/views/claims/_ineligibility_reason_ineligible_claim_school.html.erb
+++ b/app/views/claims/_ineligibility_reason_ineligible_claim_school.html.erb
@@ -10,6 +10,9 @@
 
 <p class="govuk-body">
   If you taught at more than one school during this period, you can
-  <%= link_to "start your claim again", new_claim_path, class: "govuk-link" %> with another school.
+  search again with the next school.
   You only need to have worked at one eligible school to claim.
 </p>
+
+<%= link_to "Enter another school", claim_path(policy: current_policy_routing_name, slug: "claim-school", additional_school: true), class: "govuk-button" %>
+<%= link_to "I've tried all of my schools", claim_path(policy: current_policy_routing_name, slug: "ineligible", generic: true), class: "govuk-button govuk-button--secondary" %>
--- a/app/views/claims/_ineligibility_reason_ineligible_claim_school.html.erb
+++ b/app/views/claims/_ineligibility_reason_ineligible_claim_school.html.erb
@@ -1,18 +1,22 @@
-<h1 class="govuk-heading-xl">
-  This school is not eligible
-</h1>
+<% if params[:no_more_schools] %>
+  <%= render partial: "ineligibility_reason_no_more_schools" %>
+<% else %>
+  <h1 class="govuk-heading-xl">
+    This school is not eligible
+  </h1>
 
-<p class="govuk-body">
-  <%= claim_school_name %> is not an eligible school. You can only get this
-  payment if you  were employed at an eligible school between 6 April 2018
-  and 5 April 2019.
-</p>
+  <p class="govuk-body">
+    <%= claim_school_name %> is not an eligible school. You can only get this
+    payment if you  were employed at an eligible school between 6 April 2018
+    and 5 April 2019.
+  </p>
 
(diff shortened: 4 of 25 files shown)


app/helpers/claims_helper.rb
```diff
@@ -73,4 +73,12 @@ def subject_list(subjects)
       two_words_connector: connector
     )
   end
+
+  def claim_school_title(searching_for_additional_school)
```
tekin: Feels like this method could have a more appropriate name than "claim school title". Is it more a question that happens to be used in the title?

app/helpers/claims_helper.rb
```diff
@@ -73,4 +73,12 @@ def subject_list(subjects)
...
+    searching_for_additional_school ? t("student_loans.questions.additional_school") : t("student_loans.questions.claim_school")
+  end
+
+  def school_search_value(searching_for_additional_school, current_claim)
```
tekin: Maybe there's a more intention-revealing name for this helper? how about `school_search_prefill_value`?

app/models/student_loans/eligibility.rb
```diff
@@ -21,8 +21,8 @@ class Eligibility < ApplicationRecord
       SUBJECT_ATTRIBUTES,
     ].flatten.freeze
     ATTRIBUTE_DEPENDENCIES = {
-      "claim_school_id" => "employment_status",
-      "had_leadership_position" => "mostly_performed_leadership_duties",
+      "claim_school_id" => ["employment_status", "taught_eligible_subjects", *SUBJECT_ATTRIBUTES],
```
tekin: 👍 I wonder if we should switch the order of these so that they follow the same order as the pages in the sequence?

app/models/student_loans/eligibility.rb
```diff
@@ -71,7 +71,9 @@ def ineligible?
         not_taught_enough?
     end
 
-    def ineligibility_reason
+    def ineligibility_reason(tried_all_schools = nil)
```
tekin: I feel like this change subverts this model method and couples it a bit more to the journey than before, which might not be that helpful. Rather than modifying this method on the model, how about if we just have an `if` in the ineligibility template, something like:

```ruby
if params[:tried_all_schools]
  # render the generic reason
else
  # render the specific one
end
```

That keeps the nitty gritty of the evolving journey out of this method, which to know has only been about telling us why the thing is ineligible without coupling it to the thing the user sees in the UI.

app/views/claims/_ineligibility_reason_all_schools_ineligible.html.erb
```diff
@@ -0,0 +1,3 @@
+<h1 class="govuk-heading-xl">
```
tekin: If we rename this to something more generic it could be more easily reused in other circumstances where a generic ineligible reason needs to be given.

app/views/claims/_ineligibility_reason_ineligible_claim_school.html.erb
```diff
@@ -10,6 +10,9 @@
 
 <p class="govuk-body">
   If you taught at more than one school during this period, you can
-  <%= link_to "start your claim again", new_claim_path, class: "govuk-link" %> with another school.
+  search again with the next school.
   You only need to have worked at one eligible school to claim.
 </p>
+
+<%= link_to "Enter another school", claim_path(policy: current_policy_routing_name, slug: "claim-school", additional_school: true), class: "govuk-button" %>
+<%= link_to "I've tried all of my schools", claim_path(policy: current_policy_routing_name, slug: "ineligible", tried_all_schools: true), class: "govuk-button govuk-button--secondary" %>
```
tekin: As with the template, if we use a more generic name than `:tried_all_schools` then we make the code for displaying a generic ineligible message more re-useable.

app/views/claims/_school_search.html.erb
```diff
@@ -37,7 +37,7 @@
           </span>
         </p>
 
-        <% if school_id_param == :claim_school_id %>
+        <% if school_id_param == :claim_school_id && params[:additional_school].blank? %>
```
tekin: The use case for `blank?` is when the variable might be a blank string. In this case, it's either a parameter that is present and `true` or it's not, so `params[:additional_school]` will suffice.

pezholio:
[Implement the new Multiple Schools journey](https://trello.com/c/q9NotGIu/871-implement-the-new-multiple-schools-journey)

app/helpers/claims_helper.rb
pezholio: Yeah, you're right on that. This came as part of a refactor, and I may have got a little helper-happy

app/views/claims/_ineligibility_reason_all_schools_ineligible.html.erb
pezholio: Cool - I've just renamed it to "_ineligible" and then render it in the schools ineligibility partial if the `tried_all_schools` param is set, which also solves some of the issues you highlighted below.

spec/features/subjects_taught_spec.rb
```diff
@@ -3,6 +3,8 @@
 RSpec.feature "Choosing subjects taught during Teacher Student Loan Repayments claims" do
   before do
     start_claim
+    choose_school schools(:penistone_grammar_school)
+
     visit claim_path("subjects-taught")
```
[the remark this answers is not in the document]
pezholio: Ah yeah, good call. I initially added the above line because we needed a claim_school to be able to show the subjects-taught page, but then we get redirected to that page anyway.

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- Allow users to add another school that they may also have worked at, after
+  previously entering an ineligible school or subjects.
```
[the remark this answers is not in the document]
pezholio: Much better, ta. I did this at the end of the day and my brain was failing to form words 😆

spec/features/multiple_claim_schools_spec.rb
```diff
@@ -4,37 +4,84 @@
...
-  scenario "first claim school is ineligible, but subsequent school is eligible" do
-    claim = start_claim
-    choose_school schools(:hampstead_school)
+  let!(:claim) { start_claim }
 
-    expect(claim.eligibility.reload.claim_school).to eq schools(:hampstead_school)
-    expect(page).to have_text("This school is not eligible")
-    expect(page).to have_text("Hampstead School is not an eligible school.")
+  context "first claim school is ineligible" do
+    let!(:ineligible_school) { create(:school) }
```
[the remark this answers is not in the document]
pezholio: Yeah, fair enough. I think you're right. It also means we don't have to have expectations in the `before` block, which feels like an anti-pattern.

spec/features/multiple_claim_schools_spec.rb
```diff
@@ -4,37 +4,84 @@
...
 
-  scenario "first claim school is ineligible, but subsequent school is eligible" do
-    claim = start_claim
-    choose_school schools(:hampstead_school)
+  let!(:claim) { start_claim }
 
-    expect(claim.eligibility.reload.claim_school).to eq schools(:hampstead_school)
-    expect(page).to have_text("This school is not eligible")
-    expect(page).to have_text("Hampstead School is not an eligible school.")
+  context "first claim school is ineligible" do
```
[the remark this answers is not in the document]
pezholio: Yeah, fair enough. I think you're right. It also means we don't have to have expectations in the `before` block, which feels like an anti-pattern.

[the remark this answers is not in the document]
tekin: 👍🏽 as a general rule, I think it's good to optimise specs for comprehension and ease of use, even if that means more verbose specs with duplication in them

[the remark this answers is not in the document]
pezholio: Cool - I'm pretty new to using Rspec feature specs. I've always used Cucumber / Turnip in the past

app/views/claims/_ineligibility_reason_ineligible_claim_school.html.erb
```diff
@@ -1,15 +1,22 @@
-<h1 class="govuk-heading-xl">
-  This school is not eligible
-</h1>
+<% if params[:tried_all_schools] %>
```
tekin: I was actually thinking this would go in the main ineligible view, rather than in these two partials. So like this basically: https://github.com/DFE-Digital/dfe-teachers-payment-service/commit/f59a0c2461a9627db039dfd3e694a3143015b77b

spec/features/multiple_claim_schools_spec.rb
```diff
@@ -0,0 +1,92 @@
+require "rails_helper"
+
+RSpec.feature "Applicant worked at multiple schools" do
+  let(:eligibility) { claim.eligibility }
+
+  let!(:claim) { start_claim }
+
+  context "first claim school is ineligible" do
```
tekin: So I think we can go a step further and get rid of these contexts altogether.

spec/features/multiple_claim_schools_spec.rb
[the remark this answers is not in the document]
tekin: So I think we can go a step further and actually get rid of these contexts altogether. 4 totally independent and carefully worded `scenario`s would be easy to follow and understand without having to jump around and read the outer contexts.

app/views/claims/_ineligibility_reason_no_more_schools.erb
```diff
@@ -0,0 +1,7 @@
+<h1 class="govuk-heading-xl">
+  You're not eligible for this payment
+</h1>
+
+<p class="govuk-body">
+  You must have been employed by at least one eligible school and taught eligible subjects to claim between 6 April 2018 and 5 April 2019
```
tekin:
```suggestion
  You must have been employed by at least one eligible school and taught eligible subjects to claim between 6 April 2018 and 5 April 2019.
```

app/views/claims/_school_search.html.erb
```diff
@@ -33,7 +33,7 @@
         Enter the school name. Use at least four characters.
       </span>
 
-      <% if school_id_param == :claim_school_id %>
+      <% if school_id_param == :claim_school_id && !params.has_key?(:additional_school) %>
```
tekin: Technically, `has_key?` is deprecated in favour of `key?`, but doesn't look like it'll be be removed any time soon according to https://github.com/rubocop-hq/rubocop/issues/3224.

pezholio: Yeah, I clocked that, but I think `has_key?` is more expressive

tekin: Fine, if you want to argue with Matz's opinion... ;)

tekin: Also, we currently call `key?` 9 times in the codebase and `has_key?` zero... just saying!

PR #588, opened 2019-11-06
Fix flakey current school search spec

robbiepaul:
The diff for this PR isn't very helpful as we included a small refactor to untangle things slightly.

The diff for the fix can be seen in 93578335ef27fc2cbe86d659dcc1350fc17a7528

We're not sure what's causing the error to happen, some combination of headless chrome not liking the `<p class="govuk-body">` tag wrapping the hint. Changing the paragraph's class to anything else fixes the error. Adding a sleep before the failing expectation also fixes the error.

We felt that this solution was smallest change that we could get to pass consistently.

We had concerns about the way that screen readers might read the two span tags that are back to back, but we checked it in VoiceOver and it made sense.

### DIFF
--- a/app/views/claims/_school_search.html.erb
+++ b/app/views/claims/_school_search.html.erb
@@ -1,114 +1,54 @@
-<% if @schools.blank? %>
+<%= render("shared/error_summary", instance: current_claim, errored_field_id_overrides: { school_search: "school_search" }) if current_claim.errors.any? %>
 
-  <%= render("shared/error_summary", instance: current_claim, errored_field_id_overrides: { school_search: "school_search" }) if current_claim.errors.any? %>
+<%= form_for current_claim,
+             url: claim_path,
+             method: :get,
+             data: {
+               "school-id-param": "claim_eligibility_attributes_#{school_id_param}",
+               "exclude-closed": exclude_closed
+             },
+             html: { class: "school_search_form" } do |form|
+%>
+  <%= hidden_field_tag :_method, "get", id: nil %>
+  <%= hidden_field_tag :authenticity_token, form_authenticity_token %>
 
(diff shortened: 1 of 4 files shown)


mec:
[Fix broken spec in test suite](https://trello.com/c/GsFazH47/931-fix-broken-spec-in-test-suite)

app/views/claims/_school_search.html.erb
```diff
@@ -31,18 +31,14 @@
           <strong>No results match that search term. Try again.</strong>
         </p>
       <% else %>
-        <p class="govuk-body">
-          <span class="govuk-hint">
-            Enter the school name. Use at least four characters.
-          </span>
-        </p>
+        <span class="govuk-hint">
```
pezholio: Do these still show on seperate lines? Or are they munged together?

robbiepaul: Yeah we double checked and it appears exactly the same as two paragraph tags (with a gap between) 👍

mec: I thought I'd get the definative answer on this, set to `display: block` here:

https://github.com/alphagov/govuk-frontend/blob/2f29eadef4850627c0b76cb57fb05cc51be2fce3/src/govuk/components/hint/_hint.scss#L9

lawrence-dxw:
I wonder whether this could be considered a bug with the Design System, or, least, something which is going to catch out other developers using it? If so, would we be able to produce a minimal failing example (or point them to our code, if we can't do that) and let them know about it?

app/views/claims/_school_search.html.erb
pezholio: Cool 👍

PR #589, opened 2019-11-06
Upcase existing National Insurance numbers

lawrence-dxw:
This is the data migration mentioned in 7e50b1e.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- Change all of the existing claims' National Insurance numbers to upper case
 - Add a link to our service's satisfaction survey to the confirmation page
 
 ## [Release 028] - 2019-11-06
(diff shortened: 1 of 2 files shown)


lawrence-dxw:
I wanted to run a single SQL query which updates all the records, so I just used raw SQL, and I `ActiveRecord::Base.connection.execute` is the only way that I know of executing a raw SQL query. But I didn't look into whether there is way of writing it with Active Record. I'll see if I can find anything.

mec:
No! I like it, I just wondered you reached for it specifically or arrived there by another route.

lawrence-dxw:
Ah, okay! Well, anyway, I did find a more Rails-y way to write it, so thanks for the question. I've updated the PR.

mec:
Wow! Yeah that is a rails-y way! 👍 You got the standardrb blues though…

lawrence-dxw:
D'oh, fixed 🙂

PR #592, opened 2019-11-07
Redirect to start of claim journey in development

pezholio:
At the moment, in development, when a dev goes to `http://localhost:3000`, or a claim times out, they're redirected to the GOV.UK start page, which is disorienting. This redirects the user to the start of the claim journey if we're in development.

### DIFF
--- a/app/models/student_loans.rb
+++ b/app/models/student_loans.rb
@@ -1,6 +1,10 @@
 module StudentLoans
   def self.start_page_url
-    "https://www.gov.uk/guidance/teachers-claim-back-your-student-loan-repayments"
+    if Rails.env.production?
+      "https://www.gov.uk/guidance/teachers-claim-back-your-student-loan-repayments"
+    else
+      "/#{routing_name}/claim"
+    end
   end
 
   def self.routing_name
--- a/README.md
+++ b/README.md
@@ -36,8 +36,7 @@ Architecture decision records can be found in the
      section.
 2. Run `bundle install` and `yarn install` to install the dependencies
 3. Run `bundle exec rails db:setup` to set up the database development
-4. Run `bundle exec foreman start` to launch the app on
-   https://localhost:3000/student-loans/claim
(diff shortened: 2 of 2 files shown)


lawrence-dxw:
It's certainly a good idea to not send the user to a production site in development. However, I wonder if it might be clearer to set up a development-only route to a fake start page hosted by the app, so that the flow is consistent between the environments.

pezholio:
Yeah, it's a good point, but I dunno if it's worth maintaining too much code that is only for use in development. It'll mean setting up a new template and controller action too

lawrence-dxw:
I think it's worth it, because ❤️ consistency. I don't think that having an extra controller is a very high price to pay. But I'd like to hear other opinions because everyone values different things.

tekin:
When Maths & Physics is introduced we won't be able to redirect to a start page from the root URL as there will be two start pages, at which point this becomes redundant. At that point we'll likely introduce a new page at the root with some content directing the user to either policy's start page. Given that, it's probably not worth sinking too much time into this right now.

app/models/student_loans.rb
```diff
@@ -1,6 +1,10 @@
 module StudentLoans
   def self.start_page_url
-    "https://www.gov.uk/guidance/teachers-claim-back-your-student-loan-repayments"
+    if Rails.env.production?
+      "https://www.gov.uk/guidance/teachers-claim-back-your-student-loan-repayments"
+    else
+      "student-loans/claim"
```
robbiepaul:
```suggestion
      "/student-loans/claim"
```
Otherwise you'll get a `No route matches [GET] "/student-loans/student-loans/claim"` error when clicking the service link at the top of the page

robbiepaul: We could even do `"/#{self.routing_name}/claim"` to make it a little more dry

pezholio: Like it 👍

PR #593, opened 2019-11-07
Set appropriate reply to address when sending notify emails

robbiepaul:
We're about to add the Maths and Physics policy to the service and will need to set a different reply-to address for each policy.

GOV.UK Notify allows us to create a whitelist of "Reply-to" addresses under `Settings > Reply-to email addresses`.

### DIFF
--- a/app/models/student_loans.rb
+++ b/app/models/student_loans.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module StudentLoans
   def self.start_page_url
     "https://www.gov.uk/guidance/teachers-claim-back-your-student-loan-repayments"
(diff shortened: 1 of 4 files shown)


app/models/student_loans.rb
```diff
@@ -6,4 +8,8 @@ def self.start_page_url
   def self.routing_name
     "student-loans"
   end
+
+  def self.notify_reply_to_id
+    "962b3044-cdd4-4dbe-b6ea-c461530b3dc6"
```
pezholio: Good place for it 👍

spec/mailers/claim_mailer_spec.rb
```diff
@@ -1,13 +1,24 @@
 require "rails_helper"
 
+RSpec.shared_examples "a claim mailer" do |policy|
```
pezholio: 🌟

PR #594, opened 2019-11-07
Release 029

mec:
- Make admin claim search case insensitive
- Change all of the existing claims' National Insurance numbers to upper case
- Add a link to our service's satisfaction survey to the confirmation page
- Improve journey for users that worked at more than one school during the claim
  period

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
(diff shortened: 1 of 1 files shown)


mec:
We are going to do some acceptance tests on the mulitple schools journey before releasing the code to production.

mec:
@robbiepaul yeah, I am worried about that one as well - I'll check in with everyone once we think the feature is passed accepatance tests.

PR #596, opened 2019-11-07
Speed up the changing answers spec

pezholio:
Before, we were completing the student flow completely before each scenario. I've tweaked this now, so we create a submittable claim, and then visit the check-your-answers page. I've also cut down on the amount of nested scenarios, which speeds things up further.

This cuts down the amount of time the spec takes to run from ~20 seconds to ~5 seconds, and takes our total test suite run time down from ~1 minute 20s seconds, to ~57 seconds (so under 1 minute! 🙌 )

### DIFF
--- a/spec/features/changing_answers_spec.rb
+++ b/spec/features/changing_answers_spec.rb
@@ -5,7 +5,10 @@ RSpec.feature "Changing the answers on a submittable claim" do
   let(:eligibility) { claim.eligibility }
 
   before do
-    answer_all_student_loans_claim_questions
+    claim = start_claim
+    claim.update!(attributes_for(:claim, :submittable))
+    claim.eligibility.update!(attributes_for(:student_loans_eligibility, :eligible))
+    visit claim_path("check-your-answers")
   end
 
   scenario "Teacher can edit a field" do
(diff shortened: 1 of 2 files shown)


pezholio:
[Speed up changing answers feature spec](https://trello.com/c/GWTQ2f3U/974-speed-up-changing-answers-feature-spec)

spec/features/changing_answers_spec.rb
```diff
@@ -36,215 +36,142 @@
     expect(current_path).to eq(claim_path("check-your-answers"))
   end
 
-  context "when changing subjects taught" do
-    before do
-      find("a[href='#{claim_path("subjects-taught")}']").click
-    end
+  scenario "Teacher changes their reference number" do
```
robbiepaul:
```suggestion
  scenario "Teacher changes the subjects they taught" do
```

pezholio:
@mec Yeah, good call - it seems kinda irelevant given that there's only two choices now

spec/features/changing_answers_spec.rb
mec: Is this is actually checking subjects taught?

spec/support/feature_helpers.rb
```diff
@@ -1,33 +1,9 @@
...
-    click_on "Continue"
-
-    fill_in I18n.t("questions.email_address"), with: "name@example.tld"
-    click_on "Continue"
-
-    fill_in "Name on the account", with: "Jo Bloggs"
-    fill_in "Sort code", with: "123456"
-    fill_in "Account number", with: "87654321"
-    click_on "Continue"
+  def submittable_claim
```
robbiepaul: I wonder if this needs to be a feature helper now? It could just be inlined in the `let!` block

mec: Could we have used the `:submittable` factory instead?

pezholio: Good call - it's only used in this spec, so saves looking around 👍

pezholio: That was my inital thought, but we need the claim to be present in the session, so we need to start a claim, and then update that claim. I've got a better solution coming now though, thanks to a prompt from @robbiepaul

robbiepaul: I also ran a quick test to see if there was any improvements using a factory then doing `allow_any_instance_of(ClaimController).to receive(:current_claim) { claim }` 

I know we're not keen on stubbing controller methods on feature specs but I was curious if there would be a performance boost, but there was none over the proposed implementation 👍

pezholio: Ah, that's good to know 👍 I do prefer staying as close as possible reality in feature specs, so it's nice that we're not slowing anything down too much doing it the right way

spec/features/changing_answers_spec.rb
```diff
@@ -28,223 +31,151 @@
...
 
-    choose I18n.t("student_loans.questions.qts_award_years.on_or_after_september_2013")
+    choose I18n.t("student_loans.questions.qts_award_years.before_september_2013")
     click_on "Continue"
 
-    expect(eligibility.reload.qts_award_year).to eq("on_or_after_september_2013")
+    expect(eligibility.reload.qts_award_year).to eq("before_september_2013")
 
-    expect(current_path).to eq(claim_path("check-your-answers"))
+    expect(page).to have_text("You’re not eligible")
```
robbiepaul: nice change 👍

spec/support/feature_helpers.rb
mec: Gotya! Thanks!

PR #597, opened 2019-11-07
Re-organise routing in preparation for Maths & Physics

tekin:
This should be the last PR laying the groundwork for introducing the Maths & Physics policy to the service. Several changes have been made to make it possible to drop in additional policies:

- Introduced the `Policies` so that we have a global list of available policies to work with
- `routing.rb` now generates routes for multiple policies, based on what is defined in `Policies`
- views can be organised by policy, with policy specific ones living in their own folder
- the hardcoded links to the Student Loans start page have been made dynamic


config/routes.rb
```diff
@@ -21,11 +21,31 @@
 
   get "refresh-session", to: "sessions#refresh", as: :refresh_session
 
-  scope path: ":policy", constraints: {policy: %r{student-loans}} do
-    constraints slug: %r{#{StudentLoans::SlugSequence::SLUGS.join("|")}} do
-      resources :claims, only: [:show, :update], param: :slug, path: "/"
+  # Used to constrain claim journey routing so only slugs
+  # that are part of a policy’s slug sequence are routed.
+  class RestrictToSequenceSlugs
```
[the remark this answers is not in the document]
tekin: You can define a class anywhere! The code for it just gets evaluated, just like every other line of Ruby code. And if the place where you've defined it is the only place you use it, then it's probably fine.

The reason we generally put classes in their own files and name those files based on the class name is a) so we have a conventional way of locating the source file, and b) so that the [Rails autoloading magic](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html) knows how to load a class if it hasn't been explicitly `required (by default, Ruby doesn't magically load classes for you; that's entirely a Rails thing).

In this case, this class is very specifically being used here in the routes. It's never going to be called/used outside of this file, so there isn't much to be gained by sticking it in a separate file. Another reason why it might have made sense to put this class in its own file would be if we had unit tests around it, but I opted to rely on the routing specs to verify the behaviour, and because of the higher-level coverage those provide, it didn't feel useful to have lower level unit tests for this class.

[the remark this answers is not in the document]
mec: Excellent! Thanks for explaining!

app/models/policies.rb
```diff
@@ -0,0 +1,19 @@
...
+  def self.all
+    POLICIES
+  end
+
+  # Returns a policy that matches the provided routing name
+  # For example:
+  #
+  #   Policies["student-loans"] #=> StudentLoans
+  #
+  def self.[](routing_name)
```
pezholio: Super useful, this 👍

config/routes.rb
```diff
@@ -21,11 +21,30 @@
 
   get "refresh-session", to: "sessions#refresh", as: :refresh_session
 
-  scope path: ":policy", defaults: {policy: "student-loans"}, constraints: {policy: %r{student-loans}} do
-    constraints slug: %r{#{StudentLoans::SlugSequence::SLUGS.join("|")}} do
-      resources :claims, only: [:show, :update], param: :slug, path: "/"
+  # Used to constrain claim journey routing so only slugs
+  # that are part of a policy’s slug sequence are routed.
+  class RestrictToSequenceSlugs
```
pezholio: I'm a bit iffy about having the class inline in the routing block. Would it be OK to have it up top, so it makes the routes easier to read?

tekin: Personally I like having the class near to where it is being used as it places it closer to the context in which it needs to be understood.

pezholio: No drama. Happy for it to stay where it is in that case 👍

PR #600, opened 2019-11-11
Upgrade to Rails 6.0.1 and Sprockets 4

pezholio:
This upgrades Rails to the latest version (6.0.1), after a fix was merged in and released that prevented us from upgrading. 

Sprockets needed upgrading to version 4 manually, as Rails has switched to using webpack for assets by default.

I've given this a good test in dev and it seems to be working as expected, but I think this will need a thorough test in staging before we push it live.


pezholio:
[Upgrade to Rails 6](https://trello.com/c/JatvkEHa/807-upgrade-to-rails-6)

PR #607, opened 2019-11-11
A service operator can download a csv file to generate a report from DQT

mec:
This PR adds a simple way for service operators to download a csv file that contains the `reference` and `teacher_reference_number` for claims that are `awaiting checking`.

The file is sent securely to the DQT team who return a report file with all the relevent details for claims checking, this saves a service operator from having to sign in to DQT and check each record manually as they can just reference the report.

There is the opportunity to extend this feature to ingest the report, but this is not part of this PR and would need to be scoped and prioritised.

I wanted this to be as simple as possible so I just used a new `format` on the admin#index action of the controller as this already loads the correct claims.

From there I introduced a new `Claim::DatabaseOfQualifiedTeachersReportRequest` class which converts `Claim` attributes to csv for exporting. I wanted to keep the knowledge for which attributes are exported inside this class as opposed to the controller as this felt more natural.

## Screenshot
Screenshot_2019-11-12 Claim additional payments for teaching – GOV UK

### DIFF
--- a/app/models/claim/database_of_qualified_teachers_report_request.rb
+++ b/app/models/claim/database_of_qualified_teachers_report_request.rb
@@ -1,0 +1,24 @@
+require "csv"
+
+class Claim
+  class DatabaseOfQualifiedTeachersReportRequest
+    ATTRIBUTES = {
+      reference: "Claim reference",
+      teacher_reference_number: "Teacher reference number",
+    }.freeze
+
+    def initialize(claims)
+      @claims = claims
+    end
+
+    def to_csv
+      CSV.generate do |csv|
+        csv << ATTRIBUTES.values
+
+        @claims.each do |claim|
+          csv << ATTRIBUTES.keys.map { |attribute| claim.read_attribute(attribute) }
+        end
+      end
+    end
+  end
+end
--- a/spec/features/admin_dqt_report_request_spec.rb
+++ b/spec/features/admin_dqt_report_request_spec.rb
@@ -1,0 +1,20 @@
+require "rails_helper"
+
+RSpec.feature "DQT Report request" do
+  scenario "Service operator can download CSV for the Database of Qualified Teachers report request" do
+    sign_in_to_admin_with_role(AdminSession::SERVICE_OPERATOR_DFE_SIGN_IN_ROLE_CODE)
+
+    claims = create_list(:claim, 3, :submitted)
+
+    click_on "View claims"
+
+    click_on "Download DQT report request file"
+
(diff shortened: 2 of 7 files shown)


mec:
[A Service Operator can generate a list of claim references and their TRNs to get a report from DQT](https://trello.com/c/NewbQfPh/923-a-service-operator-can-generate-a-list-of-claim-references-and-their-trns-to-get-a-report-from-dqt)

app/models/claim/export.rb
```diff
@@ -0,0 +1,31 @@
+class Claim
+  class Export
+    ATTRIBUTES_FOR_DATABSE_OF_QUALIFIED_TEACHERS_REPORT = {
+      reference: "Reference number",
+      teacher_reference_number: "Teacher reference number",
+      policy: "Policy name",
```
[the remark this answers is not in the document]
mec: The file we have been sending to the DQT team included it so I decided to also include it as well, with hindsight we probably should not as it work/complexity for nothing.

app/models/claim/database_of_qualified_teachers_report_request.rb
```diff
@@ -0,0 +1,24 @@
...
+      reference: "Claim reference",
+      teacher_reference_number: "Teacher reference number",
+    }.freeze
+
+    def initialize(claims)
+      @claims = claims
+    end
+
+    def to_csv
+      CSV.generate(headers: true) do |csv|
```
tekin: Does the `:headers` option actually do anything on the `generate` method?

```suggestion
      CSV.generate do |csv|
```

spec/features/admin_dqt_report_request_spec.rb
```diff
@@ -0,0 +1,20 @@
+require "rails_helper"
+
+RSpec.feature "User is logged in as a service operator can" do
```
tekin:
```suggestion
RSpec.feature "DQT Report request" do
```

spec/features/admin_dqt_report_request_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  scenario "download a csv for the Database of Qualified Teachers report request" do
```
tekin:
```suggestion
  scenario "Service operator can download CSV for the Database of Qualified Teachers report request" do
```

app/models/claim/database_of_qualified_teachers_report_request.rb
mec: I thought it did, but it clearly doesn't here.

PR #608, opened 2019-11-11
Introduce the Maths & Physics journey

tekin:
Finally, after the many preparatory PRs, here is the beginnings of the Maths & Physics journey.

It introduces the beginnings of the `Eligibility` model and`SlugSequence` as well as a temporary start page that can be used during private beta. Only the first question is implemented in the PR, the rest of the journey can be implemented iteratively from this base.

The Maths & Physics journey will be protected using basic auth; all that needs to happen is to set the appropriate environment variables.

### DIFF
--- a/app/views/maths_and_physics/claims/teaching_maths_or_physics.html.erb
+++ b/app/views/maths_and_physics/claims/teaching_maths_or_physics.html.erb
@@ -1,0 +1,45 @@
+<% content_for(:page_title, page_title(t("maths_and_physics.questions.teaching_maths_or_physics"), policy: current_policy_routing_name, show_error: current_claim.errors.any?)) %>
+<% path_for_form = current_claim.persisted? ? claim_path(current_policy_routing_name) : claims_path(current_policy_routing_name) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render("shared/error_summary", instance: current_claim, errored_field_id_overrides: { "eligibility.teaching_maths_or_physics": "claim_eligibility_attributes_teaching_maths_or_physics_true" }) if current_claim.errors.any? %>
+    <%= form_for current_claim, url: path_for_form  do |form| %>
+      <%= form_group_tag current_claim do %>
+        <%= form.fields_for :eligibility, include_id: false do |fields| %>
+
+          <%= fields.hidden_field :teaching_maths_or_physics %>
+
+          <fieldset class="govuk-fieldset">
+
+            <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
+              <h1 class="govuk-fieldset__heading">
+                <%= t("maths_and_physics.questions.teaching_maths_or_physics") %>
+              </h1>
+            </legend>
+
+            <%= errors_tag current_claim.eligibility, :teaching_maths_or_physics %>
+
+            <div class="govuk-radios govuk-radios--inline">
+
+              <div class="govuk-radios__item">
+                <%= fields.radio_button(:teaching_maths_or_physics, true, class: "govuk-radios__input") %>
+                <%= fields.label :teaching_maths_or_physics_true, "Yes", class: "govuk-label govuk-radios__label" %>
+              </div>
+
+              <div class="govuk-radios__item">
+                <%= fields.radio_button(:teaching_maths_or_physics, false, class: "govuk-radios__input") %>
+                <%= fields.label :teaching_maths_or_physics_false, "No", class: "govuk-label govuk-radios__label" %>
+              </div>
+
+            </div>
+
+          </fieldset>
+
+        <% end %>
+      <% end %>
+
+      <%= form.submit "Continue", class: "govuk-button", data: {module: "govuk-button"} %>
+    <% end %>
+  </div>
+</div>
--- a/app/views/static_pages/maths_and_physics.html.erb
+++ b/app/views/static_pages/maths_and_physics.html.erb
@@ -1,0 +1,52 @@
+<% content_for(:page_title, t("maths_and_physics.policy_name")) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">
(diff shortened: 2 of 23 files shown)


config/routes.rb
```diff
@@ -62,6 +62,8 @@ def matches?(request)
       resources :school_search, only: [:create]
     end
   end
+  # Temporary start page for Maths & Physics
+  get "maths-and-physics/start", to: "static_pages#maths_and_physics", defaults: {policy: "maths-and-physics"}
```
pezholio: Do we need to set up a controller action for this?

app/views/static_pages/maths_and_physics.html.erb
```diff
@@ -0,0 +1,52 @@
...
+      </li>
+      <li>the date you completed your initial teacher training</li>
+      <li>
+        sign in details for ‘Verify’ — the government’s secure way of proving
+        who you are, if you do not have a Verify account, you’ll need your
+        passport or photocard driving licence to sign up
+      </li>
+    </ul>
+
+    <%= link_to "Start now", new_claim_path("maths-and-physics"), class: "govuk-button govuk-button--start", role: "button" %>
```
mec: Does this need the arrow? I think it's inline now so the styles don't add it.

https://design-system.service.gov.uk/components/button/

It could also do with `data-module="govuk-button"` like the other comment.

app/views/maths_and_physics/claims/teaching_maths_or_physics.html.erb
```diff
@@ -0,0 +1,45 @@
...
+              </div>
+
+            </div>
+
+          </fieldset>
+
+        <% end %>
+      <% end %>
+
+      <%= form.submit "Continue", class: "govuk-button" %>
```
mec: I am still convinced we should include `data-module="govuk-button"` on these to properly fire the JS accessibility shim. 

I believe it works by building a list of govuk-buttons on the page from the `data-module` attribute:

https://github.com/alphagov/govuk-frontend/blob/2529efcbdcc66294694d3cd7d86246c0083414f4/src/govuk/all.js#L22

and then checks for `role-button` as well.

https://github.com/alphagov/govuk-frontend/blob/2529efcbdcc66294694d3cd7d86246c0083414f4/src/govuk/components/button/button.js#L21

config/routes.rb
tekin: Technically no, because of [default rendering of controller actions](https://guides.rubyonrails.org/layouts_and_rendering.html#rendering-by-default-convention-over-configuration-in-action). We've defined the methods for the other static pages mainly to aid with visibility and awareness, but they're not technically necessary either. I skipped it for this one because it is a temporary page that will be removed in the next couple months. I can add it in though if you think it would be helpful?

pezholio: Yeah, I think that's what threw me. I'd rather we went with one approach or the other TBH

spec/features/maths_and_physics_claim_spec.rb
```diff
@@ -0,0 +1,12 @@
+require "rails_helper"
+
+RSpec.feature "Maths & Physics claims" do
+  [true, false].each do |javascript_enabled|
```
mec: Could this also be:`%w[enabled disabled].each do |javascript_enabled|`?

app/views/maths_and_physics/claims/teaching_maths_or_physics.html.erb
```diff
@@ -0,0 +1,45 @@
+<% content_for(:page_title, page_title(t("maths_and_physics.questions.teaching_maths_or_physics"), policy: current_policy_routing_name, show_error: current_claim.errors.any?)) %>
+<% path_for_form = current_claim.persisted? ? claim_path(current_policy_routing_name) : claims_path(current_policy_routing_name) %>
```
mec: This was so subtle it confused me for a while!

app/views/static_pages/maths_and_physics.html.erb
tekin: I figured it wasn't worth using the arrow here given the [actual start pages we end up with](https://www.gov.uk/guidance/teachers-claim-back-your-student-loan-repayments) don't have them.

app/views/maths_and_physics/claims/teaching_maths_or_physics.html.erb
tekin: Thanks, will add it!

tekin: Do we need something in the backlog to update all the rest of the buttons in the same way?

spec/features/maths_and_physics_claim_spec.rb
tekin: We still need a `true` or `false` value for lie 6 (i.e. `js: javascript_enabled`)

app/views/maths_and_physics/claims/teaching_maths_or_physics.html.erb
tekin: yeah, it is subtle... any suggestions for how to make it less confusing?

app/views/maths_and_physics/claims/teaching_maths_or_physics.html.erb
mec: Would be good to check, maybe as part of an accessibility sweep – what do you think @titlescreen

spec/features/maths_and_physics_claim_spec.rb
mec: Ah. I see, thank you.

app/views/maths_and_physics/claims/teaching_maths_or_physics.html.erb
mec: Not sure, I think it's one of the Rails things you need to get used to decoding.

app/views/maths_and_physics/claims/teaching_maths_or_physics.html.erb
titlescreen: Yeah that is a good idea, I think it would be worthwhile making something in the backlog and look at this and accessibility when we have an end to end journey. What do you think?

PR #612, opened 2019-11-12
Improvements to the generate payroll interface

robbiepaul:
- Service operators are shown when the next payroll run file is expected to be uploaded by
- Only allow payroll runs to be created if none exist already for the current month

## Screenshots

Screenshot 2019-11-12 at 16 31 38
Screenshot 2019-11-12 at 16 32 05

### DIFF
--- a/app/helpers/admin/payroll_run_helper.rb
+++ b/app/helpers/admin/payroll_run_helper.rb
@@ -1,0 +1,14 @@
+module Admin
+  module PayrollRunHelper
+    def next_payroll_file_to_cantium_due_date
+      next_payrollable_month = PayrollRun.this_month.exists? ? Date.today.next_month.all_month : Date.today.all_month
+      all_fridays_in_month(next_payrollable_month)[-3]
+    end
+
+    private
+
+    def all_fridays_in_month(month)
+      month.select { |day| day.friday? }
+    end
+  end
+end
--- a/app/models/payroll_run.rb
+++ b/app/models/payroll_run.rb
@@ -4,6 +4,10 @@ class PayrollRun < ApplicationRecord
 
   validates :created_by, presence: true
 
+  validate :ensure_no_payroll_run_this_month, on: :create
+
+  scope :this_month, -> { where(created_at: DateTime.now.all_month) }
+
   def total_award_amount
     claims.sum(&:award_amount)
   end
@@ -16,5 +20,11 @@ class PayrollRun < ApplicationRecord
         end
       end
     end
+  end
+
+  private
+
+  def ensure_no_payroll_run_this_month
+    errors.add(:base, "There has already been a payroll run for #{Date.today.strftime("%B")}") if PayrollRun.this_month.any?
   end
 end
--- a/app/views/admin/payroll_runs/index.html.erb
+++ b/app/views/admin/payroll_runs/index.html.erb
@@ -5,9 +5,13 @@
     </h1>
   </div>
   <div class="govuk-grid-column-two-thirds">
-    <p class="govuk-body">Make a payroll file containing all the relevant information for approved claims to be paid by a payroll provider.</p>
+    <p class="govuk-body">Make a payroll file containing all the relevant information for approved claims to be paid by Cantium.</p>
 
-    <%= link_to 'Prepare payroll', new_admin_payroll_run_path, class: "govuk-button", data: {module: "govuk-button"} if PayrollRun.this_month.empty? %>
+    <div class="govuk-inset-text">
+      The next payroll file must be with Cantium on or before <%= l(next_payroll_file_to_cantium_due_date) %>
+    </div>
+
+    <%= link_to 'Prepare payroll file', new_admin_payroll_run_path, class: "govuk-button", data: {module: "govuk-button"} if PayrollRun.this_month.empty? %>
   </div>
   <div class="govuk-grid-column-full">
     <table class="govuk-table">
--- a/app/views/admin/payroll_runs/index.html.erb
+++ b/app/views/admin/payroll_runs/index.html.erb
@@ -15,7 +15,7 @@
       </strong>
     </div>
 
-    <%= link_to 'Prepare payroll', new_admin_payroll_run_path, class: "govuk-button", data: {module: "govuk-button"} %></li>
+    <%= link_to 'Prepare payroll', new_admin_payroll_run_path, class: "govuk-button", data: {module: "govuk-button"} if PayrollRun.this_month.empty? %>
   </div>
   <div class="govuk-grid-column-full">
     <table class="govuk-table">
--- a/app/views/admin/payroll_runs/index.html.erb
+++ b/app/views/admin/payroll_runs/index.html.erb
@@ -7,14 +7,6 @@
   <div class="govuk-grid-column-two-thirds">
     <p class="govuk-body">Make a payroll file containing all the relevant information for approved claims to be paid by a payroll provider.</p>
 
-    <div class="govuk-warning-text">
-      <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
-      <strong class="govuk-warning-text__text">
-        <span class="govuk-warning-text__assistive">Warning</span>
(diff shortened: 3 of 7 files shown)


robbiepaul:
[Improvements to the "generate payroll" interface](https://trello.com/c/hNr5Z9Yh/875-improvements-to-the-generate-payroll-interface)

app/views/admin/payroll_runs/index.html.erb
```diff
@@ -7,15 +7,11 @@
...
     <p class="govuk-body">Make a payroll file containing all the relevant information for approved claims to be paid by a payroll provider.</p>
 
-    <div class="govuk-warning-text">
-      <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
-      <strong class="govuk-warning-text__text">
-        <span class="govuk-warning-text__assistive">Warning</span>
-        Only one payroll file should be generated and submitted per calendar month
-      </strong>
+    <div class="govuk-inset-text">
+      The next payroll file should be uploaded by <%= l(next_payroll_file_upload_date) %>
```
tekin: The word "Upload" also appears in the payroll history, maybe we should use different words here to not confuse sending the file to Cantium with uploading the payment confirmation report?

```suggestion
      The next payroll file should be sent to Cantium by <%= l(next_payroll_file_upload_date) %>
```

tekin: Also, if the file has been generated for the month, does it still make sense to show this message?

mec: If the button is showing on the first of the month I may be tempted to (or accidentally) press it, not knowing that would have an impact on that months payroll run. 

Perhaps we should show the button at a sensible date __before__ the cantium deadline? If that was the behaviour, it would be helpful to show the user why there is no button (it's not time yet) and when there will be a button to press (something like two working days before the canitum deadline, which I am believe is the third friday of the month) or do we feel like that is too restrictive?

tekin: I think this is over-thinking it a bit. Pressing the button early would only have an impact if the user then also clicked the "Create payroll file" button on the subsequent page and actually created the payroll run. That screen also shows the number of approved claims and the value of those claims, which should give a strong signal that it might be too early.

mec: Okay, well I've updated it to calculate the third friday of the next month, which makes it this:

Screenshot_2019-11-13 Claim additional payments for teaching – GOV UK

There was a comment on the card about showing something around the schedule. Do we want to keep it or loose it?

mec: Looks like it might be the second Friday:

https://drive.google.com/file/d/0B4DnBBYceUWQd2xXZDlxVjI4RC1fS2xzMXEwWGJmWHhKS19R/view

tekin: Ah yeah, it looks slightly more subtle than that even. It looks like it's the end of the second full week of the month. So some month's it's the third Friday, and some the second. Perhaps we should speak to Cantium to get this nailed down?

mec: Sounds good. I just need to get the schedule clear in my head so I know what I am asking.

tekin: Let me know off I can help with that.

mec: I spoke with Chris and he is pretty clear:

- the 'pay day Friday' for claimants is the second to last friday of the month
- Cantium get five full workings days to perform the payroll
- the file is due to Cantium the Friday before the 'pay day Friday'
- I am not worrying about bank holidays as I feel like service operators can deal with those and things could change by then

app/helpers/admin/payroll_run_helper.rb
```diff
@@ -0,0 +1,13 @@
+module Admin
+  module PayrollRunHelper
+    def next_payroll_file_to_cantium_due_date(current_date = Date.today)
+      all_fridays_next_month(current_date)[-3]
+    end
+
+    private
+
+    def all_fridays_next_month(current_date)
+      current_date.next_month.beginning_of_month.step(current_date.next_month.end_of_month).select { |day| day.friday? }
```
pezholio: I always think `(start_date..end_date)` is more expressive than using `step`. Might be a bit more readable to store next month in a variable too I reckon.

```ruby
next_month = current_date.next_month
(next_month.beginning_of_month..next_month.end_of_month).select { |day| day.friday? }
```

app/models/payroll_run.rb
```diff
@@ -4,6 +4,10 @@ class PayrollRun < ApplicationRecord
 
   validates :created_by, presence: true
 
+  validate :ensure_no_payroll_run_this_month, on: :create
+
+  scope :this_month, -> { where(created_at: DateTime.now.all_month) }
```
pezholio: I didn't know about `all_month`. That's cool

mec: All @robbiepaul! 😄

app/helpers/admin/payroll_run_helper.rb
mec: Nice, these are the ruby things I like to learn, thank you!

mec:
As discussed with @tekin I'm updating this so the date shown to service operators is dependant on whether a payroll run has happened in the current month:

- if there is no payroll run this month, the due date is the third to last Friday of **this month**
- if a payroll run run happend this month, the due date is the third to last Friday of **next month**

I've chosen not to do anything to the due date if it's in the past as I do not feel we have any real knowlege of that as a process yet, the date is just informational for service operators and what the 'due' means is up to them at the moment.

mec:
One of the things I've found hard on this is not calling methods inside another method, something like:

`all_fridays_in_month(next_payrollable_month(Date.today))`

I don't even know if this is cool or not in Ruby/Rails land or what the alternative is?

The other has been passing around a date that really represents a month, but I can't see a way for the api to better reflect months without adding complexity where I don't think it is really needed?

So:

`all_fridays_in_month(date)` would make more sense as `all_fridays_in_month(month)`

but the argument is a date not a month so feels wrong.

Appeciate any thoughts or advice on these.

app/helpers/admin/payroll_run_helper.rb
```diff
@@ -0,0 +1,14 @@
...
+  module PayrollRunHelper
+    def next_payroll_file_to_cantium_due_date
+      next_payrollable_month = PayrollRun.this_month.exists? ? Date.today.next_month : Date.today
+      all_fridays_in_month(next_payrollable_month)[-3]
+    end
+
+    private
+
+    def all_fridays_in_month(date)
+      (date.beginning_of_month..date.end_of_month).select { |day| day.friday? }
```
pezholio: For this I was thinking about putting `next_payrollable_month` in a method, as you had before

```ruby
def next_payrollable_month
      target_date = PayrollRun.this_month.any? ? Date.today.next_month : Date.today
      (target_date.beginning_of_month..target_date.end_of_month)
end
```

That way, you end up returning a `Range`, which probably more accurately represents a whole month, then you can get all the fridays in the month like so:

```ruby
def all_fridays_in_month(month)
  month.select { |day| day.friday? }
end
```

mec: I do like it actually being a month that is passed, I'm now here:

````
  module PayrollRunHelper
    def next_payroll_file_to_cantium_due_date
      next_payrollable_month = PayrollRun.this_month.exists? ? Date.today.next_month.all_month : Date.today.all_month
      all_fridays_in_month(next_payrollable_month)[-3]
    end

    private

    def all_fridays_in_month(month)
      month.select { |day| day.friday? }
    end
  end
````

pezholio: Ah, yeah, that old fella `all_month`, good call 👍

mec:
@tekin @pezholio thanks both for all the suggestions and feedback!

@pezholio if you are happy, this can go in?

PR #613, opened 2019-11-12
Add config to Dockerfile to make sassc portable

pezholio:
We had some issues in staging that meant the Docker image fell over during boot. The solution outlined in https://github.com/sass/sassc-ruby/issues/146#issuecomment-534197168 should fix it so we're not relying on the underlying system that built the image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,6 +36,7 @@ COPY Gemfile ${DEPS_HOME}/Gemfile
 COPY Gemfile.lock ${DEPS_HOME}/Gemfile.lock
 
 RUN gem install bundler
+ENV BUNDLE_BUILD__SASSC=--disable-march-tune-native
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -36,6 +36,7 @@ COPY Gemfile ${DEPS_HOME}/Gemfile
 COPY Gemfile.lock ${DEPS_HOME}/Gemfile.lock
 
 RUN gem install bundler
+RUN bundle config --local build.sassc --disable-march-tune-native
```
tekin: Would setting the an environment variable, as outlined in https://github.com/sass/sassc-ruby/issues/146#issuecomment-542288556, be a neater solution?

```suggestion
ENV BUNDLE_BUILD__SASSC=--disable-march-tune-native
```

PR #615, opened 2019-11-13
Service operator can configure maintenance mode per policy

lawrence-dxw:
# Overview

This removes the global maintenance mode and replaces it with a per-policy maintenance configuration, which a service operator can change within a new Maintenance section in the admin interface.

We also remove the default maintenance message (see commit message for more details).

# Screenshots

Admin maintenance page:

image

Admin edit policy configuration page:

image

Admin maintenance page after policy put into maintenance mode:

image

Maintenance mode for a policy (this looks the same as the existing screen; just including it for context):

image


app/views/admin/policy_configurations/edit.html.erb
```diff
@@ -0,0 +1,54 @@
...
+      </div>
+
+      <div class="govuk-form-group">
+        <fieldset class="govuk-fieldset">
+          <legend class="govuk-fieldset__legend">
+            Service status
+          </legend>
+          <div class="govuk-radios">
+            <div class="govuk-radios__item">
+              <%= radio_button_tag :maintenance_mode, "0", !@policy_configuration.maintenance_mode?, name: "policy_configuration[maintenance_mode]", id: "policy_configuration_maintenance_mode_false", class: "govuk-radios__input" %>
```
tekin: The Rails form object has a more smart `radio_button` method that handles selecting the appropriate input based on the value of the record itself:

```suggestion
              <%= f.radio_button :maintenance_mode, false, class: "govuk-radios__input" %>
```

app/views/admin/policy_configurations/edit.html.erb
```diff
@@ -0,0 +1,54 @@
...
+              <%= f.label :maintenance_mode_false, "Open", class: "govuk-label govuk-radios__label" %>
+            </div>
+            <div class="govuk-radios__item">
+              <%= radio_button_tag :maintenance_mode, "1", @policy_configuration.maintenance_mode?, name: "policy_configuration[maintenance_mode]", id: "policy_configuration_maintenance_mode_true", class: "govuk-radios__input" %>
```
tekin:
```suggestion
              <%= f.radio_button :maintenance_mode, true, class: "govuk-radios__input" %>
```

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- Remove the default maintenance mode availability message
+- Remove the global maintenance mode, now that we can put individual services
+  into maintenance mode
+- A service operator can put an individual service into maintenance mode from
+  the admin interface
```
tekin: Can we reduce this to one entry to keep things short and sweet.

app/controllers/admin/policy_configurations_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+      policy_configuration.update!(policy_configuration_params)
+
+      redirect_to admin_policy_configurations_url
+    end
+
+    private
+
+    def policy_configuration_params
+      params.require(:policy_configuration).permit(:maintenance_mode_availability_message, :maintenance_mode).tap do |params|
+        params[:maintenance_mode_availability_message] = nil if params[:maintenance_mode_availability_message].blank?
```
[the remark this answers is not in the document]
lawrence-dxw: Yeah, I was wondering how this kind of thing is usually dealt with. But in that case, if the only time that the `availability_message` could be `nil` is before it is updated for the first time, then it probably makes sense to make the column non-nullable and give it a default value of `""`. What do you think?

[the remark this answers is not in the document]
tekin: I'd say it's fine leaving it as nullable in the database, `present?` and `blank?` mean that we can treat `nil` and a blank string as the same thing.

lawrence-dxw:
Good point re. the Azure key vault. I've created a [card](https://trello.com/c/8xK61GL6/1003-remove-maintenance-mode-variables-from-azure-key-vault) for it.

app/views/admin/policy_configurations/edit.html.erb
lawrence-dxw: Oh, lovely! I remember looking into the form helpers for radio buttons and am not sure why I ended up believing this wasn't possible. Thanks, changed it.

spec/models/policy_configuration_spec.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe PolicyConfiguration do
+  describe "#policy" do
+    it "returns the policy class" do
+      expect(policy_configurations(:student_loans).policy).to eq(StudentLoans)
```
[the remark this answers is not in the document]
lawrence-dxw: Sure thing, changed.

db/data/20191111115415_create_policy_configuration_records.rb
```diff
@@ -0,0 +1,10 @@
+class CreatePolicyConfigurationRecords < ActiveRecord::Migration[5.2]
+  def up
+    PolicyConfiguration.create!(policy_type: "StudentLoans")
```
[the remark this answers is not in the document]
lawrence-dxw: Ah, nice, thanks. Changed.

CHANGELOG.md
lawrence-dxw: Sure thing.

app/views/admin/policy_configurations/edit.html.erb
```diff
@@ -0,0 +1,54 @@
...
+        <h3 class="govuk-panel__title">Service closed</h3>
+        <p class="govuk-panel__body">This service is currently closed. Users can not submit claims.</p>
+      <% else %>
+        <h3 class="govuk-panel__title">Service open</h3>
+        <p class="govuk-panel__body">This service is currently open. Users can submit claims.</p>
+      <% end %>
+    </div>
+
+    <%= form_with model: [:admin, @policy_configuration] do |f| %>
+      <div class="govuk-form-group">
```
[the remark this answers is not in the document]
lawrence-dxw: I think that you're talking about ["conditionally revealing content"](https://design-system.service.gov.uk/components/radios/#conditionally-revealing-content) within the Radios pattern. I'll run it by Rich and see what he thinks, but it seems a sensible change to me.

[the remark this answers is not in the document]
titlescreen: Yeah I'm all in favour of conditionality, and showing the box only when it becomes relevant to what the user is trying to do. I honestly can't remember if we discussed this when we were working in Miro and didn't do it right now to keep it as simple as possible. 

However, it seems like a low cost thing to do that will make it nicer 👍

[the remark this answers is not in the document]
lawrence-dxw: Thanks, both. I've made that change. I think it looks better:

image

lawrence-dxw:
> One thing I am wondering is would the code look simpler/less surprising if we handled rendering the maintenance mode page in controller code, rather than as a routing constraint. Would it be worth seeing what that looks like?

As discussed, I've made this change, and we agree it looks nicer!

app/controllers/concerns/part_of_claim_journey.rb
```diff
@@ -12,6 +13,16 @@ def current_policy_routing_name
...
 
+  def check_for_maintenance_mode
+    policy = Policies[current_policy_routing_name]
+    policy_configuration = PolicyConfiguration.find_by(policy_type: policy.name)
+
+    if policy_configuration.maintenance_mode?
+      @availability_message = policy_configuration.maintenance_mode_availability_message
+      render "static_pages/maintenance", status: :service_unavailable
+    end
+  end
```
tekin: Now that we're no longer handling maintenance mode in the router, this action method is redundant as we only ever render maintenance mode from the `check_for_maintenance_mode` before action.

app/controllers/static_pages_controller.rb
```diff
@@ -18,7 +18,10 @@ def terms_conditions
   end
 
   def maintenance
```
tekin: Now that we're no longer handling maintenance mode in the router, this action method is redundant as we only ever render maintenance mode from the `check_for_maintenance_mode` before action.

lawrence-dxw: Good point! Fixed.

lawrence-dxw:
@tekin as discussed on Slack, I've changed the interface and the code to remove the concept of "maintenance" and to replace it with "close a service" (or "close a service for submissions"):

image

PR #617, opened 2019-11-14
Send statistics to Geckoboard

pezholio:
This PR sets up the necessary plumbing for us to record statistics in Geckoboard, using the [Geckoboard Datasets API](https://www.geckoboard.com/datasets/). We have two datasets:

* `claims.submitted.$ENV` - This gives us a record of all the claims that have been submitted, which will allow us to get a count of claims submitted during a particular time period, as well as track activity over time. This dataset is updated every time a user submits a claim.
* `claims.paid.$ENV` - This gives us a record of all the claims that have been paid, which will allow us to get a count of claims that have been paid during a particular time period, as well as track activity over time. This is updated every time a claim is paid.

PR #620, opened 2019-11-18
Send payment stats to Geckoboard

pezholio:
This is a follow-up PR to send data about claims that have been paid to Geckoboard. It uses the same class introduced in #617 to give us a record of all the claims that have been paid, which will allow us to get a count of claims that have been paid during a particular time period, as well as track activity over time. This is updated every time a claim is paid.

### DIFF
--- a/spec/jobs/record_payment_job_spec.rb
+++ b/spec/jobs/record_payment_job_spec.rb
@@ -1,0 +1,27 @@
+require "rails_helper"
+require "geckoboard"
+
+RSpec.describe RecordPaymentJob do
+  let(:claim) { build(:claim) }
+  let(:payment) { build(:payment, :with_figures, updated_at: DateTime.now, claim: claim) }
+
+  subject { described_class.new }
+
+  it "sends the claim reference, policy and payment date to claims.submitted.ENV dataset" do
+    ClimateControl.modify ENVIRONMENT_NAME: "environment_name" do
+      claim_data = {
+        reference: claim.reference,
+        policy: claim.policy.to_s,
+        performed_at: payment.updated_at.strftime("%Y-%m-%dT%H:%M:%S%:z"),
+      }
+
+      stub_geckoboard_dataset_find_or_create("claims.paid.environment_name")
+
+      dataset_post_stub = stub_geckoboard_dataset_post("claims.paid.environment_name")
+
+      subject.perform(claim)
+
+      expect(dataset_post_stub.with(body: {data: [claim_data]})).to have_been_requested
+    end
+  end
+end
--- a/spec/jobs/record_payment_job_spec.rb
+++ b/spec/jobs/record_payment_job_spec.rb
@@ -15,9 +15,7 @@ RSpec.describe RecordPaymentJob do
         performed_at: payment.updated_at.strftime("%Y-%m-%dT%H:%M:%S%:z"),
       }
 
-      stub_geckoboard_dataset_find_or_create("claims.paid.environment_name")
-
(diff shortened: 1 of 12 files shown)


spec/models/payment_confirmation_spec.rb
```diff
@@ -64,6 +65,21 @@
         expect(second_email.body.raw_source).to include(the_following_friday)
       end
     end
+
+    it "sends the each claim's reference, policy and payment date to Geckoboard" do
```
lawrence-dxw:
```suggestion
    it "sends each claim's reference, policy and payment date to Geckoboard" do
```

spec/support/geckoboard_helpers.rb
```diff
@@ -2,9 +2,9 @@ module GeckoboardHelpers
   # Use this to stub out the API calls that will be made to Geckoboard when a
   # Claim is submitted. It will return the webmock POST request, which can be
   # used to write expectations about the request being made.
-  def stub_geckoboard_submitted_event
-    stub_geckoboard_dataset_find_or_create("claims.submitted.test")
-    stub_geckoboard_dataset_post("claims.submitted.test")
+  def stub_geckoboard_submitted_event(dataset_id = "claims.submitted.test")
```
lawrence-dxw: The comment is outdated here, and I think that the method name might be too (although I'm not sure if `submitted` in the method name refers to a claim being submitted or data being submitted to Geckoboard).

spec/jobs/record_payment_job_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  it "sends the claim reference, policy and payment date to claims.submitted.ENV dataset" do
+    ClimateControl.modify ENVIRONMENT_NAME: "environment_name" do
+      claim_data = {
+        reference: claim.reference,
+        policy: claim.policy.to_s,
+        performed_at: payment.updated_at.strftime("%Y-%m-%dT%H:%M:%S%:z"),
+      }
+
+      stub_geckoboard_dataset_find_or_create("claims.paid.environment_name")
```
lawrence-dxw: `stub_geckoboard_dataset_post` already calls `stub_geckoboard_dataset_find_or_create`, so this call isn't necessary. (There are a few other places like this in the codebase where this call can be removed.)

spec/jobs/record_payment_job_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+require "geckoboard"
+
+RSpec.describe RecordPaymentJob do
+  let(:claim) { build(:claim) }
+  let(:payment) { build(:payment, :with_figures, updated_at: DateTime.now, claim: claim) }
+
+  subject { described_class.new }
+
+  it "sends the claim reference, policy and payment date to claims.submitted.ENV dataset" do
+    ClimateControl.modify ENVIRONMENT_NAME: "environment_name" do
```
lawrence-dxw: In some other specs, instead of fixing the `ENVIRONMENT_NAME` variable, you just check that the `claims.submitted.test` dataset is being used. I'm curious why you chose a different approach here?

pezholio: With this one, we're stubbing out the env to make sure that it can be changed, and there's not some side-effect somewhere else that's hardcoding the dataset name to have `test` in there. It's also trickier to stub the ENV in the feature specs, and we don't need to do it, as we've already got the confidence here.

pezholio:
Ready for another look now @lawrence-dxw

PR #623, opened 2019-11-18
Add current school question to maths and physics

mec:
This is the first PR in a series for maths and physics eligibility.

Obtains the `current_school` using the existing school search. Although I looked at the specs around the school search, I decided not to add additional tests for maths and physics as the school search is independant of which journey, but let me know if that was a bad call. 

Most of this is taken straight from student loans.

### DIFF
--- a/db/migrate/20191114141607_add_current_school_to_maths_and_physics_eligibilities.rb
+++ b/db/migrate/20191114141607_add_current_school_to_maths_and_physics_eligibilities.rb
@@ -1,0 +1,5 @@
+class AddCurrentSchoolToMathsAndPhysicsEligibilities < ActiveRecord::Migration[6.0]
+  def change
+    add_reference :maths_and_physics_eligibilities, :current_school, type: :uuid, foreign_key: {to_table: :schools}
+  end
+end
--- a/spec/models/maths_and_physics/eligibility_spec.rb
+++ b/spec/models/maths_and_physics/eligibility_spec.rb
@@ -31,6 +31,13 @@ RSpec.describe MathsAndPhysics::Eligibility, type: :model do
     end
   end
 
+  context "when saving in the “current-school” context" do
+    it "validates the presence of the current_school" do
+      expect(MathsAndPhysics::Eligibility.new).not_to be_valid(:"current-school")
+      expect(MathsAndPhysics::Eligibility.new(current_school: schools(:penistone_grammar_school))).to be_valid(:"current-school")
+    end
+  end
+
   context "when saving in the “submit” context" do
     it "is valid when all attributes are present" do
       expect(build(:maths_and_physics_eligibility, :eligible)).to be_valid(:submit)
@@ -39,6 +46,11 @@ RSpec.describe MathsAndPhysics::Eligibility, type: :model do
     it "is not valid without a value for teaching_maths_or_physics" do
       expect(build(:maths_and_physics_eligibility, :eligible, teaching_maths_or_physics: nil)).not_to be_valid(:submit)
       expect(build(:maths_and_physics_eligibility, :eligible, teaching_maths_or_physics: true)).to be_valid(:submit)
+    end
+
+    it "is not valid without a value for current_school" do
+      expect(build(:maths_and_physics_eligibility, :eligible, current_school: nil)).not_to be_valid(:submit)
+      expect(build(:maths_and_physics_eligibility, :eligible, current_school: schools(:penistone_grammar_school))).to be_valid(:submit)
     end
   end
 end
--- a/spec/models/maths_and_physics/eligibility_spec.rb
+++ b/spec/models/maths_and_physics/eligibility_spec.rb
@@ -10,6 +10,11 @@ RSpec.describe MathsAndPhysics::Eligibility, type: :model do
       expect(MathsAndPhysics::Eligibility.new(teaching_maths_or_physics: false).ineligible?).to eql true
       expect(MathsAndPhysics::Eligibility.new(teaching_maths_or_physics: true).ineligible?).to eql false
     end
+
+    it "returns true when teaching at an ineligble school" do
+      expect(MathsAndPhysics::Eligibility.new(current_school: schools(:hampstead_school)).ineligible?).to eql true
+      expect(MathsAndPhysics::Eligibility.new(current_school: schools(:penistone_grammar_school)).ineligible?).to eql false
+    end
   end
 
   describe "#ineligibility_reason" do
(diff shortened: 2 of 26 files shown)


mec:
[Build out the Eligibility part of the Maths & Physics claim journey](https://trello.com/c/0ly8FQov/924-build-out-the-eligibility-part-of-the-maths-physics-claim-journey)

mec:
[Build out the Eligibility part of the Maths & Physics claim journey](https://trello.com/c/0ly8FQov/924-build-out-the-eligibility-part-of-the-maths-physics-claim-journey)

app/views/claims/_school_search.html.erb
```diff
@@ -33,7 +33,7 @@
         Enter the school name. Use at least four characters.
       </span>
 
-      <% if school_id_param == :claim_school_id && !params.has_key?(:additional_school) %>
+      <% if current_policy_routing_name == "student-loans" && school_id_param == :claim_school_id && !params.has_key?(:additional_school) %>
```
pezholio: Do we want to have this added conditional here? I would have thought the journey for multiple schools would be the same for Maths and Physics as for Student Loans?

mec: The policy for Maths and Physics is that you are currently employed at an eligible school, so we don't need to ask for multiple schools, you have to still be working at one to claim.

The school search partial is re-used across the policies so we need this conditional to prevent mis-informing maths and physics users.

As an aside, I do think we could look to making the school search partials a bit neater, but I am not sure it should be part of this work?

tekin:
Should the user actually be providing their "current school", rather than their "claim school"? i.e. the school they are currently employed at?

tekin:
There's also already a perhaps entirely re-useable `claims/current_school` template.

app/views/claims/_school_search.html.erb
lawrence-dxw: @mec and I were talking about this a bit earlier. I noted that if we called this field `current_school` instead of `claim_school`, then the flow would indeed be the same as the Student Loans flow, and we wouldn't need the conditional.

But then, I wonder whether it even makes sense to prefix the school in Maths and Physics. There is only one school, so why not just call it `school`, I wonder?

mec:
Feel like I got the completely wrong end of the stick on this one, apologies @lawrence-dxw @pezholio.

Switched this to use `current_school` now.

mec:
For the ineligible feature spec I wanted a helper which meant re-naming the existing one, as part of this I updated a similar reuest spec helper even though I was not using it, it felt like it was part of the renaming though.

mec:
@lawrence-dxw I am hopeful I have them all now as well as the ineligiblity coverage. Thanks for the pointers and patience.

spec/models/maths_and_physics/eligibility_spec.rb
```diff
@@ -19,6 +24,29 @@
...
+      eligibility = MathsAndPhysics::Eligibility.new(current_school: schools(:penistone_grammar_school))
+      expect(eligibility.current_school_name).to eq schools(:penistone_grammar_school).name
+    end
+
+    it "does not error if the claim school is not set" do
+      expect(MathsAndPhysics::Eligibility.new.current_school_name).to be_nil
+    end
+  end
+
+  describe "#current_school_name" do
```
lawrence-dxw: The same method is being tested twice.

spec/models/maths_and_physics/eligibility_spec.rb
```diff
@@ -19,6 +24,29 @@
 
     it "returns a symbol indicating the reason for ineligibility" do
       expect(MathsAndPhysics::Eligibility.new(teaching_maths_or_physics: false).ineligibility_reason).to eq :not_teaching_maths_or_physics
+      expect(MathsAndPhysics::Eligibility.new(current_school: schools(:hampstead_school)).ineligibility_reason).to eq :ineligible_current_school
+    end
+  end
+
+  describe "#current_school_name" do
+    it "returns the name of the claim school" do
```
lawrence-dxw:
```suggestion
    it "returns the name of the current school" do
```

spec/models/maths_and_physics/eligibility_spec.rb
```diff
@@ -19,6 +24,29 @@
...
+      eligibility = MathsAndPhysics::Eligibility.new(current_school: schools(:penistone_grammar_school))
+      expect(eligibility.current_school_name).to eq schools(:penistone_grammar_school).name
+    end
+
+    it "does not error if the claim school is not set" do
```
lawrence-dxw:
```suggestion
    it "does not error if the current school is not set" do
```

spec/models/maths_and_physics/eligibility_spec.rb
```diff
@@ -19,6 +24,29 @@
 
     it "returns a symbol indicating the reason for ineligibility" do
       expect(MathsAndPhysics::Eligibility.new(teaching_maths_or_physics: false).ineligibility_reason).to eq :not_teaching_maths_or_physics
+      expect(MathsAndPhysics::Eligibility.new(current_school: schools(:hampstead_school)).ineligibility_reason).to eq :ineligible_current_school
+    end
+  end
+
+  describe "#current_school_name" do
```
lawrence-dxw: This method isn't actually used at the moment, because we don't display the name of the ineligible school. (But probably we should, same as student loans, I wonder why we don't?)

mec: I did have it in the `_ineligibility_reason_ineligible_current_school.html.erb` but decided to take it out in favour of replicating the mockup exactly. Do we think the copy on student loans it right?

https://github.com/DFE-Digital/dfe-teachers-payment-service/blob/master/app/views/student_loans/claims/_ineligibility_reason_ineligible_current_school.html.erb

spec/models/maths_and_physics/eligibility_spec.rb
mec: 🤦‍♂

spec/models/maths_and_physics/eligibility_spec.rb
mec: I thought `current_school_name` was used somewhere else, I remember looking:

https://github.com/DFE-Digital/dfe-teachers-payment-service/blob/master/app/views/claims/current_school.html.erb

But still happy to change the ineligble screen.

lawrence-dxw: Ah, sorry, you're totally right, the method needs to be there. And yeah, let's just replicate the wireframes for now.

lawrence-dxw: (Although @tekin suggests [here](https://github.com/DFE-Digital/dfe-teachers-payment-service/pull/623#issuecomment-555113496) we could reuse the question...)

mec: Yeah,  I think we've reused the question because the attribute is `current_school` so Rails looks for the `current_school.html.erb` template file.

The ineligibility reason is part of the `MathsAndPhysics::Eligibility` class though, so I think it has to live in `maths_and_physics/claim`. We might be able to share the partial in `claim` but would we be happy with that as there is seperation of eligibility/other claim views?

lawrence-dxw: Ah, yes, of course, I forgot about the ineligibility reason. I don't think we're aiming to share ineligibility partials at the moment.

db/migrate/20191114141607_add_current_school_to_maths_and_physics_eligibilities.rb
```diff
@@ -0,0 +1,5 @@
+class AddCurrentSchoolToMathsAndPhysicsEligibilities < ActiveRecord::Migration[6.0]
+  def change
+    add_reference :maths_and_physics_eligibilities, :current_school, type: :uuid, foreign_key: {to_table: :schools}
```
pezholio: Can we change the commit name to something like `Add current_school to Maths and Physics eligibility`?

PR #624, opened 2019-11-19
Backfill claim data in Geckoboard

pezholio:
This follows up from #617 and #620 to backfill existing data about historic claims in Geckoboard. As we don't currently have a way of running arbitary rake tasks on production, I've used Data Migrations for this. I'm unsure if this is a misuse of tools, but as we're using this to update data (even though said data doesn't exist in the database), I think this is fine. Totally up to have a discussion about alternatives though!


pezholio:
[Stats dashboard](https://trello.com/c/YHOpTt4x/934-stats-dashboard)

PR #626, opened 2019-11-19
Wire in the post-eligiblity Qs for Maths & Physics

tekin:
This will get the user as far as the check-your-answers page. For now,
the check-your-answers page is just a holding page as the real thing
is planned to be implemented as a separate story.

This commit adds a temporary check-your-answers template as the existing
one will break if given a Maths & Physics claim. A separate story is
planned that will make check-your-answers work with any type of claim.

Note that there is now a fair amount of commonality/duplication between
the two SlugSequence classes but it feels too early to try and extract
any kind of abstraction.

### DIFF
--- a/app/views/maths_and_physics/claims/eligibility_confirmed.html.erb
+++ b/app/views/maths_and_physics/claims/eligibility_confirmed.html.erb
@@ -4,5 +4,7 @@
   <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-xl">You are eligible to claim a payment for teaching maths or physics</h1>
     <p class="govuk-body">Based on your answers you can claim £2,000 for teaching maths or physics.</p>
+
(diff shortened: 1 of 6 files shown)


app/views/maths_and_physics/claims/eligibility_confirmed.html.erb
```diff
@@ -4,5 +4,7 @@
   <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-xl">You are eligible to claim a payment for teaching maths or physics</h1>
     <p class="govuk-body">Based on your answers you can claim £2,000 for teaching maths or physics.</p>
+
+    <%= link_to "Continue", claim_path(current_policy_routing_name, next_slug), class: "govuk-button", role: "button" %>
```
mec: Does this need `data: { module: "govuk-button" }` as it is a link that has button behaviour?

tekin: 👍🏽

PR #629, opened 2019-11-19
Backfill claim data in Geckoboard

pezholio:
**Not to be merged until we have the Geckoboard account set up**

This follows up from #617 and #620 to backfill existing data about historic claims in Geckoboard. As we don't currently have a way of running arbitary rake tasks on production, I've used Data Migrations for this.

This has also resulted in a bunch of refactors that should hopefully make integrating with Geckoboard more generic for when we start to introduce other data types.


app/models/claim/geckoboard_event.rb
```diff
@@ -2,32 +2,34 @@
 
 class Claim
   class GeckoboardEvent
-    attr_reader :claim, :event_type, :performed_at
+    attr_reader :claims, :event_type, :date_field
```
mec: I had to dig quite a bit to find out what date_field is expected to be, I wonder if a different name could help, something like `event_timestamp`?

app/models/payment_confirmation.rb
```diff
@@ -24,10 +24,13 @@ def ingest
...
       if errors.empty?
-        payroll_run.claims.each do |claim|
+        claims = payroll_run.claims
+
+        claims.each do |claim|
           ClaimMailer.payment_confirmation(claim, payment_date_timestamp).deliver_later
-          RecordPaymentJob.perform_later(claim)
         end
+
+        RecordPaymentJob.perform_later(claims.pluck(:id))
```
mec: `pluck` looks awesome! 👍

app/models/claim/geckoboard_event.rb
pezholio: Like it 👍

app/models/claim.rb
```diff
@@ -240,6 +240,10 @@ def policy
     eligibility&.class&.module_parent
   end
 
+  def paid_at
+    payment&.updated_at
```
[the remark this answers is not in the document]
pezholio: How about we have a `payment_date` method that calls `payment.updated_at.next_occurring(:friday)`?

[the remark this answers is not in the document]
tekin: How would you feel about that as a solution?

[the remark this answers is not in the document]
pezholio: Works for me 👍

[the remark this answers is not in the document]
tekin: I can think of a couple of downsides:

- `updated_at` is not entirely reliable: if the payment is updated for some reason the payment is updated the reported `payment_date` will change and report an incorrect value
- the logic for determining the actual date the payment (i.e. the Friday after the report is ingested) will be duplicated in two places

I'm now thinking that the only reliable and safe way to report the actual payment date is to record a `paid_on` datestamp on the payment at the point we ingest the report. Anything else will be brittle and potentially inaccurate.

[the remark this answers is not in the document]
tekin: Maybe we should update the existing code and back-fill the existing payments. What do you think?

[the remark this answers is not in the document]
pezholio: Yeah, I think this might be better done as a seperate PR though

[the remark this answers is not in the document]
pezholio: @tekin PR is here https://github.com/DFE-Digital/dfe-teachers-payment-service/pull/771

spec/jobs/record_claim_event_job_spec.rb
```diff
@@ -1,8 +1,8 @@
 require "rails_helper"
 require "geckoboard"
 
-RSpec.describe RecordSubmittedClaimJob do
-  let(:claim) { build(:claim, :submitted) }
+RSpec.describe RecordClaimEventJob do
```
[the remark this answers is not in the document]
pezholio: That particular use case is covered in the PaymentConfirmation spec, but I think it's probably best to add another spec to the generic job spec too, to make it clearer, yep.

db/data/20200115145845_backfill_geckoboard_data.rb
```diff
@@ -0,0 +1,13 @@
+class BackfillGeckoboardData < ActiveRecord::Migration[6.0]
+  def up
+    submitted_claim_ids = Claim.submitted.pluck(:id)
+    paid_claim_ids = Claim.where.not(payment: nil).pluck(:id)
+
+    RecordClaimEventJob.perform_later(submitted_claim_ids, :submitted, :submitted_at)
+    RecordClaimEventJob.perform_later(paid_claim_ids, :paid, :paid_at)
```
tekin: Should this now be this?

```suggestion
    RecordClaimEventJob.perform_later(paid_claim_ids, :paid, : scheduled_payment_date)
```

Reckon it's worth re-running this data migration _and_ having the jobs run to verify it all actually marry's up.

db/data/20200115145845_backfill_geckoboard_data.rb
```diff
@@ -0,0 +1,13 @@
+class BackfillGeckoboardData < ActiveRecord::Migration[6.0]
+  def up
+    submitted_claim_ids = Claim.submitted.pluck(:id)
+    paid_claim_ids = Claim.where.not(payment: nil).pluck(:id)
```
tekin: I don't think this scope is correct. This will return all claims that have a payment _even those that have not actually been scheduled for payment_, e.g. any that are part of the current payroll run will be included, even though we haven't received or processed the payment confirmation report. We only want to backfill the claims that actually have a `scheduled_payment_date` set, right?

spec/jobs/record_claim_event_job_spec.rb
```diff
@@ -0,0 +1,20 @@
+require "rails_helper"
+require "geckoboard"
```
tekin: Is this `require` necessary? If the gecko board code isn't autoloaded by the gem, then it's better practice to put the dependency closer to where it is actually required, i.e. the file where the `Claim:: GeckoboardEvent` class is defined.

spec/jobs/record_claim_event_job_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+RSpec.describe RecordClaimEventJob do
+  let(:claims) { create_list(:claim, 5, :submitted) }
+
+  subject { described_class.new }
+
+  it "sends each claim's reference, policy and submitted date to claims.submitted.ENV dataset" do
```
tekin: This makes it sound like `"claims.submitted.ENV"` is a constant in the behaviour, when in fact the `"submitted"` part comes from the second parameter passed to the `perform` method. Can you rewrite this spec to make it clearer that the second parameter is used to determine the dataset?

spec/support/geckoboard_helpers.rb
```diff
@@ -39,4 +39,17 @@ def stub_geckoboard_dataset_find_or_create(dataset_id)
   def stub_geckoboard_dataset_post(dataset_id)
     stub_request(:post, "https://api.geckoboard.com/datasets/#{dataset_id}/data")
   end
+
+  def body_matching_geckoboard_data_for_claims(request, claims, date_field)
```
tekin: This method returns `true` or `false` depending on the result of the comparison. Perhaps worth calling it something like `request_body_matches_geckoboard_data_for_claims?`.

app/jobs/record_claim_event_job.rb
```diff
@@ -0,0 +1,6 @@
+class RecordClaimEventJob < ApplicationJob
+  def perform(claim_ids, event_type, date_field)
```
tekin: The parameter is called `date_field` here but `performed_at_method` in the `Claim::GeckoboardEvent` class. Worth making them consistent? Also in the `body_matching_geckoboard_data_for_claims` method.

spec/support/geckoboard_helpers.rb
pezholio: Yeah, nice. I was struggling with a good name for this method!

db/data/20200115145845_backfill_geckoboard_data.rb
pezholio: Well spotted 👍 All seems to be in order when running it

db/data/20200115145845_backfill_geckoboard_data.rb
pezholio: Good call - I've changed this

PR #633, opened 2019-11-20
Add ITT specialisation question

lawrence-dxw:
We did not want an abbreviation in the attribute, which made it very
long, but we felt the price was worth paying to be clear what the
attribute represents.

The question template is a copy-and-paste of an existing one.

image


lawrence-dxw:
This question never determines eligibility - if you say that you didn't specialise in maths or physics, then you are taken to a secondary flow that asks whether you have a maths or physics degree. This question and the implications on eligibility will come in a future PR.

PR #635, opened 2019-11-20
Make transactional emails work with any policy

tekin:
This updates the transactional claim emails so that they will work with any policy. There are a lot of commits here, but it's not as bad as it looks:

- The first half dozen update the mailer and the views such that they will handle either type of claim
- The rest are small iterative steps that rework the mailer specs until the point where they can all be run with a claim for either policy

**NOTE: the rejection reasons for Maths & Physics are currently being reviewed and I'll be fixing them up in the translation file once the final content has been agreed**

### DIFF
--- a/spec/factories/claims.rb
+++ b/spec/factories/claims.rb
@@ -1,6 +1,13 @@
 FactoryBot.define do
   factory :claim do
-    association(:eligibility, factory: :student_loans_eligibility)
+    transient do
+      policy { StudentLoans }
+      eligibility_factory { "#{policy.to_s.underscore}_eligibility".to_sym }
+    end
+
+    after(:build) do |claim, evaluator|
+      claim.eligibility = build(*evaluator.eligibility_factory) unless claim.eligibility
+    end
 
     trait :submittable do
       first_name { "Jo" }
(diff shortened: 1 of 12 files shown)


pezholio:
@tekin Have you checked the emails are generated OK by Notify in the previews? Just seeing those bullet lists in the translations made me a touch nervous 😟

spec/factories/claims.rb
```diff
@@ -1,6 +1,13 @@
 FactoryBot.define do
   factory :claim do
-    association(:eligibility, factory: :student_loans_eligibility)
+    transient do
+      policy { StudentLoans }
+      eligibility_factory { "#{policy.to_s.underscore}_eligibility".to_sym }
+    end
+
+    after(:build) do |claim, evaluator|
+      claim.eligibility = build(*evaluator.eligibility_factory) unless claim.eligibility
```
robbiepaul: This is very cool!

spec/mailers/claim_mailer_spec.rb
```diff
@@ -1,102 +1,105 @@
...
+
+        it "includes the NET pay amount and the student loan deduction in the body" do
+          expect(mail.body.encoded).to include("You will receive £500.00 on or after 1 January 2019")
+          expect(mail.body.encoded).to include("Student loan (deducted): £60.00")
+        end
+
+        context "when user does not currently have a student loan" do
+          let(:payment) { build(:payment, :with_figures, student_loan_repayment: nil, claim: claim) }
+
+          it "does not mention the content relating to student loan dedictions" do
```
robbiepaul:
```suggestion
          it "does not mention the content relating to student loan deductions" do
```

tekin: 👍🏽

PR #636, opened 2019-11-20
Clear claim session before prepending view path

pezholio:
Currently, if a user starts a claim session for one policy, and then attempts to switch policies, they're confronted with a Template error. This is because `prepend_view_path_for_policy` runs before the claim session is cleared, so the view path is set for the old policy. This tweak changes this behaviour so `clear_claim_session` runs before the view path is set.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog]
 ## [Unreleased]
 
 - Update transational emails with improved content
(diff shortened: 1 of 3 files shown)


pezholio:
[Claim session not cleared when you swap policies?](https://trello.com/c/FWaA7TTh/1045-claim-session-not-cleared-when-you-swap-policies)

tekin:
Just needs a changelog!

PR #640, opened 2019-11-20
Display the gross pay instead of the gross value in the payment confirmation email

robbiepaul:
We've been using the gross value from the Cantium confirmation report in our payment advice to the claimants.

This is incorrect, the 'gross value' as they call it is in fact the total outlay (how much it is costing DfE). Instead we need to subtract the employer's (DfE) NI contributions from this value to get the gross pay that will be communicated to the claimant.

I'm not sure that having both `gross_pay` and `gross_value` is a good idea. If either of them should change it should be `gross_value` but since that's the name that Cantium have given it I've decided to leave it for now.

I decided against having `gross_pay` as a computed method as we will probably want to query against it in the future.

We will also need to notify all previously paid claimants of their new payment advice, but I thought it probably best to tackle that separately.


robbiepaul:
[BUG: Pay advice email doesn't show the correct gross value](https://trello.com/c/6PUNFXnh/1044-bug-pay-advice-email-doesnt-show-the-correct-gross-value)

app/models/payment_confirmation.rb
```diff
@@ -84,6 +84,7 @@ def update_payment_fields(payment, row)
     payment.student_loan_repayment = row["Student Loans"]
     payment.tax = row["Tax"]
     payment.net_pay = row["Net Pay"]
+    payment.gross_pay = row["Gross Value"].to_d - row["Employers NI"].to_d
```
tekin: If we used the values from the `payment` instance here would we be able to avoid the typecasting?

robbiepaul: I tried that first, but I got some test failures (NoMethodError on nil) on the spec that checks there are errors if the the values are missing. I could maybe do something like:

```ruby
payment.gross_pay = payment.gross_value - payment.employers_national_insurance if payment.gross_value.present?
```

tekin: Fair enough. You'd also need to check that `employers_national_insurance` is present also. Probably not worth it.

PR #642, opened 2019-11-21
Set cache headers so pages are refreshed each time

pezholio:
This fixes two issues - the first being an issue where a user who has finished, or was part way through their claim and has navigated away, can have their personal data surfaced by someone using the back button, even after the session has expired.

The second issue is if a user starts the Verify journey and then uses the back button, clicking continues results in an error, because the browser sends the same authentication request, rather than a newly minted one. This is slightly harder to test, because we stub out a lot of the Verify stuff in feature tests (and we  don't even see the error in the development version of the VSP), so the only thing we can test is that the cache headers get successfully set in the part of the journey where the authentication request is made.

### DIFF
--- a/spec/features/claim_journey_does_not_get_cached_spec.rb
+++ b/spec/features/claim_journey_does_not_get_cached_spec.rb
@@ -1,0 +1,21 @@
+require "rails_helper"
+
+RSpec.feature "Claim journey does not get cached", js: true do
+  it "redirects the user to the start of the claim journey if they go back after the claim is completed" do
+    claim = start_student_loans_claim
+    claim.update!(attributes_for(:claim, :submittable))
+    claim.eligibility.update!(attributes_for(:student_loans_eligibility, :eligible))
+    visit claim_path(claim.policy.routing_name, "check-your-answers")
+
+    expect(page).to have_text(claim.first_name)
+
+    click_on "Confirm and send"
+
+    expect(current_path).to eq(claim_confirmation_path(claim.policy.routing_name))
+
+    page.evaluate_script("window.history.back()")
+
+    expect(page).to_not have_text(claim.first_name)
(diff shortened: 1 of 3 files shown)


spec/features/claim_journey_does_not_get_cached_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  it "redirects the user to the start of the claim journey if they go back after the claim is completed" do
+    claim = start_student_loans_claim
+    claim.update!(attributes_for(:claim, :submittable))
+    claim.eligibility.update!(attributes_for(:student_loans_eligibility, :eligible))
+    visit claim_path(claim.policy.routing_name, "check-your-answers")
+
+    expect(page).to have_text(claim.first_name)
+
+    click_on "Confirm and send"
+
```
tekin: I'd perhaps say it would be worth having an assertion here so that we can be confident we're on the final confirmation screen (and thus the session has been cleared) before the back button is hit. Makes the spec a bit safer and more robust because it will fail early with a clear reason why, and also make it less susceptible to false-positives.

spec/requests/verify_authentications_spec.rb
```diff
@@ -19,6 +19,12 @@
       it "stores the request_id in the user’s session" do
         expect(session[:verify_request_id]).to eql(stubbed_auth_request_response["requestId"])
       end
+
+      it "returns cache headers so the page gives a newsly minted VSP request on every visit" do
```
tekin:
```suggestion
      it "returns cache headers so the page gives a newly minted VSP request on every visit" do
```

spec/features/claim_journey_does_not_get_cached_spec.rb
pezholio: Good call 👍

pezholio: Done!

PR #644, opened 2019-11-21
Add supply teacher questions to Maths and Physics journey

lawrence-dxw:
This adds the "Are you currently employed as a supply teacher?" question, which acts as a branching point in the Maths and Physics eligibility journey. It then adds all the questions for this branch of the journey. (After these questions, the main journey will resume.)

image

image

image

image

image

### DIFF
--- a/app/views/maths_and_physics/claims/_ineligibility_reason_no_entire_term_contract.html.erb
+++ b/app/views/maths_and_physics/claims/_ineligibility_reason_no_entire_term_contract.html.erb
@@ -1,0 +1,8 @@
+<h1 class="govuk-heading-xl">
+  You’re not eligible for this payment
+</h1>
+
+<p class="govuk-body">
+  You can only get this payment if you are employed directly by your school for
+  at least one term.
+</p>
--- a/app/views/maths_and_physics/claims/entire_term_contract.html.erb
+++ b/app/views/maths_and_physics/claims/entire_term_contract.html.erb
@@ -1,0 +1,44 @@
+<% content_for(:page_title, page_title(t("maths_and_physics.questions.has_entire_term_contract"), policy: current_policy_routing_name, show_error: current_claim.errors.any?)) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render("shared/error_summary", instance: current_claim, errored_field_id_overrides: { "eligibility.has_entire_term_contract": "claim_eligibility_attributes_has_entire_term_contract_true" }) if current_claim.errors.any? %>
+    <%= form_for current_claim, url: claim_path(current_policy_routing_name) do |form| %>
+      <%= form_group_tag current_claim do %>
+        <%= form.fields_for :eligibility, include_id: false do |fields| %>
+
+          <%= fields.hidden_field :has_entire_term_contract %>
+
+          <fieldset class="govuk-fieldset">
+
+            <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
+              <h1 class="govuk-fieldset__heading">
+                <%= t("maths_and_physics.questions.has_entire_term_contract") %>
+              </h1>
+            </legend>
+
+            <%= errors_tag current_claim.eligibility, :has_entire_term_contract %>
+
+            <div class="govuk-radios govuk-radios--inline">
+
+              <div class="govuk-radios__item">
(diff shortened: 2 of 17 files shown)


app/views/maths_and_physics/claims/_ineligibility_reason_no_entire_term_contract.html.erb
```diff
@@ -0,0 +1,8 @@
+<h1 class="govuk-heading-xl">
+  You’re not eligible for this payment
+</h1>
+
+<p class="govuk-body">
+  You can only get this payment if you are employed directly by your school for
+  at least 1 term.
```
pezholio: The [GOV.UK Style Guide](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#numbers) seems to suggest we use 'one' in this case.

spec/features/maths_and_physics_claim_spec.rb
```diff
@@ -165,6 +170,55 @@
     expect(page).to have_text(I18n.t("questions.qts_award_year"))
   end
 
+  scenario "Supply teacher claims for Maths and Physics, employed directly by school with a contract to teach for an entire term" do
```
pezholio: I'm wondering with these 'alernative journey' specs if we do something along the lines of what we do for the Changing Answers specs (https://github.com/DFE-Digital/dfe-teachers-payment-service/blob/master/spec/features/changing_answers_spec.rb#L8-L11) in that we start the claim (so we have a claim in memory), then update the claim's details up to the part we're testing, and then visit that path directly. This would solve a lot of duplication, and speed up the tests too. 

Also, rather than testing up until the end of the journey we can just test whether the claim is still eligible with something like:

```ruby
expect(claim.reload.ineligbile?).to eq(false)
```

app/views/maths_and_physics/claims/entire_term_contract.html.erb
```diff
@@ -0,0 +1,45 @@
+<% content_for(:page_title, page_title(t("maths_and_physics.questions.has_entire_term_contract"), policy: current_policy_routing_name, show_error: current_claim.errors.any?)) %>
+<% path_for_form = current_claim.persisted? ? claim_path(current_policy_routing_name) : claims_path(current_policy_routing_name) %>
```
robbiepaul: Do we need this? Won't the `current_claim` always be persisted at this point?

app/views/maths_and_physics/claims/supply_teacher.html.erb
```diff
@@ -0,0 +1,45 @@
+<% content_for(:page_title, page_title(t("maths_and_physics.questions.employed_as_supply_teacher"), policy: current_policy_routing_name, show_error: current_claim.errors.any?)) %>
+<% path_for_form = current_claim.persisted? ? claim_path(current_policy_routing_name) : claims_path(current_policy_routing_name) %>
```
robbiepaul: Same question as above

app/views/maths_and_physics/claims/entire_term_contract.html.erb
lawrence-dxw: Good spot, thanks. It's because I copied-and-pasted the template for the `teaching_maths_or_physics` question, which is a special case because it's the first question. I'll sort it out.

app/views/maths_and_physics/claims/supply_teacher.html.erb
lawrence-dxw: Yep, thanks. Will fix.

app/views/maths_and_physics/claims/_ineligibility_reason_no_entire_term_contract.html.erb
lawrence-dxw: Yeah, it rankled a bit but I didn't want to change it just because of personal preference. Good to know about the Style Guide, thanks! Will fix.

spec/features/maths_and_physics_claim_spec.rb
lawrence-dxw: I think it's a good idea. I think that perhaps we could leave it as it is here, and revisit these specs once we see how much of these additional journeys we explore in the (yet to come) changing answers specs for Maths and Physics, what do you think?

pezholio: Yeah, makes sense 👍

PR #646, opened 2019-11-21
Change feedback link depending on policy

pezholio:
This adds a helper method to get the feedback link for each policy. I've also refactored the `ApplicationHelper` to get the current policy based on the routing name given that we're using it in two places now. 

I've also updated the StudentLoans and Maths and Physics modules to extend from themselves, to avoid the repetition of typing `self` at the beginning of every method name, and given that we're unlikely to be including these modules in any other classes.

I've also added a simple test in the Student Loan and Maths and Physics happy paths to give us the confidence that the correct feedback URLs are being returned for each journey.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -28,10 +28,14 @@ module ApplicationHelper
   end
 
   def feedback_url
-    "https://docs.google.com/forms/d/e/1FAIpQLSdAyOxHme39E8lMnD2qY029mmk4Lpn84soYg2vLrT5BV9IUSg/viewform?usp=sf_link"
+    current_policy.feedback_url
   end
 
   def start_page_url
-    Policies[current_policy_routing_name].start_page_url
+    current_policy.start_page_url
+  end
+
+  def current_policy
+    Policies[current_policy_routing_name]
   end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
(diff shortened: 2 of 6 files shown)


pezholio:
[Feedback form link varies by policy](https://trello.com/c/GgJj2Gr1/1047-feedback-form-link-varies-by-policy)

app/helpers/application_helper.rb
```diff
@@ -28,10 +28,14 @@ def policy_service_name(policy = nil)
   end
 
   def feedback_url
-    "https://docs.google.com/forms/d/e/1FAIpQLSdAyOxHme39E8lMnD2qY029mmk4Lpn84soYg2vLrT5BV9IUSg/viewform?usp=sf_link"
+    current_policy&.feedback_url
   end
 
   def start_page_url
-    Policies[current_policy_routing_name].start_page_url
+    current_policy&.start_page_url
```
tekin: Is there a reason why the safe navigation operator has snuck in here? I feel like the service is currently set up such that anywhere we use these helpers to link to the start page or feedback url we are in the context of a policy. Generally if we can avoid a method returning `nil` then we should as it is one of the causes of strange and subtle bugs. For example, say a page _was_ introduced where there is no `current_policy`. Would it be better for these methods to return `nil`, and therefor create links to nowhere, or would it be better for a spec for that page to blow up so that we notice the missing dependency?

pezholio: Good thinking, I um-ed and ah-ed about this too, but yeah, I think blowing up would be the preferred behaviour, so we could fix it one way or the other.

pezholio:
That's a really useful link, ta 👍

PR #647, opened 2019-11-21
The "claim submitted" page should not link to the done page for Maths and Physics

pezholio:
This does two things:

* Moves the done page URL to a method in each policy (Maths and Physics currently has none, so returns `nil`)
* Only link to a done page on the confirmation page if it exists for that particular policy


pezholio:
[The "Claim submitted" page should not link to the "done page" for Maths & Physics](https://trello.com/c/dv9UDeaJ/1048-the-claim-submitted-page-should-not-link-to-the-done-page-for-maths-physics)

PR #648, opened 2019-11-21
Add final two Maths and Physics eligibility questions

lawrence-dxw:
This PR completes the eligibility journey for Maths and Physics.

# Disciplinary action

image
image

# Formal action for poor performance

image
image


lawrence-dxw:
I've just tacked on an additional commit which adds some missing hint text for the first question:

image

PR #649, opened 2019-11-21
Make sure payroll process works with a Maths & Physics claim

tekin:
Prior to this things would have blown up if a payroll run was attempted with an approved Maths & Physics claim.

This doesn't deal with the bigger issue of when there is a claim for each policy from the same person in the same payroll run, but that is being tackled elsewhere in another story. This at least means we can approve and pay Maths & Physics claims, as long as the person doesn't have an approved Student Loans claim in the same run.

### DIFF
--- a/spec/models/maths_and_physics/eligibility_spec.rb
+++ b/spec/models/maths_and_physics/eligibility_spec.rb
@@ -52,6 +52,12 @@ RSpec.describe MathsAndPhysics::Eligibility, type: :model do
     end
   end
 
+  describe "#award_amount" do
+    it "returns the £2,000 amount that Maths & Physics claimants are eligible for" do
+      expect(MathsAndPhysics::Eligibility.new.award_amount).to eq(BigDecimal("2000"))
+    end
+  end
+
   describe "#current_school_name" do
     it "returns the name of the current school" do
       eligibility = MathsAndPhysics::Eligibility.new(current_school: schools(:penistone_grammar_school))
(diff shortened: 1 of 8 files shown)


spec/models/maths_and_physics/eligibility_spec.rb
```diff
@@ -52,6 +52,12 @@
     end
   end
 
+  describe "#award_amount" do
+    it "returns the £2,000 amount that Maths & Physics claimants are eligible for" do
+      expect(MathsAndPhysics::Eligibility.new.award_amount).to eq(BigDecimal("2000"))
```
pezholio: Any reason why this `BigDecimal` doesn't have the decimal places, whereas the code does?

tekin: So the value returned by `BigDecimal("2000")` is equivalent to the value returned by `BigDecimal("2000.00")` (which is also equivalent to the value returned by `BigDecimal("2000.00000000")`) because the class just converts the string into its decimal equivalent and that's what we're comparing here. I suppose it is a little inconsistency in the way the value is represented in the test vs the actual code, but I don't feel it has a big enough material impact to warrant a fixup/CI-rerun. What you reckon?

pezholio: Yeah, I figured they were equivalent, it just stuck out a bit. Fine for it to go in as is 👍

tekin: cool, just need an approval on the review

PR #650, opened 2019-11-21
Add new domains and certificates to template

pezholio:
This adds the GOV.UK service domains (with the canonical one first), as well as the certificate names and mapings to the template. With these in place we will be able to start serving requests from the service domain, and any requests to the old domain will be redirected to the new one.


pezholio:
[Set up GOV.UK service domain](https://trello.com/c/gSh36Q1V/824-set-up-govuk-service-domain)

PR #656, opened 2019-11-25
Make service.gov.uk domain canonical

pezholio:
Now the certificate is set up an confirmed working, we can now redirect all requests to the service domain.

**NOTE** Before this goes to production, we'll need to put the app into maintenance mode, as the sessions are stored by domain, so any in-progress claims will break.


pezholio:
[Set up GOV.UK service domain](https://trello.com/c/gSh36Q1V/824-set-up-govuk-service-domain)

PR #657, opened 2019-11-25
Signed-out admin users redirected to orginal request url after signing in

mec:
Not only is this a better user experience in general, Cantium users will only be
given a single endpoint to download the payroll run file. By redirecting them
after sign in we save them from having to navigate to the right place.

### DIFF
--- a/app/controllers/admin/base_admin_controller.rb
+++ b/app/controllers/admin/base_admin_controller.rb
@@ -31,7 +31,10 @@ module Admin
     end
 
     def ensure_authenticated_user
-      redirect_to admin_sign_in_path unless admin_signed_in?
+      unless admin_signed_in?
+        session[:requested_admin_path] = request.fullpath
+        redirect_to admin_sign_in_path
+      end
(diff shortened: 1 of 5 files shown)


app/controllers/admin/base_admin_controller.rb
```diff
@@ -31,7 +31,10 @@ def end_expired_admin_sessions
     end
 
     def ensure_authenticated_user
-      redirect_to admin_sign_in_path unless admin_signed_in?
+      unless admin_signed_in?
+        session[:user_signed_out_path] = request.fullpath
```
robbiepaul: Without reading the code further I was a bit confused by the naming of this. I thought it was going to be something we would redirect to when signing out. Maybe something like `requested_admin_path` would be more intention revealing?

pezholio: Yeah, this confused me too. I agree with Robbie, `requested_admin_path` would work better I think 👍

PR #660, opened 2019-11-25
Stop running workers in the same instance as the web app

robbiepaul:
The main changes in this PR:
- Replace the default subnet (unused) with the worker subnet
- Add ACI and associated resources to run workers
- Change the runtime stack to use "Single container mode" instead of "Docker Compose"


robbiepaul:
[Stop running workers in the same instance as the web app](https://trello.com/c/JRVQLjIg/433-stop-running-workers-in-the-same-instance-as-the-web-app)

robbiepaul:
Yeah it deployed to the tmp-app successfully 👍

pezholio:
Siiiiick, punch it in that case 👍

PR #662, opened 2019-11-26
Refactor check-your-answers helper for multiple policies and add answer formatting for Maths and Physics

lawrence-dxw:
This PR is the first step in building the `check-your-answers` page for Maths and Physics. There's a first refactor which lets us provide a per-policy class to format the answers for playback to the claimant. After that, I create this class for M&P.

I'm submitting this PR now, before building the whole `check-your-answers` feature, because:
1. I would like feedback on the approach (@tekin and I did talk about it a bit already though);
2. We will probably want to use a similar approach when [adapting the admin claims interface for M&P](https://trello.com/c/hgPHPmX4/991-admin-is-updated-to-accommodate-maths-physics-claims), so getting this in early means we can be consistent in our approach.

### DIFF
--- a/app/helpers/claims_helper.rb
+++ b/app/helpers/claims_helper.rb
@@ -10,15 +10,8 @@ module ClaimsHelper
     fields.to_sentence
   end
 
-  def eligibility_answers(eligibility)
-    [].tap do |a|
-      a << [t("questions.qts_award_year"), I18n.t("student_loans.questions.qts_award_years.#{eligibility.qts_award_year}"), "qts-year"]
-      a << [t("student_loans.questions.claim_school"), eligibility.claim_school_name, "claim-school"]
-      a << [t("questions.current_school"), eligibility.current_school_name, "still-teaching"]
-      a << [t("student_loans.questions.subjects_taught", school: eligibility.claim_school_name), subject_list(eligibility.subjects_taught), "subjects-taught"]
-      a << [t("student_loans.questions.leadership_position"), (eligibility.had_leadership_position? ? "Yes" : "No"), "leadership-position"]
-      a << [t("student_loans.questions.mostly_performed_leadership_duties"), (eligibility.mostly_performed_leadership_duties? ? "Yes" : "No"), "mostly-performed-leadership-duties"] if eligibility.had_leadership_position?
-    end
+  def eligibility_answers(claim)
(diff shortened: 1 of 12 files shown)


app/helpers/claims_helper.rb
```diff
@@ -11,14 +11,7 @@ def verified_fields(claim)
...
   def eligibility_answers(eligibility)
-    [].tap do |a|
-      a << [t("questions.qts_award_year"), I18n.t("student_loans.questions.qts_award_years.#{eligibility.qts_award_year}"), "qts-year"]
-      a << [t("student_loans.questions.claim_school"), eligibility.claim_school_name, "claim-school"]
-      a << [t("questions.current_school"), eligibility.current_school_name, "still-teaching"]
-      a << [t("student_loans.questions.subjects_taught", school: eligibility.claim_school_name), subject_list(eligibility.subjects_taught), "subjects-taught"]
-      a << [t("student_loans.questions.leadership_position"), (eligibility.had_leadership_position? ? "Yes" : "No"), "leadership-position"]
-      a << [t("student_loans.questions.mostly_performed_leadership_duties"), (eligibility.mostly_performed_leadership_duties? ? "Yes" : "No"), "mostly-performed-leadership-duties"] if eligibility.had_leadership_position?
-    end
+    eligibility.class.module_parent::EligibilityAnswersPresenter.new(eligibility).answers
```
tekin: I wonder if we should use `eligibility.claim.policy` here instead of `eligibility.class.module_parent`? I know under the hood that's what happens, but `claim.policy` feels like the API we've defined for identifying the policy namespace for a claim so the code may be more intention-revealing if we stick to using that?

lawrence-dxw: Makes sense - I'll change the helper's `#eligibility_answers` to take a claim instead of an eligibility.

lawrence-dxw:
> Is there a sense of how this pattern will be used in admin?

I'm thinking that the best option might be to use the same presenter class, but with a different instance method, so that the admin / public can share formatting logic (e.g. the `subject_list` method which currently lives in that slightly awkward new module that I created). But I'll leave that for @pezholio to think about.

> Is there a plan for handling the student loan amount question, which is student loans policy specific but currently sits in the more general student_loan_answers helper?

No, not yet 😃

lawrence-dxw:
Thanks for the feedback; I added documentation comments. Ready for re-review.

PR #663, opened 2019-11-26
Allow admin to filter by policy

pezholio:
This shows the policy in the admin view against each claim, and also allows users to filter by policy.

There will be some further work needed on this story, once the Check Your Answers work is done.

image

image

### DIFF
--- a/app/controllers/admin/claims_controller.rb
+++ b/app/controllers/admin/claims_controller.rb
@@ -3,6 +3,8 @@ class Admin::ClaimsController < Admin::BaseAdminController
 
   def index
     @claims = Claim.includes(:check, eligibility: [:claim_school, :current_school]).awaiting_checking.order(:submitted_at)
+    @claims = @claims.by_policy(filtered_policy) if filtered_policy
+
     respond_to do |format|
       format.html
       format.csv {
@@ -28,5 +30,11 @@ class Admin::ClaimsController < Admin::BaseAdminController
     else
       flash.now[:notice] = "Cannot find a claim with reference \"#{params[:reference]}\""
     end
+  end
+
+  private
+
+  def filtered_policy
+    Policies[params[:policy]]
   end
 end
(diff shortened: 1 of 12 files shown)


pezholio:
[Admin is updated to accommodate Maths & Physics claims](https://trello.com/c/hgPHPmX4/991-admin-is-updated-to-accommodate-maths-physics-claims)

app/models/policies.rb
```diff
@@ -9,6 +9,10 @@ def self.all
     POLICIES
   end
 
+  def self.options_for_select
+    all.collect { |c| [c.short_name, c.to_s.underscore] }
```
robbiepaul: We could maybe utilise `c.routing_name` here? It might make matching a bit easier as well

app/controllers/admin/claims_controller.rb
```diff
@@ -29,4 +30,15 @@ def search
...
     end
   end
+
+  private
+
+  def policy_from_param
+    @policy_from_param ||= begin
+      return nil if params[:policy].blank?
+
+      policy = Policies.all.map(&:to_s).find { |p| p == params[:policy].camelize }
```
robbiepaul: If we use `routing_name` we could potentially do:
```
policy = Policies[params[:policy]]
```

pezholio: Waaaaay ahead of ya - I just realised there was a simpler way of doing this 😄

app/controllers/admin/claims_controller.rb
```diff
@@ -3,6 +3,8 @@ class Admin::ClaimsController < Admin::BaseAdminController
 
   def index
     @claims = Claim.includes(:check, eligibility: [:claim_school, :current_school]).awaiting_checking.order(:submitted_at)
+    @claims = @claims.by_policy(filter_policy) if filter_policy
```
robbiepaul: Small naming nitpick: `filter_policy` sounds like an action. Maybe `selected_policy` or `filtered_policy`

PR #665, opened 2019-11-26
Hardcode the VSP entity ID

pezholio:
Since we switched to the service domain, we accidentally switched the VSP entity ID too, as it's tied to the canonical domain. This hardcodes the ID, so it's no longer tied to the domain, as Verify don't really care if it's related to the domain (it's a abitary string)

### DIFF
--- a/azure/resource_groups/app/parameters/production.template.json
+++ b/azure/resource_groups/app/parameters/production.template.json
@@ -210,6 +210,9 @@
     },
     "VSP.EUROPEAN_IDENTITY_ENABLED": {
       "value": "true"
+    },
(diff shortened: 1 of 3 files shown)


azure/resource_groups/app/parameters/production.template.json
```diff
@@ -210,6 +210,9 @@
     },
     "VSP.EUROPEAN_IDENTITY_ENABLED": {
       "value": "true"
+    },
+    "VSP.VERIFY_ENTITY_ID": {
+      "value": "additional-teaching-payment.education.gov.uk"
```
robbiepaul: I think this needs to be prefixed with `https://`

PR #673, opened 2019-11-27
Add maths and physics seeds

lawrence-dxw:
Local development quality-of-life improvement.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -16,12 +16,14 @@ if Rails.env.development?
   class Seeds
     extend FactoryBot::Syntax::Methods
 
-    create(:payroll_run, :confirmation_report_uploaded, claims_count: 32, created_at: 3.months.ago - 10.days)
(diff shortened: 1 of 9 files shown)


lawrence-dxw:
> I wonder if we could improve our test coverage if some of the places where we're using the payroll_run factory where updated to include Maths & Physics claims too. Perhaps the default should create at least one of each type of policy?

It'd be a good idea to update some of the places where we explicitly specify `claims_counts: {StudentLoans => <something>}`, probably. As for changing the default, it probably is good just from a "student loans isn't special" symmetry point of view, but given that most of the specs which care about the _payments_ of a payroll run (other than just the payroll run's existence) explicitly specify `claims_counts`, it might not end up testing anything more than it already does.

> Is it possible that at certain days of the month the seeds file would fail because the date arithmetic would result in attempts to create payroll runs in the same month?

I don't think so. The `created_at` dates that we choose for the three payroll runs in the seeds each have more than one calendar month between them.

PR #679, opened 2019-11-28
Information provided page works for both policies

robbiepaul:


app/views/claims/information_provided.html.erb
```diff
@@ -10,7 +10,7 @@
     </p>
 
     <ul class="govuk-list govuk-list--bullet">
-      <li>any schools you have worked for</li>
+      <li>the school you currently employed to teach at</li>
```
tekin:
```suggestion
      <li>the school you are currently employed to teach at</li>
```

config/locales/en.yml
```diff
@@ -43,6 +43,8 @@ en:
         unit: "£"
   service_name: "Claim additional payments for teaching"
   support_email_address: "additionalteachingpayment@digital.education.gov.uk"
+  information_provided:
+    student_loan_repayment_line: "which is deducted from your claim payment if applicable"
```
tekin: should this also be updated to include the ` and credited to your Student Loans Account` bit?

robbiepaul: Yeah you're right, the lines are so similar I've now extracted out an `award_description` translation which is just `£2,000 payment` or `claim payment`

PR #683, opened 2019-11-28
Update check-your-answers feature specs for M&P

lawrence-dxw:
This does three things:

- Slims down the check-your-answers specs, reducing it to a
representative set of scenarios instead of a separate scenario for
changing each answer from each policy, which would not be a sustainable
approach as we add new policies.

- Introduces some maths and physics questions into the
check-your-answers specs (as part of the aforementioned set of
scenarios).

- Rearranges the file so that it's clear which specs are related to
eligibility (and hence are only representative of a broader set of
policy-specific questions), and which specs represent questions that are
present across all policies.

### DIFF
--- a/spec/features/changing_answers_spec.rb
+++ b/spec/features/changing_answers_spec.rb
@@ -1,32 +1,30 @@
 require "rails_helper"
 
 RSpec.feature "Changing the answers on a submittable claim" do
-  let(:claim) { Claim.order(:created_at).last }
-  let(:eligibility) { claim.eligibility }
+  scenario "Teacher changes an answer which is not a dependency of any of the other answers they’ve given, remaining eligible" do
+    claim = start_maths_and_physics_claim
+    claim.update!(attributes_for(:claim, :submittable))
+    claim.eligibility.update!(attributes_for(:maths_and_physics_eligibility, :eligible, initial_teacher_training_subject: "maths"))
+    visit claim_path(MathsAndPhysics.routing_name, "check-your-answers")
 
-  before do
+    find("a[href='#{claim_path(MathsAndPhysics.routing_name, "initial-teacher-training-subject")}']").click
+
+    expect(find("#claim_eligibility_attributes_initial_teacher_training_subject_maths").checked?).to eq(true)
+
+    choose "Physics"
+    click_on "Continue"
+
+    expect(claim.eligibility.reload.initial_teacher_training_subject).to eq("physics")
+
+    expect(current_path).to eq(claim_path(MathsAndPhysics.routing_name, "check-your-answers"))
+  end
+
+  scenario "Teacher changes an answer which is not a dependency of any of the other answers they’ve given, becoming ineligible" do
     claim = start_student_loans_claim
     claim.update!(attributes_for(:claim, :submittable))
     claim.eligibility.update!(attributes_for(:student_loans_eligibility, :eligible))
     visit claim_path(StudentLoans.routing_name, "check-your-answers")
-  end
 
-  scenario "Teacher can edit a field" do
-    old_number = claim.national_insurance_number
-    new_number = "AB123456C"
-
-    expect {
-      find("a[href='#{claim_path(StudentLoans.routing_name, "national-insurance-number")}']").click
-      fill_in "National Insurance number", with: new_number
-      click_on "Continue"
-    }.to change {
-      claim.reload.national_insurance_number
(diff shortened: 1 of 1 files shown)


spec/features/changing_answers_spec.rb
```diff
@@ -1,284 +1,230 @@
...
-    let!(:new_claim_school) { create(:school, :student_loan_eligible, name: "Claim School") }
+  context "User changes fields that aren't related to eligibility" do
+    let(:claim) { Claim.order(:created_at).last }
+    let(:eligibility) { claim.eligibility }
 
     before do
-      find("a[href='#{claim_path(StudentLoans.routing_name, "claim-school")}']").click
-
-      choose_school new_claim_school
+      claim = start_student_loans_claim
```
mec: I feel like `start_student_loans_claim` could be inside the let as it is the same thing and this just seems to re-assign it here?

spec/features/changing_answers_spec.rb
```diff
@@ -1,284 +1,230 @@
...
+      claim.update!(attributes_for(:claim, :submittable))
+      claim.eligibility.update!(attributes_for(:student_loans_eligibility, :eligible))
```
mec: Can this be `eligibility.update!(attributes_for(:student_loans_eligibility, :eligible))` as we've let it?

spec/features/changing_answers_spec.rb
```diff
@@ -1,284 +1,230 @@
...
-
-  before do
-    claim = start_student_loans_claim
+  scenario "Teacher changes an answer which is not a dependency of any of the other answers they’ve given, remaining eligible" do
+    claim = start_maths_and_physics_claim
     claim.update!(attributes_for(:claim, :submittable))
-    claim.eligibility.update!(attributes_for(:student_loans_eligibility, :eligible))
-    visit claim_path(StudentLoans.routing_name, "check-your-answers")
-  end
+    claim.eligibility.update!(attributes_for(:maths_and_physics_eligibility, :eligible, initial_teacher_training_subject: "maths"))
```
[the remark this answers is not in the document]
lawrence-dxw: Yeah, that was my reasoning.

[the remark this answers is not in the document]
lawrence-dxw: (and to be 100% about the starting state and make sure that the test continues to test the same thing even if the factory is changed.)

spec/features/changing_answers_spec.rb
```diff
@@ -1,284 +1,230 @@
...
 
-  scenario "Teacher changes their leadership position to no" do
-    claim.eligibility.had_leadership_position = true
+  scenario "Teacher changes an answer which is a dependency of some of the subsequent answers they’ve given, making them ineligible" do
+    claim = start_student_loans_claim
+    claim.update!(attributes_for(:claim, :submittable))
+    claim.eligibility.update!(attributes_for(:student_loans_eligibility, :eligible))
+    visit claim_path(StudentLoans.routing_name, "check-your-answers")
+
+    claim.eligibility.had_leadership_position = false
```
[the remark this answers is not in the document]
lawrence-dxw: It sure could. Thanks!

spec/features/changing_answers_spec.rb
lawrence-dxw: Yep, thanks!

spec/features/changing_answers_spec.rb
lawrence-dxw: You're right. I've had to change it to a `let!` so that `start_student_loans_claim` happens before the `it` starts.

spec/features/changing_answers_spec.rb
```diff
@@ -1,284 +1,226 @@
...
 
     find("a[href='#{claim_path(StudentLoans.routing_name, "leadership-position")}']").click
 
-    choose "No"
+    choose "Yes"
     click_on "Continue"
 
-    expect(eligibility.reload.had_leadership_position).to eq(false)
+    expect(claim.eligibility.reload.had_leadership_position).to eq(true)
+    expect(claim.eligibility.reload.mostly_performed_leadership_duties).to be_nil
```
robbiepaul: We could save a database call here as the record has already been reloaded
```suggestion
    expect(claim.eligibility.mostly_performed_leadership_duties).to be_nil
```

lawrence-dxw: Thanks, good catch. Fixed it.

PR #685, opened 2019-12-02
Service operators can see Maths and Physics claim eligibility

robbiepaul:
## Screenshot

Screenshot 2019-12-02 at 13 24 32


robbiepaul:
[Admin is updated to accommodate Maths & Physics claims](https://trello.com/c/hgPHPmX4/991-admin-is-updated-to-accommodate-maths-physics-claims)

lawrence-dxw:
When I run it with one of the M&P claims from the seeds, the "ITT field" shows a string representation of a Ruby hash:

image

robbiepaul:
Interesting... I'll fix it

robbiepaul:
I can't see to recreate it, they all say "Maths" for me. What's the value of `initial_teacher_training_subject` for that claim?

lawrence-dxw:
Ah, don't worry. I had some old seed data from before we introduced the new ITT journey. `rails db:reset` and all good.

config/locales/en.yml
```diff
@@ -140,6 +140,16 @@ en:
       employed_directly:
         "yes": "Yes, I’m employed by my school"
         "no": "No, I’m employed by a private agency"
+    admin:
+      teaching_maths_or_physics: "Teach any maths or physics?"
```
lawrence-dxw: The other questions here are written in the third person e.g. "Has entire term contract?", so I think this would be more consistent:

```suggestion
      teaching_maths_or_physics: "Teaches any maths or physics?"
```

app/presenters/maths_and_physics/eligibility_admin_answers_presenter.rb
```diff
@@ -0,0 +1,45 @@
...
+        a << [I18n.t("admin.current_school"), display_school(eligibility.current_school)]
+        a << [I18n.t("maths_and_physics.admin.initial_teacher_training_subject"), I18n.t("maths_and_physics.answers.initial_teacher_training_subject.#{eligibility.initial_teacher_training_subject}")]
+        a << [I18n.t("maths_and_physics.admin.initial_teacher_training_subject_specialism"), I18n.t("maths_and_physics.answers.initial_teacher_training_subject_specialism.#{eligibility.initial_teacher_training_subject_specialism}")] if eligibility.initial_teacher_training_subject_specialism.present?
+        a << [I18n.t("maths_and_physics.admin.has_uk_maths_or_physics_degree"), degree_answer] if eligibility.has_uk_maths_or_physics_degree.present?
+        a << [I18n.t("admin.qts_award_year"), I18n.t("maths_and_physics.questions.qts_award_years.#{eligibility.qts_award_year}")]
+        a << [I18n.t("maths_and_physics.admin.employed_as_supply_teacher"), (eligibility.employed_as_supply_teacher? ? "Yes" : "No")]
+        a << [I18n.t("maths_and_physics.admin.has_entire_term_contract"), (eligibility.has_entire_term_contract? ? "Yes" : "No")] if eligibility.employed_as_supply_teacher?
```
lawrence-dxw: @tekin took a different approach (see the ITT questions above); display the answers if they're not nil. This means that this helper doesn't need to know about which answers are dependent on other answers.

```suggestion
        a << [I18n.t("maths_and_physics.admin.has_entire_term_contract"), (eligibility.has_entire_term_contract? ? "Yes" : "No")] unless eligibility.employed_as_supply_teacher.nil?
```

(It's a pattern that I hadn't thought about when writing the claim eligibility presenter, which I'm guessing these lines are copied from.)

app/presenters/maths_and_physics/eligibility_admin_answers_presenter.rb
```diff
@@ -0,0 +1,45 @@
...
+        a << [I18n.t("maths_and_physics.admin.employed_directly"), I18n.t("maths_and_physics.answers.employed_directly.#{eligibility.employed_directly? ? "yes" : "no"}")] if eligibility.employed_as_supply_teacher?
```
lawrence-dxw: Same as the above comment re. not adding knowledge of question dependencies here.

spec/presenters/maths_and_physics/eligibility_admin_answers_presenter_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+        [I18n.t("maths_and_physics.admin.disciplinary_action"), "Yes"],
+        [I18n.t("maths_and_physics.admin.formal_performance_action"), "Yes"],
+      ]
+
+      expect(presenter.answers).to eq expected_answers
+    end
+
+    it "excludes questions skipped from the flow" do
+      eligibility.initial_teacher_training_subject_specialism = :biology
+      eligibility.has_uk_maths_or_physics_degree = nil
```
lawrence-dxw: It's not possible to submit a claim with this combination of data; if you have a biology ITT then you need a maths or physics degree. I know it's only test data but still probably better that it be realistic.

robbiepaul:
It's a fairly new concept to me, but as far as I understand it a domain model can be anything in the business domain with data or behaviour.

There's a bit of prior art here, previously (in 3210bcaa872e5313a48bcae2de8dc120421ec7a8) we refactored out services to the models directory.

app/models/maths_and_physics/eligibility_admin_answers_presenter.rb
```diff
@@ -0,0 +1,45 @@
...
+        a << [I18n.t("admin.current_school"), display_school(eligibility.current_school)]
+        a << [I18n.t("maths_and_physics.admin.initial_teacher_training_subject"), I18n.t("maths_and_physics.answers.initial_teacher_training_subject.#{eligibility.initial_teacher_training_subject}")]
+        a << [I18n.t("maths_and_physics.admin.initial_teacher_training_subject_specialism"), I18n.t("maths_and_physics.answers.initial_teacher_training_subject_specialism.#{eligibility.initial_teacher_training_subject_specialism}")] if eligibility.initial_teacher_training_subject_specialism.present?
+        a << [I18n.t("maths_and_physics.admin.has_uk_maths_or_physics_degree"), degree_answer] if eligibility.has_uk_maths_or_physics_degree.present?
+        a << [I18n.t("admin.qts_award_year"), I18n.t("maths_and_physics.questions.qts_award_years.#{eligibility.qts_award_year}")]
+        a << [I18n.t("maths_and_physics.admin.employed_as_supply_teacher"), (eligibility.employed_as_supply_teacher? ? "Yes" : "No")]
+        a << [I18n.t("maths_and_physics.admin.has_entire_term_contract"), (eligibility.has_entire_term_contract? ? "Yes" : "No")] if eligibility.has_entire_term_contract.present?
```
lawrence-dxw: I think that for a boolean, `#nil?` communicates intent more clearly than `#present?`, which returns true when called on `false`. In this particular example, it probably doesn't make a difference, because you can't submit a claim if this value is false. But I think it would set a better example for future, similar, questions.

robbiepaul: Good point, I've updated the code to use `unless nil?`

PR #688, opened 2019-12-04
Redirect root url to an all-encompassing page

pezholio:
Now Maths and Physics is in private beta, it no longer makes sense to redirect requests to the root url to the Maths and Physics start page. There is now a [generic page on GOV.UK](https://www.gov.uk/government/publications/additional-payments-for-teaching-eligibility-and-payment-details) about teacher payments, so we're now redirecting to that. I've stored it as a config var in the Rails application.rb, rather than exposing it in the routes file, as it a) may be useful in other places and b) I'm not a huge fan of having a big URL like that in the routes.rb file itself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog]
 - Service operators can see Maths and Physics claim eligibility
 - Fix a bug that would occassionally redirect users even after they'd continued
   their session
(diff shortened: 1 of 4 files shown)


pezholio:
[Update page that domain root redirects to](https://trello.com/c/r9GfiQqy/1006-update-page-that-domain-root-redirects-to)

pezholio:
Good catch 👍

PR #689, opened 2019-12-04
Remove the Azure web tests to fix deploys

lawrence-dxw:
This removes the resources of type "Microsoft.Insights/webtests", which
provided HTTP health checks for the app and VSP.

We are doing this because our deploys started failing with a message
like:

> You cannot create or modify classic metric alerts for this subscription
> as this subscription 8655985a-2f87-44d7-a541-0be9a8c2779d is being
> migrated or has been migrated to use new metric alerts. Learn more -
> aka.ms/alertclassicretirement

Craig Gumbley told us that this is because our Azure subscriptions have
been migrated from "classic" alerts to a new alerting format
(https://ukgovernmentdfe.slack.com/archives/CKYNU0VG9/p1575410074005300),
and attempts to create classic alerts are now failing.

Following the guidance in the link provided by the the failure message,
I looked at the Azure portal and saw that the only classic alerts that
we have are the webtest-alert and vsp-webtest-alert.

So, to unblock deploys, we remove these alerts. I've created a Trello
card for figuring out what to replace it with:
https://trello.com/c/cw3nWKfZ/1102-figure-out-what-to-replace-azure-webcheck-with

I tested this on development with `bin/azure-deploy development` and it
succeeded.


lawrence-dxw:
> we may want to do some clearing up of parameters/variables as a follow up

I kept the existing variables there, because I assumed they would be useful for whatever new Azure concept we replace the web test with.

lawrence-dxw:
(But if that doesn't exist, then yeah we should clean them up.)

lawrence-dxw:
[Figure out what to replace Azure web test with](https://trello.com/c/cw3nWKfZ/1102-figure-out-what-to-replace-azure-web-test-with)

PR #693, opened 2019-12-05
Bug: fix special school age range and add Alternative provision schools

mec:
Special schools with a `phase` of `not_applicable` are still eligble when they teach students over the age of 11. We use the `statututoryHighAge` column in the Get information about schools data to test the elgiblity of this type of school.

Alternative provision schools that teach students over the age of 11 are also eligible. Alternative provision schools are those of the `school_type`:

- pupil_referral_unit
- secure_unit
- free_school_alternative_provider
- academy_alternative_provision_converter
- academy_alternative_provision_sponsor_led

The same `statututoryHighAge` can be used along with the school type to determine eligibility.

Secue units do not have a recognised state funded `school_type_group` so I had to add an exception for it in the `state_funded?` method as the group they are in contains non-state funded school types - I still need to confirm this with policy before merging this change.

## Significant spec refactoring ahead
The school eligibility specs have undergone significant refactoring as part of this work. The goal was to make it clearer what attributes make a school eligible and provide appropriate test coverage.

## When merging
Please check the trello card where there are two zendesk tickets that need to be updated once this work is in production.

### DIFF
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -136,24 +136,16 @@ class School < ApplicationRecord
     (STATE_FUNDED_SCHOOL_TYPE_GROUPS.include?(school_type_group) && school_type != "other_independent_special_school") || secure_unit?
   end
 
-  def special?
-    SPECIAL_SCHOOL_TYPES.include?(school_type)
-  end
-
-  def alternative_provision?
-    ALTERNATIVE_PROVISION_TYPES.include?(school_type)
-  end
-
-  def has_statutory_high_age_over_eleven?
-    statutory_high_age.present? && statutory_high_age > 11
-  end
-
   def secondary_phase?
     SECONDARY_PHASES.include?(phase)
   end
 
   def secondary_equivalent_special?
     special? && school_type != "special_post_16_institutions" && has_statutory_high_age_over_eleven?
+  end
+
+  def secondary_equivalent_alternative_provision?
+    alternative_provision? && has_statutory_high_age_over_eleven?
   end
 
   def open?
@@ -169,5 +161,17 @@ class School < ApplicationRecord
 
   def secure_unit?
     school_type == "secure_unit"
+  end
+
+  def alternative_provision?
+    ALTERNATIVE_PROVISION_TYPES.include?(school_type)
+  end
+
+  def has_statutory_high_age_over_eleven?
+    statutory_high_age.present? && statutory_high_age > 11
+  end
+
+  def special?
+    SPECIAL_SCHOOL_TYPES.include?(school_type)
   end
 end
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -152,6 +152,10 @@ class School < ApplicationRecord
     SECONDARY_PHASES.include?(phase)
   end
 
+  def secondary_equivalent_special?
+    special? && school_type != "special_post_16_institutions" && has_statutory_high_age_over_eleven?
+  end
+
   def open?
     close_date.nil?
   end
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -93,6 +93,14 @@ class School < ApplicationRecord
     academy_special_converter
   ].freeze
 
+  ALTERNATIVE_PROVISION_TYPES = %w[
+    pupil_referral_unit
+    secure_unit
+    free_school_alternative_provider
+    academy_alternative_provision_converter
+    academy_alternative_provision_sponsor_led
+  ].freeze
+
   enum phase: PHASES
   enum school_type_group: SCHOOL_TYPE_GROUPS
   enum school_type: SCHOOL_TYPES
@@ -123,11 +131,15 @@ class School < ApplicationRecord
   end
 
   def state_funded?
-    STATE_FUNDED_SCHOOL_TYPE_GROUPS.include?(school_type_group) && school_type != "other_independent_special_school"
+    (STATE_FUNDED_SCHOOL_TYPE_GROUPS.include?(school_type_group) && school_type != "other_independent_special_school") || secure_unit?
   end
 
   def special?
     SPECIAL_SCHOOL_TYPES.include?(school_type)
+  end
+
+  def alternative_provision?
+    ALTERNATIVE_PROVISION_TYPES.include?(school_type)
   end
 
   def has_statutory_high_age_over_eleven?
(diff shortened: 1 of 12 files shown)


mec:
[BUG : Fix special school age range](https://trello.com/c/QeP17cG7/1056-bug-fix-special-school-age-range)

mec:
[BUG : Fix special school age range](https://trello.com/c/QeP17cG7/1056-bug-fix-special-school-age-range)

spec/models/maths_and_physics/school_eligibility_spec.rb
```diff
@@ -64,6 +64,16 @@
         end
       end
 
+      context "and it is a alternative provision school" do
+        let(:school_attributes) { {school_type: :pupil_referral_unit, phase: :not_applicable} }
+        it { is_expected.to be false }
+
+        context "and it has a statutory high age of over 11" do
```
tekin: How do you feel about these nested contexts in terms of how they document and communicate about the behaviour?

app/models/student_loans/school_eligibility.rb
```diff
@@ -58,7 +58,11 @@ def eligible_phase?
     end
 
     def eligible_special_school?
-      @school.phase == "not_applicable" && @school.special? && @school.school_type != "special_post_16_institutions"
+      @school.phase == "not_applicable" && @school.has_statutory_high_age_over_eleven? && @school.special? && @school.school_type != "special_post_16_institutions"
+    end
+
+    def eligible_alternative_provision_school?
+      @school.phase == "not_applicable" && @school.has_statutory_high_age_over_eleven? && @school.alternative_provision?
```
tekin: There's a part of me that feels like the combination of these particular checks define a specific "type" of school. If we could come up with a name for that, then we could move this logic onto the `School` model and better communicate what this represents in terms of the characteristics of the school, rather than "these things make it an eligible alternative provision school". Same applies to the `eligible_special_school?` method. Does that make sense?

db/schema.rb
```diff
@@ -137,6 +137,8 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.string "confirmation_report_uploaded_by"
+    t.datetime "downloaded_at"
```
tekin: Still got these rogue changes. Have you rebased with master?

spec/models/maths_and_physics/school_eligibility_spec.rb
lawrence-dxw: When I looked at the school eligibility specs in the past, I wasn't able to get a sense of what all these different combinations of attributes meant, and whether they occur in real life. When I wrote further tests for school eligibility, I included an [example of a real-life school for each one](https://github.com/DFE-Digital/dfe-teachers-payment-service/blob/master/spec/models/student_loans/school_eligibility_spec.rb#L200). It might be worth doing here, too.

app/models/student_loans/school_eligibility.rb
```diff
@@ -38,13 +37,11 @@ def eligible_claim_school?
       !closed_before_policy_start? &&
         eligible_local_authority? &&
         @school.state_funded? &&
-        (eligible_phase? || eligible_special_school?)
+        @school.secondary_or_equivalent?
```
robbiepaul: 👍

spec/models/maths_and_physics/school_eligibility_spec.rb
```diff
@@ -64,6 +64,16 @@
...
+          let(:school_attributes) { {statutory_high_age: 16} }
```
robbiepaul: The `school_attributes` here don't reflect the nested context we're in. I would expect them to contain a `school_type` belonging to an alternative provision school, at the moment the `school_type` will be the one from the factory.

app/models/school.rb
```diff
@@ -119,15 +129,11 @@ def eligible_for_student_loans_as_current_school?
...
+    MathsAndPhysics::SchoolEligibility.new(self).eligible_current_school?
   end
 
   def state_funded?
-    STATE_FUNDED_SCHOOL_TYPE_GROUPS.include?(school_type_group) && school_type != "other_independent_special_school"
-  end
-
-  def special?
-    SPECIAL_SCHOOL_TYPES.include?(school_type)
+    (STATE_FUNDED_SCHOOL_TYPE_GROUPS.include?(school_type_group) && school_type != "other_independent_special_school") || secure_unit?
```
tekin: I don't think this is the right place to include secure units. `state_funded?` should always return true of false based on the funding type of the school. If we want to include non-state funded secure units in the list of eligible schools then we should have that logic in the eligibility check, not here as the exception is for eligibility, not for funding.

app/models/school.rb
```diff
@@ -140,4 +146,38 @@ def dfe_number
...
   end
+
+  def secondary_or_equivalent?
+    secondary_phase? || secondary_equivalent_special? || secondary_equivalent_alternative_provision?
+  end
+
+  private
+
+  def secure_unit?
+    school_type_group == "other" && school_type == "secure_unit"
```
tekin: I would suggest that the system is more robust if we only check `school_type` here. It might be the case today that all secure units also happen to have a `school_type_group` of `"other"`, but there is nothing to guarantee that will be the case in the future.

```suggestion
    school_type == "secure_unit"
```

app/models/school_data_importer.rb
```diff
@@ -45,6 +45,7 @@ def row_to_school(row)
     school.local_authority_district = local_authority_district
     school.close_date = row.fetch("CloseDate")
     school.establishment_number = row.fetch("EstablishmentNumber")
+    school.statutory_high_age = row.fetch("StatutoryHighAge")
```
tekin: Once this code is deployed do we want to trigger a school import to get these values set or will it be OK to wait for the nightly import?

app/models/school.rb
```diff
@@ -140,4 +146,38 @@ def dfe_number
       establishment_number,
     ].join("/")
   end
+
+  def secure_unit?
+    school_type_group == "other" && school_type == "secure_unit"
```
tekin: This is still checking `school_type_group`, which I think we agreed should be removed.

app/models/school_data_importer.rb
mec: Now we have access I think it would be good to run the import. We can then let the claimants who are waiting on this to be resolved know they can claim.

app/models/school.rb
mec: yeah, must be a rebase goof, sorry.

app/models/school_data_importer.rb
tekin: Sounds good 👍🏽

PR #694, opened 2019-12-05
Change payment <-> claim associations to enable multiple payments per claim

lawrence-dxw:
This is technical groundwork for some feature work that will come later. The main change also allows us to optimise the `PayrollRun#total_award_amount` method, which comes in its own commit.

### DIFF
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -1,5 +1,5 @@
 class Payment < ApplicationRecord
-  belongs_to :claim
+  has_one :claim
   belongs_to :payroll_run
 
   validates :award_amount, presence: true
--- a/app/models/payroll_run.rb
+++ b/app/models/payroll_run.rb
@@ -11,7 +11,7 @@ class PayrollRun < ApplicationRecord
   scope :this_month, -> { where(created_at: DateTime.now.all_month) }
 
   def total_award_amount
-    claims.sum(&:award_amount)
+    payments.sum(:award_amount)
   end
 
   def self.create_with_claims!(claims, attrs = {})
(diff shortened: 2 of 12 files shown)


app/models/payment.rb
```diff
@@ -1,9 +1,17 @@
 class Payment < ApplicationRecord
-  belongs_to :claim
+  has_many :claims
```
tekin: Did you consider using a `has_one` association instead of this and the temporary `claim` method?

app/models/payment.rb
```diff
@@ -1,9 +1,17 @@
...
   belongs_to :payroll_run
 
+  validates :claims, length: {maximum: 1}
```
tekin: What situation in the running app code would this validation prevent?

app/models/payroll_run.rb
```diff
@@ -16,7 +16,7 @@ def self.create_with_claims!(claims, attrs = {})
     ActiveRecord::Base.transaction do
       PayrollRun.create!(attrs).tap do |payroll_run|
         claims.each do |claim|
-          Payment.create!(payroll_run: payroll_run, claim: claim, award_amount: claim.award_amount)
+          Payment.create!(payroll_run: payroll_run, claims: [claim], award_amount: claim.award_amount)
```
tekin: Would using a `has_one` association on `Payment` mean this change (and a bunch of similar changes in the specs) could come later when mulit-claim support is actually implemented?

app/models/payment.rb
lawrence-dxw: Nope, it didn't occur to us. But I think it's a good idea, in the spirit of small changes. I've changed it.

app/models/payment.rb
lawrence-dxw: None; it served more as documentation than anything else. I've removed it.

app/models/payroll_run.rb
lawrence-dxw: Yep. Changed it.

PR #699, opened 2019-12-06
Make payments CSV classes use payments, not claims

lawrence-dxw:
This is prep work for updating the CSV to work with multiple claims,
where each line of the CSV will correspond to a single _payment_, not
necessarily a single claim. We modify and rename the CSV classes to
reflect this.

When we move to multiple claims, there will be some data fields which
are duplicated across a payment's claims, for example address, payroll
gender, etc. We add these fields as methods on the Payment class. At the
moment, they just delegate to the payment's claim. Once we move to
multiple claims, they will probably delegate to the first claim of the
payment. This allows us to keep the CSV generation code the same, and
abstract away the process of getting e.g. an address for a payment.

### DIFF
--- a/spec/models/payroll/payment_csv_row_spec.rb
+++ b/spec/models/payroll/payment_csv_row_spec.rb
@@ -1,13 +1,13 @@
 require "rails_helper"
 
-RSpec.describe Payroll::ClaimCsvRow do
-  subject { described_class.new(claim) }
+RSpec.describe Payroll::PaymentCsvRow do
+  subject { described_class.new(payment) }
   let(:claim) { build(:claim) }
 
   describe "to_s with an approved claim that has a payment" do
     let(:row) { CSV.parse(subject.to_s).first }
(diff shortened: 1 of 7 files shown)


spec/models/payroll/payment_csv_row_spec.rb
```diff
@@ -1,13 +1,13 @@
...
-RSpec.describe Payroll::ClaimCsvRow do
-  subject { described_class.new(claim) }
+RSpec.describe Payroll::PaymentCsvRow do
+  subject { described_class.new(payment) }
   let(:claim) { build(:claim) }
 
   describe "to_s with an approved claim that has a payment" do
     let(:row) { CSV.parse(subject.to_s).first }
     let(:payment_award_amount) { BigDecimal("1234.56") }
-    let!(:payment) { build(:payment, award_amount: payment_award_amount, claim: claim) }
```
tekin: 👍 this bothered me ever so slightly in the last PR, glad to see this refactoring means it can go.

PR #703, opened 2019-12-09
Schools data import tweaks

tekin:
A couple of quality of life improvements for developers around schools data import


pezholio:
Tests here may need tweaking because at the moment we're only testing that a job gets queued

PR #706, opened 2019-12-09
Make payment confirmation classes use payments, not claims

lawrence-dxw:
This is groundwork for allowing a claimant to be paid for multiple claims with
a single payment.

This updates the PaymentConfirmation class, and the mailer and Geckoboard
reporting job that it uses, to use a payment as their primary concept instead
of claims. I did a similar thing in #699, for the payroll CSV classes.


app/mailers/payment_mailer.rb
```diff
@@ -0,0 +1,30 @@
...
+    view_mail_with_payment_and_subject(payment, "We’re paying your #{@claim_description}, reference number: #{payment.claim.reference}")
+  end
+
+  private
+
+  def claim_description(claim)
+    I18n.t("#{claim.policy.routing_name.underscore}.claim_description")
+  end
+
+  def view_mail_with_payment_and_subject(payment, subject)
```
tekin: Would it make the code easier to follow if the code in this method was simply inlined in the `confirmation` method? I get why this is here (it was present in the `ClaimMailer`), but over there it (kinda) justifies its existence as it DRYs up the setting of the display name and policy. I see fewer benefits where it is called once vs the indirection it creates.

app/models/payment_confirmation.rb
```diff
@@ -25,7 +25,7 @@ def ingest
 
       if errors.empty?
         payroll_run.claims.each do |claim|
-          ClaimMailer.payment_confirmation(claim, payment_date_timestamp).deliver_later
+          PaymentMailer.confirmation(claim.payment, payment_date_timestamp).deliver_later
```
tekin: Given that we only use the `claim` to access its `payment`, does it make sense to loop through `payroll_run.payments` above instead of the claims?

spec/mailers/payment_mailer_spec.rb
```diff
@@ -0,0 +1,69 @@
+require "rails_helper"
+
+RSpec.shared_examples "an email related to a payment" do |policy|
```
tekin: Shared examples are useful for DRYing up specs where there are many things that behave the same way – using a shared example removes the need to duplicate the same specs over and over. This shared example is only used once below, which makes it hard to justify. A couple of alternatives I can think of: 

a) get rid of the shared example here and inline the examples below in the `describe "#payment_confirmation"` block
b) figure out how to change the existing shared example such that it can be re-used in the specs below, rather than duplicating a near identical one. The shared example doesn't need to live in a spec file. There are some conventions for where to store them in the docs — https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples

What do you reckon?

spec/jobs/record_payment_job_spec.rb
```diff
@@ -7,7 +7,7 @@
 
   subject { described_class.new }
 
-  it "sends the claim reference, policy and payment date to claims.submitted.ENV dataset" do
+  it "sends the claim reference, policy and payment date to claims.paid.ENV dataset" do
```
tekin: 👍

app/models/payment_confirmation.rb
lawrence-dxw: Yep - it's in the final commit.

tekin: 👍

lawrence-dxw:
I've also realised that I need to fix the mailer preview. Will do that.

app/mailers/payment_mailer.rb
lawrence-dxw: Agreed. I'll change it.

spec/mailers/payment_mailer_spec.rb
lawrence-dxw: I don't think we should make the shared example reusable, because the behaviour that it's testing is going to diverge in the payment mailer shortly, once we introduce multiple claims. I think handling both behaviours in a single set of examples would not be worth the complexity.

lawrence-dxw: Which is to say: I'll inline them 😄

lawrence-dxw:
Ready for re-review 👍

PR #708, opened 2019-12-10
Add a .gitkeep to the bin/vsp directory

jacksonj04:
Adds a `.gitkeep` file to the `bin/vsp` directory so that download/unzip/copy of VSP doesn't fail.


tekin:
One suggestion: the PR description explains "why" this change. Including that "why" in the commit message as well would mean it was there in the history for a future developer that might want to understand that.

PR #711, opened 2019-12-10
Don't match bank account sort code when identifying potentially similar claims

pezholio:
This changes the `ATTRIBUTES_TO_MATCH` array to be an array of groups to match against in the database. If there is more than one attribute in the group, we concatenate the subjects and  predicates together, so we can make sure both things match. At the moment, we only need this for bank details, but we may need this for other things (such as maybe firstname / lastname / DOB etc)

### DIFF
--- a/app/models/claim/matching_attribute_finder.rb
+++ b/app/models/claim/matching_attribute_finder.rb
@@ -1,12 +1,10 @@
 class Claim
   class MatchingAttributeFinder
-    ATTRIBUTES_TO_MATCH = %w[
-      teacher_reference_number
-      email_address
-      national_insurance_number
-      bank_account_number
-      bank_sort_code
-      building_society_roll_number
+    ATTRIBUTE_GROUPS_TO_MATCH = [
+      ["teacher_reference_number"],
+      ["email_address"],
+      ["national_insurance_number"],
+      ["bank_account_number", "bank_sort_code", "building_society_roll_number"],
     ].freeze
 
     def initialize(source_claim, claims_to_compare = Claim.submitted)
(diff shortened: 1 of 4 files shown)


pezholio:
[Don't match bank account sort code when identifying potentially similar claims](https://trello.com/c/TbZJK4Kc/1050-dont-match-bank-account-sort-code-when-identifying-potentially-similar-claims)

app/models/claim/matching_attribute_finder.rb
```diff
@@ -17,15 +15,26 @@ def initialize(source_claim, claims_to_compare = Claim.submitted)
...
-        value = @source_claim.read_attribute(attribute)
-        next if value.blank?
+      ATTRIBUTE_GROUPS_TO_MATCH.each do |attributes|
+        vals = values_for_attributes(attributes)
 
-        predicates << "LOWER(#{attribute}) = LOWER(?)"
-        values << value
+        next if vals.blank?
+
+        subject = "concat(#{attributes.join(",")})"
```
mec: As `concat` here is part of the SQL statement, should it be capitalised?

app/models/claim/matching_attribute_finder.rb
```diff
@@ -17,15 +15,26 @@ def initialize(source_claim, claims_to_compare = Claim.submitted)
...
+
+        predicates << "LOWER(#{subject}) = LOWER(?)"
```
mec: We were wondering if these two variables could have more intention revealing names? It took a while for use to un-pick what was going on and we thought that might help? Something like: `columns_to_concat` and `comparison_statement`?

app/models/claim/matching_attribute_finder.rb
```diff
@@ -15,17 +13,29 @@ def initialize(source_claim, claims_to_compare = Claim.submitted)
...
-
-        predicates << "LOWER(#{attribute}) = LOWER(?)"
-        values << value
+      claims = @claims_to_compare.where.not(id: @source_claim.id)
+      match_queries = nil
+
+      ATTRIBUTE_GROUPS_TO_MATCH.each do |attributes|
+        vals = values_for_attributes(attributes)
+
+        next if vals.blank?
```
tekin: `empty?` would be more appropriate for an array as it will signal to the reader what the type of `vals` is. `blank?` is more for values that are expected to be a string.

app/models/claim/matching_attribute_finder.rb
```diff
@@ -15,17 +13,29 @@ def initialize(source_claim, claims_to_compare = Claim.submitted)
...
-        value = @source_claim.read_attribute(attribute)
-        next if value.blank?
-
-        predicates << "LOWER(#{attribute}) = LOWER(?)"
-        values << value
+      claims = @claims_to_compare.where.not(id: @source_claim.id)
+      match_queries = nil
+
+      ATTRIBUTE_GROUPS_TO_MATCH.each do |attributes|
+        vals = values_for_attributes(attributes)
```
tekin: Unless there's a compelling reason to abbreviate, full variable names are generally better as they are easier to read and understand.

PR #713, opened 2019-12-11
Fix test flicker on payroll run spec

pezholio:
This test seems to ocassionally fail because claims aren't always in the order we expect them. I've got around this by mapping the subjects and addressee emails to an array and then matching the array, so we don't worry about the order.

### DIFF
--- a/spec/features/admin_payroll_runs_spec.rb
+++ b/spec/features/admin_payroll_runs_spec.rb
@@ -123,14 +123,18 @@ RSpec.feature "Payroll" do
 
     expect(ActionMailer::Base.deliveries.count).to eq(2)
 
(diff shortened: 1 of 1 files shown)


mec:
I like the fix, but I am left wondering if we need the level of coverage on the emails in this feature?!

For almost all the other features that result in a email we just look for the count to change, I couldn't see a mailer spec for this email either - if we had coverage there that checks the reference number, would this bit of the feature still be needed?

pezholio:
It's a good shout, but if we just check that en email gets sent, we lose the confidence (for me anyway) that the right emails are being sent with the right content. I think also if we add an extra mailer spec that adds to the maintenance load with another bunch of tests to look after.

PR #716, opened 2019-12-16
Import users from DfE Signin API

pezholio:
Currently, when users perform an action when logged in as an admin user, we record their DfE Sign-in ID against the action. This is just a plain UUID as a string, and, if we wanted to find out exactly who'd done something, we'd have to access the database, get the UUID and then check the DfE Signin API to get their user details. 

As part of the process of displaying a user's name against actions that have been performed, this does three things:

* Sets up a `DfeSignIn::User` model
* Adds an endpoint to get all users from the DfE Sign-in API
* Adds a class and job to add/update users at 2am every day

It also runs the add/update class as a data migration to ensure that we have all the users available when we deploy.

Once this is merged and deployed, we can then start referencing the UUIDs against records.


pezholio:

mec:
The way the DfE sigin UUID is saved as 'our' UUID is making me a little nervous - I struggled to see how we were storing it until you pointed it out to me, which made it a bit magical for a simpleton like me!

I think it's that we are using a third party bit of data as our own UUID here that worries me, I've tried to think through why that is and can't come up with too much concrete, I guess it's just in my `jellies`!? 

I checked the UUID docs for Postgres and had a look at the DfE signin UUIDs and they seem to follow the RFC standard, but should we rely on that in our own system, what happens if for whatever reason we get a malformed or duplicate UUID back from DfE sigin, are we relying too heavily on DfE signin by baking this in to our application?

My train of thought led me to ask 'what is the alternative' and I guess it would be storing the DfE siginin UUID in it's own column and letting Postgres take care of our own UUIDs. I think I feel like although not as clever, it would make the code more accessible for someone like myself.

I was also thinking about the advantage of using the DfE UUID as our own, I can see how it would make fetching the name from the UUID super simple and it is really neat - but I do wonder if the 'magic' could happen in fetching the name rather than in how we store the DfE UUID? 

What are your thoughts?

pezholio:
Nope, I'm totally with you. I don't think it makes much sense at all to use the same UUID, especially given they should be Universally Unique (the clue's in the name 😄). I'll tweak it so we have two IDs.

spec/lib/dfe_sign_in/api/user_spec.rb
```diff
@@ -1,72 +1,112 @@
...
+        }
+      end
 
-    describe "#role_codes" do
-      it "returns the role codes" do
-        expect(subject.role_codes).to eq(["my_role"])
+      describe "has_role?" do
+        let(:has_role?) { subject.has_role?(role) }
+
+        context "when a role exists" do
```
mec: I am battliing with how I feel about conexts like this at the moment, I suspect some people like them and it reflects their mental model, whilst other people, the added processing and nesting burdens them, I do wonder if in a simple case like this they bring complexity where it may not be needed, how would something like 

````ruby
describe "#has_role?" do
  it "returns true when the user has the role" do
    expect(subject.has_role?("my_role")).to eq(true)
    expect(subject.has_role?("other_role")).to eq(false)
  end
end
````
feel to you?

File: db/data/20191216123406_import_users_from_dfe_signin.rb [hunk 2]
```diff
@@ -0,0 +1,9 @@
+class ImportUsersFromDfeSignin < ActiveRecord::Migration[6.0]
```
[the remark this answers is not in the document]
pezholio: Yeah, actually, that might be safer. I was worried about later data migrations (when we move to referencing a user) breaking if a user isn't yet present, but I can guard against this.

app/jobs/user_data_importer_job.rb
```diff
@@ -0,0 +1,10 @@
+class UserDataImporterJob < CronJob
```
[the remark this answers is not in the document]
pezholio: Yeah, we could actually namespace this, as I have with the users

db/schema.rb
```diff
@@ -85,6 +85,16 @@
     t.index ["priority", "run_at"], name: "delayed_jobs_priority"
   end
 
+  create_table "dfe_sign_in_users", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.string "dfe_sign_in_id"
+    t.string "given_name"
+    t.string "family_name"
+    t.string "email"
+    t.string "organisation_name"
+    t.datetime "created_at", precision: 6, null: false
```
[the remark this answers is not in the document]
pezholio: Yeah, it was introduced in this PR https://github.com/rails/rails/pull/34970, to solve problems with timestamps in certain versions of MySQL. There's an open issue talking about it here https://github.com/rails/rails/issues/37031. I don't think there's any danger in keeping it.

spec/models/cron_job_scheduler_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  def queue_adapter_for_test
+    DelayedJobTestAdapter.new
+  end
+
+  it "schedules SchoolDataImporterJob" do
+    expect {
+      described_class.new.schedule
+    }.to change {
+      Delayed::Job
+        .where("handler LIKE ?", "%job_class: SchoolDataImporterJob%")
```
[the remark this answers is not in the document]
pezholio: Ah yeah, that works perfectly. I couldn't get that approach to work in the Rake tasks, but here it works perfectly

spec/lib/dfe_sign_in/api/user_spec.rb
```diff
@@ -1,72 +1,112 @@
...
+              "value" => "identifier-value",
+            },
+          ],
+        }
+      end
 
-    describe "#role_codes" do
-      it "returns the role codes" do
-        expect(subject.role_codes).to eq(["my_role"])
+      describe "has_role?" do
```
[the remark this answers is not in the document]
pezholio: It's a convention that's documented [here](http://www.betterspecs.org/#describe). We do seem to apply it in a scattergun fashion though. It doesn't always get used. I need to remember to do it consistently.

spec/lib/dfe_sign_in/api/user_spec.rb
pezholio: 👍

spec/lib/dfe_sign_in/api/user_spec.rb
```diff
@@ -1,72 +1,112 @@
...
+    context "with one page" do
+      before do
+        stub_dfe_sign_in_user_list_request
+      end
 
-  before do
-    stub_request(:get, url)
-      .to_return(body: response.to_json, status: status)
-  end
+      it "gets all pages" do
```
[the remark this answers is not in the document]
pezholio: 👍

spec/lib/dfe_sign_in/api/user_spec.rb
```diff
@@ -1,72 +1,112 @@
...
+    context "with multiple pages" do
+      let!(:first_page_stub) { stub_dfe_sign_in_user_list_request(number_of_pages: 3) }
+      let!(:second_page_stub) { stub_dfe_sign_in_user_list_request(number_of_pages: 3, page_number: 2) }
+      let!(:third_page_stub) { stub_dfe_sign_in_user_list_request(number_of_pages: 3, page_number: 3) }
 
-      context "when a role exists" do
-        let(:role) { "my_role" }
-        it { expect(has_role?).to eq(true) }
-      end
+      it "requests multiple pages" do
```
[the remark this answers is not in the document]
pezholio: 👍

app/jobs/dfe_sign_in/user_data_importer_job.rb
```diff
@@ -0,0 +1,10 @@
+class DfeSignIn::UserDataImporterJob < CronJob
+  self.cron_expression = "0 2 * * *"
+
+  queue_as :user_data
+
+  def perform
+    Rails.logger.info "Importing user data..."
```
tekin:
```suggestion
    Rails.logger.info "Importing DfE Sign-in user data..."
```

app/models/cron_job_scheduler.rb
```diff
@@ -0,0 +1,13 @@
+class CronJobScheduler
+  def initialize
+    Dir.glob(glob).each { |f| require f }
+  end
+
+  def schedule
+    CronJob.subclasses.each { |job| job.schedule }
+  end
+
+  def glob
```
tekin: It's good to limit the public interface of classes we create to only those methods we expect to get called publicly, so maybe this method should be `private`/

app/models/dfe_sign_in/user_data_importer.rb
```diff
@@ -0,0 +1,15 @@
...
+    users = DfeSignIn::Api::User.all
+    users.each do |u|
+      user = DfeSignIn::User.find_or_initialize_by(dfe_sign_in_id: u.user_id)
+
+      user.given_name = u.given_name
+      user.family_name = u.family_name
+      user.email = u.email
+      user.organisation_name = u.organisation_name
+
+      user.save
```
tekin: We should probably call `save!` so this fails noisily if unexpectedly a user record cannot be saved or updated, otherwise we'll silently fail to update/save records.

lib/dfe_sign_in/api/user.rb
```diff
@@ -22,19 +66,7 @@ def role_codes
...
-
-          response = Net::HTTP.start(uri.hostname, uri.port, use_ssl: true) { |http|
-            http.request(request)
-          }
-
-          raise ExternalServerError, "#{response.code}: #{response.body}" unless response.code.eql?("200")
-
-          JSON.parse(response.body)
-        end
+        @body ||= self.class.get(uri)
```
tekin: because the module with the `get` method has been included and extended, could this be written as:

```suggestion
        @body ||= get(uri)
```

mec:
👍

lib/dfe_sign_in/api/user.rb
```diff
@@ -2,13 +2,57 @@ module DfeSignIn
...
+
+        def get_users
+          uri = all_users_endpoint
+          body = get(uri)
+          page_number = 2
+          users = body["users"]
+          while page_number <= body["numberOfPages"]
+            uri.query = "page=#{page_number}"
+            page = get(uri)
+            users << page["users"]
```
tekin: Would using [`Array#concat`](https://apidock.com/ruby/Array/concat) remove the need for the flatten call further down?

```suggestion
            users.concat page["users"]
```

db/migrate/20191211160017_create_dfe_sign_in_users.rb
```diff
@@ -0,0 +1,13 @@
+class CreateDfeSignInUsers < ActiveRecord::Migration[6.0]
+  def change
+    create_table :dfe_sign_in_users, id: :uuid do |t|
+      t.string :dfe_sign_in_id
```
tekin: We'll want to make this unique (as it should always be unique) and also indexed (because it's the column we'll be performing the SELECT on when we look up a user record)

spec/factories/dfe_sign_in/users.rb
```diff
@@ -0,0 +1,9 @@
+FactoryBot.define do
+  factory :dfe_signin_user, class: "DfeSignIn::User" do
+    dfe_sign_in_id { "7516e8f7-ab90-4eb6-a998-b798e835ca3d" }
```
tekin: This should be generated rather than a fixed string, otherwise all the factory-created users will have the same unique identifier.

spec/lib/dfe_sign_in/api/user_spec.rb
```diff
@@ -1,72 +1,108 @@
...
+          "identifiers" => [
+            {
+              "key" => "identifier-key",
+              "value" => "identifier-value",
+            },
+          ],
+        }
+      end
+
+      describe "#has_role?" do
```
tekin: It looks like the nesting has gone awry here: this is inside the `describe ".new"` block, which feels wrong. Maybe we should discuss this nesting in meatspace.

spec/models/dfe_sign_in/user_data_importer_spec.rb
```diff
@@ -0,0 +1,28 @@
+require "rails_helper"
+
+RSpec.describe DfeSignIn::UserDataImporter, type: :model do
+  before { stub_dfe_sign_in_user_list_request }
+
+  it "imports all user data" do
+    DfeSignIn::UserDataImporter.new.run
+
+    expect(DfeSignIn::User.count).to eq(3)
```
tekin: What about verifying the actual values of the user records are being correctly set? Seems like a whole heap of behaviour isn't being tested here.

app/models/dfe_sign_in/user.rb
```diff
@@ -0,0 +1,9 @@
+class DfeSignIn::User < ApplicationRecord
```
tekin: Should we be validating the presence of things like name, email, organisation, etc?

PR #717, opened 2019-12-16
Unpin Bundler version in Dockerfile

tekin:
We locked the version of Bundler to 2.0.2 in the Dockerfile because with
the latest version (2.1.0) the app was failing to start – the calls to
the `rails` command in docker-entrypoint were resulting in "command not
found" errors. We traced the issue to the fact that bundler 2.1.0 when
installed in the Docker container doesn't appear to be creating any
binstubs for commands like `rake` or `rails`. Instead of relying on the
presence of binstubs we call the commands with `bundle exec` instead.

The second commit gets rid of the deprecation warnings introduced by bundler 2.1.0.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,7 @@ WORKDIR ${DEPS_HOME}
 COPY Gemfile ${DEPS_HOME}/Gemfile
 COPY Gemfile.lock ${DEPS_HOME}/Gemfile.lock
 
-RUN gem install bundler -v 2.0.2
+RUN gem install bundler
 ENV BUNDLE_BUILD__SASSC=--disable-march-tune-native
 
 RUN if [ ${RAILS_ENV} = "production" ]; then \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -35,17 +35,21 @@ WORKDIR ${DEPS_HOME}
...
-  bundle install --frozen --retry 3 --without development; \
+  bundle config set without 'development'; \
   else \
-  bundle install --frozen --retry 3 --without test; \
+   bundle config set without 'test'; \
   fi
 # End
+RUN bundle config
+
+RUN bundle install --retry 3
```
pezholio: Do we no longer care about the `--frozen` flag now?

mec: Should we use the `--deployment` flag - this would set 'frozen` to true as well? I am not sure we care about the optimisations it makes and it will move the gems, not sure where GEM_HOME is currently?

https://bundler.io/v2.0/man/bundle-install.1.html#DEPLOYMENT-MODE

tekin: @pezholio we do, getting rid of the deprecation warning meant setting it with `bundle config` instead, which is being done on line 41.

tekin: @mec Oh, I didn't know about `--deployment`. I'm finding the docs for it to be a little ambiguous. It does look on the surface like it would be equivalent to `--frozen --without development test` but I'm not 100% on that. OK if we stick with this more predictable/safer change to get bundler unpinned, rather than switch to a potentially different config for production?

pezholio: Arrrgh. I missed that. I was expecting to see it either in or after the conditional

PR #720, opened 2019-12-17
Fix Brakeman

pezholio:
Brakeman currently does not exit with a non-zero exit code, so if we get any warnings, the test suite continues and we don't see any issues (unless we scroll back an read the output). This updates the Rake task to exit appropriately if there are any warnings. 

We also previously had a warning, which we're safe to ignore.

### DIFF
--- a/lib/tasks/brakeman.rake
+++ b/lib/tasks/brakeman.rake
@@ -3,11 +3,14 @@ namespace :brakeman do
   task :run do
     require "brakeman"
 
-    Brakeman.run(
+    result = Brakeman.run(
       app_path: ".",
       quiet: true,
       pager: false,
       print_report: true
     )
+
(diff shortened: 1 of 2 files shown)


lib/tasks/brakeman.rake
```diff
@@ -3,11 +3,13 @@ namespace :brakeman do
...
 
-    Brakeman.run(
+    result = Brakeman.run(
       app_path: ".",
       quiet: true,
       pager: false,
       print_report: true
     )
+
+    exit Brakeman::Warnings_Found_Exit_Code unless result.filtered_warnings.empty?
```
tekin: Would this also exit if there brakeman errors?

pezholio: Ah, right you are. I copied this out of the (now deprecated and removed) Brakeman Rake task, which didn't account for errors for some reason. I've fixed this now

PR #721, opened 2019-12-17
Update SchoolDataImporter to handle redirects

tekin:
The GIAS data file moved at some point recently but we didn't know about it because the import code carried on and processed the blank response. This PR updates to the new download URL, but also enhances the code such that:

- it will automatically handle any redirects for the file
- it will raise an exception if download fails for any other reason

This way we should find out about any failures in the schools data import as we'll see an exception in Rollbar.


app/models/school_data_importer.rb
```diff
@@ -2,6 +2,12 @@
 require "csv"
 
 class SchoolDataImporter
+  # Returns the URL for today's schools data CSV file
+  def self.gias_schools_csv_url
```
pezholio: What's the rationale behind making this a class method rather than an instance method?

lib/file_download.rb
```diff
@@ -0,0 +1,36 @@
...
+    download_file(url)
+  end
+
+  private
+
+  def download_file(url)
+    response = Net::HTTP.get_response(URI(url))
+
+    case response
+      when Net::HTTPSuccess
```
pezholio: At a risk of bikeshedding over style - is it OK if we unindent the `when`s here? Everywhere else we've used `case` statements, the `when`s have been indented at the same level as the `case`, and it just (IMO) makes the statement a bit more readable.

lib/file_download.rb
```diff
@@ -1,8 +1,12 @@
 require "net/http"
 
-# Used to download a remote file. Handles redirects.
+# Used to download a remote file. Handles redirects up to a maximum of 5.
```
pezholio: Is it worth mentioning that this can be configured?

lib/file_download.rb
tekin: 👍🏽

lib/file_download.rb
tekin: It can't be configured. The value of 5 comes from the `FileDownload::MAX_REDIRECTS` constant and the place it is used is a private method.

pezholio: 👍

app/models/school_data_importer.rb
tekin: Mostly because it makes the specs for this class simpler – we can stub the requests with this value without having to instantiate an instance, or replicate the implementation to get the correct time-stamped URL.

pezholio: Ah, yeah, of course. I think that became clearer as my review progressed. Good thinking 👍

tekin: I mean it's not a great reason to be fair... Might have made more sense to keep it private and freeze time in the specs...

tekin: OK, I've changed my mind and pushed up an additional commit to make this method private again.

PR #723, opened 2019-12-17
Update student loan question

mec:
Users found the previous wording for this question confusing and we saw
support tickets about it. This change aims to make it clearer what the
question is asking.


mec:
[Change how we ask users if we need to make student loan deductions](https://trello.com/c/mytKXV4n/1123-change-how-we-ask-users-if-we-need-to-make-student-loan-deductions)

PR #726, opened 2019-12-18
Users can submit a claim, bypassing GOV.UK Verify

tekin:
Part one of allowing users to make a claim without completing GOV.UK Verify. Further PRs will follow to meet all the acceptance criteria, but this gets the foundations in place without letting the PR get too big.

Not all users are able to complete the GOV.UK Verify journey, particularly if they are new in the country or are lacking a comprehensive digital footprint. So that all eligible users can make a claim, we are introducing a process that allows users to bypass GOV.UK Verify and submit their claim, at which point service operators will perform an offline check to confirm the claimant's identity. This PR introduces the first part of that process: Allowing users to claim without completing GOV.UK Verify.

For now, that involves coming back to the service at a specific URL to pick up the journey at the first identity question. Eventually we will have a single endpoint that we will link back to from GOV.UK Verify's failure page, but for now we can test the process by manually visiting that page (in this case, it will be `/student-loans/name` or `/maths-and-physics/name`, depending on the type of claim being made).

This PR also updates the back office such that claims that have skipped GOV.UK Verify will be flagged to service operators and they will be prevented from approving such claims. The work to allow service operators to confirm the identity on such claims will follow in a separate story.

Full acceptance criteria, including an indication of the work still outstanding on this feature can be seen here: https://trello.com/c/Lgj5pKUw/1081-a-user-can-complete-their-claim-having-failed-to-get-through-govuk-verify


spec/models/check_spec.rb
```diff
@@ -14,4 +14,12 @@
     expect(build(:check, result: "approved")).to be_valid
     expect(build(:check, result: nil)).not_to be_valid
   end
+
+  it "prevents an unapprovable claim from being approved" do
+    claim = create(:claim, :unverified)
+    check = build(:check, claim: claim, result: "approved")
+
+    expect(check).not_to be_valid
+    expect(check.errors.messages[:base]).to eq ["This claim cannot be approved"]
```
pezholio: Can we add brackets to the call to `eq` here? Seeing the start of the array threw me a bit.

tekin: 👍🏽

PR #731, opened 2019-12-19
Display the name of the service operator who approved a claim

pezholio:
Following on from #716 - this now fetches or creates a user from the database when a user approves a claim. This is then displayed on the claim decision details like so:

image

If the user's details have not yet been imported, the decision details show like so:

image


app/models/admin_session.rb
```diff
@@ -5,6 +5,10 @@ class AdminSession < DfeSignIn::AuthenticatedSession
   SUPPORT_AGENT_DFE_SIGN_IN_ROLE_CODE = "teacher_payments_support"
   PAYROLL_OPERATOR_DFE_SIGN_IN_ROLE_CODE = "teacher_payments_payroll"
 
+  def user
+    @user ||= DfeSignIn::User.find_or_create_by(dfe_sign_in_id: user_id)
```
tekin: Didn't we discuss doing this at the point of user sign-in and storing our own UID as the session key?

app/controllers/admin/auth_controller.rb
```diff
@@ -12,9 +12,10 @@ def sign_out
 
     def callback
       admin_session = AdminSession.from_auth_hash(request.env.fetch("omniauth.auth"))
+      dfe_sign_in_user = DfeSignIn::User.find_or_create_by(dfe_sign_in_id: admin_session.user_id)
```
tekin: This should only really happen `if admin_session.has_admin_access?`, otherwise we'll create user records even for those users that don't have access.

db/migrate/20191216135816_add_checked_by_reference_to_checks.rb
```diff
@@ -0,0 +1,11 @@
+class AddCheckedByReferenceToChecks < ActiveRecord::Migration[6.0]
+  def up
+    add_column :checks, :checked_by_id, :uuid, using: "checked_by::uuid"
```
tekin: I'm pretty sure we could use `add_reference` to do both things in one go and use `change` instead to get the automagic migrate down.

db/schema.rb
```diff
@@ -23,6 +23,7 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.text "notes"
+    t.uuid "checked_by_id"
```
tekin: Have we captured somewhere that we'll want to drop the `checked_by` column that this work makes redundant?

db/migrate/20191216135816_add_checked_by_reference_to_checks.rb
pezholio: We could, but this would result in a field called `dfe_sign_in_user_id`, which doesn't really reflect the intention of the column.

tekin: Does the `add_reference` API not allow the foreign key to be overridden?

tekin: Had a look at the [API](https://apidock.com/rails/ActiveRecord/ConnectionAdapters/SchemaStatements/add_reference), I think this would do it:

```suggestion
  def change
    add_reference :checks, :checked_by, type: :uuid, foreign_key: {to_table: :dfe_sign_in_users}
  end
```

pezholio: Arrrrgh. That's got it. I was looking at the wrong version of the docs.

db/schema.rb
pezholio: I've added a checkbox in the card in Trello

db/schema.rb
```diff
@@ -46,10 +48,10 @@
     t.boolean "has_student_loan"
     t.integer "student_loan_country"
     t.integer "student_loan_courses"
-    t.integer "student_loan_start_date"
-    t.integer "student_loan_plan"
     t.integer "payroll_gender"
     t.text "verified_fields", default: [], array: true
+    t.integer "student_loan_start_date"
+    t.integer "student_loan_plan"
```
robbiepaul: Did the order of these mean to change in this commit?

pezholio: Gaaaah. This keeps happening. I've tried dropping and recreating the DB from scratch and it still happens on every migration 😠

app/models/dfe_sign_in/user.rb
```diff
@@ -1,11 +1,42 @@
...
+
+    def is_payroll_operator?
+      role_codes.include?(PAYROLL_OPERATOR_DFE_SIGN_IN_ROLE_CODE)
+    end
+
+    def has_admin_access?
+      is_service_operator? || is_support_agent? || is_payroll_operator?
+    end
+
+    def update_role_codes_from_session(session)
```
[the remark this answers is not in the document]
pezholio: Good catch 👍

pezholio:

PR #732, opened 2019-12-19
Ensure first migration will run without error

tekin:
Because of the timestamp ordering of the first two migrations, running
`rails db:migrate` on a fresh install will fail. This is because the first
timestamped migration tries to create a table with a UUID primary key
_before_ the PostgreSQL extension (`pgcrypto`) required to support UUID
primary keys has been enabled (happens in the second timestamped
migration).

This wasn't an issue when the app was first set up as the second
migration was merged before the first. But this makes install difficult
to do things like run a Heroku pipeline for review apps where we're
running `rails db:migrate` as part of the release process.

Modifying the existing migration to enable the extension feels like the
least bad option as it won't impact existing deployments (no timestamps
are changed and thus as far as Rails is concerned, all existing
migrations have been run), and `enable_extension` is idempotent so
calling it twice will have no negative impact.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,3 +1,3 @@
 web: bin/rails server -p $PORT -e $RAILS_ENV
 worker: bundle exec rake jobs:work
-release: rake db:migrate
+release: yarn install && rake db:migrate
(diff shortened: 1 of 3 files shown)


tekin:
Closing this as there is more work to do before we can run the application on Heroku.

PR #734, opened 2019-12-19
Adjust content when GOV.UK Verify hasn't been completed

tekin:
Users that do not complete GOV.UK Verify see different content explaining how their claim will be handled.

This PR also introduces the endpoint that we will link back to from GOV.UK Verify's failure screen so that users can pick up where they left off.

### DIFF
--- a/docs/govuk-verify.md
+++ b/docs/govuk-verify.md
@@ -50,6 +50,17 @@ flow from here:
 These steps are explained in more detail in the
 [Verify documentation](https://www.docs.verify.service.gov.uk/get-started/set-up-successful-verification-journey/#run-the-identity-verified-response-scenario).
 
+### How to simulate skipping the Verify user journey in local development
+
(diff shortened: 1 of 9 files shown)


docs/govuk-verify.md
```diff
@@ -50,6 +50,17 @@ flow from here:
...
+### How to simulate skipping the Verify user journey in local development
+
+Users that are unable to complete the GOV.UK Verify process are still able to
+submit a claim. To do so they are linked back to the service on the GOV.UK
+Verify failure screen so they can provide their identity information and
+complete their claim. Such claims are then manually check to confirm the
+claimants identity. To simulate such a claim, visit the following URL after you
+are redirected to the Verify JSON page:
+
+/verify/authentications/skip
```
pezholio: I think it might be easier to direct devs to visit the page _before_ clicking on `Continue` at the "information-provided" stage to save having to enter the full URL

PR #735, opened 2019-12-20
Add BUILD_NUMBER env var to containers

pezholio:
Currently the workers aren't getting the up to date code when we do a production / development deploy. After a lot of chat with other ops people we've realised that this is because we always the the `development` Docker tag, so the container doesn't know that anything has changed, so doesn't pull down the latest image.

This adds a new environment variable to the web and worker containers, set by setting the `AZURE_BUILD_NUMBER` in the build environment when deploying. This means that the worker container knows something has changed and restarts itself, pulling down the latest version of the `development` or `production` image from DockerHub.

Once this is merged, we'll also need to set an `AZURE_BUILD_NUMBER` environment variable in the Azure DevOps config to the value of the build we're deploying.

### DIFF
--- a/azure/resource_groups/app/template.json
+++ b/azure/resource_groups/app/template.json
@@ -100,6 +100,10 @@
     "alertEmailAddress": {
       "type": "string"
     },
+    "azureBuildNumber": {
+      "type": "string",
+      "defaultValue": ""
+    },
     "RAILS_ENV": {
       "type": "string",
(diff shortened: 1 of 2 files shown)


pezholio:
[Workers not getting up to date code](https://trello.com/c/4fuyNiwr/1135-workers-not-getting-up-to-date-code)

PR #736, opened 2019-12-24
1116 inconsistent error messages

titlescreen:
Updated error messages to make them consistent and provide some more guidance to users. To be clear these are the messages shown when you don't select an option/input any text and try to continue. 

Messages have been changed to follow the pattern e.g. `select yes if you teach any maths or physics` rather than telling the user to select an option. Messages should now also be consistent in their use of case and punctuation.

Miro board of new error messages can be found here https://miro.com/app/board/o9J_kvlNgOg=/

### DIFF
--- a/app/models/maths_and_physics/eligibility.rb
+++ b/app/models/maths_and_physics/eligibility.rb
@@ -47,17 +47,17 @@ module MathsAndPhysics
 
     belongs_to :current_school, optional: true, class_name: "School"
 
-    validates :teaching_maths_or_physics, on: [:"teaching-maths-or-physics", :submit], inclusion: {in: [true, false], message: "Select either Yes or No"}
-    validates :current_school, on: [:"current-school", :submit], presence: {message: "Select a school from the list"}
-    validates :initial_teacher_training_subject, on: [:"initial-teacher-training-subject", :submit], presence: {message: "Choose a subject, or select None of these subjects"}
-    validates :initial_teacher_training_subject_specialism, on: [:"initial-teacher-training-subject-specialism", :submit], presence: {message: "Choose a subject, or select I'm not sure"}, if: :itt_subject_science?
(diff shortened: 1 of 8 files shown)


app/models/maths_and_physics/eligibility.rb
```diff
@@ -47,17 +47,17 @@ class Eligibility < ApplicationRecord
...
+    validates :initial_teacher_training_subject_specialism, on: [:"initial-teacher-training-subject-specialism", :submit], presence: {message: "Select the subject your initial teacher training specialised in or select I'm not sure"}, if: :itt_subject_science?
+    validates :has_uk_maths_or_physics_degree, on: [:"has-uk-maths-or-physics-degree", :submit], presence: {message: "Select yes if you have a UK degree specialising in maths or physics"}, unless: :initial_teacher_training_specialised_in_maths_or_physics?
+    validates :qts_award_year, on: [:"qts-year", :submit], presence: {message: "Select whether you completed your initial teacher training before or after September 2014"}
+    validates :employed_as_supply_teacher, on: [:"supply-teacher", :submit], inclusion: {in: [true, false], message: "Select yes if you are currently employed as a supply teacher"}
+    validates :has_entire_term_contract, on: [:"entire-term-contract", :submit], inclusion: {in: [true, false], message: "Select yes if you have a contract to teach at the same school for one term or longer"}, if: :employed_as_supply_teacher?
```
pezholio: I think for these, when we're prompting for a user to select an option, I'd expect it to be in quotes (i.e. `Select "yes"`) or at least capitalized (`Select Yes`), but I can't find any guidance about this. The Miro boards seem to have the options capitalized.

titlescreen: Good question. I was going the design system, the guidance I could find gave this as an example `Select yes if Sarah normally lives with you`. So based on that and that we were pretty inconsistent in regards to capitalising the word yes, I changed them all to lower case.

There are some examples for radios at the bottom of this page. https://design-system.service.gov.uk/components/radios/

pezholio: Ah cool. I'm happy in that case. As long as we're consistent 👍

PR #742, opened 2020-01-02
Add aria-describedby to elements with hints

pezholio:
This adds an `aria-describedby` to elements that have hints. In some cases the hints describe a group of elements (such as, for example a bunch of radio buttons), in this case, the `aria-describedby` element is added to the enclosing `fieldset`, which will also have a `role` of `group` (which is a pattern I've seen elsewhere, but hasn't until now, been consistent)

### DIFF
--- a/app/views/claims/bank_details.html.erb
+++ b/app/views/claims/bank_details.html.erb
@@ -6,14 +6,14 @@
       <%= render("shared/error_summary", instance: current_claim) if current_claim.errors.any? %>
 
       <%= form_for current_claim, url: claim_path(current_policy_routing_name) do |form| %>
-        <fieldset class="govuk-fieldset">
+        <fieldset class="govuk-fieldset" aria-describedby="bank_details-hint" role="group">
           <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
             <h1 class="govuk-fieldset__heading">
               <%= t("questions.bank_details") %>
             </h1>
           </legend>
 
-          <span class="govuk-hint">
+          <span class="govuk-hint" id="bank_details-hint">
             The account you want us to send your payment to.
           </span>
 
--- a/app/views/student_loans/claims/leadership_position.html.erb
+++ b/app/views/student_loans/claims/leadership_position.html.erb
@@ -10,7 +10,7 @@
 
           <%= fields.hidden_field :had_leadership_position %>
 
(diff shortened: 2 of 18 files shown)


pezholio:
[Add `aria-describedby` to input elements with hints](https://trello.com/c/4vGvgVmZ/882-add-aria-describedby-to-input-elements-with-hints)

app/views/claims/bank_details.html.erb
```diff
@@ -6,14 +6,14 @@
...
-        <fieldset class="govuk-fieldset">
+        <fieldset class="govuk-fieldset" aria-describedby="bank_details-hint">
           <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
             <h1 class="govuk-fieldset__heading">
               <%= t("questions.bank_details") %>
             </h1>
           </legend>
 
-          <span class="govuk-hint">
+          <span class="govuk-hint" id="bank_details-hint">
```
titlescreen: Hey, I'm not sure this is needed. Screen readers will read the legend when users navigate to the fieldset. If the intention is to get screenreaders to say `Name on the account. The account you want us to send your payment to.` when a user clicks on the banking name field then I think we need to move the aria described by to the `banking-name` form.

app/views/student_loans/claims/leadership_position.html.erb
```diff
@@ -10,15 +10,15 @@
...
+          <fieldset class="govuk-fieldset" aria-describedby="had_leadership_position-hint" role="group">
 
             <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
               <h1 class="govuk-fieldset__heading">
                 <%= t("student_loans.questions.leadership_position") %>
               </h1>
             </legend>
 
-            <span class="govuk-hint">
+            <span class="govuk-hint" id="leadership_position-hint">
```
titlescreen: I think the id on this should be `had_leadership_position-hint`

app/views/claims/bank_details.html.erb
pezholio: I think my assumption here would be this hint would describe the whole group, rather than an individual field. I guess in this case, I'd need to add `role="group"` to the fieldset right?

titlescreen: Yeah now when you go to an input it will say e.g. “enter bank account detail. Name on the account“.

PR #744, opened 2020-01-02
Display the name of a user who created a payroll run

pezholio:
This updates the PayrollRun model to refer to a user in the database, rather than a simple Dfe Sign-In user ID. This also allows the name of a user to be shown like so:

image

### DIFF
--- a/spec/factories/payroll_runs.rb
+++ b/spec/factories/payroll_runs.rb
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :payroll_run do
-    created_by { "123" }
+    association :created_by, factory: :dfe_signin_user
 
     transient do
       claims_counts { {StudentLoans => 1} }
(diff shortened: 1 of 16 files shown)


pezholio:

spec/factories/payroll_runs.rb
```diff
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :payroll_run do
-    created_by { "123" }
+    created_by { build(:dfe_signin_user) }
```
robbiepaul: Should this be explicitly be defined as an association? (just looking at Tekin's comment here: https://github.com/DFE-Digital/dfe-teachers-payment-service/pull/731#discussion_r359816862)

pezholio: Good catch. Done! 👍

PR #747, opened 2020-01-06
Display the name of the user who downloaded a payroll report

pezholio:
This is the final part of https://trello.com/c/fr9sNCXk/877-display-the-name-of-the-service-operator-who-took-an-action, and this displays the name of a user who has downloaded a Payroll Report. As with #731 and #744, if a user's details have been imported, it shows their full name, but if we haven't yet hit the API to import some user data, it shows the user's DfE Sign-In ID:

image

image

There is also a data migration to import any existing user IDs that are set on the `downloaded_by` column and convert them into a proper reference to a user object.

### DIFF
--- a/app/helpers/dfe_sign_in/user_helper.rb
+++ b/app/helpers/dfe_sign_in/user_helper.rb
@@ -1,16 +1,18 @@
 module DfeSignIn
   module UserHelper
-    def user_details(user)
-      user.full_name.presence || unknown_user(user)
+    def user_details(user, include_line_break: true)
+      user.full_name.presence || unknown_user(user, include_line_break)
     end
 
     private
 
-    def unknown_user(user)
+    def unknown_user(user, include_line_break)
+      join_chr = include_line_break == true ? "<br/>" : " "
+
       [
         "Unknown user",
         unknown_user_details(user),
-      ].join("<br/>").html_safe
+      ].join(join_chr).html_safe
     end
 
     def unknown_user_details(user)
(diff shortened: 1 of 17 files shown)


pezholio:

app/views/admin/payroll_run_downloads/show.html.erb
```diff
@@ -13,7 +13,7 @@
       <p class="govuk-body">
         This month's payroll file was downloaded on
         <%= l(@payroll_run.downloaded_at) %> by
-        <%= @payroll_run.downloaded_by %>.
+        <%= user_details(@payroll_run.downloaded_by, include_line_break: false) %>.
```
jacksonj04: The option which is used here isn't added until the [next commit](https://github.com/DFE-Digital/dfe-teachers-payment-service/pull/747/commits/b6c9c562bb9f4743a3dad61d00a6702722e526e0) - are these out of order?

app/helpers/dfe_sign_in/user_helper.rb
```diff
@@ -1,16 +1,18 @@
 module DfeSignIn
   module UserHelper
-    def user_details(user)
-      user.full_name.presence || unknown_user(user)
+    def user_details(user, options = {})
```
jacksonj04: Would this be better as a single named parameter instead of an options hash?

pezholio: Good call, that actually makes cycling down to that private method work a bit better too

PR #748, opened 2020-01-06
Fix verify timeout bug

jacksonj04:
Fix a bug where users who spent more than 90 minutes in Verify would trigger a routing exception and not receive the session timeout message.


spec/features/govuk_verify_spec.rb
```diff
@@ -73,53 +73,38 @@
...
-  context "cancelled response" do
-    before do
-      stub_vsp_translate_response_request("no-authentication")
-    end
-
-    scenario "shows the no_authentication error", js: true do
-      click_on "Continue"
-      # Submit the form generated by our FakeSso
-      click_on "Perform identity check"
+  scenario "Cancelling the Verify process shows a process incomplete message to the user", js: true do
```
pezholio: 👍 for cleaning this up

PR #751, opened 2020-01-08
Handling the situation where a user tries to claim for both policies at once

pezholio:
When a user is in the middle of a claim, and tries to access the initial page of a new claim, they're now confronted with a warning telling them that they are in the middle of an application, and giving them the opportunity to start a new claim, or continue the claim they are currently making:

image

We've used the interruption card pattern, as outlined in the [Home Office Design System](https://home-office-digital-patterns.herokuapp.com/patterns/interruption-card), so this may need some looking over by @titlescreen before merge too.

### DIFF
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -3,6 +3,7 @@ class ClaimsController < BasePublicController
 
   skip_before_action :send_unstarted_claiments_to_the_start, only: [:new, :create, :timeout]
   before_action :check_page_is_in_sequence, only: [:show, :update]
+  before_action :check_claim_not_in_progress, only: [:new]
   before_action :clear_claim_session, only: [:new]
   before_action :prepend_view_path_for_policy
 
@@ -40,6 +41,9 @@ class ClaimsController < BasePublicController
   def timeout
   end
 
+  def existing_session
+  end
+
   private
 
   helper_method :next_slug
@@ -70,6 +74,14 @@ class ClaimsController < BasePublicController
 
   def check_page_is_in_sequence
     raise ActionController::RoutingError.new("Not Found") unless page_sequence.in_sequence?(params[:slug])
+  end
+
+  def check_claim_not_in_progress
+    redirect_to(existing_session_path(policy: params[:policy])) if claim_in_progress?
+  end
+
+  def claim_in_progress?
+    session[:claim_id].present? && !current_claim.eligibility.ineligible?
   end
 
   def page_sequence
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -3,6 +3,7 @@ class ClaimsController < BasePublicController
 
   skip_before_action :send_unstarted_claiments_to_the_start, only: [:new, :create, :timeout]
   before_action :check_page_is_in_sequence, only: [:show, :update]
+  before_action :update_session_with_current_slug, only: [:show]
   before_action :check_claim_not_in_progress, only: [:new]
   before_action :clear_claim_session, only: [:new]
   before_action :prepend_view_path_for_policy
@@ -74,6 +75,10 @@ class ClaimsController < BasePublicController
 
   def check_page_is_in_sequence
     raise ActionController::RoutingError.new("Not Found") unless page_sequence.in_sequence?(params[:slug])
+  end
+
+  def update_session_with_current_slug
+    session[:current_slug] = params[:slug]
   end
 
   def check_claim_not_in_progress
--- a/app/views/claims/existing_session.html.erb
+++ b/app/views/claims/existing_session.html.erb
@@ -1,0 +1,15 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <div class="govuk-panel govuk-panel--interruption">
+      <h1 class="govuk-heading-xl">
+        You have a claim in progress <%= policy_description(current_claim.policy.routing_name) %>.
+      </h1>
+
+      <h2 class="govuk-heading-m">
+        You will lose your progress on this claim if you start a different claim before you send it.
+      </h2>
+    </div>
+
+    <%= button_to "Start claim #{policy_description(params[:policy])}", session_path(policy: params[:policy]), method: :delete, class: "govuk-button govuk-button--secondary" %>
+  </div>
+</div>
--- a/app/views/claims/existing_session.html.erb
+++ b/app/views/claims/existing_session.html.erb
@@ -10,6 +10,10 @@
       </h2>
     </div>
 
+    <p class="govuk-body govuk-!-margin-top-7 govuk-!-margin-bottom-0">
+      <%= link_to "Finish claim in progress", claim_path(current_claim.policy.routing_name, slug: session[:current_slug]), class: "govuk-button" %>
+    </p>
+
     <%= button_to "Start claim #{policy_description(params[:policy])}", session_path(policy: params[:policy]), method: :delete, class: "govuk-button govuk-button--secondary" %>
   </div>
 </div>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -104,7 +104,7 @@ en:
   maths_and_physics:
     policy_name: "Claim a payment for teaching maths or physics"
     policy_short_name: "Maths and Physics"
-    claim_description: "claim for a payment for teaching maths or physics"
+    claim_description: "for a payment for teaching maths or physics"
     claim_amount_description: "Payment for teaching maths or physics"
     claim_action: "claim £2,000 for teaching maths or physics"
     award_description: "£2,000 payment"
@@ -157,7 +157,7 @@ en:
   student_loans:
     policy_name: "Teachers: claim back your student loan repayments"
     policy_short_name: "Student Loans"
-    claim_description: "claim to get back your student loan repayments"
+    claim_description: "to get back your student loan repayments"
     claim_amount_description: "Student loan repayments you’ve claimed back"
     claim_action: "claim back student loan repayments"
     award_description: "claim payment"
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -14,6 +14,7 @@ Rails.application.routes.draw do
   end
 
   get "refresh-session", to: "sessions#refresh", as: :refresh_session
+  delete "session", to: "sessions#destroy", as: :session
 
   # Used to constrain claim journey routing so only slugs
   # that are part of a policy’s slug sequence are routed.
@@ -45,6 +46,7 @@ Rails.application.routes.draw do
     get "claims/confirmation", as: :claim_confirmation, to: "submissions#show"
 
     get "timeout", to: "claims#timeout", as: :timeout_claim
+    get "existing-session", as: :existing_session, to: "claims#existing_session"
 
     %w[privacy_notice terms_conditions contact_us cookies accessibility_statement].each do |page_name|
       get page_name.dasherize, to: "static_pages##{page_name}", as: page_name
--- a/spec/features/ineligible_student_loans_claims_spec.rb
+++ b/spec/features/ineligible_student_loans_claims_spec.rb
@@ -88,6 +88,8 @@ RSpec.feature "Ineligible Teacher Student Loan Repayments claims" do
 
     visit new_claim_path(StudentLoans.routing_name)
 
+    expect(page).to_not have_content("You have a claim in progress")
+
     expect(page).to have_content("When did you complete your initial teacher training?")
     expect(page).not_to have_css("input[checked]")
     choose_qts_year
--- a/spec/features/switching_policies_spec.rb
+++ b/spec/features/switching_policies_spec.rb
@@ -14,4 +14,10 @@ RSpec.feature "Switching policies" do
 
     expect(page).to have_text(I18n.t("maths_and_physics.questions.teaching_maths_or_physics"))
   end
+
+  scenario "a user can choose to continue their claim" do
+    click_on "Finish claim in progress"
+
+    expect(page).to have_text(I18n.t("student_loans.questions.claim_school"))
+  end
 end
--- a/spec/features/switching_policies_spec.rb
+++ b/spec/features/switching_policies_spec.rb
@@ -1,17 +1,17 @@
 require "rails_helper"
 
-RSpec.feature "A user can switch policies" do
-  it "a user can switch to maths and physics after starting a student loan claim" do
+RSpec.feature "Switching policies" do
+  before do
     start_student_loans_claim
     visit new_claim_path(MathsAndPhysics.routing_name)
-
-    expect(page).to have_text(I18n.t("maths_and_physics.questions.teaching_maths_or_physics"))
   end
 
-  it "a user can switch to student loans after starting a maths and physics claim" do
-    start_maths_and_physics_claim
-    visit new_claim_path(StudentLoans.routing_name)
+  scenario "a user can switch to a different policy after starting a claim on another" do
+    expect(page.title).to have_text(I18n.t("maths_and_physics.policy_name"))
+    expect(page.find("header")).to have_text(I18n.t("maths_and_physics.policy_name"))
 
-    expect(page).to have_text(I18n.t("questions.qts_award_year"))
+    click_on "Start claim for a payment for teaching maths or physics"
+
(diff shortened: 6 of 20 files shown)


pezholio:
[Handling the situation where a user tries to claim for both policies at once...](https://trello.com/c/xLbbk0CH/1046-handling-the-situation-where-a-user-tries-to-claim-for-both-policies-at-once)

app/views/claims/existing_session.html.erb
```diff
@@ -0,0 +1,21 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <div class="govuk-panel govuk-panel--interruption">
+      <h1 class="govuk-heading-xl">
+        You have an application in progress for <%= policy_service_name(current_policy_routing_name) %>.
+      </h1>
+
+      <h2 class="govuk-heading-m">
```
titlescreen: I think we just need to add govuk-!-margin-bottom-9 to force it to have a decent space between the text and the button.

jacksonj04: localhost_3000_maths-and-physics_existing-session

Duly added

app/views/claims/existing_session.html.erb
```diff
@@ -0,0 +1,21 @@
...
+      <h1 class="govuk-heading-xl">
+        You have an application in progress for <%= policy_service_name(current_policy_routing_name) %>.
+      </h1>
+
+      <h2 class="govuk-heading-m govuk-!-margin-bottom-9">
+        If you start a new application then your progress will be lost.
+      </h2>
+
+      <p class="govuk-body">
+        <%= link_to "Continue application", claim_path(current_policy_routing_name, slug: session[:current_slug]), class: "govuk-button" %>
```
tekin: This will send the user to the last question that they answered, not the last one they saw.

tekin: Also, do we need to make the call to action clearer that "Continue application" means continue the one that is already in progress, and not continue with the application they have just clicked on to start?

spec/features/ineligible_student_loans_claims_spec.rb
```diff
@@ -88,6 +88,8 @@
 
     visit new_claim_path(StudentLoans.routing_name)
 
+    click_on "Start new application for"
```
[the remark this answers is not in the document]
pezholio: Yeah, I'm totally on board with that principle, it was just easier to show progress to have these commits seperated out

app/controllers/claims_controller.rb
```diff
@@ -3,9 +3,11 @@ class ClaimsController < BasePublicController
 
   skip_before_action :send_unstarted_claiments_to_the_start, only: [:new, :create, :timeout]
   before_action :check_page_is_in_sequence, only: [:show, :update]
+  before_action :update_session_with_current_slug, only: [:show]
   before_action :check_claim_not_in_progress, only: [:new]
   before_action :clear_claim_session, only: [:new]
   before_action :prepend_view_path_for_policy
+  helper_method :page_sequence
```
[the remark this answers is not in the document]
pezholio: Nowhere - was left over from some previous work we were trying I think. We don't need it anymore

app/views/claims/existing_session.html.erb
jacksonj04: "Continue current application" makes more sense, agreed. Do we need to ask a content designer for the exact wording ("Continue with in-progress application"?)

This doesn't look to be taking the user back to the last answered question, just running through it now it definitely takes me to the last seen question, and (I think) `spec/features/switching_policies_spec.rb:35` checks this behaviour (the test has selected the QTS year but not current school, and expects to see the current school question).

spec/features/ineligible_student_loans_claims_spec.rb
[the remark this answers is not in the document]
pezholio: I've completely rebased this branch out now, so there are just three commits. It's a lot clearer now I think.

spec/requests/switching_policy_spec.rb
```diff
@@ -1,21 +0,0 @@
-require "rails_helper"
-
-RSpec.describe "Switching Policies", type: :request do
```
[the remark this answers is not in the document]
pezholio: Yeah, I was just being lazy. Thanks for keeping me honest 😃

app/views/claims/existing_session.html.erb
pezholio: Yep, the `before_action` here stores the current slug: https://github.com/DFE-Digital/dfe-teachers-payment-service/blob/dffb5e01ddc95153c57e39a8fcb66bbce9932d88/app/controllers/claims_controller.rb#L6. I've rejigged the commits a bit now, so it should make things a bit clearer

app/controllers/claims_controller.rb
```diff
@@ -72,6 +73,10 @@ def check_page_is_in_sequence
     raise ActionController::RoutingError.new("Not Found") unless page_sequence.in_sequence?(params[:slug])
   end
 
+  def check_claim_not_in_progress
+    redirect_to(existing_session_path(policy: params[:policy])) && return if session[:claim_id].present?
```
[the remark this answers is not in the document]
pezholio: Yeah, the `&& return` was me misremembering about using `redirect_to` in a conditional. I don't think it needs to be there if there's no other logic in the method

spec/features/switching_policies_spec.rb
```diff
@@ -1,21 +1,41 @@
...
-    start_student_loans_claim
-    visit new_claim_path(MathsAndPhysics.routing_name)
+  context "after starting a student loan claim" do
+    before do
+      start_student_loans_claim
+      visit new_claim_path(MathsAndPhysics.routing_name)
+    end
 
-    click_on "Start new application"
+    it "a user can switch to maths and physics after starting a student loan claim" do
```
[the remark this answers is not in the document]
pezholio: Yeah, right you are, just one of those things I forgot in the white heat of pairing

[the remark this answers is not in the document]
lawrence-dxw: 🔥 😄

app/views/claims/existing_session.html.erb
tekin: RE the call to action button text and elsewhere, yes, Faye should definitely be walked through this flow and asked to provide feedback on the content.

tekin: > "Continue current application" makes more sense, agreed. Do we need to ask a content designer for the exact wording ("Continue with in-progress application"?)
> 
> This doesn't look to be taking the user back to the last answered question, just running through it now it definitely takes me to the last seen question, and (I think) `spec/features/switching_policies_spec.rb:35` checks this behaviour (the test has selected the QTS year but not current school, and expects to see the current school question).

Hmmm, I must have done something weird when testing this locally on my machine.

app/controllers/claims_controller.rb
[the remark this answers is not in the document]
jacksonj04: I think it would make sense for the title to match the user's intent when they hit the warning, which would be the policy in the params (ie url and title match, error message tells you what the problem is)

pezholio:
New copy has been added, as per https://miro.com/app/board/o9J_kvy6CH0=/?moveToWidget=3074457347139380513

image

image

app/views/claims/existing_session.html.erb
```diff
@@ -0,0 +1,21 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <div class="govuk-panel govuk-panel--interruption">
+      <h1 class="govuk-heading-xl">
+        You have an claim in progress <%= policy_description(current_policy_routing_name) %>.
```
tekin:
```suggestion
        You have a claim in progress <%= policy_description(current_policy_routing_name) %>.
```

lawrence-dxw:
@pezholio @jacksonj04 did you have any thoughts about this?

> Also, the current behaviour is that the warning is displayed even if the user navigates to the start page of the same policy for which they're currently claiming (either explicitly or by using the back button). Is this intentional? If so, I think it would be good to make it explicit with a test.

pezholio:
@lawrence-dxw It's covered in this test https://github.com/DFE-Digital/dfe-teachers-payment-service/blob/db5bfd925c1a606d45e24a6bb9beb65f1828d50e/spec/features/ineligible_student_loans_claims_spec.rb#L84

app/controllers/claims_controller.rb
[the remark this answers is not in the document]
lawrence-dxw: > I think it would make sense for the title to match the user's intent when they hit the warning, which would be the policy in the params (ie url and title match, error message tells you what the problem is)

Yeah, I think this would be the best.

config/locales/en.yml
```diff
@@ -104,7 +104,7 @@ en:
   maths_and_physics:
     policy_name: "Claim a payment for teaching maths or physics"
     policy_short_name: "Maths and Physics"
-    claim_description: "claim for a payment for teaching maths or physics"
+    claim_description: "for a payment for teaching maths or physics"
```
lawrence-dxw: This string is also used by `app/views/payment_mailer/confirmation_for_multiple_claims.text.erb`, so that will need updating.

spec/features/ineligible_student_loans_claims_spec.rb
```diff
@@ -88,6 +88,8 @@
 
     visit new_claim_path(StudentLoans.routing_name)
 
+    click_on "Start claim"
```
lawrence-dxw: I'm still trying to think about what would be the best user journey here. As it stands, when the user gets told they're ineligible, they can visit the start page for that policy again, where they will see the "claim already in progress" message. Clicking "Finish claim in progress" on that screen will take them to the ineligible screen again, so it's not a super-useful action. I wonder whether it would be better to not show the "claim already in progress" screen if the current claim is ineligible?

spec/features/switching_policies_spec.rb
```diff
@@ -1,17 +1,29 @@
...
+      click_on "Start claim"
+
+      expect(page).to have_text(I18n.t("maths_and_physics.questions.teaching_maths_or_physics"))
+    end
   end
 
-  it "a user can switch to student loans after starting a maths and physics claim" do
-    start_maths_and_physics_claim
-    visit new_claim_path(StudentLoans.routing_name)
+  context "after starting a maths and physics claim" do
```
lawrence-dxw: I wonder whether we need to test with two different initial policies. I think we could probably just test switching from, say, Student Loans to Maths and Physics.

spec/requests/claims_spec.rb
```diff
@@ -11,6 +11,7 @@
 
     it "redirects to the new claim confirmation page if a Student Loans claim is already in progress" do
       start_student_loans_claim
+      follow_redirect!
```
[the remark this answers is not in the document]
lawrence-dxw: Not sure if this comment is still valid.

config/routes.rb
```diff
@@ -45,6 +45,8 @@ def matches?(request)
     get "claims/confirmation", as: :claim_confirmation, to: "submissions#show"
 
     get "timeout", to: "claims#timeout", as: :timeout_claim
+    get "existing-session", as: :existing_session, to: "claims#existing_session"
+    get "clear-session-and-redirect", to: "claims#clear_session_and_redirect", as: :clear_session_and_redirect
```
lawrence-dxw: I thought I had asked this in my first review, but couldn't find it - how do you feel about using a `GET` for a destructive action? Might a `POST` be more appropriate?

spec/features/switching_policies_spec.rb
```diff
@@ -1,17 +1,29 @@
 require "rails_helper"
 
 RSpec.feature "A user can switch policies" do
-  it "a user can switch to maths and physics after starting a student loan claim" do
-    start_student_loans_claim
-    visit new_claim_path(MathsAndPhysics.routing_name)
+  context "after starting a student loan claim" do
```
lawrence-dxw:
```suggestion
  context "when trying to start a maths and physics claim, with a student loans claim already in progress" do
```

(and then `a user can switch to maths and physics` for the `scenario`)

lawrence-dxw:
Commits are looking way nicer, thanks!

spec/features/ineligible_student_loans_claims_spec.rb
pezholio: Yeah, good point. I didn't think about that particular case 👍

config/routes.rb
pezholio: Yeah, I wondered about using `DELETE` at first, but discounted it, so just switch to `GET`, but yeah, `POST` is a good shout

jacksonj04: @pezholio @lawrence-dxw Using `method: :post` looks to rely on JavaScript - is this something we need to run past accessibility?

pezholio: Hmmm - good shout. Maybe `button_to` would be a better choice, which would then generate a form

tekin: > Yeah, I wondered about using `DELETE` at first, but discounted it, so just switch to `GET`, but yeah, `POST` is a good shout

What is it about `DELETE` that made you discount it?

pezholio: It's not acting on a specific resource, so didn't really make sense.

jacksonj04: I'm just asking in the design system Slack as to if they have any prior art on styling buttons as links, since then `button_to` would solve it neatly.

`DELETE` means we'd be stuck relying on JavaScript as well, since we can't `DELETE` from a standard form.

tekin: It is acting on a singular resource, i.e. the current session.

Pretty sure Rails sets a `_method` param set to `DELETE` and under the hood sets up routing so that it will all work as you'd expect, even without JS.

jacksonj04: Some styling thoughts at https://github.com/alphagov/govuk-design-system-backlog/issues/159. One from Slack which I don't completely hate is the idea of moving all the buttons outside of the interruption dialog (although we should still really have them on the same line as per the [design system](https://design-system.service.gov.uk/components/button/).

localhost_3000_student-loans_existing-session

It looks like `_method` is some magic which is [added by JS at runtime](https://edgeguides.rubyonrails.org/working_with_javascript_in_rails.html#data-method) vs when the view is rendered. Disabling JS means that the link just performs a `GET` (and 404s).

`DELETE` would make sense if we reworked the URLs so that it was `DELETE`ing something like `student-loans/session`, but that still runs into the 'needs JS' problem.

tekin: Ah I meant was if we use a form, it will just work. Rails will automatically set a hidden input like this: 

```ruby
<input type="hidden" value="delete" name="_method">
```

which will route correctly without JS.

tekin: So suddenly we're reeaaaaally close to having an actual RESTful resource for the session... What would this code look like if the actions were on the existing sessions controller, rather than the claims controller?

pezholio: That's what I had originally, but it was becoming problematic as there were a whole bunch of helpers etc that I needed for the views, and it didn't seem right to inherit the sessions controller from the base controller

jacksonj04: I think it would make sense to move `#clear_session_and_redirect` to the sessions controller since that's the domain it's acting on. I'm less certain about `#existing-session`, since it's the _claim_ which is informing the user it cannot continue.

I am, however, not entirely sold on this view.

tekin: I think that makes sense. So move the action to clear the session to the sessions controller and keep the existing action on claims?

pezholio: That'll work. Then a `DELETE` makes much more sense

jacksonj04: I remember why this was a minor pain - we kept it in the claims flow so that:

1. It could use `BasePublicController#clear_claim_session`, and
2. We didn't have to pass the target policy into the controller (which is needed to fulfil the redirect), although I guess it's not too much faff to just sling a query parameter along like `?redirect_to_policy`

tekin: 1. Reckon `SessionsController` should probably be inheriting from `BasePublicController` anyway. It may or may not mean skipping some `before_action`s...
2. Yep, I reckon make it as much a RESTful `destroy` action as possible to lean on conventions. A query param for the redirect sounds like a good shout. I wondered if a more generic `redirect_to` param that accepts a path would make sense, but actually it's probably safer to expect a policy name and always redirect to the start of claim form as you suggest.

PR #753, opened 2020-01-08
Eligible schools in the West Somerset opportunity area should be able to apply

jacksonj04:
The policy document says that schools in the West Somerset opportunity
area are eligible, but this is defined as contiguous with the (now
defunct) West Somerset local authority district. When the local
authority districts of West Somerset and Taunton were merged this local
authority district ceased to exist, and the new local authority district
which schools were moved to is not eligible.

Schools in the `EXPLICITLY_ELIGIBLE_SCHOOL_URNS` list are now treated as
being in an eligible local authority district. This is a list of schools
which are in the West Somerset opportunity area and which are eligible.

### DIFF
--- a/app/models/maths_and_physics/school_eligibility.rb
+++ b/app/models/maths_and_physics/school_eligibility.rb
@@ -45,13 +45,19 @@ module MathsAndPhysics
       "E06000014", # York
     ].freeze
 
+    SCHOOL_URNS_CONSIDERED_AS_ELIGIBLE_LOCAL_AUTHORITY_DISTRICT = [
+      136774, # Minehead Middle School
+      136791, # West Somerset College
+      140631, # Danesfield Church of England Voluntary Controlled Community Middle School
+    ].freeze
+
     def initialize(school)
       @school = school
     end
 
     def eligible_current_school?
       @school.open? &&
-        eligible_local_authority_district? &&
+        (eligible_local_authority_district? || considered_as_eligible_local_authority_district?) &&
         (@school.state_funded? || @school.secure_unit?) &&
         @school.secondary_or_equivalent?
     end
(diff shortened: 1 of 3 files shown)


app/models/maths_and_physics/school_eligibility.rb
```diff
@@ -42,10 +42,15 @@ class SchoolEligibility
       "E06000021", # Stoke-on-Trent
       "E08000024", # Sunderland
       "E08000036", # Wakefield
-      "E07000191", # West Somerset
       "E06000014", # York
     ].freeze
 
+    EXPLICITLY_ELIGIBLE_SCHOOL_URNS = [
+      136774, # Minehead Middle School
```
pezholio: Good call with the commenting to make it clear what refers to what 👍

app/models/maths_and_physics/school_eligibility.rb
```diff
@@ -60,7 +65,8 @@ def eligible_current_school?
     private
 
     def eligible_local_authority_district?
-      ELIGIBLE_LOCAL_AUTHORITY_DISTRICT_CODES.include?(@school.local_authority_district.code)
+      EXPLICITLY_ELIGIBLE_SCHOOL_URNS.include?(@school.urn) ||
```
pezholio: Would it be better for this to be in a method of its own, like `explicity_eligible_school?`. Just avoids polluting this method with edge cases. You could then change the `eligible_current_school?` method to look a bit like this:

```ruby
def eligible_current_school?
  @school.open? &&
    (eligible_local_authority_district? || explicity_eligible_school?) &&
    (@school.state_funded? || @school.secure_unit?) &&
    @school.secondary_or_equivalent?
end
```

I know it complicates that method a little bit more, but it makes the intention clearer, for me anyhoo

lawrence-dxw: Yeah, I think it's a good idea. I'll change it.

jacksonj04: Another quick thought, it might be worth renaming this to something like `school_considered_in_eligible_district`, since we're not actually overriding any other eligibility tests.

lawrence-dxw: Just pushed a change, what do you think?

jacksonj04: Looks sane to me.

PR #754, opened 2020-01-08
Claims factory doesn't create potential duplicates

tekin:
We use the factories to seed our local development databases so we want to make sure they don't create claims that look the same and get flagged by our duplicate claim matching code. Reseting the sequences between each spec ensures they don't get so large that we end up generating
invalid TRNs or NI numbers.


tekin:
Pushed up some changes: one flakey spec fix and one spec that needed updating now that claims don't match by default.

PR #756, opened 2020-01-09
Service operators can approve GOV.UK Verify-skipped claims

tekin:
This re-enables the ability to approve claims that did not complete the GOV.UK Verify process.

This is different to how the story was initially outlined. For now we are keeping this super simple and expecting the identity check process to be carried out and managed as part of the existing spreadsheet-based claim checking. Unverified claims can be approved or rejected as before, but we now provide additional information to service operators for these claims:

**Highlight unverified claims on the claim list**
Screenshot 2020-01-09 at 10 52 07

**Unverified claims include an info panel relating to the manual identity check**
Screenshot 2020-01-09 at 11 00 36

**An additional warning is shown in the claim approve/reject form for unverified claims**
Screenshot 2020-01-09 at 10 52 24

This also updates the `School` model and the import code to save school phone numbers so that we can show that to service operators on the info panel. Phone numbers will get imported overnight on the next school import.

### DIFF
--- a/app/helpers/admin/claims_helper.rb
+++ b/app/helpers/admin/claims_helper.rb
@@ -53,6 +53,10 @@ module Admin
       content_tag(:strong, pluralize(days_until_check_deadline, "day"), class: "govuk-tag #{check_deadline_warning_class}")
     end
 
+    def id_verification_status(claim)
+      claim.identity_confirmed? ? "GOV.UK Verify" : content_tag(:strong, "Unverified", class: "govuk-tag tag--warning")
+    end
+
     def matching_attributes(first_claim, second_claim)
       first_attributes = first_claim.attributes.slice(*Claim::MatchingAttributeFinder::ATTRIBUTE_GROUPS_TO_MATCH.flatten).to_a
       second_attributes = second_claim.attributes.slice(*Claim::MatchingAttributeFinder::ATTRIBUTE_GROUPS_TO_MATCH.flatten).to_a
--- a/app/helpers/admin/claims_helper.rb
+++ b/app/helpers/admin/claims_helper.rb
@@ -2,6 +2,10 @@ module Admin
   module ClaimsHelper
     include StudentLoans::PresenterMethods
 
+    def confirming_identity_playbook_url
+      "https://docs.google.com/document/d/1wZh68_RV_FTJLxXIDPr3XFtJHW3vRgiXGaBDUo1Q1ZU"
+    end
+
     def admin_eligibility_answers(claim)
       claim.policy::EligibilityAdminAnswersPresenter.new(claim.eligibility).answers
     end
--- a/app/models/school_data_importer.rb
+++ b/app/models/school_data_importer.rb
@@ -44,6 +44,7 @@ class SchoolDataImporter
     school.close_date = row.fetch("CloseDate")
     school.establishment_number = row.fetch("EstablishmentNumber")
     school.statutory_high_age = row.fetch("StatutoryHighAge")
+    school.phone_number = row.fetch("TelephoneNum")
     school
   end
 end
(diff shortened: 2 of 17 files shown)


app/models/school_data_importer.rb
```diff
@@ -44,6 +44,7 @@ def row_to_school(row)
     school.close_date = row.fetch("CloseDate")
     school.establishment_number = row.fetch("EstablishmentNumber")
     school.statutory_high_age = row.fetch("StatutoryHighAge")
+    school.phone_number = row.fetch("TelephoneNum")
```
pezholio: This is an excellent shout 👍

app/helpers/admin/claims_helper.rb
```diff
@@ -2,6 +2,10 @@ module Admin
   module ClaimsHelper
     include StudentLoans::PresenterMethods
 
+    def confirming_identity_playbook_url
+      "https://docs.google.com/document/d/1wZh68_RV_FTJLxXIDPr3XFtJHW3vRgiXGaBDUo1Q1ZU"
```
pezholio: Not relevant at all to the code, and not a blocker, but should this be in Confluence, rather than as a Google Doc? Just concious of the risk of documentation being all over the place.

app/views/admin/claims/_info_panel_identity_unconfirmed.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  </h2>
+
+  <p class="govuk-body">
+    Refer to the <%= link_to "Confirming a claimant's identity playbook", confirming_identity_playbook_url, class: "govuk-link" %>
+    and contact <%= school.name %> on <strong><%= school.phone_number %></strong>.
+  </p>
+
+  <p class="govuk-body">
+    Once you have completed all the steps outlined in the playbook, you can
+    <%= link_to "approve or reject this claim", "#new_check", class: "govuk-link" %>.
```
pezholio: I know this `new_check` ID is autogenerated by Rails, but I'm wondering if it might be clearer if we link to an ID we define ourselves, to guard against the risk of something inside Rails changing and the form IDs being generated differently

spec/features/admin_claim_check_spec.rb
```diff
@@ -93,12 +93,17 @@
...
       let!(:claim_without_identity_confirmation) { create(:claim, :unverified) }
 
-      scenario "the service operator should be told the identity hasn't been confirmed" do
+      scenario "the service operator is told the identity hasn't been confirmed and can approve the claim" do
         click_on "View claims"
         find("a[href='#{admin_claim_path(claim_without_identity_confirmation)}']").click
 
-        expect(page).to have_content("Identity has not been confirmed")
-        expect(page).to have_field("Approve", disabled: true)
+        expect(page).to have_content("The claimant did not complete GOV.UK Verify")
```
pezholio: Is it worth checking for the presence of the correct school phone number here too?

app/helpers/admin/claims_helper.rb
tekin: That's where the service operations team have put this and other docs. If we think they should be storing those elsewhere then we should tackle that separately and with their OK. I'm not even sure they have Confluence access tbh...

app/views/admin/claims/_info_panel_identity_unconfirmed.html.erb
tekin: Yep, good shout, better to be explicit, especially as [`form_with`](https://apidock.com/rails/ActionView/Helpers/FormHelper/form_with) doesn't magically set an `id` attribute.

app/helpers/admin/claims_helper.rb
pezholio: Aaaah, if they don't even have Confluence access, then it's probably fine to have this stuff in Google docs.

spec/features/admin_claim_check_spec.rb
tekin: 👍

tekin:
@pezholio should have addressed your comments now.

PR #757, opened 2020-01-09
Group multiple claims from the same person into single payment

robbiepaul:
This currently uses National Insurance number to identify a claimant.

Claims from the same claimant are grouped in to one payment in the payroll run. If any of the personal details on the claim don't match the payroll run will fail and will need some intervention from a developer. In the future we plan to catch these discrepancies earlier in the process so the user should never see the error when creating payroll run. The error message doesn't give much detail as to what claims don't match, however as it is just a stop-gap solution it should be ok.

**Open for discussion: We need to decide which of the personal details we care about not matching, currently we check them all but there may be some that legitimately differ like address.**

# Screenshot
Screenshot 2020-01-09 at 11 35 12

### DIFF
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -7,9 +7,26 @@ class Payment < ApplicationRecord
   validates :payroll_reference, :gross_value, :national_insurance, :employers_national_insurance, :tax, :net_pay, :gross_pay, presence: true, on: :upload
   validates :gross_value, :national_insurance, :employers_national_insurance, :student_loan_repayment, :tax, :net_pay, :gross_pay, numericality: true, allow_nil: true
 
-  delegate :first_name, :middle_name, :surname, :national_insurance_number, :payroll_gender, :date_of_birth, :email_address, :address_line_1, :address_line_2, :address_line_3, :address_line_4, :postcode, :has_student_loan, :student_loan_plan, :banking_name, :bank_sort_code, :bank_account_number, :building_society_roll_number, to: :claim_for_personal_details
+  validate :personal_details_must_be_consistent
+
+  PERSONAL_DETAILS_ATTRIBUTES_PERMITTING_DISCREPANCIES = %i[first_name middle_name surname national_insurance_number payroll_gender email_address address_line_1 address_line_2 address_line_3 address_line_4 postcode has_student_loan banking_name]
+  PERSONAL_DETAILS_ATTRIBUTES_FORBIDDING_DISCREPANCIES = %i[national_insurance_number date_of_birth student_loan_plan bank_sort_code bank_account_number building_society_roll_number]
+
+  delegate(*(PERSONAL_DETAILS_ATTRIBUTES_PERMITTING_DISCREPANCIES + PERSONAL_DETAILS_ATTRIBUTES_FORBIDDING_DISCREPANCIES), to: :claim_for_personal_details)
 
   private
+
+  def personal_details_must_be_consistent
+    mismatching_attributes = PERSONAL_DETAILS_ATTRIBUTES_FORBIDDING_DISCREPANCIES.select { |attribute|
+      claims.map(&attribute).uniq.count > 1
+    }
+
+    if mismatching_attributes.any?
+      claims_sentence = claims.map(&:reference).to_sentence
+      attributes_sentence = mismatching_attributes.map { |attribute| Claim.human_attribute_name(attribute).downcase }.to_sentence
+      errors.add(:claims, "#{claims_sentence} have different values for #{attributes_sentence}")
+    end
+  end
 
   def claim_for_personal_details
     @claim_for_personal_details ||= claims.first
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -29,6 +29,6 @@ class Payment < ApplicationRecord
   end
 
   def claim_for_personal_details
-    @claim_for_personal_details ||= claims.first
+    @claim_for_personal_details ||= claims.order(:submitted_at).last
   end
 end
--- a/spec/factories/payments.rb
+++ b/spec/factories/payments.rb
@@ -5,8 +5,15 @@ FactoryBot.define do
     end
 
     claims do
+      personal_details = {
+        national_insurance_number: generate(:national_insurance_number),
+        teacher_reference_number: generate(:teacher_reference_number),
+        email_address: "email@example.com",
+        bank_sort_code: "220011",
+        bank_account_number: "12345678",
+      }
       claim_policies.map do |policy|
-        association(:claim, :approved, policy: policy)
+        association(:claim, :approved, personal_details.merge(policy: policy))
       end
     end
     association(:payroll_run, factory: :payroll_run)
--- a/spec/models/payment_spec.rb
+++ b/spec/models/payment_spec.rb
@@ -44,4 +44,99 @@ RSpec.describe Payment do
       end
     end
   end
+
+  context "when the payment is for more than one claim" do
+    subject { build(:payment, claims: claims) }
+    let(:claims) do
+      build_list(:claim, 2, :approved,
+        national_insurance_number: "JM603818B",
+        teacher_reference_number: "1234567",
+        email_address: "email@example.com",
+        bank_sort_code: "112233",
+        bank_account_number: "95928482",
+        building_society_roll_number: nil,
+        address_line_1: "64 West Lane",
+        student_loan_plan: StudentLoan::PLAN_1,
+        date_of_birth: 30.years.ago.to_date,
+        first_name: "Jennifer",
+        middle_name: "Poppy",
+        surname: "Blake",
+        payroll_gender: :female)
+    end
+
+    it "is valid when all claims have matching personal details" do
+      expect(subject).to be_valid
+    end
+
+    it "is invalid when claims’ National Insurance numbers do not match" do
+      claims[0].national_insurance_number = "JM102019D"
+
+      expect(subject).not_to be_valid
+      expect(subject.errors[:claims]).to eq(["#{claims[0].reference} and #{claims[1].reference} have different values for national insurance number"])
+    end
+
+    it "is invalid when claims’ dates of birth do not match" do
+      claims[0].date_of_birth = 20.years.ago.to_date
+
+      expect(subject).not_to be_valid
+      expect(subject.errors[:claims]).to eq(["#{claims[0].reference} and #{claims[1].reference} have different values for date of birth"])
+    end
+
+    it "is invalid when claims’ bank account numbers do not match" do
+      claims[0].bank_account_number = "34192192"
+
+      expect(subject).not_to be_valid
+      expect(subject.errors[:claims]).to eq(["#{claims[0].reference} and #{claims[1].reference} have different values for bank account number"])
+    end
+
+    it "is invalid when claims’ bank sort codes do not match" do
+      claims[0].bank_sort_code = "024828"
+
+      expect(subject).not_to be_valid
+      expect(subject.errors[:claims]).to eq(["#{claims[0].reference} and #{claims[1].reference} have different values for bank sort code"])
+    end
+
+    it "is invalid when claims’ building society roll numbers do not match" do
+      claims[0].building_society_roll_number = "123456789/ABCD"
+
+      expect(subject).not_to be_valid
+      expect(subject.errors[:claims]).to eq(["#{claims[0].reference} and #{claims[1].reference} have different values for building society roll number"])
+    end
+
+    it "is invalid when claims’ student loan plans do not match" do
+      claims[0].student_loan_plan = StudentLoan::PLAN_2
+
+      expect(subject).not_to be_valid
+      expect(subject.errors[:claims]).to eq(["#{claims[0].reference} and #{claims[1].reference} have different values for student loan plan"])
+    end
+
+    it "is invalid when its claims have multiple forbidden discrepancies in personal details" do
+      claims[0].bank_account_number = "34192192"
+      claims[0].bank_sort_code = "013028"
+
+      expect(subject).not_to be_valid
+      expect(subject.errors[:claims]).to eq(["#{claims[0].reference} and #{claims[1].reference} have different values for bank sort code and bank account number"])
+    end
+
+    it "remains valid when claims’ names do not match" do
+      claims[0].first_name = "David"
+      claims[0].middle_name = "Michael"
+      claims[0].surname = "Rollins"
+      claims[0].banking_name = "MR DM ROLLINS"
+
+      expect(subject).to be_valid
+    end
+
+    it "remains valid when claims’ payroll gender do not match" do
+      claims[0].payroll_gender = :male
+
+      expect(subject).to be_valid
+    end
+
+    it "remains valid when claims’ addresses do not match" do
+      claims[0].address_line_1 = "129 Brookland Drive"
+
+      expect(subject).to be_valid
+    end
+  end
 end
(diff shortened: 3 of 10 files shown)


spec/features/admin_payroll_runs_spec.rb
```diff
@@ -9,9 +9,9 @@
 
     click_on "Payroll"
 
-    create(:claim, :approved, policy: MathsAndPhysics)
-    create(:claim, :approved, policy: StudentLoans)
-    create(:claim, :approved, policy: StudentLoans)
+    create(:claim, :approved, policy: MathsAndPhysics, national_insurance_number: generate(:national_insurance_number))
```
tekin: Is there a reason why `national_insurance_number` is being set explicitly here? Normally I see an attribute explicitly set when the value is important to the substance of the spec, but here a) we're setting it to be the same as what the factory would give us anyway, and b) nothing specific in this spec references anything to do with the NI number.

spec/models/payroll_run_spec.rb
```diff
@@ -43,16 +43,36 @@
   end
 
   describe ".create_with_claims!" do
-    let(:claims) { Policies.all.map { |policy| create(:claim, :approved, policy: policy) } }
+    let(:claims) do
+      Policies.all.map do |policy|
+        create(:claim, :approved, policy: policy, national_insurance_number: generate(:national_insurance_number))
```
tekin: as above, what is the thinking behind being explicit here with the NI number?

spec/models/payroll_run_spec.rb
```diff
@@ -43,16 +43,36 @@
...
+        [
+          create(:claim, :approved, policy: StudentLoans, national_insurance_number: national_insurance_number),
+          create(:claim, :approved, policy: MathsAndPhysics, national_insurance_number: national_insurance_number),
+        ]
+      end
+      let(:other_claim) { create(:claim, :approved) }
+      let(:claims) { matching_claims + [other_claim] }
+
+      it "groups them into a single payment" do
+        expect(matching_claims[0].reload.payment.claims).to match_array(matching_claims)
```
tekin: I think this spec would be more confidence-inducing if it checked `payroll_run.payments`, that there were two, and that one was against the two matching claims and one was the other claim.

spec/requests/admin_payroll_runs_spec.rb
```diff
@@ -21,7 +21,7 @@
 
     describe "admin_payroll_runs#create" do
       it "creates a payroll run with payments and redirects to it" do
-        claims = create_list(:claim, 2, :approved)
+        claims = Array.new(2) { create(:claim, :approved, national_insurance_number: generate(:national_insurance_number)) }
```
tekin: As above re: explicitness on ni number, what do we gain?

spec/factories/payments.rb
```diff
@@ -5,8 +5,14 @@
     end
 
     claims do
+      claim_attributes = {
```
tekin: So we're setting these explicitly so that we can ensure the claims share the same personal details. Is there a better, more expressive name for this variable to help the reader understand that?

spec/models/payment_spec.rb
```diff
@@ -44,4 +44,46 @@
       end
     end
   end
+
+  describe "validations of claims’ personal details" do
```
tekin: Can we make this better express the business logic, rather than the code's behaviour? Say something like:

```suggestion
  describe "when the payment is for more than one claim" do
```

spec/models/payment_spec.rb
```diff
@@ -44,4 +44,46 @@
...
+        national_insurance_number: "JM603818B",
+        teacher_reference_number: "1234567",
+        email_address: "email@example.com",
+        bank_sort_code: "112233",
+        bank_account_number: "95928482",
+        address_line_1: "64 West Lane",
+        student_loan_plan: StudentLoan::PLAN_1)
+    end
+
+    it "is valid when claims’ personal details match" do
```
tekin:
```suggestion
    it "is valid when all claims have matching personal details" do
```

spec/models/payment_spec.rb
```diff
@@ -44,4 +44,46 @@
...
+    end
+
+    it "is invalid when claims’ National Insurance numbers do not match" do
+      claims[0].national_insurance_number = "JM102019D"
+
+      expect(subject).not_to be_valid
+    end
+
+    it "is invalid when claims’ bank details do not match" do
+      claims[0].bank_account_number = "34192192"
```
tekin: I feel like we're missing a bunch of coverage here for sort code, role number, etc.

spec/models/payroll_run_spec.rb
```diff
@@ -43,16 +43,46 @@
...
     it "creates a payroll run with payments and populates the award_amount" do
-      payroll_run = PayrollRun.create_with_claims!(claims, created_by: user)
-
       expect(payroll_run.reload.created_by.id).to eq(user.id)
       expect(payroll_run.claims).to match_array(claims)
       expect(claims[0].payment.award_amount).to eq(claims[0].award_amount)
       expect(claims[1].payment.award_amount).to eq(claims[1].award_amount)
     end
+
+    context "with multiple claims from the same National Insurance number" do
```
tekin:
```suggestion
    context "with multiple claims with the same National Insurance number" do
```

spec/requests/admin_payroll_runs_spec.rb
```diff
@@ -32,6 +32,20 @@
 
         expect(response).to redirect_to(admin_payroll_run_path(payroll_run))
       end
+
+      context "when a payroll run contains claims from the same person with different details" do
```
tekin:
```suggestion
      context "when a payroll run contains claims from the same person with different personal details" do
```

spec/models/payroll_run_spec.rb
lawrence-dxw: :+1:

spec/factories/payments.rb
lawrence-dxw: :+1: changed it to `personal_details`.

spec/models/payment_spec.rb
lawrence-dxw: Changed it to a `context` plus your suggested text.

spec/models/payment_spec.rb
lawrence-dxw: :+1:

spec/requests/admin_payroll_runs_spec.rb
lawrence-dxw: :+1:

spec/models/payroll_run_spec.rb
lawrence-dxw: :+1:

spec/models/payment_spec.rb
lawrence-dxw: I'd thought it was perhaps enough to just have some specs for a few representative examples of attributes. I've added those two in now.

spec/features/admin_payroll_runs_spec.rb
lawrence-dxw: So, my thinking was something like this. When we changed the claims factory so that instead of using a constant National Insurance number it generated a new one each time, I worried about how we could assess the impact of this change on all of our existing tests. In the sense of, how could we be sure that the change wasn't affecting the setup of those tests in a way that means that they are no longer testing what their author intended them to test?

So I then wondered a bit about how to write tests that are resilient against changes in the factory, and decided that it would be best, in the places where we want to test what happens when the National Insurance number is either the same or different, to be explicit about this in the test setup.

I'd like to understand how people usually handle this problem of changing factories / fixtures.

lawrence-dxw: Just realised something that you can't see from this PR:

> When we changed the claims factory so that instead of using a constant National Insurance number it generated a new one each time

We originally made this change as part of our work on this PR, and then coincidentally you added it in dc6868a too. So the explicit NI numbers originally accompanied our changes to the factory. I guess they look a bit weird in isolation, now I look at it.

tekin: I think you are spot on in that these specs should ultimately be resilient as much as possible to changes made to the factories. That's why it's important when setting up a record in a spec to be explicit with setting values of attributes that are important to the meat and potatoes of the spec itself.

In this case though, I think we can be pretty confident that the change to the factory was a refinement that made the factory more representative of reality. i.e. that a claim coming in will 99% of the time have a unique NI number, that should be default, and that if you want two claims with the same NI number, _that's when you should be explicit_. I think being explicit with setting a random NI number will mostly be noise, unless we have a strong feeling that the factory will change in the future to make NI number the same by default.

lawrence-dxw: That makes a lot of sense, thanks. I'll change it to get rid of the explicit random NI numbers.

lawrence-dxw:
> I think it will be much more useful for the service operator if the error message when there are discrepancies includes the references of the offending claims, rather than list the fields that don't match. That way they can easily find the claims and decide what needs doing.

Agreed. I've made this change.

> I do wonder if we should reduce the number of things that need to match though. Address definitely sounds like one that could legitimately come back different from Verify in the same month, which would be something that the user might find difficult change/fix.

It's a tricky one. The truth is that _most_ of the non-bank fields listed there could legitimately change between two claims:

- first / last / middle name
- payroll gender
- email address
- address lines / postcode
- `has_student_loan` (once they stop paying it off)

And if we decide we just want to pick one, for these values, I wonder whether instead of picking the value from the first (i.e. earliest) claim we should actually be picking the most up-to-date value, i.e. that from the latest claim. Do you have any thoughts on this?

tekin:
> > I think it will be much more useful for the service operator if the error message when there are discrepancies includes the references of the offending claims, rather than list the fields that don't match. That way they can easily find the claims and decide what needs doing.
> 
> Agreed. I've made this change.
> 
> > I do wonder if we should reduce the number of things that need to match though. Address definitely sounds like one that could legitimately come back different from Verify in the same month, which would be something that the user might find difficult change/fix.
> 
> It's a tricky one. The truth is that _most_ of the non-bank fields listed there could legitimately change between two claims:
> 
> * first / last / middle name
> * payroll gender
> * email address
> * address lines / postcode
> * `has_student_loan` (once they stop paying it off)
> 
> And if we decide we just want to pick one, for these values, I wonder whether instead of picking the value from the first (i.e. earliest) claim we should actually be picking the most up-to-date value, i.e. that from the latest claim. Do you have any thoughts on this?

I guess my broader point was more: does it actually matter if some of these fields don't match/change between the claims? i.e. would it make a material difference if we used one or the other? Some thoughts on individual fields:

- name: probably not? Whatever version of the name we use will get sent to HMRC for them to match against the tax record and they will already be set up to handle individual discrepancies in an indivuals name
- payroll gender: as above, I reckon HMRC will just deal with it
- email address: probably no impact. We do send this to Cantium, but I have a feeling we don't actually need to. Their system is set up to give the payroll individual their own sign-in to access payslips, and this is why they originally asked for email address, but we asked them to effectively disable that part of their service and not send an activation link to teachers.
- address: as for HMRC, I don't believe it will make a difference.
- `has_student_loan`: this one will have a material impact as it will impact how much student loan repayment is taken from their payout

And secondarily my point was: if we block a claim from being approved because the address has changed, how does that get unblocked? In theory we plan to reject the claim and ask the user to claim again, but if the address came back from Verify, then the user may have no way to re-claim with a different address.

spec/factories/payments.rb
```diff
@@ -5,8 +5,14 @@
     end
 
     claims do
+      personal_details = {
+        national_insurance_number: generate(:national_insurance_number),
```
tekin: although we don't technically check it for matching claims, I wonder if we should also set an explicit `teacher_reference_number` here as otherwise the two claims against the payment will have different ones, which feels off.

spec/models/payroll/payment_csv_row_spec.rb
```diff
@@ -7,21 +7,25 @@
     let(:row) { CSV.parse(subject.to_s).first }
     let(:payment_award_amount) { BigDecimal("1234.56") }
     let(:payment) { create(:payment, award_amount: payment_award_amount, claims: claims) }
+    let(:personal_details) do
+      {
+        national_insurance_number: generate(:national_insurance_number),
```
tekin: if we do it in the `payments factory, we may consider also being explicit with the TRN here too.

spec/factories/payments.rb
lawrence-dxw: :+1:

spec/models/payroll/payment_csv_row_spec.rb
lawrence-dxw: I feel that putting TRN here might be misleading, as it could imply that the TRN needs to be consistent when in fact the TRN is neither validated for consistency nor used in the class that is being tested here. What do you think?

tekin: We may not explicitly check/validate it in code, but does it better represent what the claims would look like in reality?

lawrence-dxw: > does it better represent what the claims would look like in reality?

Definitely, but I suppose that I didn't consider this to always be something worth aiming for in test data. i.e. that sometimes the only requirement of the test data for a unit test is that it be realistic enough for our class to be able to do its job. I'm happy to add it in, though.

lawrence-dxw:
I've changed it so that:

- we allow mismatching values for some attributes (choice of attributes explained in commit message);
- added an extra commit to take the value from the latest claim instead of the first.

app/models/payment.rb
```diff
@@ -7,11 +7,28 @@ class Payment < ApplicationRecord
...
+    if mismatching_attributes.any?
+      claims_sentence = claims.map(&:reference).to_sentence
+      attributes_sentence = mismatching_attributes.map { |attribute| Claim.human_attribute_name(attribute).downcase }.to_sentence
+      errors.add(:claims, "#{claims_sentence} have different values for #{attributes_sentence}")
+    end
+  end
+
   def claim_for_personal_details
-    @claim_for_personal_details ||= claims.first
+    @claim_for_personal_details ||= claims.max_by(&:created_at)
```
tekin: We can lean on ActiveRecord for this:

```suggestion
    @claim_for_personal_details ||= claims.last
```

lawrence-dxw: That was what I originally wrote, but it didn't work when I set up the payment like this in the tests:

```ruby
let(:claims) do
  [
    build(:claim, :approved, first_name: "Margaret", address_line_1: "17 Green Road", payroll_gender: :female, created_at: 5.days.ago),
    build(:claim, :approved, first_name: "John", address_line_1: "64 West Lane", payroll_gender: :male, created_at: 10.days.ago),
  ]
end
```

It returns the claim from `10.days.ago`, which isn't what I want.

tekin: So the reason using `last` will fail in that spec is because the records aren’t persisted – they’re all new records, so last is falling back to returning the last in the array. Do you think the method will ever be called against unpersisted claims? Also we should probably be returning the most recently submitted claim, i.e. check `submitted_at`.

lawrence-dxw: These things about what does and doesn't work for an unsaved relation continue to catch me out! It isn't being used with unpersisted claims at the moment, no. I can't really speculate about how it might be used in the future, though. I'll change it to use `submitted_at` and use persisted records in the specs.

lawrence-dxw: I've made that change. Had to change a few other specs to persist their records too, for them to continue working. Does it look alright to you?

tekin: Yep, working with the persisted records makes sense, new changes look like they cover reality better than before 👍🏽

PR #760, opened 2020-01-09
Reduce the chance of user submitting Verify SAML form twice

lawrence-dxw:
As we have observed in the past (e.g. commit 0f225f8), submitting the
same SAML request to Verify twice causes Verify to show a "Sorry,
something went wrong" error screen.

It was possible to submit the request twice by clicking the "Continue"
button on the `/verify/authentications/new` page after the JavaScript
auto-submit had been triggered. (I was able to reproduce this in a
browser, in our development environment, although not consistently.)

To prevent this, we take advantage of Rails' built-in prevention of
double-clicks (the `data-disable-with` attribute, which is set by
default), which disables the submit button once it is clicked. We change
the JavaScript so that it clicks the submit button instead of submitting
the form element.

We also change the page message so that once button is auto-clicked, it
no longer tells the user to click the "Continue" button, which is now
disabled.

I'm not sure if there is anything we can do to prevent a double
submission in the non-JS case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ The format is based on [Keep a Changelog]
 - Bank account numbers must be exactly 8 digits long (6- and 7-digit numbers are
   no longer accepted)
 - Multiple approved claims from the same person are grouped in to one payment
+- Fix "Sorry, something went wrong" message displayed by GOV.UK Verify when
+  claimant clicks "Continue" button on /verify/authentications/new page
 
 ## [Release 043] - 2020-01-08
 
--- a/app/assets/javascripts/components/govuk_verify_auth_request_auto_submit.js
+++ b/app/assets/javascripts/components/govuk_verify_auth_request_auto_submit.js
@@ -7,5 +7,14 @@ document.addEventListener("DOMContentLoaded", function() {
     return;
   }
 
+  var submit_button = form.querySelector("#submit_verify_request");
+  submit_button.style.display = "none";
+
+  var heading = form.querySelector("#verify_heading");
+  heading.innerText = "Continuing to next step…";
+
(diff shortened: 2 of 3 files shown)


pezholio:
I'm wondering if a better way to do this would be to hide the form altogether is JS is no disabled, in a similar way to the Verify docs - https://www.docs.verify.service.gov.uk/get-started/set-up-successful-verification-journey/#send-the-authentication-request-to-the-testing-service

tekin:
> I'm wondering if a better way to do this would be to hide the form altogether is JS is no disabled, in a similar way to the Verify docs - https://www.docs.verify.service.gov.uk/get-started/set-up-successful-verification-journey/#send-the-authentication-request-to-the-testing-service

Note that the way that is implement in the example wouldn't be possible in our app because we employ a strict content policy that prevents inline JavaScript from executing.

lawrence-dxw:
I've taken a look at a few Verify-using services and they seem hide the form submit button as you suggested, @pezholio. Some of them use a very basic page for the Verify form (e.g. the HMRC services just display a blank page if JavaScript is enabled, with "Processing..." in the page's title), some have a more fully-fledged page as we have (but still with the Continue button hidden). Some of them are also explicit about _why_ the user is seeing the Continue button (because JavaScript is disabled).

I don't understand why the Verify docs example re-displays the form after 5 seconds, though. I'm guessing it's in case the network is slow or something, but I think it's better to just have the message telling the user to refresh the page if they aren't directed.

lawrence-dxw:
I've changed it so that it hides the button, and tweaked some messages (see commit message for more details). What do you think, @pezholio?

## Without JavaScript enabled:
image

## With JavaScript enabled:
image

PR #770, opened 2020-01-14
Data report request CSV includes claimant's full name

tekin:
So that service operators can easily make requests to TPS for salary and employment data, we modify the existing report extract code to also include the claimants' full names.


app/models/claim/database_of_qualified_teachers_report_request.rb
```diff
@@ -2,21 +2,22 @@
...
 
     def initialize(claims)
       @claims = claims
     end
 
     def to_csv
-      CSV.generate do |csv|
-        csv << ATTRIBUTES.values
-
+      CSV.generate(write_headers: true, headers: HEADERS) do |csv|
```
robbiepaul: 👍

PR #771, opened 2020-01-14
Add a payment date to payroll runs

pezholio:
As discussed in #629, we thought it best to add a reliable way to see when a claim has / will be paid. Rather than gunk up the other PR, I've opened a new PR to do this, as there's extra work to be done with a data migration and refactoring the mailers.

### DIFF
--- a/app/mailers/payment_mailer.rb
+++ b/app/mailers/payment_mailer.rb
@@ -1,9 +1,9 @@
 class PaymentMailer < ApplicationMailer
   helper :application
 
-  def confirmation(payment, payment_date_timestamp)
+  def confirmation(payment)
     @payment = payment
-    @payment_date = Time.at(payment_date_timestamp).to_date
+    @payment_date = payment.scheduled_payment_date
     @display_name = [payment.first_name, payment.surname].join(" ")
 
     if payment.claims.size == 1
--- a/app/models/payment_confirmation.rb
+++ b/app/models/payment_confirmation.rb
@@ -21,11 +21,14 @@ class PaymentConfirmation
         update_payment_fields(payment, row) if payment
       end
 
-      payroll_run.update!(confirmation_report_uploaded_by: admin_user_id)
+      payroll_run.update!(
+        confirmation_report_uploaded_by: admin_user_id,
+        scheduled_payment_date: scheduled_payment_date
+      )
 
       if errors.empty?
         payroll_run.payments.each do |payment|
-          PaymentMailer.confirmation(payment, payment_date_timestamp).deliver_later
+          PaymentMailer.confirmation(payment, scheduled_payment_date.to_time.to_i).deliver_later
           RecordPaymentJob.perform_later(payment)
         end
       else
@@ -36,8 +39,8 @@ class PaymentConfirmation
 
   private
 
-  def payment_date_timestamp
-    Date.today.next_occurring(:friday).to_time.to_i
+  def scheduled_payment_date
+    Date.today.next_occurring(:friday)
   end
 
   def validate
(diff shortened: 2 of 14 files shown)


app/mailers/payment_mailer.rb
```diff
@@ -1,9 +1,9 @@
 class PaymentMailer < ApplicationMailer
   helper :application
 
-  def confirmation(payment, payment_date_timestamp)
+  def confirmation(payment)
     @payment = payment
-    @payment_date = Time.at(payment_date_timestamp).to_date
+    @payment_date = payment.payroll_run.payment_date
```
tekin: How about we encapsulate this by adding a delegated `Payment#payment_date` (or rather `Payment#scheduled_payment_date`) method?

app/models/payment_confirmation.rb
```diff
@@ -36,8 +39,8 @@ def ingest
 
   private
 
-  def payment_date_timestamp
-    Date.today.next_occurring(:friday).to_time.to_i
+  def payment_date
```
tekin: If we're renaming the column maybe we should also rename this method to `scheduled_payment_date`?

spec/models/claim_spec.rb
```diff
@@ -621,4 +621,23 @@
       expect(described_class.payrollable).to match_array([first_unpayrolled_claim, second_unpayrolled_claim])
     end
   end
+
+  describe "#scheduled_payment_date" do
+    let(:claim) { build(:claim, :submitted) }
+
+    context "when a claim has a payroll run associated with it" do
+      let!(:payroll_run) { create(:payroll_run, scheduled_payment_date: Date.parse("2019-01-01")) }
+      let!(:payment) { create(:payment, :with_figures, net_pay: 500.00, student_loan_repayment: 60, claims: [claim], payroll_run: payroll_run) }
```
tekin: Could the two lines above now be simplified to just the one?

```
      let!(:payment) { create(:payment, claims: [claim], scheduled_payment_date: Date.parse("2019-01-01")) }
```

Also when a value is important to the meat of the test, i.e. in this case the date, I sometimes find it helpful to see it pulled out to a named variable, perhaps with a let:

```
  let(:scheduled_payment_date) { Date.parse("2019-01-01") }
```

spec/models/claim_spec.rb
```diff
@@ -621,4 +621,23 @@
...
+
+      it "returns the payment date" do
+        expect(claim.scheduled_payment_date).to eq(payroll_run.scheduled_payment_date)
+      end
+    end
+
+    context "when a claim does not have a payroll run associated with it" do
+      it "returns nil" do
+        expect(claim.scheduled_payment_date).to eq(nil)
```
tekin:
```suggestion
        expect(claim.scheduled_payment_date).to be_nil
```

spec/models/payment_confirmation_spec.rb
```diff
@@ -43,8 +43,19 @@
       expect(second_payment.tax).to eq("162.8".to_d)
       expect(second_payment.net_pay).to eq("534".to_d)
       expect(second_payment.gross_pay).to eq("814.64".to_d)
+    end
+
+    it "populates the payroll run's confirmation_report_uploaded_by and sets the correct payment date" do
+      a_tuesday = Date.parse("2019-01-01")
+      the_following_friday = Date.parse("2019-01-04")
```
tekin: these named variables make the spec really expressive 👍🏽. Maybe we should say "the following Friday" in the spec description. When I see "correct" in a spec description it feels like a smell as the spec description is an opportunity to tell me what "correct" in this case is without forcing me having to read the entire spec to figure that out.

pezholio: Yesh, that's a bad habit of mine. I'm trying to get out of it

PR #773, opened 2020-01-14
Use most recent address from Verify even if not verified

robbiepaul:
When we get a users details back from GOV.UK Verify, we store their most recent "verified" address against their claim and play that back to them. However a user may have a more recent address that isn't "verified".


robbiepaul:
There's a couple of failing specs I need to fix

PR #775, opened 2020-01-15
Use most recent address from GOV.UK Verify even if not verified

robbiepaul:
When we get a users details back from GOV.UK Verify, we store their most recent "verified" address against their claim and play that back to them. However a user may have a more recent address that isn't "verified".


db/migrate/20200115092652_rename_verified_fields_to_gov_uk_verify_fields.rb
```diff
@@ -0,0 +1,5 @@
+class RenameVerifiedFieldsToGovUkVerifyFields < ActiveRecord::Migration[6.0]
+  def change
+    rename_column :claims, :verified_fields, :govuk_verify_fields
```
pezholio: 👍

app/models/claim.rb
```diff
@@ -202,23 +202,23 @@ def student_loan_country_with_one_plan?
...
   end
 
   def name_verified?
-    verified_fields.include?("first_name")
+    govuk_verify_fields.include?("first_name")
   end
 
-  def address_verified?
-    (ADDRESS_ATTRIBUTES & verified_fields).any?
+  def address_from_govuk_verify?
```
pezholio: Wondering if this should be moved below all the `*_verified?` methods, just so it's clearer the intent is different

app/helpers/claims_helper.rb
```diff
@@ -1,12 +1,12 @@
 module ClaimsHelper
-  def verified_fields(claim)
+  def govuk_verify_fields(claim)
```
pezholio: Maybe a better name for this would be `fieldnames_from_govuk_verify` to differentiate it from `claim.verified_fields`

tekin:
Will we need to take any extra care when we deploy this given we're changing a column name?

pezholio:
Yep, good call. It'll probably be fine, but probably best to make sure we deploy at a quiet time (so any time during the school day) and disable the service.

robbiepaul:
Comments addressed and rebased, ready for another look

robbiepaul:
Yep will do 👍

robbiepaul:
[Use "current address" from Verify, even if not verified](https://trello.com/c/2tsLhLyA/927-use-current-address-from-verify-even-if-not-verified)

PR #776, opened 2020-01-15
Add log aggregation to the service

jacksonj04:
Send _application_ logs to a Logstash instance hosted on LogIt.io, so that aggregated logs are searchable and more easily monitored.

This uses the [Rails Semantic Logger](http://rocketjob.github.io/semantic_logger/rails) gem to make the information which is pushed up to Logstash more structured and useful, as well as reducing log noise by turning a request from many log entries into one.

kibana logit io_app_kibana

## Docker Logs: Here Be Dragons
@lawrence-dxw  and I spent a good couple of hours dredging through Azure documentation and blog posts from across the web. Getting Docker logs into LogIt.io is a massive pain - the best we could come up with involves configuring Azure logging to write to a Storage Account, and then having a new worker with the Storage Account mounted and running FileBeat to pipe them into Logstash.

Find Teacher Training advises that they also do not log Docker output, and that they ran into the exact same problems and thinking. For the time being I recommend that it is not worth tackling Docker logs further unless we can identify a specific need, and sticking purely to the Rails application logs.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -60,6 +60,9 @@ gem "jwt"
 # Send app telemetry to Azure Application Insights
 gem "application_insights", git: "https://github.com/microsoft/ApplicationInsights-Ruby.git", ref: "5db6b4ad65262d23f26b678143a4a1fd7939e5c2"
 
+# Semantic Logger provides more useful application log entries
+gem "rails_semantic_logger"
+
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem "byebug", platforms: [:mri, :mingw, :x64_mingw]
--- a/Gemfile
+++ b/Gemfile
@@ -63,6 +63,9 @@ gem "application_insights", git: "https://github.com/microsoft/ApplicationInsigh
 # Semantic Logger provides more useful application log entries
 gem "rails_semantic_logger"
 
+# Use logstash-logger to send to Logit.io
+gem "logstash-logger", "~> 0.26"
+
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem "byebug", platforms: [:mri, :mingw, :x64_mingw]
--- a/azure/resource_groups/app/parameters/development.template.json
+++ b/azure/resource_groups/app/parameters/development.template.json
@@ -156,6 +156,17 @@
         "secretName": "GeckoboardAPIKey"
       }
     },
+    "LOGSTASH_HOST": {
+      "reference": {
+        "keyVault": {
+          "id": "${keyVaultId}"
+        },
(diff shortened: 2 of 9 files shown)


robbiepaul:
It's a shame about the Docker logs, another black flag against Azure! We've ran in to a few errors previously that couldn't be explained by Rails application logs

Gemfile
```diff
@@ -60,6 +60,12 @@ gem "jwt"
 # Send app telemetry to Azure Application Insights
 gem "application_insights", git: "https://github.com/microsoft/ApplicationInsights-Ruby.git", ref: "5db6b4ad65262d23f26b678143a4a1fd7939e5c2"
 
+# Semantic Logger provides more useful application log entries
+gem "rails_semantic_logger"
+
+# Use logstash-logger to send to Logit.io
+gem "logstash-logger", "~> 0.26.1"
```
robbiepaul: I think this will restrict us to patch versions of `0.26`. Is there a particular reason why we should restrict the version? If the gem follows semver we could probably get away with using `~> 0.26` which allows us to use any version `< 1.0`

azure/resource_groups/app/template.json
```diff
@@ -492,6 +498,10 @@
                 "name": "GECKOBOARD_API_KEY",
                 "value": "[parameters('GECKOBOARD_API_KEY')]"
               },
+              {
+                "name": "LOGSTASH_HOST",
+                "value": "[parameters('LOGSTASH_HOST')]"
+              },
```
robbiepaul: We need to define the `LOGSTASH_PORT` value here

azure/resource_groups/app/parameters/development.template.json
```diff
@@ -156,6 +156,22 @@
...
+        "secretName": "LogstashHost"
+      }
+    },
+    "LOGSTASH_PORT": {
+      "reference": {
+        "keyVault": {
+          "id": "${keyVaultId}"
+        },
+        "secretName": "LogstashPort"
+      }
```
robbiepaul: I understand why the host needs to be a secret, but can the port just be defined in the parameters? Things in the KeyVault have a higher maintenance burden for the team so we should try to keep the amount of them down to a minimum.

jacksonj04: My logic was that it keeps the port and the host in the same place, but I'm easy on this if you think it makes more sense otherwise.

robbiepaul: Yeah that makes sense. Although I think having the value of the environment variables defined in the code where we can get away with it is probably better than keeping them together. If I was a future developer picking up a card to change the environment variables I would start with these ARM templates.

PR #777, opened 2020-01-15
1108 2.4.4 link purpose in context every links purpose is clear from its context

titlescreen:
This PR will amend the change links on the check your answers page so they will have an ARIA label to provide context for the link.

E.g. `Change which subject did you complete your initial teacher training in?`

This is important as the links will now provide context and let accessibility users know what question they are changing the answer to. We did consider adding some words before the question to help explain the possible janky-ness of the sentence being read out, but when testing with voice over, it was just adding extra words that didn't have much value and made it harder to focus on which question we are changing.

### DIFF
--- a/app/views/claims/_check_your_answers_section.html.erb
+++ b/app/views/claims/_check_your_answers_section.html.erb
@@ -16,7 +16,7 @@
       </dd>
 
       <dd class="govuk-summary-list__actions">
-        <%= link_to('Change', claim_path(current_policy_routing_name, slug), class: 'govuk-link') unless slug.nil? %>
+        <%= link_to('Change', claim_path(current_policy_routing_name, slug), class: 'govuk-link', 'aria-label': "Change #{label.downcase}") unless slug.nil? %>
       </dd>
     </div>
   <% end %>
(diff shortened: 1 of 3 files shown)


app/views/claims/_check_your_answers_section.html.erb
```diff
@@ -16,7 +16,7 @@
       </dd>
 
       <dd class="govuk-summary-list__actions">
-        <%= link_to('Change', claim_path(current_policy_routing_name, slug), class: 'govuk-link') unless slug.nil? %>
+        <%= link_to('Change' , claim_path(current_policy_routing_name, slug), class: 'govuk-link', 'aria-label': "Change  #{label.downcase}") unless slug.nil? %>
```
jacksonj04: Are the extra spaces significant to the change, or did they sneak in when removing the extra words? I'm not up to speed with how screen-readers may interpret them so may be entirely legit!

Also, an errant space has snuck in after the first parameter of `link_to()` which could probably be removed in the interests of cleanliness.

titlescreen: Thanks Nick, nope those spaces shouldn't have been in there, I've amended it now :)

PR #778, opened 2020-01-15
Return worker container to virtual network

pezholio:
Now the issues with Azure's virtual networks and containers have been solved, we can now put the container back in the VN. We've had to rename the VN too, as the original network had some issues. This new name seems more descriptive now though.


tekin:
Should this PR have got rid of the `bin/update-db-firewall-rules` script?

pezholio:
I'd rather keep it around for now, until we're absolutely sure we don't need it

tekin:
You gotta get over that habit: we have something called a revision history which means it never goes away 😉

If we've tested the vnet-based approach and are happy that it works and don't expect to run the script unless we revert back to the non-vnet-based solution, then there is no reason why this script should exist as a file in the latest version of our codebase.

PR #781, opened 2020-01-16
Change links that look like buttons to be buttons

titlescreen:
We have a couple of links that are styled to look like buttons and have
the appropriate aria role so that screen readers will put them up
however, they do not function as a keyboard only user wold expect them
to. Users were unable to give the buttons focus and press space to
trigger them.

This PR swaps the links to be actual buttons so they will function as
expected.

### DIFF
--- a/app/views/maths_and_physics/claims/eligibility_confirmed.html.erb
+++ b/app/views/maths_and_physics/claims/eligibility_confirmed.html.erb
@@ -5,6 +5,7 @@
     <h1 class="govuk-heading-xl">You are eligible to claim a payment for teaching maths or physics</h1>
     <p class="govuk-body">Based on your answers you can claim £2,000 for teaching maths or physics.</p>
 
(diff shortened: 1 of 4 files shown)


app/views/maths_and_physics/claims/eligibility_confirmed.html.erb
```diff
@@ -5,6 +5,7 @@
     <h1 class="govuk-heading-xl">You are eligible to claim a payment for teaching maths or physics</h1>
     <p class="govuk-body">Based on your answers you can claim £2,000 for teaching maths or physics.</p>
 
-    <%= link_to "Continue", claim_path(current_policy_routing_name, next_slug), class: "govuk-button", role: "button", data: { module: "govuk-button" } %>
+    <%= button_to "Continue", claim_path(current_policy_routing_name, next_slug), method: :get, class: "govuk-button", data: { module: "govuk-button" } %>
```
tekin: Can we drop the `data: { module: "govuk-button" }` on this now?

titlescreen: Facepalm, yes, yes we can and we will! 👍

PR #784, opened 2020-01-16
Enable the firewall on our KeyVaults and storage accounts

robbiepaul:
To harden up our infrastructure a firewall has been added to our storage account and KeyVault.

This will affect developers viewing secrets in our KeyVault. It will require an extra step of adding your IP to the allow list on the firewall. For production this will require a PIM request. I've added some documentation to [confluence].(https://dfedigital.atlassian.net/wiki/spaces/TP/pages/1264418844/Secrets)

The storage account used by Postgres' security alerts is no longer needed so I've removed it.

### DIFF
--- a/azure/resource_groups/ProjectCore/parameters/development.json
+++ b/azure/resource_groups/ProjectCore/parameters/development.json
@@ -5,6 +5,9 @@
     "vnetAddressSpaceCIDR": {
       "value": "11.0.1.0/24"
     },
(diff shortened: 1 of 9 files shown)


robbiepaul:
[Enable the firewall on our KeyVaults and Storage accounts](https://trello.com/c/LyvE2qQg/932-enable-the-firewall-on-our-keyvaults-and-storage-accounts)

PR #785, opened 2020-01-16
Purge unsubmitted claims nightly.

tekin:
This will run every day at midnight and remove any unsubmitted claims
that are more than 24 hours old so we don't hang on to user data that we
don't need and can't use.

Because this inheritts from `CronJob` it will automatically get picked
up and scheduled by `CronJobScheduler` when the `jobs:schedule` rake
task is run.

**Note: this PR delivers one part of the larger story around data purging and anonymising.**

### DIFF
--- a/app/jobs/purge_unsubmitted_claims_job.rb
+++ b/app/jobs/purge_unsubmitted_claims_job.rb
@@ -1,0 +1,17 @@
+# Runs nightly at midnight and deletes any unsubmitted claims that are more than
+# 24 hours old, purging our database of data we don't want or need to hang on
+# to.
+class PurgeUnsubmittedClaimsJob < CronJob
+  self.cron_expression = "0 0 * * *"
+
+  def perform
+    Rails.logger.info "Purging #{old_unsubmitted_claims.count} old and unsubmitted claims from the database"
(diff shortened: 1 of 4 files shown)


app/jobs/purge_unsubmitted_claims_job.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  def perform
+    Rails.logger.info "Purging #{old_unsubmitted_claims.count} old and unsubmitted claims from the database"
+    old_unsubmitted_claims.destroy_all
+  end
+
+  private
+
+  def old_unsubmitted_claims
+    Claim.unsubmitted.where("updated_at < ?", 24.hours.ago)
```
pezholio: Should this be a scope as well?

tekin: I don't believe so, no. I think creating a named scope would only make sense if the concept of an "old" claim was more firmly defined in our domain, and was also useful outside of the context of this specific job. In this case the 24 hour value is mostly arbitrary. The important part here is that it is higher than the session timeout. I think naming an `old` scope in `Claim` just for this would be a premature abstraction and would lend it more weight and meaning than it actually has right now.

PR #788, opened 2020-01-20
Batch Geckoboard requests

pezholio:
After deploying #629, we got an error because there are now more than 500 sumbitted claims, causing Geckoboard to hit us with an error, as we can only submit 500 datapoints at a time. This adds batching functionality to the `GeckoboardEvent` class, so claims are batched into groups of 500.

### DIFF
--- a/spec/models/claim/geckoboard_event_spec.rb
+++ b/spec/models/claim/geckoboard_event_spec.rb
@@ -1,7 +1,7 @@
 require "rails_helper"
 
 RSpec.describe Claim::GeckoboardEvent, type: :model do
-  let(:claim) { create(:claim, :submitted, created_at: DateTime.now) }
+  let(:claim) { build(:claim, :submitted, created_at: DateTime.now) }
   let(:event_name) { "some_event" }
 
   it "sends the claim's reference, policy, along with the timestamp to the dataset indicated by the event name" do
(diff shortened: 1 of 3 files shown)


spec/models/claim/geckoboard_event_spec.rb
```diff
@@ -17,4 +17,24 @@
       }).to have_been_requested
     end
   end
+
+  context "with lots of claims" do
+    let(:claims) { build_list(:claim, 520, :submitted, created_at: DateTime.now) }
```
robbiepaul: We could use `stub_const` to set `Claim::GeckoboardEvent:: BATCH_SIZE` to a smaller amount so we don't need to build so many objects here. It might speed up the spec a little bit... I have mixed feelings about stubbing constants so I'll leave it up to you.

spec/models/claim/geckoboard_event_spec.rb
```diff
@@ -17,4 +17,24 @@
...
+
+    it "batches the claims into groups" do
+      event = Claim::GeckoboardEvent.new(claims, event_name, :created_at)
+
+      dataset_post_stub = stub_geckoboard_dataset_update("claims.#{event_name}.test")
+
+      event.record
+
+      expect(dataset_post_stub.with { |request|
+        request_body_matches_geckoboard_data_for_claims?(request, claims[0, 500], :created_at)
```
robbiepaul: I had to look up the syntax for using a 2nd argument with `Array[]`. Would `claims.first(500)` and `claims.last(20)` be more expressive?

pezholio: Good call 👍

spec/models/claim/geckoboard_event_spec.rb
pezholio: Good call, I don't think there's a huge performance benefit from doing this though, given that we're only building the claims, not creating them. I think I'd rather live with the slight extra fraction of a second than polluting our tests with `stub_const`

PR #789, opened 2020-01-20
Replace multiple Geckoboard datasets with a single one

jacksonj04:
To simplify the creation of dashboards in Geckoboard, these changes replace maintaining multiple datasets (one per type of event) with a single one, containing (for each claim):

* The claim reference
* The policy the claim is part of
* The date and time the claim was submitted
* If the claim has been checked, and if so:
  * The date and time the claim was checked
  * The outcome of the check
* If the claim has been paid (or, more accurately, scheduled to be paid), and if so:
  * The date and time the claim is scheduled to be paid

Once this is merged you can backfill the Geckoboard dataset with:

``` ruby
claim_ids = Claim.submitted.pluck(:id)
RecordOrUpdateGeckoboardDatasetJob.perform_now(claim_ids)
```


app/models/claim/geckoboard_claim.rb
```diff
@@ -0,0 +1,71 @@
...
+      claims.map do |claim|
+        {
+          reference: claim.reference,
+          policy: claim.policy.to_s,
+          submitted_at: claim.submitted_at,
+          checked: claim.check.present?.to_s,
+          checked_at: claim.check.present? ? claim.check.created_at : placeholder_date_for_nil_value,
+          check_result: claim.check.present? ? claim.check.result : "",
+          paid: claim.scheduled_payment_date.present?.to_s,
+          paid_at: claim.scheduled_payment_date.present? ? claim.scheduled_payment_date : placeholder_date_for_nil_value,
```
robbiepaul: This could be simplified to
```suggestion
          paid_at: claim.scheduled_payment_date.presence || placeholder_date_for_nil_value,
```

spec/jobs/record_claim_event_job_spec.rb
```diff
@@ -1,22 +1,21 @@
 require "rails_helper"
 
-RSpec.describe RecordClaimEventJob do
+RSpec.describe RecordOrUpdateGeckoboardClaimJob do
```
robbiepaul: I think this filename needs updating too

spec/models/claim/geckoboard_event_spec.rb
```diff
@@ -1,19 +1,18 @@
 require "rails_helper"
 
-RSpec.describe Claim::GeckoboardEvent, type: :model do
+RSpec.describe Claim::GeckoboardClaim, type: :model do
```
robbiepaul: I think this filename also needs updating

PR #791, opened 2020-01-20
Update next steps when not using Verify

pezholio:
This better reflects the next steps that actually get taken when a user can't Verify using GOV.UK Verify.


pezholio:
[Change next steps in Verify without Verify confirmation email](https://trello.com/c/a9vVxwdL/1198-change-next-steps-in-verify-without-verify-confirmation-email)

PR #792, opened 2020-01-20
Disable application alerts in the test environment

robbiepaul:
Most of the alerts are intended to let us know that the application isn't working as expected. For the test environment we mainly care about if the infrastructure has deployed or not. eg. Alerts letting us
know that the CPU usage is high during a deployment are just noise.

I've left the service health alerts in there as they only happen when there's a problem with the infrastructure.

### DIFF
--- a/azure/resource_groups/app/template.json
+++ b/azure/resource_groups/app/template.json
@@ -103,6 +103,10 @@
     "alertEmailAddress": {
       "type": "string"
     },
+    "enableWebTestAlerts": {
+      "type": "bool",
+      "defaultValue": true
+    },
     "azureBuildNumber": {
       "type": "string",
       "defaultValue": ""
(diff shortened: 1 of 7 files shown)


azure/resource_groups/app/template.json
```diff
@@ -342,6 +346,9 @@
           },
           "webTestLocations": {
             "value": "[variables('azureAvailabilityTestLocations')]"
+          },
+          "enableAlerts": {
```
pezholio: Would it be better to name this variable `enableWebTestAlerts`?

robbiepaul: Sure, my reasoning was that this is a local parameter to the `availability_test` deployment. I've changed it to `enableWebTestAlerts`

pezholio: Cool, just makes it a bit more easy to follow

PR #794, opened 2020-01-21
Feedback link says it opens in new tab

titlescreen:
As the link opens in a new window or tab, we should tell the user this
so that they are not sent off somewhere unexpected.

image


titlescreen:
Yeah, I thought the same but like you say this seems to be how everyone is doing it...

PR #796, opened 2020-01-21
Allow soft-deleted KeyVaults to be recovered with a flag

robbiepaul:
We considered making this automatic with another step of the script that checks to see if a soft-deleted KeyVault exists and change the `createMode` depending on the result.

However we decided against it. Deleting a KeyVault is quite a purposeful action, the expected result from the next deployment should error loudly to the user to let them know they need to either delete the KeyVault permanently or recover it explicitly.

The KeyVault can be recovered using the `--recover-keyvault` flag when calling the azure-deploy script.

### DIFF
--- a/azure/resource_groups/secrets/template.json
+++ b/azure/resource_groups/secrets/template.json
@@ -17,6 +17,9 @@
       "defaultValue": 365,
       "minValue": 0,
       "maxValue": 365
+    },
+    "keyVaultCreateMode": {
+      "type": "string"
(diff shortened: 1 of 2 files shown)


robbiepaul:
[Recover deleted key vaults if there's a soft deleted version still available](https://trello.com/c/QUEu66C7/504-recover-deleted-key-vaults-if-theres-a-soft-deleted-version-still-available)

PR #797, opened 2020-01-21
Add environment name tags to Logstash entries

jacksonj04:
This allows for aggregated logs from different deployments to be distinguished in Logstash.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -50,14 +50,5 @@ module DfeTeachersPaymentService
     config.action_view.form_with_generates_remote_forms = false
 
     config.guidance_url = "https://www.gov.uk/government/publications/additional-payments-for-teaching-eligibility-and-payment-details"
-
-    if ENV["LOGSTASH_HOST"].present?
-      tcp_logger = LogStashLogger.new(type: :tcp,
-                                      host: ENV.fetch("LOGSTASH_HOST"),
-                                      port: ENV.fetch("LOGSTASH_PORT"),
-                                      ssl_enable: true)
-
-      SemanticLogger.add_appender(logger: tcp_logger, level: :info, formatter: :json)
-    end
   end
 end
--- a/config/application.rb
+++ b/config/application.rb
@@ -50,5 +50,11 @@ module DfeTeachersPaymentService
     config.action_view.form_with_generates_remote_forms = false
 
     config.guidance_url = "https://www.gov.uk/government/publications/additional-payments-for-teaching-eligibility-and-payment-details"
+
+    # Additional information which is passed in the logs for each request
+    # See https://rocketjob.github.io/semantic_logger/rails.html#named-tags
+    config.log_tags = {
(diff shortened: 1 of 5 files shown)


config/application.rb
```diff
@@ -51,13 +51,8 @@ class Application < Rails::Application
...
-    if ENV["LOGSTASH_HOST"].present?
-      tcp_logger = LogStashLogger.new(type: :tcp,
-                                      host: ENV.fetch("LOGSTASH_HOST"),
-                                      port: ENV.fetch("LOGSTASH_PORT"),
-                                      ssl_enable: true)
-
-      SemanticLogger.add_appender(logger: tcp_logger, level: :info, formatter: :json)
-    end
+    config.log_tags = {
+      environment: ENV.fetch("ENVIRONMENT_NAME"),
```
pezholio: It might be best to add a default option to `.fetch` here, so `ENV.fetch("ENVIRONMENT_NAME", "local")` so it doesn't raise an error if `ENVIRONMENT_NAME` isn't set

robbiepaul: Do we need to add `ENVIRONMENT_NAME` to the `.env.development` file? The app won't start up without this environment variable now.

config/application.rb
```diff
@@ -51,13 +51,8 @@ class Application < Rails::Application
...
 
-    if ENV["LOGSTASH_HOST"].present?
-      tcp_logger = LogStashLogger.new(type: :tcp,
-                                      host: ENV.fetch("LOGSTASH_HOST"),
-                                      port: ENV.fetch("LOGSTASH_PORT"),
-                                      ssl_enable: true)
-
-      SemanticLogger.add_appender(logger: tcp_logger, level: :info, formatter: :json)
-    end
+    config.log_tags = {
```
tekin: Where does this magic config come from and where does it get used? Is it picked up by SemanticLogger? Something that LogStash reads? It looks quite mysterious to me know in the context of this PR. Future developer will likely struggle even more to understand it.

config/application.rb
```diff
@@ -51,13 +51,10 @@ class Application < Rails::Application
...
-      tcp_logger = LogStashLogger.new(type: :tcp,
-                                      host: ENV.fetch("LOGSTASH_HOST"),
-                                      port: ENV.fetch("LOGSTASH_PORT"),
-                                      ssl_enable: true)
-
-      SemanticLogger.add_appender(logger: tcp_logger, level: :info, formatter: :json)
-    end
+    # Additional information which is passed in the logs for each request
+    # See https://rocketjob.github.io/semantic_logger/rails.html#named-tags
+    config.log_tags = {
```
pezholio: I wonder if it makes sense to put this in with the rest of the logstash config in that initializer?

jacksonj04: This is picked up by Semantic Logger in all contexts, not just when sending to Logstash. I debated putting it in there since that's where we use it, but actually this information would be useful no matter where logs might end up being sent so I figured it would work better being applied to across the board, even if Logstash wasn't initialised.

PR #798, opened 2020-01-21
Record total claims currently outside of their checking deadline

robbiepaul:
The number of claims passed their checking deadline will fluctuate depending on the state of their claim and the date they submitted their application.

We can combine `passed_checking_deadline == true` and `checked == false` in Geckoboard to get the number of unchecked claims that are passed their deadline.

This means that we need to ensure each claim dataset is updated when a check has been submitted and everyday on a cronjob to check if the deadline has passed.


robbiepaul:
[Record Total claims currently outside of their checking SLA in Geckoboard](https://trello.com/c/DsGeiz7Z/1019-record-total-claims-currently-outside-of-their-checking-sla-in-geckoboard)

PR #801, opened 2020-01-22
Record the mean time to reach a decision on a claim

pezholio:
This updates the Geckoboard dataset to add a `number_of_days_to_check` value, which gets updated when a claim is approved. This will then allow us to show the average number of days it takes a claim to be approved / rejected on our dashboard(s)

### DIFF
--- a/app/models/check.rb
+++ b/app/models/check.rb
@@ -14,6 +14,10 @@ class Check < ApplicationRecord
     persisted?
   end
 
+  def number_of_days_since_claim_submitted
+    (claim.check.created_at.to_date - claim.submitted_at.to_date).to_i
+  end
+
   private
 
   def claim_must_be_approvable
--- a/app/models/claim/geckoboard_dataset.rb
+++ b/app/models/claim/geckoboard_dataset.rb
@@ -18,14 +18,18 @@ class Claim
     ]
     BATCH_SIZE = 500
 
-    def initialize(claim_or_claims)
-      @claims = Array(claim_or_claims)
+    def initialize(claims: [])
+      @claims = claims
     end
 
     def save
       batched_claims.each do |batch|
         dataset.post(data_for_claims(batch))
       end
+    end
+
+    def delete
+      client.datasets.delete(dataset_name)
     end
 
     private
--- a/app/models/claim/geckoboard_dataset.rb
+++ b/app/models/claim/geckoboard_dataset.rb
@@ -12,6 +12,7 @@ class Claim
       Geckoboard::StringField.new(:checked, name: "Checked"),
       Geckoboard::DateTimeField.new(:checked_at, name: "Checked at"),
       Geckoboard::StringField.new(:check_result, name: "Check result"),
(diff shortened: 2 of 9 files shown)


app/models/check.rb
```diff
@@ -14,6 +14,10 @@ def readonly?
     persisted?
   end
 
+  def number_of_days_to_check
```
jacksonj04: This method name doesn't quite fit to me – since it's on the `Check` model it implies it's the number of days the check took to complete.

Maybe `number_of_days_since_claim_submitted`?

app/models/claim/geckoboard_dataset.rb
```diff
@@ -39,6 +40,7 @@ def data_for_claims(claims)
           checked: claim.check.present?.to_s,
           checked_at: claim.check.present? ? claim.check.created_at : placeholder_date_for_nil_value,
           check_result: claim.check.present? ? claim.check.result : "",
+          number_of_days_to_check: claim.check&.number_of_days_to_check,
```
jacksonj04: Should this be behind a `claim.check.present?`, or `claim&.check.number_of_days_to_check`?

spec/support/geckoboard_helpers.rb
```diff
@@ -75,6 +80,7 @@ def request_body_matches_geckoboard_data_for_claims?(request, claims)
         "checked" => claim.check.present?.to_s,
         "checked_at" => (claim.check.present? ? claim.check.created_at : DateTime.parse("1970-01-01")).strftime("%Y-%m-%dT%H:%M:%S%:z"),
         "check_result" => claim.check.present? ? claim.check.result : "",
+        "number_of_days_to_check" => claim.check&.number_of_days_to_check,
```
jacksonj04: Ditto with safe navigation operator position.

lib/tasks/geckoboard.rake
```diff
@@ -0,0 +1,7 @@
+namespace :geckoboard do
+  desc "Reset and recreate Geckoboard claims dataset"
+  task reset: :environment do
+    Claim::GeckoboardDataset.new(nil).delete
```
jacksonj04: Maybe a comment explaining why we're passing `nil` as the argument (eg "we don't have any data to send, we just need the object to manipulate")? It's not immediately clear what's going on here.

app/models/check.rb
pezholio: 👍

app/models/claim/geckoboard_dataset.rb
pezholio: Nah, we only need the safe navigation operator on the the `check`, as it could potentially be nil

lib/tasks/geckoboard.rake
pezholio: Good call 👍

pezholio: I've actually tweaked the class now to make the claims an optional parameter, which makes more sense, rather than passing in `nil`

pezholio:
Whoops! That should've been one commit!

PR #804, opened 2020-01-22
Prevent service operator from approving an inconsistent claim

lawrence-dxw:
This implements the work hinted at in 2d7bf9b. If a service operator
views a claim for which there is a claim from the same claimant which
has been approved but not yet payrolled, and which has inconsistent
personal details which would prevent the two claims from being combined
into a single payment, then inform the service operator that this has
happened and prevent them from approving the claim.

This means that it is now unlikely that we will ever violate the
validations introduced on Payment in 2d7bf9b, but we keep them as an
additional safety net.

We still need to decide how a service operator can resolve this situation
of an inconsistent claim. For example, one option would be to contact
the claimant and ask them whether they are happy to be paid in two
different months or to ask them to re-submit their claim. If we wanted
to enable this sort of conversation, we may need to think about exposing
a subset of the claimant's bank details to the service operator (in a
privacy-preserving manner).

The implementation and user interface follow the same patterns that we
use for the "payroll gender missing" situation.

* Screenshot

Screenshot 2020-01-22 at 14 40 23

### DIFF
--- a/app/controllers/admin/claim_checks_controller.rb
+++ b/app/controllers/admin/claim_checks_controller.rb
@@ -3,6 +3,7 @@ class Admin::ClaimChecksController < Admin::BaseAdminController
   before_action :load_claim
   before_action :reject_checked_claims
   before_action :reject_missing_payroll_gender
+  before_action :reject_if_claims_preventing_payment
 
   def create
     @check = @claim.build_check(check_params.merge(checked_by: admin_user))
@@ -11,6 +12,7 @@ class Admin::ClaimChecksController < Admin::BaseAdminController
       RecordOrUpdateGeckoboardDatasetJob.perform_later([@claim.id])
       redirect_to admin_claims_path, notice: "Claim has been #{@claim.check.result} successfully"
     else
+      @claims_preventing_payment = claims_preventing_payment_finder.claims_preventing_payment
       render "admin/claims/show"
     end
   end
@@ -22,6 +24,10 @@ class Admin::ClaimChecksController < Admin::BaseAdminController
     @matching_claims = Claim::MatchingAttributeFinder.new(@claim).matching_claims
   end
 
+  def claims_preventing_payment_finder
+    @claims_preventing_payment_finder ||= Claim::ClaimsPreventingPaymentFinder.new(@claim)
+  end
+
   def reject_checked_claims
     if @claim.check.present?
       redirect_to admin_claim_path(@claim), notice: "Claim already checked"
@@ -31,6 +37,12 @@ class Admin::ClaimChecksController < Admin::BaseAdminController
   def reject_missing_payroll_gender
     if check_params[:result] == "approved" && @claim.payroll_gender_missing?
       redirect_to admin_claim_path(@claim), alert: "Claim cannot be approved"
+    end
+  end
+
+  def reject_if_claims_preventing_payment
+    if check_params[:result] == "approved" && claims_preventing_payment_finder.claims_preventing_payment.any?
+      redirect_to admin_claim_path(@claim), alert: "Claim cannot be approved because there are inconsistent claims"
     end
   end
 
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -174,7 +174,7 @@ class Claim < ApplicationRecord
   end
 
   def approvable?
-    submitted? && !payroll_gender_missing? && !checked?
+    submitted? && !payroll_gender_missing? && !checked? && !payment_prevented_by_other_claims?
   end
 
   def checked?
@@ -183,6 +183,10 @@ class Claim < ApplicationRecord
 
   def payroll_gender_missing?
     %w[male female].exclude?(payroll_gender)
+  end
+
+  def payment_prevented_by_other_claims?
+    ClaimsPreventingPaymentFinder.new(self).claims_preventing_payment.any?
   end
 
   def check_deadline_date
--- a/app/models/claim/claims_preventing_payment_finder.rb
+++ b/app/models/claim/claims_preventing_payment_finder.rb
@@ -1,0 +1,38 @@
+class Claim
+  class ClaimsPreventingPaymentFinder
+    attr_reader :claim
+
+    def initialize(claim)
+      @claim = claim
+    end
+
+    # Returns a list of claims which prevent us from adding `claim` to the
+    # payroll run for the current month.
+    #
+    # These claims are payrollable, and submitted by the same claimant as
+    # `claim`, as identified by National Insurance number.
+    #
+    # The returned claims have different payment or tax details to those
+    # provided by `claim`, and hence `claim` cannot be paid in the same payment
+    # as the returned claims.
+    def claims_preventing_payment
+      @claims_preventing_payment ||= find_claims_preventing_payment
+    end
+
+    private
+
+    def find_claims_preventing_payment
+      payrollable_claims_from_same_claimant = Claim.payrollable.where(national_insurance_number: claim.national_insurance_number)
+
+      payrollable_claims_from_same_claimant.select do |other_claim|
+        Payment::PERSONAL_DETAILS_ATTRIBUTES_FORBIDDING_DISCREPANCIES.any? do |attribute|
+          attribute_does_not_match?(other_claim, attribute)
+        end
+      end
+    end
+
+    def attribute_does_not_match?(claim_to_compare, attribute)
+      claim_to_compare.read_attribute(attribute) != claim.read_attribute(attribute)
+    end
+  end
+end
--- a/app/views/admin/claims/show.html.erb
+++ b/app/views/admin/claims/show.html.erb
@@ -66,6 +66,16 @@
       </table>
     <% end %>
 
+    <% if @claims_preventing_payment.any? %>
+      <div class="govuk-warning-text">
+        <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
+        <strong class="govuk-warning-text__text">
+          <span class="govuk-warning-text__assistive">Warning</span>
+          <%= t('admin.claims_preventing_payment_message', count: @claims_preventing_payment.count, claim_references: @claims_preventing_payment.map(&:reference).to_sentence) %>
+        </strong>
+      </div>
+    <% end %>
+
     <% if @check.persisted? %>
       <%= render "admin/claims/answer_section", {heading: "Claim decision details", answers: admin_check_details(@check)} %>
     <% else %>
--- a/spec/models/claim/claims_preventing_payment_finder_spec.rb
+++ b/spec/models/claim/claims_preventing_payment_finder_spec.rb
@@ -1,0 +1,56 @@
+require "rails_helper"
+
+RSpec.describe Claim::ClaimsPreventingPaymentFinder do
+  subject(:finder) { described_class.new(claim) }
+
+  describe "#claims_preventing_payment" do
+    let(:personal_details) do
+      {
+        national_insurance_number: generate(:national_insurance_number),
+        bank_account_number: "32828838",
+        bank_sort_code: "183828",
+        first_name: "Boris",
+      }
+    end
+    let(:claim) { create(:claim, :submitted, personal_details) }
+    subject(:claims_preventing_payment) { finder.claims_preventing_payment }
+
+    context "when there is another claim with the same National Insurance number, with inconsistent personal details that would prevent us from running payroll" do
+      let(:inconsistent_personal_details) do
+        personal_details.merge(
+          bank_account_number: "87282828",
+          bank_sort_code: "388183",
+        )
+      end
+
+      it "does not include the other claim when the other claim is not yet approved" do
+        create(:claim, :submitted, inconsistent_personal_details)
+        expect(claims_preventing_payment).to be_empty
+      end
+
+      it "includes the other claim when the other claim is approved but not yet payrolled" do
+        other_claim = create(:claim, :approved, inconsistent_personal_details)
+        expect(claims_preventing_payment).to eq([other_claim])
+      end
+
+      it "does not include the other claim when the other claim is already payrolled" do
+        other_claim = create(:claim, :approved, inconsistent_personal_details)
+        create(:payment, claims: [other_claim])
+        expect(claims_preventing_payment).to be_empty
+      end
+    end
+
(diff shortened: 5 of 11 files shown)


app/controllers/admin/claim_checks_controller.rb
```diff
@@ -34,6 +37,12 @@ def reject_missing_payroll_gender
     end
   end
 
+  def reject_inconsistent_claims
```
pezholio: This reads to me as if we are rejecting the inconsistent claims, rather than rejecting if there are inconsistent claims. Maybe something like `reject_if_inconsistent_claims`?

pezholio: In fact, there now seem to be quite a few validations in the before filters now. How about combining them all into one? Something like:

```ruby
before_action :validate_check

...

def validate_check
  reject_checked_claims
  reject_missing_payroll_gender
  reject_if_inconsistent_claims
end
```

app/models/claim.rb
```diff
@@ -246,6 +246,27 @@ def school
     eligibility&.current_school
   end
 
+  def inconsistent_claims
```
pezholio: There's quite a lot of logic in here, and I found it tricky to follow, and I struggled to see the rationale for some things that were going on (for example the call to `attributes.delete(:national_insurance_number)`). I'm wondering if it might be better to wrap this in a class (something like `InconsistentClaimFinder`). This finder could also have a method to return the inconsistent attributes too, to avoid having to pass the inconsistent claims in again in the ClaimsController.

app/views/admin/claims/show.html.erb
```diff
@@ -66,6 +66,18 @@
       </table>
     <% end %>
 
+    <% if @inconsistent_claims.any? %>
+      <div class="govuk-warning-text">
+        <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
+        <strong class="govuk-warning-text__text">
+          <span class="govuk-warning-text__assistive">Warning</span>
+          This claim cannot be approved because <%= @inconsistent_claims.count == 1 ? 'claim' : 'claims' %> <%= @inconsistent_claims.map(&:reference).to_sentence %>
```
pezholio: How about using translations for this? There's a lot of checking counts to work out plurals, and this is supported out of the box with Rails' i18n stuff. For example you could do something like:

```yml
inconsistency_message:
  one: This claim cannot be approved because claim %{reference_numbers} from the same claimant has different values for %{inconsistent_values}
  other: This claim cannot be approved because claims %{reference_numbers} from the same claimant have different values for %{inconsistent_values}
```

You can then use it in the template like so:

```erb
<%= t("inconsistency_message", count: @inconsistent_claims.count, reference_numbers: @inconsistent_claims.map(&:reference).to_sentence, inconsistent_values: @inconsistent_attributes.map { |attribute| Claim.human_attribute_name(attribute).downcase }.to_sentence %>
```

tekin:
A bit of a fly-by review, but one thing I noticed scanning the code and UX is there isn't much mention of the fact that the inconsistency only matters because of the presence of another _**approved claim already set to go in the payroll run this month**_. I feel like that is quite an important point to explain both to the user that is prevented from approving, and to the developer trying to understand what "inconsistent" means. I know there is a plan to do future work to figure out how a service operator would resolve this, but we should still do enough now such that it at least communicates to the user as clearly as possible what the issue relates to. Suggestion for the user side of things: show this to a service operator and see if they would understand what it means; show it to Faye to get her input. On the code side, I'd expect to see the words like "payroll" more in the code that deals with this issue.

app/views/admin/claims/show.html.erb
titlescreen: I was just about to write what Tekin said about mentioning the month.  I also wonder if there should be some guidance on how they can recover from this. Would it be worthwhile mentioning the service operator playbook?

app/controllers/admin/claim_checks_controller.rb
lawrence-dxw: `reject_if_inconsistent_claims` is a better name, I'll change that.

As for combining the multiple filters into a single filter, I think that this would modify the behaviour of this controller in a way that we don't want, because we'd lose Rails's "end filter chain if a filter renders or redirects" behaviour. Consider, for example, if we have a claim that has already been rejected because its payroll gender is missing. At the moment, the `reject_checked_claims` filter would redirect, and the filter chain would end. If we combine all the filters into a single method, then we'd end up doing a redirect in `reject_checked_claims`, and then a _second_ redirect in `reject_missing_payroll_gender`, which isn't allowed.

pezholio: We could do `redirect ... and return` which would work though right?

lawrence-dxw: No, that would just return from that method, it wouldn't break the filter chain.

app/views/admin/claims/show.html.erb
lawrence-dxw: @pezholio Cool! Didn't know that you could use I18n to handle counts. I've changed that and it looks neater.

@titlescreen Yeah, I'm definitely going to get some feedback from a content designer on this. I think that the message could be improved, for sure. It probably would be worth mentioning the playbook, although at the moment we don't really know how service operators are going to handle this situation so there's not going to be anything in the playbook 😄

pezholio: Yeah, it's super useful :)

app/models/claim.rb
lawrence-dxw: So, I think the best thing is if I explain what the code is _meant_ to be doing, and then I can understand whether you think that my approach is unnecessarily complicated, or whether I just need to figure out how to make the current implementation more understandable.

The aim of the method is the following: given a claim `c`, we want to find all the claims which:
- are payrollable;
- have the same National Insurance number as our claim;
- have a different value to `c` for at least one of the attributes in `Payment::PERSONAL_DETAILS_ATTRIBUTES_FORBIDDING_DISCREPANCIES`.

The first thing I had to decide was: given two claims, what's the best way of comparing their values for these attributes?
- I initially tried writing code like `.where.not(student_loan_plan_type: read_attribute(:student_loan_plan))`, but this didn't work because `read_attribute(:student_loan_plan)` returns a string `"plan_1"` but the database needs the value to be specified as the integer enum value. So instead of trying to figure out how to correctly re-serialize each attribute for insertion into the equality query, I decided to let the database to directly compare the two records. This is why I join the `claims` table to itself, so that I can compare the other claims to claim `c` (represented by the `this_claim` in the `JOIN`). I'm not very happy about the join, though.
- I wanted a comparision which treated `NULL` as "equal" to `NULL` (for the case of building society roll number, which is nullable), hence using `IS DISTINCT FROM` instead of `=`.

What do you think?

(And as for putting it into a separate class, happy to try that but the need may fall away if it turns out that the method can be hugely simplified.)

lawrence-dxw: Ah, and the `delete(:national_insurance_number)` is because when building up the list of inconsistent attributes to look for, we know that there's no point looking for National Insurance Number mismatch, because we use NI number to pick the records that we want to compare with, so there will never be a mismatch.

app/controllers/admin/claim_checks_controller.rb
pezholio: Oh yeah, I remember trying something similar and being massively confused 😆

lawrence-dxw:
@tekin, thanks for your feedback. I agree with everything you've said.

I've renamed:
- `inconsistent_claims` to `claims_preventing_payment`
- `inconsistent_attributes(other_claims)` to `attributes_preventing_combined_payment(other_claims)`

What do you think of the new naming? I've put it into separate commits, will squash if you're happy with it.

Here's a new wording that I've come up with, which I think makes things clearer.

Screenshot 2020-01-23 at 12 04 45

I'm going to run things by the people you suggested, too.

app/models/claim.rb
pezholio: I reckon it's looking a lot better now, but I still think it'd be best to break out into a seperate class, partly because there's quite a lot of logic there, and also partly because the Claim class is getting rather overloaded. How about something like this?

```ruby
  class ClaimsPreventingPaymentFinder
    attr_reader :claim

    def initialize(claim)
      @claim = claim
    end

    def claims
      @claims ||= begin
        Claim.payrollable
          .joins("INNER JOIN claims AS this_claim ON claims.national_insurance_number = this_claim.national_insurance_number \
              AND (#{distinct_attributes_predicate})")
          .where("this_claim.id = ?", claim.id)
      end
    end

    def attributes_preventing_combined_payment
      Payment::PERSONAL_DETAILS_ATTRIBUTES_FORBIDDING_DISCREPANCIES.select do |attribute|
        claims.any? { |claim| claim.read_attribute(attribute) != read_attribute(attribute) }
      end
    end

    private

    def distinct_attributes_predicate
      distinct_attributes_predicate = attributes_to_search_for.map { |attribute|
        "claims.#{attribute} IS DISTINCT FROM this_claim.#{attribute}"
      }.join(" OR ")
    end

    def attributes_to_search_for
      attributes = Payment::PERSONAL_DETAILS_ATTRIBUTES_FORBIDDING_DISCREPANCIES.dup
      attributes.delete(:national_insurance_number)
      attributes
    end
  end
```

lawrence-dxw: Moved it to a separate class, as suggested.

So, after getting lost in a failing test rabbit hole that turned out to be because factory_bot was silently failing to set up some associations, and thinking that this was because of the wacky SQL (which it turned out it wasn't), I took your advice and moved some of this comparison out of the SQL and into the code. I've added some justification to the commit message.

lawrence-dxw:
All comments addressed, just awaiting some feedback from the people that you mentioned, @tekin.

app/controllers/admin/claim_checks_controller.rb
```diff
@@ -22,6 +25,10 @@ def load_claim
     @matching_claims = Claim::MatchingAttributeFinder.new(@claim).matching_claims
   end
 
+  def claims_preventing_payment_finder
+    @claims_preventing_payment_finder ||= Claim::ClaimsPreventingPaymentFinder.new(@claim)
```
pezholio: Given this is used in all sorts of other places, I wonder if `claims_preventing_payment_finder` could be a method in the model?

app/models/claim/claims_preventing_payment_finder.rb
```diff
@@ -0,0 +1,29 @@
...
+    def claims_preventing_payment
+      @claims_preventing_payment ||= find_claims_preventing_payment
+    end
+
+    def find_claims_preventing_payment
+      payrollable_claims_from_same_claimant = Claim.payrollable.where(national_insurance_number: claim.national_insurance_number)
+
+      payrollable_claims_from_same_claimant.select do |other_claim|
+        Payment::PERSONAL_DETAILS_ATTRIBUTES_FORBIDDING_DISCREPANCIES.any? do |attribute|
+          other_claim.read_attribute(attribute) != claim.read_attribute(attribute)
```
pezholio: Given that this logic is used is two places, is it worth breaking this into a method? Something like:

```ruby
def attribute_does_not_match?(claim_to_compare, attribute)
  claim_to_compare.read_attribute(attribute) != claim.read_attribute(attribute)
end
```

spec/models/claim/claims_preventing_payment_finder_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+
+    context "when there is another claim with the same National Insurance number, with inconsistent personal details that would prevent us from running payroll" do
+      let(:inconsistent_personal_details) do
+        personal_details.merge(
+          bank_account_number: "87282828",
+          bank_sort_code: "388183",
+        )
+      end
+
+      context "when the other claim is not yet approved" do
```
pezholio: The embedded contexts here are a bit tricky to follow, do you reckon you could flatten these, so they look a bit like this?

```ruby
it "does not include the other claim when the other claim is not yet approved" do
  other_claim = create(:claim, :submitted, inconsistent_personal_details)
  expect(claims_preventing_payment).to be_empty
end
```

app/models/claim/claims_preventing_payment_finder.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    def initialize(claim)
+      @claim = claim
+    end
+
+    def claims_preventing_payment
+      @claims_preventing_payment ||= find_claims_preventing_payment
+    end
+
+    def find_claims_preventing_payment
```
pezholio: This could probably be a private method, as it doesn't get used outside the class

app/controllers/admin/claim_checks_controller.rb
lawrence-dxw: I'm not sure it's an appropriate thing to memoize on the model, since it can be invalidated by e.g. deleting another claim. So then it would just be a method on the model which instantiates this object and returns it, and I'm not sure it's worth it?

app/models/claim/claims_preventing_payment_finder.rb
lawrence-dxw: Nice spot!

app/models/claim/claims_preventing_payment_finder.rb
lawrence-dxw: Sure thing, will change.

spec/models/claim/claims_preventing_payment_finder_spec.rb
lawrence-dxw: Sure thing - changed.

app/controllers/admin/claim_checks_controller.rb
pezholio: Yeah, fair enough 👍

app/models/claim/claims_preventing_payment_finder.rb
```diff
@@ -0,0 +1,35 @@
+class Claim
+  class ClaimsPreventingPaymentFinder
```
tekin: I think it would be useful to future-reader if we start adding Tomdoc-style comments to classes like this to give the reader some context about what it is, what it does and where it's used without having to splunk through the codebase and commit history.

lawrence-dxw: Good idea, added.

lawrence-dxw:
I just spoke to Phil, Faye, and Rich about this, and we've agreed on a message (below). In particular it removes the list of inconsistent attributes because people felt that was too much information at the moment when the service operator can't resolve the situation anyway. So I've removed the related code.

image

PR #808, opened 2020-01-23
Put Maths and Physics in public beta

pezholio:
In preparation for public beta, this PR does the following:

- [x] Updates the "Claim submitted" page so that we link to the done page on GOV.UK
- [x] Removes basic auth from the service
- [x] Removes the temporary start page and redirects to the one on GOV.UK _(waiting to find out the URL of the permanent start page)_
- [x] Updates the Maths and Physics feedback form link


pezholio:
[Put Maths and Physics in public beta](https://trello.com/c/5Ea7RBWU/1205-put-maths-and-physics-in-public-beta)

app/models/maths_and_physics.rb
```diff
@@ -18,11 +18,11 @@ def notify_reply_to_id
   end
 
   def feedback_url
-    "https://docs.google.com/forms/d/e/1FAIpQLSeJcp50-eA5H_twIFKTUX8Z1ATDnj63wZaSlcnBCN-idJ7Ztg/viewform?usp=sf_link"
+    "https://docs.google.com/forms/d/e/1FAIpQLSfwPUxmNHqSnQ6RJ-0nedu5F2FRibBF5UIJ_EciTPcWQg581A/viewform"
   end
 
   def done_page_url
```
tekin: This method was only introduced because we wanted to have different behaviour between the two policies. Now that we've settled on the decision that there will only be a single done page for both/all policies, rather than changing this method we can go further and remove all the code that was introduced to split the logic: https://github.com/DFE-Digital/dfe-teachers-payment-service/pull/647/files

pezholio: I did wonder about that, given that there's likely to be other policies coming down the pipe, but there's no reason why this functionality can't be put back if/when that happens.

lawrence-dxw: My understanding of your reply is that you would remove the logic for having an optional done page, @pezholio. Is that correct? It seems to still be there.

app/helpers/application_helper.rb
```diff
@@ -32,7 +32,7 @@ def policy_description(policy)
   end
 
   def feedback_url
-    current_policy.feedback_url
+    "https://docs.google.com/forms/d/e/1FAIpQLSfwPUxmNHqSnQ6RJ-0nedu5F2FRibBF5UIJ_EciTPcWQg581A/viewform"
```
lawrence-dxw: I just tried this link and it goes to a form titled "Claim a payment for teaching maths or physics - feedback form". Are we sure that this is the right link?

app/models/maths_and_physics.rb
pezholio: Ah, I thought I'd fixed that

azure/resource_groups/app/parameters/development.template.json
```diff
@@ -84,22 +84,6 @@
         "secretName": "SecretKeyBase"
       }
     },
-    "BASIC_AUTH_USERNAME": {
```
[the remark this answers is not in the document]
pezholio: development.json is gitignore'd, so isn't present in the Github source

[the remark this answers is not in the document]
lawrence-dxw: Ah, sorry!

app/helpers/application_helper.rb
pezholio: I've asked in Slack

app/models/maths_and_physics.rb
tekin: This code is still here.

pezholio: Ach. Must be a bad rebase. I'll try again 🙄

app/controllers/static_pages_controller.rb
```diff
@@ -8,9 +8,6 @@ def contact_us
   def cookies
   end
 
-  def maths_and_physics
```
tekin: The actual view file can also be removed.

pezholio: Arrrrgh. Again, I thought I'd done this!

pezholio: Yeah, was another bad rebase, because we'd edited the view file recently

tekin: One habit I've got into is to always eyeball my PRs on GitHub after any changes/rebases just to make sure I've not missed anything obvious. It saves this sort of back and forth with a reviewer for stuff that I could/should have spotted.

PR #811, opened 2020-01-23
Changes to achieve AAA status

titlescreen:
The GDS design system targets AA status due to the complexities and practicalities of what some services are. This PR tweaks parts of the design system and some of the custom components that we have added to provide the necessary contrast for us to achieve AAA status. 🤞

The main focus is the contrast on small text (anything under 24px) and mainly involves selecting darker colours for button and links etc to provide that contrast. Whilst the hover state of the govuk-buttons has been changed there is no contrast requirement between normal, hover and disabled states.

Screenshot 2020-01-27 at 12 41 34

image

image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog]
 ## [Unreleased]
 
 - QLS questions now refer to academic year, instead of "1 September"
+- Styling tweaks to make the service AAA accessible. Accessiblity updated to
+  reflect AAA status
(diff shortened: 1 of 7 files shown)


tekin:
This looks fantastic, such a big achievement to get full AAA status! Plus thanks for making the changes in small atomic commits, it makes for a much easier review and captures a bunch of useful info for future devs. As discussed this needs looking at on Safari as it looks like the `:not` selector is throwing it off a bit.

PR #815, opened 2020-01-27
Auto email teachers 3 weeks after a claim is submitted

jacksonj04:
Send emails every morning for claims which have been submitted three weeks ago (to the day).

This does not include any logic for retrying emails should a job fail, on the grounds that the emails themselves are purely advisory and don't form an integral part of the claim flow.

### DIFF
--- a/app/jobs/send_emails_after_three_weeks_job.rb
+++ b/app/jobs/send_emails_after_three_weeks_job.rb
@@ -1,0 +1,20 @@
+# Runs every morning at 8am. Finds claims which are not yet checked and which
+# are three weeks old, and sends an email saying we are still in the process of
+# checking.
+
+class SendEmailsAfterThreeWeeksJob < CronJob
+  self.cron_expression = "0 8 * * *"
+
+  def perform
+    Rails.logger.info "Sending three-week emails to #{three_week_old_unchecked_claims.count} unchecked claims from the database"
+    three_week_old_unchecked_claims.each do |claim|
+      ClaimMailer.update_after_three_weeks(claim).deliver_later
+    end
+  end
+
+  private
+
+  def three_week_old_unchecked_claims
+    Claim.awaiting_checking.where(submitted_at: (21.days.ago.beginning_of_day...21.days.ago.end_of_day))
+  end
+end
--- a/app/mailers/claim_mailer.rb
+++ b/app/mailers/claim_mailer.rb
@@ -17,6 +17,11 @@ class ClaimMailer < ApplicationMailer
     view_mail_with_claim_and_subject(claim, "Your claim #{@claim_description} has been rejected, reference number: #{claim.reference}")
   end
 
+  def update_after_three_weeks(claim)
+    @claim_description = claim_description(claim)
+    view_mail_with_claim_and_subject(claim, "We are still reviewing your claim #{@claim_description}, reference number: #{claim.reference}")
+  end
+
   private
 
   def claim_description(claim)
--- a/app/views/claim_mailer/update_after_three_weeks.text.erb
+++ b/app/views/claim_mailer/update_after_three_weeks.text.erb
@@ -1,0 +1,11 @@
+Dear <%= @display_name %>,
+
+We're still reviewing your claim <%= @claim_description %>, which we received on <%= @claim.submitted_at %>. Claim reference: <%= @claim.reference %>.
+
+We'll notify you as your claim progresses. We do not expect any delays to the processing of your claim.
+
+You do not need to do anything.
+
+Regards,
+
+The Claim additional payments for teaching team
--- a/spec/jobs/send_emails_after_three_weeks_job_spec.rb
+++ b/spec/jobs/send_emails_after_three_weeks_job_spec.rb
@@ -1,0 +1,21 @@
+require "rails_helper"
+
+RSpec.describe SendEmailsAfterThreeWeeksJob do
+  before do
+    create_list(:claim, 2, :submitted, submitted_at: 21.days.ago)
+    create_list(:claim, 2, :submitted, submitted_at: 40.days.ago)
+    create_list(:claim, 2, :submitted, submitted_at: 3.days.ago)
+    create_list(:claim, 2, :approved, submitted_at: 21.days.ago)
+    create_list(:claim, 2)
+  end
+
+  describe "#perform" do
+    it "sends the expected number of emails" do
+      expect {
+        perform_enqueued_jobs do
+          SendEmailsAfterThreeWeeksJob.new.perform
+        end
+      }.to change { ActionMailer::Base.deliveries.count }.by(2)
+    end
+  end
+end
--- a/spec/mailers/claim_mailer_spec.rb
+++ b/spec/mailers/claim_mailer_spec.rb
@@ -74,6 +74,18 @@ RSpec.describe ClaimMailer, type: :mailer do
           expect(mail.body.encoded).to include("not been able to approve")
         end
       end
+
+      describe "#update_after_three_weeks" do
+        let(:claim) { build(:claim, :submitted, policy: policy) }
+        let(:mail) { ClaimMailer.update_after_three_weeks(claim) }
+
+        it_behaves_like "an email related to a claim", policy
+
+        it "mentions that the claim is still being reviewed in the subject and body" do
+          expect(mail.subject).to include("still reviewing your claim")
+          expect(mail.body.encoded).to include("still reviewing your claim")
+        end
+      end
     end
   end
 end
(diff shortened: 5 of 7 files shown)


app/models/claim.rb
```diff
@@ -246,6 +246,10 @@ def school
     eligibility&.current_school
   end
 
+  def days_since_submission
+    ((Time.now - submitted_at) / 1.day).to_i if submitted?
```
jacksonj04: I'm not entirely happy about this line, my gut says there should be a cleaner way to do it.

app/jobs/send_emails_after_three_weeks_job.rb
```diff
@@ -0,0 +1,20 @@
...
+  self.cron_expression = "0 4 * * *"
+
+  def perform
+    Rails.logger.info "Sending three-week emails to #{three_week_old_unchecked_claims.count} unsubmitted claims from the database"
+    three_week_old_unchecked_claims.each do |claim|
+      ClaimMailer.update_after_three_weeks(claim).deliver_later
+    end
+  end
+
+  # private
```
pezholio: This probably doesn't need to be commented out

app/jobs/send_emails_after_three_weeks_job.rb
```diff
@@ -0,0 +1,20 @@
+# Runs every morning at 4am. Finds claims which are not yet checked and which
+# are three weeks old on the date specified (defaulting to today), and sends an
+# email saying we are still in the process of checking.
+
+class SendEmailsAfterThreeWeeksJob < CronJob
+  self.cron_expression = "0 4 * * *"
```
pezholio: I think this might be better to be sent in the day, rather than at 4am

app/jobs/send_emails_after_three_weeks_job.rb
```diff
@@ -0,0 +1,20 @@
...
+    Rails.logger.info "Sending three-week emails to #{three_week_old_unchecked_claims.count} unsubmitted claims from the database"
+    three_week_old_unchecked_claims.each do |claim|
+      ClaimMailer.update_after_three_weeks(claim).deliver_later
+    end
+  end
+
+  # private
+
+  def three_week_old_unchecked_claims
+    Claim.awaiting_checking.select { |claim| claim.days_since_submission == 21 }
```
pezholio: You can probably do this via ActiveRecord, rather than filtering through all the claims. Something like:

```
Claim.awaiting_checking.where(submitted_at: (21.days.ago.beginning_of_day...21.days.ago.end_of_day))
```

This generates the following SQL:

```
"SELECT "claims".* FROM "claims" LEFT OUTER JOIN "checks" ON "checks"."claim_id" = "claims"."id" WHERE "claims"."submitted_at" IS NOT NULL AND "checks"."claim_id" IS NULL AND "claims"."submitted_at" >= '2020-01-06 00:00:00' AND "claims"."submitted_at" < '2020-01-06 23:59:59.999999'"
```

app/models/claim.rb
pezholio: See above, you probably don't need this. For the record though, the best way to get the number of days since a particular date would be this:

```ruby
(Date.today.to_date - submitted_at.to_date).to_i
```

app/views/claim_mailer/update_after_three_weeks.text.erb
```diff
@@ -0,0 +1,11 @@
...
+
+We're still reviewing your claim <%= @claim_description %>, which we received on <%= @claim.submitted_at %>. Claim reference: <%= @claim.reference %>.
+
+We'll notify you as your claim progresses. We do not expect any delays to the processing of your claim.
+
+You do not need to do anything.
+
+Regards,
+
+The claim additional payments for teaching team
```
pezholio: I'd probably capitalise `Claim Additional Payments for Teaching` here

spec/jobs/send_emails_after_three_weeks_job_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+  end
+
+  describe "#perform" do
+    it "sends the expected number of emails" do
+      expect {
+        perform_enqueued_jobs do
+          SendEmailsAfterThreeWeeksJob.new.perform
+        end
+      }.to change { ActionMailer::Base.deliveries.count }.by(2)
```
pezholio: 👍

app/mailers/claim_mailer.rb
```diff
@@ -17,6 +17,11 @@ def rejected(claim)
     view_mail_with_claim_and_subject(claim, "Your claim #{@claim_description} has been rejected, reference number: #{claim.reference}")
   end
 
+  def update_after_three_weeks(claim)
```
pezholio: I guess the one thing we might need to guard against here is if the mailer has been queued to send, but for one reason or another, the delivery has been significantly delayed (for example, maybe Notify has had a catastrophic outage) and the claim has since been approved. Might be worth adding a guard clause to check if the claim has been approved before actually sending the mail.

app/jobs/send_emails_after_three_weeks_job.rb
```diff
@@ -0,0 +1,20 @@
+# Runs every morning at 4am. Finds claims which are not yet checked and which
+# are three weeks old on the date specified (defaulting to today), and sends an
+# email saying we are still in the process of checking.
+
+class SendEmailsAfterThreeWeeksJob < CronJob
```
pezholio: I wondering if there might be a case where a user might potentially get two emails if this job gets requeued constantly, so it might be worth adding a boolean column to the claim that gets updated once the mail is sent?

app/views/claim_mailer/update_after_three_weeks.text.erb
jacksonj04: As per Slack chat, we should be "The Claim additional payments for teaching team"

app/jobs/send_emails_after_three_weeks_job.rb
tekin: I feel like adding a column to track the state of this email is perhaps overkill. If we think this job is likely to fail whilst in the `each` loop, resulting in multiple mailer jobs getting queued for a given claim when this job retries then another way to mitigate against that would be to wrap the entire thing in a transaction, that way the jobs get rolled-back if this job blows up. But I'd say it would only be worth introducing that complexity if we think that there is a real risk of the code falling over part-way through queueing all the mailer jobs.

app/mailers/claim_mailer.rb
jacksonj04: Not a bad shout

app/jobs/send_emails_after_three_weeks_job.rb
jacksonj04: I can't see any way the loop would fall over short of something fairly catastrophic. I think the main question is around how robust this needs to be - if it fails to run fully on one day do we _need_ to be able to pick up and send the missing emails, or do we shrug and go "they were only informational anyway"?

spec/mailers/claim_mailer_spec.rb
```diff
@@ -74,6 +74,18 @@
           expect(mail.body.encoded).to include("not been able to approve")
         end
       end
+
+      describe "#update_after_three_weeks" do
```
pezholio: Can we get a test to check that an approved claim doesn't do anything?

jacksonj04: Just looking into how to implement this test makes me wonder if the claim mailer is the right place for a guard to live? It feels like it's not the concern of the mailer to check the current state of the claim - it's been told to send an email, so it should.

Thoughts?

pezholio: Yeah, you're probably right. Actually, thinking about it, the guard probably doesn't make too much sense. We're queuing the emails at 8am, and they'll get rattled through pretty quickly, so the odds on someone checking and approving claims at this time are pretty slim

app/jobs/send_emails_after_three_weeks_job.rb
```diff
@@ -0,0 +1,20 @@
+# Runs every morning at 4am. Finds claims which are not yet checked and which
```
tekin: This needs updating to reflect the new run time.

app/jobs/send_emails_after_three_weeks_job.rb
```diff
@@ -0,0 +1,20 @@
...
+# are three weeks old, and sends an email saying we are still in the process of
+# checking.
+
+class SendEmailsAfterThreeWeeksJob < CronJob
+  self.cron_expression = "0 8 * * *"
+
+  def perform
+    Rails.logger.info "Sending three-week emails to #{three_week_old_unchecked_claims.count} unsubmitted claims from the database"
```
tekin: "unsubmitted"?

PR #817, opened 2020-01-28
Fix session timeout/CSRF-related issues

tekin:
Some fixes and tweaks to deal with some of the exception noise we've been seeing the last couple of days. Full details in each commit message.

### DIFF
--- a/spec/requests/sessions_spec.rb
+++ b/spec/requests/sessions_spec.rb
@@ -10,5 +10,30 @@ RSpec.describe "Sessions", type: :request do
         expect(response).to have_http_status(:ok)
       end
     end
+
+    it "does not extend an expired public user session" do
+      start_student_loans_claim
+
+      travel(2.hours) do
+        get refresh_session_path
+        expect(session[:claim_id]).to be_nil
+
+        get claim_path(StudentLoans.routing_name, "qts-year")
+        expect(response).to redirect_to(new_claim_path(StudentLoans.routing_name))
+      end
+    end
+
+    it "does not extend an expired admin session" do
+      sign_in_to_admin_with_role(DfeSignIn::User::SERVICE_OPERATOR_DFE_SIGN_IN_ROLE_CODE, "_user_id", "_org_id")
+
+      travel(2.hours) do
+        get refresh_session_path
+
+        expect(session[:user_id]).to be_nil
+
+        get admin_root_path
+        expect(response).to redirect_to(admin_sign_in_path)
+      end
+    end
   end
 end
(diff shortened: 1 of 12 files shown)


spec/requests/sessions_spec.rb
```diff
@@ -10,5 +10,30 @@
         expect(response).to have_http_status(:ok)
       end
     end
+
+    it "does not extend an expired public user session" do
+      start_student_loans_claim
+
+      travel(2.hours) do
+        get refresh_session_path
```
lawrence-dxw: Would we like to make an assertion about the response's status? (Same below)

lawrence-dxw: Since the response from this endpoint ([namely, whether a "success" or not](https://github.com/DFE-Digital/dfe-teachers-payment-service/blob/master/app/assets/javascripts/components/timeout_dialog.js.erb), which I _think_ is derived from the response's status) determines the behaviour of the client-side code which makes the request, would we like to make an assertion about the response's status here? Same in the other example.

app/controllers/application_controller.rb
```diff
@@ -29,4 +30,25 @@ def claim_session_timed_out?
   def claim_timeout_in_minutes
     self.class::CLAIM_TIMEOUT_LENGTH_IN_MINUTES
   end
+
+  def end_expired_admin_sessions
```
lawrence-dxw: I'm a bit uneasy about moving the logic for claim and admin session expiry into the `ApplicationController` superclass. It increases the method surface area of its subclasses, introducing methods that do not make sense for that subclass (for example, should `BasePublicController` have access to a method called `admin_session_timed_out?`?)

I don't know whether this is a worry for you, and if not then no problem. But one alternative would be to introduce concerns to contain the specific expiry methods (I think is the Rails thing, right?):

- `AdminSessionTimeout`: included in `BaseAdminController` and `SessionsController`;
- `ClaimSessionTimeout`: included in `BasePublicController` and `SessionsController`.

tekin: Fair point, having them separated out like that will reduce the chance of admin session-related methods being used in the public-facing part of the site and vice versa 👍🏽

spec/requests/sessions_spec.rb
tekin: I don't think the response is important ant enough to assert on here because this is very much an edge case and we wouldn't expect a user to make a request to the refresh endpoint under normal circumstances. The important thing is to be confident an exception does't get raised and cause noise in our monitoring (which is the underlying issue being resolved) and that the session is still expired.

lawrence-dxw: 👍

PR #821, opened 2020-01-29
Identify claimants by teacher reference instead of NI number

robbiepaul:
Because of the restrictions of payroll, we can only make one payment to
an individual in a given month. Therefor if a user claims for more than
one policy, we need a way to combine their claims into a single payment.
Previously this was done by matching claims with the same
National Insurance number (NiNo), this commit switches that matching to
be based on the claimant's teacher reference number (TRN):

- Historically, users enter their TRN more reliably than NiNo so there
  are few with errors, meaning fewer claims to have to correct
- There is no technically reason for us to be check the NiNo on claims
  during claim checking except because of this matching. If we switch to
  use the TRN instead, claim checkers can ignore the NiNo during claim
  checking, giving them one less thing to worry about (HMRC will deal
  with any errors in NiNo)
- We need to guarantee that the TRN is correct anyway as that will be
  the only identifiable field we will maintain for paid claims once the
  claim window is closed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog]
 ## [Unreleased]
 
 - Claim award amount is now sent to the Geckoboard dataset
+- Identify claimants by teacher reference number instead of NI number
 
 ## [Release 049] - 2020-01-29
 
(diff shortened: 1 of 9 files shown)


robbiepaul:
[Change Payroll run claim grouping to use TRN instead of NiNo](https://trello.com/c/N3vuNPO0/1258-change-payroll-run-claim-grouping-to-use-trn-instead-of-nino)

PR #826, opened 2020-01-30
Allow claimants to search by school postcode

robbiepaul:
The current school search only works if the search query you type matches exactly to part of the school's name. We've seen some queries come in to support where the user hasn't been able to find their school because the thing they've typed doesn't match exactly to the information in Get Information About Schools (eg. They omit an apostrophe).

This PR:

- allows users to search by typing the partial or full postcode of their school
- reduces the minimum character length of the search to 3 so the user doesn't have to remember the postcode sector

## Screenshot

Screenshot 2020-01-30 at 15 21 08

### DIFF
--- a/app/controllers/school_search_controller.rb
+++ b/app/controllers/school_search_controller.rb
@@ -21,7 +21,7 @@ class SchoolSearchController < BasePublicController
     rescue ArgumentError => e
       raise unless e.message == School::SEARCH_NOT_ENOUGH_CHARACTERS_ERROR
 
-      errors.push("'query' parameter must have a minimum of four characters")
+      errors.push(School::SEARCH_NOT_ENOUGH_CHARACTERS_ERROR)
     end
   end
 
--- a/app/views/claims/_school_search.html.erb
+++ b/app/views/claims/_school_search.html.erb
@@ -30,7 +30,7 @@
       </p>
     <% else %>
       <span class="govuk-hint" id="school_search-hint">
-        Enter the school name. Use at least four characters.
+        Enter the school name or school postcode. Use at least four characters.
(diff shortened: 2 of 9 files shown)


app/controllers/school_search_controller.rb
```diff
@@ -21,7 +21,7 @@ def search_schools
     rescue ArgumentError => e
       raise unless e.message == School::SEARCH_NOT_ENOUGH_CHARACTERS_ERROR
 
-      errors.push("'query' parameter must have a minimum of four characters")
+      errors.push("'query' parameter must have a minimum of three characters")
```
pezholio: Would it make sense to use the `SEARCH_NOT_ENOUGH_CHARACTERS_ERROR` here?

app/views/claims/_school_search.html.erb
```diff
@@ -30,7 +30,7 @@
       </p>
     <% else %>
       <span class="govuk-hint" id="school_search-hint">
-        Enter the school name. Use at least four characters.
+        Enter the school name or school postcode. Use at least three characters.
```
pezholio: I feel like `Enter the school name or postcode` flows better here

robbiepaul:
[Users can search by school name or postcode](https://trello.com/c/86wVE64O/1213-users-can-search-by-school-name-or-postcode)

app/controllers/school_search_controller.rb
robbiepaul: I've reworded `SEARCH_NOT_ENOUGH_CHARACTERS_ERROR` to make more sense outside the model and reused it here 👍

pezholio: Kewl 😃

PR #827, opened 2020-01-30
Stop blank attributes showing as matching

pezholio:
Previously, the `matching_attributes` helper was returning any attributes that were a blank string as 'matching'. This strips out any blank attributes before comparing.

### DIFF
--- a/spec/helpers/admin/claims_helper_spec.rb
+++ b/spec/helpers/admin/claims_helper_spec.rb
@@ -162,16 +162,26 @@ describe Admin::ClaimsHelper do
         building_society_roll_number: "123456789/ABCD",
       )
     }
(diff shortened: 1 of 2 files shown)


pezholio:
[Blank roll numbers shows as matching when there is another claim with matching attributes](https://trello.com/c/VXN46i7g/1232-blank-roll-numbers-shows-as-matching-when-there-is-another-claim-with-matching-attributes)

spec/helpers/admin/claims_helper_spec.rb
```diff
@@ -162,16 +162,26 @@
...
+      )
+    }
     subject { helper.matching_attributes(first_claim, second_claim) }
 
     it "returns the humanised names of the matching attributes" do
-      expect(subject).to eq(["Building society roll number", "National insurance number", "Teacher reference number"])
+      expect(subject).to eq(["Bank account number", "Bank sort code", "Building society roll number", "National insurance number", "Teacher reference number"])
     end
 
     it "does not consider nil building society roll number to be a match" do
```
robbiepaul:
```suggestion
     it "does not consider a blank building society roll number to be a match" do
```

PR #829, opened 2020-02-03
Add "approaching SLA" state to Geckoboard dataset

jacksonj04:
Replaces a boolean "has this claim passed its check deadline" with a string representing the state of the claim relative to the deadline. A claim can either be "ok", "warning" or "passed".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog]
 ## [Unreleased]
 
 - Allow claimants to search by school postcode
+- Replace "passed SLA deadline" true/false with ok/warning/passed status in
+  Geckoboard dataset
 
 ## [Release 050] - 2020-01-30
 
--- a/spec/models/claim/geckoboard_dataset_spec.rb
+++ b/spec/models/claim/geckoboard_dataset_spec.rb
@@ -2,10 +2,12 @@ require "rails_helper"
 
 RSpec.describe Claim::GeckoboardDataset, type: :model do
   let(:claim) { build(:claim, :submitted, created_at: DateTime.now) }
+  let(:dataset) { Claim::GeckoboardDataset.new }
 
   it "returns an award_amount of nil for a claim with no eligibility" do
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog]
 ## [Unreleased]
 
 - Allow claimants to search by school postcode
+- Replace the boolean "has this claim passed its check deadline" which is sent
+  to Geckoboard with a string representing the state of the claim relative to
+  the deadline
```
robbiepaul: I think the aim is to keep the changelog as something that everyone (including non-devs) can read, this might be too technical? Maybe we could use the PR title `Add "approaching SLA" state to Geckoboard dataset` or `Replace "passed SLA" with "approaching SLA" state to Geckoboard dataset`?

spec/models/claim/geckoboard_dataset_spec.rb
```diff
@@ -21,6 +21,27 @@
     end
   end
 
+  it "correctly reports a claim's SLA status when the claim is not approaching the check deadline" do
+    ClimateControl.modify ENVIRONMENT_NAME: "test" do
+      claim_sla_ok = build(:claim, :submitted, submitted_at: (Claim::CHECK_DEADLINE - Claim::CHECK_DEADLINE_WARNING_POINT - 1.week).ago)
+      expect(Claim::GeckoboardDataset.new(claims: [claim_sla_ok]).data_for_claims([claim_sla_ok]).first[:sla_status]).to eq("ok")
```
pezholio: This is a bit tricky to follow. Might be better to have the dataset initialized as a variable first?

```ruby
dataset = Claim::GeckoboardDataset.new # Given that we don't have to pass in the claims
data = dataset.data_for_claims([claim_sla_ok])

expect(data.first[:sla_status]).to eq("ok")
```

pezholio: You could even have the blank `dataset` set up in a `let` earlier in the tests to DRY things up

PR #831, opened 2020-02-03
Run database migrations as part of the deployment process

pezholio:
This creates a container instance that does nothing apart from run the migrations. This comes up before the worker and the app, so we no longer need to define the entrypoint in the Dockerfile, as the migrations have already run. This means that migrations will only ever run once per deploy.


pezholio:
[Run database migrations as part of the deployment process](https://trello.com/c/7IRsxf47/1173-run-database-migrations-as-part-of-the-deployment-process)

robbiepaul:
We may need to depend on the Postgres server deployment. Probably only applies on a fresh deploy but we should try to get it working anyway

pezholio:
Just to confirm, me and @robbiepaul have created a migration in test that takes two minutes to run and the infrastructure all came up cleanly! 🙌

Dockerfile
```diff
@@ -109,7 +109,6 @@ RUN if [ ${RAILS_ENV} = "production" ]; then \
 
 EXPOSE 3000
 
-ENTRYPOINT [ "bin/docker-entrypoint" ]
```
lawrence-dxw: For this commit ("Create a new container to run migrations") it would be good to explain in the commit message:

- What's the motivation? It says migrations now only run once per deploy but why is this a good thing, what were we experiencing before?
- The delay that we rely on that makes this approach work
- The worst-case scenario, where this delay isn't long enough and database preparation takes longer than it takes the worker and web containers to start up. Yesterday we were talking and you said it's not a problem because those containers would just keep failing with a pending migrations error and trying to restart. This is useful information for future devs.

pezholio:
@lawrence-dxw Yeah, there is an `az container create` command, and I did look into it, but I think that would complicate the deployment more, because we need to database to be created before we can run this command, and the database is created in the app deployment

lawrence-dxw:
Great, thanks for the explanation @pezholio

docker-compose.test.yml
```diff
@@ -20,7 +20,7 @@ services:
       args:
         RAILS_ENV: test
     image: local/dfe-teachers-payment-service:test
-    entrypoint: bin/docker-entrypoint bundle exec rake
+    entrypoint: bundle exec rake db:prepare default
```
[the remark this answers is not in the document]
pezholio: Done!

Dockerfile
pezholio: Yeah, that makes sense. Sorted 👍

bin/docker-entrypoint
```diff
@@ -1,24 +1,8 @@
...
-  fi
-fi
-
-if [ "$RAILS_ENV" != "test" ]; then
-  echo "Scheduling Cron Jobs"
-  bundle exec rake jobs:schedule
-fi
-
-exec "$@"
+echo "Scheduling jobs..."
```
lawrence-dxw: Sorry, missed this on a previous review but you've lost the `if [ "$RAILS_ENV" != "test" ];` here.

pezholio: Yeah, we don't need this as `bin/prepare-database` doesn't get used in test.

pezholio: (See https://github.com/DFE-Digital/dfe-teachers-payment-service/pull/831/files#diff-0d4080c652c491057176df93ca5768d1R23)

lawrence-dxw: In the context of the commit where you introduce the `db:setup_or_migrate` task, `bin/docker-entrypoint` is still being used in the test Rails env, so this check needs to remain there in that commit.

pezholio: Done! 👍

lawrence-dxw: Oh, I've realised that the same commit ("Make database setup command more robust") also accidentally loses the `exec "$@"`; could you restore that too please?

pezholio: Done! 👍

PR #838, opened 2020-02-05
Remove data-migrate gem

pezholio:
Once #831 is merged, we will no longer be running data migrations as part of the deployment process. This removes the `data-migrate` gem and all the old migrations, as well as adding a much simpler manual process for running data migrations, and updates the documentation.


lib/generators/data_migration/templates/data_migration.template
```diff
@@ -0,0 +1,11 @@
+class <%= @class_name %>
```
tekin: Because we don't have the same level of complexity around running these as normal schema migrations (up, down, rollback, etc) we don't need the extra ceremony and faff of defining a named class with an `up` method that gets called on the next line. We can literally have a `# put your Ruby code here` line after the `# Run me with` line and that is enough.

lib/generators/data_migration/data_migration_generator.rb
```diff
@@ -0,0 +1,14 @@
+class DataMigrationGenerator < Rails::Generators::NamedBase
+  source_root File.expand_path("templates", __dir__)
+
+  def copy_data_migration_file
+    @class_name = file_name.camelize
```
tekin: We probably don't need to create a class, see the comment on the template.

tekin:
`docker-entrypoint` will need updating too, it's currently calling the removed `db:schema:load:with_data` task.

pezholio:
Ah, fair point. I was working under the assumption that this would do in after #831, but looks like this might go in first

PR #840, opened 2020-02-05
Update admin navigation

tekin:
Updates the admin navigation to more closely follow the standard GDS Design System main navigation.

**Before:**
Screenshot 2020-02-05 at 14 40 28

**After:**
Screenshot 2020-02-05 at 14 42 35


app/views/application/_admin_menu.html.erb
```diff
@@ -0,0 +1,22 @@
...
+        <%= link_to "Search", search_admin_claims_path, class: "govuk-header__link" %>
+      </li>
+      <li class="govuk-header__navigation-item">
+        <%= link_to "Payroll", admin_payroll_runs_path, class: "govuk-header__link" %>
+      </li>
+      <li class="govuk-header__navigation-item">
+        <%= link_to "Open/close a service", admin_policy_configurations_path, class: "govuk-header__link" %>
+      </li>
+    <% end %>
+    <li class="govuk-header__navigation-item">
```
titlescreen: Should this be in the if statement? Can they sign out if they are not signed in? 🤔

tekin: No, we always show the "sign out" link for a signed in user, but we only show the other links if the user is signed in as a "service operator". e.g. Cantium users don't get to see those links but should be able to sign-out still.

PR #841, opened 2020-02-05
Update "Some of this information is wrong" content on the verified page

robbiepaul:
Screenshot 2020-02-05 at 15 26 07

### DIFF
--- a/app/views/claims/verified.html.erb
+++ b/app/views/claims/verified.html.erb
@@ -21,9 +21,14 @@
         </span>
       </summary>
       <div class="govuk-details__text">
-        If you find that some of this information is incorrect please contact support at
-        <%= mail_to support_email_address(current_policy_routing_name), support_email_address(current_policy_routing_name), class: "govuk-link" %>
-        as this could affect your application.
+        <p class="govuk-body">
+          If this information is out of date you do not need to tell us. This will not affect your claim.
+        </p>
+        <p class="govuk-body">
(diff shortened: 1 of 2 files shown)


robbiepaul:
[Update "Some of this information is wrong" content on /verified page](https://trello.com/c/WqGUvFS7/1286-update-some-of-this-information-is-wrong-content-on-verified-page)

app/views/claims/verified.html.erb
```diff
@@ -21,9 +21,14 @@
...
       <div class="govuk-details__text">
-        If you find that some of this information is incorrect please contact support at
-        <%= mail_to support_email_address(current_policy_routing_name), support_email_address(current_policy_routing_name), class: "govuk-link" %>
-        as this could affect your application.
+        <p class="govuk-body">
+          If this information is out of date you do not need to tell us. This will not affect your claim.
+        </p>
+        <p class="govuk-body">
+          If you do not recognise some of these details please let us know by emailing
+          <%= mail_to support_email_address(current_policy_routing_name), support_email_address(current_policy_routing_name), class: "govuk-link" %>.
```
pezholio: I don't think you need the second argument if you want to show the email address as the link text do you?

robbiepaul: Yeah, you're right. If the second argument is `nil` it will use the link as the text. Updated

PR #844, opened 2020-02-06
Automatically remove PII from claims

jacksonj04:
Claims will now have PII automatically removed from the database approximately two months after they are either rejected or paid.

Admins who view a claim which has had PII removed will be informed of the change to the claim.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ The format is based on [Keep a Changelog]
 - Service operators can change the academic year a policy is accepting claims
   for
 - Service operators can view qualification and employment checks
+- PII is automatically deleted from claims which are either rejected and have
+  had no update in the last two months, or which have been paid more than two
+  months ago
 
 ## [Release 053] - 2020-02-12
 
--- a/app/helpers/admin/claims_helper.rb
+++ b/app/helpers/admin/claims_helper.rb
@@ -10,13 +10,17 @@ module Admin
       claim.policy::EligibilityAdminAnswersPresenter.new(claim.eligibility).answers
     end
 
+    def pii_removed_text
+      content_tag(:span, "Removed", class: "capt-text-quiet")
+    end
+
     def admin_personal_details(claim)
       [
         [t("admin.teacher_reference_number"), claim.teacher_reference_number],
-        [t("govuk_verify_fields.full_name").capitalize, claim.full_name],
-        [t("govuk_verify_fields.date_of_birth").capitalize, l(claim.date_of_birth, format: :day_month_year)],
-        [t("admin.national_insurance_number"), claim.national_insurance_number],
-        [t("govuk_verify_fields.address").capitalize, sanitize(claim.address("<br>").html_safe, tags: %w[br])],
+        [t("govuk_verify_fields.full_name").capitalize, claim.pii_removed? ? pii_removed_text : claim.full_name],
+        [t("govuk_verify_fields.date_of_birth").capitalize, claim.pii_removed? ? pii_removed_text : l(claim.date_of_birth, format: :day_month_year)],
+        [t("admin.national_insurance_number"), claim.pii_removed? ? pii_removed_text : claim.national_insurance_number],
+        [t("govuk_verify_fields.address").capitalize, claim.pii_removed? ? pii_removed_text : sanitize(claim.address("<br>").html_safe, tags: %w[br])],
         [t("admin.email_address"), claim.email_address],
       ]
     end
--- a/app/jobs/delete_pii_from_old_claims_job.rb
+++ b/app/jobs/delete_pii_from_old_claims_job.rb
@@ -1,0 +1,10 @@
+# Runs weekly at 00:30 on a Sunday morning, and instructs the PiiScrubber class
+# to remove PII from eligible claims.
+class DeletePiiFromOldClaimsJob < CronJob
+  self.cron_expression = "30 0 * * 0"
+
+  def perform
+    Rails.logger.info "Deleting PII from old claims which have been rejected or paid"
+    Claim::PiiScrubber.new.scrub_completed_claims
+  end
+end
--- a/app/models/claim/pii_scrubber.rb
+++ b/app/models/claim/pii_scrubber.rb
@@ -1,0 +1,51 @@
+# Removes personally identifiable information from a claim where the claim is
+# either rejected and hasn't been updated in over two months, or where the claim
+# was scheduled to be paid more than two months ago.
+#
+# Attributes are set to nil, and pii_removed_at is set to the current timestamp.
+
+class Claim
+  class PiiScrubber
+    PII_ATTRIBUTES_TO_DELETE = [
+      :first_name,
+      :middle_name,
+      :surname,
+      :date_of_birth,
+      :address_line_1,
+      :address_line_2,
+      :address_line_3,
+      :address_line_4,
+      :postcode,
+      :payroll_gender,
+      :national_insurance_number,
+      :bank_sort_code,
+      :bank_account_number,
+      :building_society_roll_number,
+    ]
+
+    MINIMUM_DATE_OF_LAST_ACTION = 2.months.ago
+
+    def scrub_completed_claims
+      old_claims_rejected_or_paid.update_all(attribute_values_to_set)
+    end
+
+    private
+
+    def attribute_values_to_set
+      PII_ATTRIBUTES_TO_DELETE.map { |attr| [attr, nil] }.to_h.merge(
+        pii_removed_at: Time.zone.now
+      )
+    end
+
+    def old_claims_rejected_or_paid
+      Claim.left_outer_joins(payment: [:payroll_run])
+        .joins(:decision)
+        .where(pii_removed_at: nil)
+        .where(
+          "(decisions.result = :rejected AND decisions.created_at < :minimum_time) OR scheduled_payment_date < :minimum_time",
+          minimum_time: MINIMUM_DATE_OF_LAST_ACTION,
+          rejected: Decision.results.fetch(:rejected)
+        )
+    end
+  end
+end
--- a/app/views/admin/claims/show.html.erb
+++ b/app/views/admin/claims/show.html.erb
@@ -10,6 +10,12 @@
       </div>
     <% end %>
 
+    <% if @claim.pii_removed? %>
+      <div class="govuk-body-l govuk-flash__notice">
+        This claim had personally identifiable information removed on <%= l(@claim.pii_removed_at.to_date) %>.
+      </div>
+    <% end %>
+
     <% if @matching_claims.any? %>
       <div class="govuk-body-l govuk-flash__warning">
         <a href="#claims-with-matches">Details in this claim match those in other claims</a>
--- a/spec/factories/claims.rb
+++ b/spec/factories/claims.rb
@@ -69,5 +69,24 @@ FactoryBot.define do
 
       eligibility_factory { ["#{policy.to_s.underscore}_eligibility".to_sym, :ineligible] }
     end
+
+    trait :pii_removed do
+      submitted
+      first_name { nil }
+      middle_name { nil }
+      surname { nil }
+      date_of_birth { nil }
+      address_line_1 { nil }
+      address_line_2 { nil }
+      address_line_3 { nil }
+      address_line_4 { nil }
+      postcode { nil }
+      payroll_gender { nil }
+      national_insurance_number { nil }
+      bank_sort_code { nil }
+      bank_account_number { nil }
+      building_society_roll_number { nil }
+      pii_removed_at { Time.zone.now }
+    end
   end
 end
--- a/spec/features/admin_claim_check_spec.rb
+++ b/spec/features/admin_claim_check_spec.rb
@@ -105,6 +105,19 @@ RSpec.feature "Admin checks a claim" do
       expect(page).to have_content(user.full_name)
     end
 
+    context "When the claim has had PII removed" do
+      let!(:claim_with_pii_removed) { create(:claim, :rejected, :pii_removed) }
+
+      scenario "User can see where a claim has had PII removed" do
+        visit admin_claim_path(claim_with_pii_removed)
+        expect(page).to have_content("personally identifiable information removed")
+        expect(page).to have_content("Full name Removed")
+        expect(page).to have_content("Date of birth Removed")
+        expect(page).to have_content("National Insurance number Removed")
+        expect(page).to have_content("Address Removed")
+      end
+    end
+
     context "When the payroll gender is missing" do
       let!(:claim_missing_payroll_gender) { create(:claim, :submitted, payroll_gender: :dont_know) }
 
--- a/spec/jobs/delete_pii_from_old_claims_job_spec.rb
+++ b/spec/jobs/delete_pii_from_old_claims_job_spec.rb
@@ -1,0 +1,21 @@
+require "rails_helper"
+
+RSpec.describe DeletePiiFromOldClaimsJob do
+  describe "#perform" do
+    let(:over_two_months_ago) { 2.months.ago - 1.day }
+
+    it "deletes the PII from eligible claims" do
+      submitted_claim = create(:claim, :submitted)
+      rejected_claim = create(:claim, :submitted)
+      create(:decision, :rejected, claim: rejected_claim, created_at: over_two_months_ago)
+      paid_claim = create(:claim, :approved)
+      create(:payment, :with_figures, claims: [paid_claim], scheduled_payment_date: over_two_months_ago)
+
+      DeletePiiFromOldClaimsJob.new.perform
+
+      expect(Claim.find(submitted_claim.id).pii_removed_at).to be_nil
+      expect(Claim.find(rejected_claim.id).pii_removed_at).to_not be_nil
+      expect(Claim.find(paid_claim.id).pii_removed_at).to_not be_nil
+    end
+  end
+end
--- a/spec/models/claim/pii_scrubber_spec.rb
+++ b/spec/models/claim/pii_scrubber_spec.rb
@@ -1,0 +1,82 @@
+require "rails_helper"
+
+RSpec.describe Claim::PiiScrubber, type: :model do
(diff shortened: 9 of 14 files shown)


app/jobs/delete_pii_from_old_claims_job.rb
```diff
@@ -0,0 +1,20 @@
...
+  MINIMUM_TIME_SINCE_LAST_UPDATE = 2.months.ago
+
+  def perform
+    Rails.logger.info "Deleting PII from #{claims_eligible_to_delete.count} claims"
+    Claim::Cleaner.new(claims_eligible_to_delete).delete_pii
+  end
+
+  private
+
+  def claims_eligible_to_delete
```
pezholio: I'd prefer to see this in one query, that way it returns an `ActiveRecord_Relation`, which we can then chain additional queries onto. I've had a bit of a play and this seems to do the trick:

```ruby
Claim.left_outer_joins(payment: [:payroll_run])
  .joins(:decision)
  .where(
    "(decisions.result = 1 AND claims.updated_at < :minumum_time) OR scheduled_payment_date < :minumum_time",
    minumum_time: MINIMUM_TIME_SINCE_LAST_UPDATE
  )
```

pezholio: It might be worth giving a more descriptive name for the method too - something like `old_rejected_or_paid_claims`?

app/models/claim/cleaner.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    private
+
+    def attribute_values_to_set
+      PII_ATTRIBUTES_TO_DELETE.map { |attr| [attr, nil] }.to_h.merge(
+        pii_removed_at: Time.now
+      )
+    end
+
+    def safe_to_clean(claim)
```
pezholio: Do we need to do this, given the only claims we're passing into the class are rejected or paid?

app/views/admin/claims/show.html.erb
```diff
@@ -4,12 +4,18 @@
   <div class="govuk-grid-column-two-thirds">
     <%= render("shared/error_summary", instance: @decision, errored_field_id_overrides: { "result": "decision_result_approved" }) if @decision.errors.any? %>
 
-    <% if @claim.payroll_gender_missing? %>
+    <% if @claim.payroll_gender_missing? and not @claim.pii_removed? %>
```
pezholio: I find it a bit more "Rubyish" to use a `!` rather than `not` when checking the reverse:

```suggestion
    <% if @claim.payroll_gender_missing? and !@claim.pii_removed? %>
```

spec/features/admin_claim_check_spec.rb
```diff
@@ -80,6 +80,15 @@
       expect(page).to have_content(user.full_name)
     end
 
+    context "When the claim has had PII removed" do
+      let!(:claim_with_pii_removed) { create(:claim, :rejected, :pii_removed) }
+
+      scenario "User can see when a claim has had PII removed" do
```
pezholio: Might be worth checking that the payroll gender stuff doesn't get mentioned in this scenario too

app/models/claim/cleaner.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    def initialize(claims)
+      claims.each do |claim|
+        safe_to_clean(claim)
+      end
+      @claims = claims
+    end
+
+    def delete_pii
+      Claim.where(id: @claims.pluck(:id)).update_all(attribute_values_to_set)
```
pezholio: If we pass an `ActiveRecord_Relation` into the initializer, rather than an array, we can just do `@claims.update_all...`

jacksonj04:
@pezholio This is a bit of a weird one as far as where the actual bits of logic go. I went with the current separation because saying "clean up these claims of PII" seemed to belong to the domain of the model, but deciding which claims should be cleaned felt a bit higher level. I don't feel strongly about it though.

app/models/claim/cleaner.rb
jacksonj04: This was to act as a sanity check in case something inadvertently passed an unsafe claim in, given that it's a (very) destructive operation.

pezholio:
I think the usual pattern we've had in the past is that we've kept the jobs quite thin, and have most of the logic in the job, so we could potentially have a `DeletePiiFromOldClaims` class and a corresponsing `DeletePiiFromOldClaimsJob` that just calls the class.

app/models/claim/cleaner.rb
pezholio: Actually, I'm wondering about moving this to the `Claim` model itself, so each `Claim` would have a `delete_pii!` method, then this class would be a very thin wrapper that loops through all the claims and calls `delete_pii!` on each.

jacksonj04: The only issue with that is that this is likely to impact a lot of claims simultaneously (ie every claim paid out in a cycle will all become eligible for cleaning at the same time), so doing an `update_all` saves running a DB query for each one.

I'm not totally convinced of the merits of doing that vs the method you suggest here though, which was my first thought.

app/jobs/delete_pii_from_old_claims_job.rb
tekin: Have a look if you can make use of the [`or` method](https://blog.bigbinary.com/2016/05/30/rails-5-adds-or-support-in-active-record.html) to chain this such that we don't need to have a hardcode the enum mapping ("decisions.result = 1") and instead pass in the `Claim.rejected` scope.

app/views/admin/claims/show.html.erb
tekin: Note that this is using the control-flow `and` and not the logical operator `&&` here. Stick with the logical operators `&&` and `||` in conditional statements. See http://www.virtuouscode.com/2010/08/02/using-and-and-or-in-ruby/ for more info on the different. TBH I'm surprised that our Ruby standard checks allowed this, I thought they disallowed them...

app/jobs/delete_pii_from_old_claims_job.rb
tekin: We'll also want to exclude claims that have already had their PII removed from the relation right?

pezholio: I tried using `or`, and this resulted in an `ArgumentError (Relation passed to #or must be structurally compatible. Incompatible values: [:left_outer_joins])`

spec/features/admin_claim_check_spec.rb
jacksonj04: It might actually make more sense to remove the payroll gender warning entirely if a claim is rejected - at that point an operator cannot take any further action, so why warn them?

app/models/claim/cleaner.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    private
+
+    def attribute_values_to_set
+      PII_ATTRIBUTES_TO_DELETE.map { |attr| [attr, nil] }.to_h.merge(
+        pii_removed_at: Time.now
+      )
+    end
+
+    def old_claims_rejected_or_paid
```
pezholio: I'm wondering if, with this, we could be more specific about how we fetch the results, to guard against any changes being made to the `Decision`s enum. Something like:

```ruby
Claim.left_outer_joins(payment: [:payroll_run])
  .joins(:decision)
  .where(pii_removed_at: nil)
  .where(
    "(decisions.result = :rejected AND claims.updated_at < :minimum_time) OR scheduled_payment_date < :minimum_time",
    minimum_time: MINIMUM_DATE_OF_LAST_ACTION,
    rejected: Decision.results.fetch(:rejected)
  )
```

(using `fetch` here would raise an error if we change the name of the `rejected` enum, and alert us that there was something wrong)

spec/models/claim/cleaner_spec.rb
```diff
@@ -0,0 +1,75 @@
+require "rails_helper"
+
+RSpec.describe Claim::Cleaner, type: :model do
+
```
pezholio: I think Standard might complain about this linebreak

spec/models/claim/cleaner_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    expect(cleaned_claim.address_line_4).to be_nil
+    expect(cleaned_claim.postcode).to be_nil
+    expect(cleaned_claim.payroll_gender).to be_nil
+    expect(cleaned_claim.national_insurance_number).to be_nil
+    expect(cleaned_claim.bank_sort_code).to be_nil
+    expect(cleaned_claim.bank_account_number).to be_nil
+    expect(cleaned_claim.building_society_roll_number).to be_nil
+  end
+
+  it "records when the claim had PII removed" do
```
pezholio: I'd probably merge this test with the previous one as we're testing the same part of the code

spec/jobs/delete_pii_from_old_claims_job_spec.rb
```diff
@@ -0,0 +1,19 @@
+require "rails_helper"
+
+RSpec.describe DeletePiiFromOldClaimsJob do
+  describe "#perform" do
+    let(:four_hours_ago) { 4.hours.ago }
+    let(:over_two_months_ago) { 2.months.ago - 1.day }
+
+    it "deletes the PII from eligible claims" do
+      rejected_claim = create(:claim, :rejected, updated_at: over_two_months_ago)
```
pezholio: It might be clearer to create a claim, and then use `travel_to` to go over two months into the future, rather than mutating the `updated_at` variable like that

spec/factories/claims.rb
```diff
@@ -68,5 +68,10 @@
 
       eligibility_factory { ["#{policy.to_s.underscore}_eligibility".to_sym, :ineligible] }
     end
+
+    trait :pii_removed do
```
tekin: To make this trait realistic, can we `nil` out the details that should be blank?

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- PII is automatically deleted from claims which are either rejected and have
+  had no update in the last two months, or which have been paid more than two
+  months ago.
```
tekin: Conventionally we do not put fullstops on these entries unless they are more than one sentence.

app/assets/stylesheets/components/text-quiet.scss
```diff
@@ -0,0 +1,3 @@
+.text__quiet {
+  color: govuk-colour("dark-grey");
```
tekin: Can we check with @titlescreen to see if there isn existing bit of CSS in the design system we can use, rather than introducing our own custom one?

app/helpers/admin/claims_helper.rb
```diff
@@ -10,13 +10,17 @@ def admin_eligibility_answers(claim)
       claim.policy::EligibilityAdminAnswersPresenter.new(claim.eligibility).answers
     end
 
+    def pii_removed_text
+      content_tag(:span, "Removed", class: "text__quiet")
+    end
+
     def admin_personal_details(claim)
```
tekin: We have no test coverage for this new behaviour. Worth having some to give us confidence this helper doesn't fall over for a scrubbed claim?

app/jobs/delete_pii_from_old_claims_job.rb
```diff
@@ -0,0 +1,11 @@
+# Runs weekly at 00:30 on a Sunday morning, and requests that eligible claims
+# (rejected or paid, and more than 60 days since their last update) have their
```
tekin: This comment description now talks about behaviour that the job has no responsibility or knowledge of, i.e. what gets scrubbed. Let's move that info to the file for `Claim::Cleaner`.

app/models/claim/cleaner.rb
```diff
@@ -0,0 +1,46 @@
+class Claim
+  class Cleaner
```
tekin: Let's add some docs to this class to help future developers.

spec/features/admin_claim_check_spec.rb
```diff
@@ -80,6 +80,15 @@
...
+        visit admin_claim_path(claim_with_pii_removed)
+        expect(page).to have_content("personally identifiable information removed")
```
tekin: Can we also verify that at least one of the scrubbed fields is listed as `"Removed"` .

spec/jobs/delete_pii_from_old_claims_job_spec.rb
tekin: I prefer to set an explicit `updated_at` as it is less cognitive load than than wrapping multiple lines inside time travel blocks.

spec/models/claim/cleaner_spec.rb
```diff
@@ -0,0 +1,75 @@
+require "rails_helper"
+
+RSpec.describe Claim::Cleaner, type: :model do
+
+  let(:over_two_months_ago) { 2.months.ago - 1.day }
+
+  it "does not delete details from a newly submitted claim" do
```
tekin: I think this spec and the spec below should create claims that are old enough to qualify for scrubbing, otherwise we are missing coverage. We should also have specs verifying that rejected and paid claims do not get scrubbed if they are not old enough.

app/models/claim/cleaner.rb
tekin: On reflection `Cleaner` feels a bit too generic. How about something more specific to communicate the behaviour more to the reader. Something like:

```ruby
Claim::PiiScrubber.scrub_completed_claims
```

We can then define "completed" as being rejected and paid claims that have not been modified in 2 months.

app/assets/stylesheets/components/text-quiet.scss
jacksonj04: I did, and (unless he misunderstood my question) there isn't. I've put a shout in X-Gov Slack as well.

spec/models/claim/cleaner_spec.rb
jacksonj04: I think it's describing two separate behaviours (one that the fields are actually nilled, the second that it records when that happened). The fact the class does that in one update is an implementation detail.

spec/models/claim/pii_scrubber_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+    claim = create(:claim, :submitted, updated_at: over_two_months_ago)
+
+    Claim::PiiScrubber.new.scrub_completed_claims
+    cleaned_claim = Claim.find(claim.id)
+
+    expect(cleaned_claim.attributes).to eq(claim.attributes)
+  end
+
+  it "does not delete details from an approved but unpaid claim" do
+    freeze_time
```
pezholio: Might be worth wrapping `freeze_time` in a block here, just to make sure the time unfreezes at the end of the spec (https://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html#method-i-freeze_time)

app/models/claim/pii_scrubber.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    def scrub_completed_claims
+      old_claims_rejected_or_paid.update_all(attribute_values_to_set)
+    end
+
+    private
+
+    def attribute_values_to_set
+      PII_ATTRIBUTES_TO_DELETE.map { |attr| [attr, nil] }.to_h.merge(
+        pii_removed_at: Time.now
```
tekin: `Time.now` returns the system time. Generally in app code we want to use `Time.zone.now` so that the timestamp accounts for the timezone configured in the application.

app/models/claim/pii_scrubber.rb
```diff
@@ -0,0 +1,51 @@
...
+      )
+    end
+
+    def old_claims_rejected_or_paid
+      Claim.left_outer_joins(payment: [:payroll_run])
+        .joins(:decision)
+        .where(pii_removed_at: nil)
+        .where(
+          "(decisions.result = :rejected AND claims.updated_at < :minimum_time) OR scheduled_payment_date < :minimum_time",
```
tekin:
```suggestion
          "(decisions.result = :rejected AND decisions.created_at < :minimum_time) OR scheduled_payment_date < :minimum_time",
```

spec/models/claim/pii_scrubber_spec.rb
```diff
@@ -0,0 +1,102 @@
+require "rails_helper"
+
+RSpec.describe Claim::PiiScrubber, type: :model do
+  let(:over_two_months_ago) { 2.months.ago - 1.day }
+
+  it "does not delete details from a submitted claim" do
+    freeze_time do
```
tekin: Are these `freeze_time` calls no these "not delete" specs entirely necessary?

spec/models/claim/pii_scrubber_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+      claim = create(:claim, :submitted, updated_at: over_two_months_ago)
+
+      Claim::PiiScrubber.new.scrub_completed_claims
+      cleaned_claim = Claim.find(claim.id)
+
+      expect(cleaned_claim.attributes).to eq(claim.attributes)
```
tekin: We could use the Rspec change matcher to rewrite these "not delete" specs like this:

```ruby
  expect { Claim::PiiScrubber.new.scrub_completed_claims }.not_to change { claim.reload.attributes }
```

spec/models/claim/pii_scrubber_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+    end
+  end
+
+  it "deletes details from an old rejected claim" do
+    claim = create(:claim, :rejected, updated_at: over_two_months_ago)
+
+    Claim::PiiScrubber.new.scrub_completed_claims
+    cleaned_claim = Claim.find(claim.id)
+
+    expect(cleaned_claim.attributes).to_not eq(claim.attributes)
```
tekin: This feels like a pretty weak assertion as any number of things might cause the attributes to be different. Instead I think I'd prefer to see the assertions from the `"deletes the fields we expect"` and the one on the timestamp in the bottom spec duplicated in this spec and the one for old paid claims so we end up with two combined `it "scrubs the PII fields on BLAH claims, setting a pii_removed_at timstamp"` examples that define what happens to specific types of claims in their entirety. It creates some duplication, but a little duplication is acceptable in specs in the service of making the specs more robust and easier to follow.

PR #851, opened 2020-02-10
Rename Check model to Decision

tekin:
We are about to embark on some iterations on the claim checking process that will involve breaking down the process into individual "checks", for example checking the claimant's qualification details against DQT data. Therefor it no longer makes sense to have a claim-wide "check" model. Instead, we rename it to "decision" so that it is easier to distinguish between the individual checks and the overall decision on the outcome of the claim. This also makes the model match the language used in the admin UI.
 
We'll need to ensure that claim checkers are not in the process of approving/rejecting claims when we release this, but otherwise the rename should be releasable without causing any disruption to the service.

### Follow-up work

The data we send to Geckoboard using `Claim::GeckoboardDataset` still uses the word "check". We held back from changing that as part of this PR as renaming the Geckoboard data fields will involve additional work to update the dashboards themselves, and we felt it was best if that was tackled separately.


db/schema.rb
```diff
@@ -68,6 +55,19 @@
     t.index ["submitted_at"], name: "index_claims_on_submitted_at"
   end
 
+  create_table "decisions", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.integer "result"
+    t.string "checked_by"
```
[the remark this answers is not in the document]
tekin: Yep, this and a couple more that were part of the story for showing the name of users performing actions – https://trello.com/c/fr9sNCXk/644-display-the-name-of-the-service-operator-who-took-an-action

We're doing that as a separate PR though as this one is large enough as it is.

db/schema.rb
```diff
@@ -68,6 +55,19 @@
...
+    t.uuid "claim_id"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.text "notes"
+    t.uuid "made_by_id"
```
lawrence-dxw: `created_by` would be consistent with the naming we use on `PayrollRun`. I think we should be consistent, given that the meaning of "created" and "made" is pretty much the same.

robbiepaul: I've updated our to be consistent with `PayrollRun`.

Part of me wonders whether we should use "Created by" on the UI or not, it might not be immediately understandable to end-users. However it's just an assumption so I've gone with the same as the payroll run UI for the time being:

Screenshot 2020-02-11 at 12 12 37

lawrence-dxw: I would say that the previous text ("Checked by") probably made more sense; do you think we need to change it?

robbiepaul: Because we're implementing this new concept of multiple checks that can be carried out by various different users I think we'd want to avoid "Checked by" for the decision. I think what we have now is ok, or perhaps "Decided by" but we can always iterate it later.

lawrence-dxw: Ah, good point. Cool, looks good to me then

PR #852, opened 2020-02-10
Don't share the last seen at session variables

pezholio:
We picked up a bug in development, where we noticed that it was possible to extend an admin session by visiting a page on the public side of the site.

The seperates out the management of `last_seen_at` into two parts - one to handle admins, and one to handle public users. When a user carries out an action on the public side, this updates `session[:last_seen_at]` and when a user carries out an action on the admin side, this updates `session[:admin_last_seen_at]`.

I've also seperated out the `SessionsController`s, which are used when a user using Javascript opts to extend their session. This means that both the admin and application layouts now pass an additional`refresh_session_path` variable, which is sent to the JS via data attributes, and tells the frontend code which URL to send a request to.

### DIFF
--- a/app/assets/javascripts/components/timeout_dialog.js
+++ b/app/assets/javascripts/components/timeout_dialog.js
@@ -8,6 +8,7 @@
   var continueButtonElement = document.getElementById("continue_session");
   var warningTimeInMinutes = dialogElement.getAttribute("data-warning-in-minutes");
   var timeoutInMinutes = dialogElement.getAttribute("data-timeout-in-minutes");
+  var refreshSessionPath = dialogElement.getAttribute("data-refresh-session-path");
   var timeoutInMilliseconds = (timeoutInMinutes - warningTimeInMinutes) * 60 * 1000;
   var timeoutPath = dialogElement.getAttribute("data-timeout-path");
   var dialog = new window.A11yDialog(dialogElement);
@@ -23,7 +24,7 @@
 
   function refreshSession() {
     Rails.ajax({
-      url: "<%= refresh_session_path %>",
+      url: refreshSessionPath,
       type: "get",
       success: function() {
         clearInterval(window.sessionTimeoutTimer);
--- a/app/controllers/admin/sessions_controller.rb
+++ b/app/controllers/admin/sessions_controller.rb
@@ -1,0 +1,9 @@
+module Admin
+  class SessionsController < BaseAdminController
+    def refresh
+      end_expired_admin_sessions
+
+      head :ok
+    end
+  end
+end
--- a/app/controllers/concerns/admin_session_timeout.rb
+++ b/app/controllers/concerns/admin_session_timeout.rb
@@ -15,7 +15,7 @@ module AdminSessionTimeout
   end
 
   def admin_session_timed_out?
-    admin_signed_in? && session[:last_seen_at] < admin_timeout_in_minutes.minutes.ago
+    admin_signed_in? && session[:admin_last_seen_at] < admin_timeout_in_minutes.minutes.ago
   end
 
   def admin_timeout_in_minutes
--- a/spec/requests/admin_sessions_spec.rb
+++ b/spec/requests/admin_sessions_spec.rb
@@ -1,0 +1,29 @@
+require "rails_helper"
+
+RSpec.describe "Admin Sessions", type: :request do
+  describe "#refresh" do
+    it "updates the last_seen_at session timestamp and responds with OK" do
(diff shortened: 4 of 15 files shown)


pezholio:
[Don't share `last_seen_at` variables between public and admin](https://trello.com/c/Sxffm2nh/1300-dont-share-lastseenat-variables-between-public-and-admin)

app/assets/javascripts/components/timeout_dialog.js
```diff
@@ -8,9 +8,11 @@
   var continueButtonElement = document.getElementById("continue_session");
   var warningTimeInMinutes = dialogElement.getAttribute("data-warning-in-minutes");
   var timeoutInMinutes = dialogElement.getAttribute("data-timeout-in-minutes");
+  var refreshSessionPath = dialogElement.getAttribute("data-refresh-session-path");
   var timeoutInMilliseconds = (timeoutInMinutes - warningTimeInMinutes) * 60 * 1000;
   var timeoutPath = dialogElement.getAttribute("data-timeout-path");
   var dialog = new window.A11yDialog(dialogElement);
+  var refreshURL;
```
tekin: Looks like this is an unused variable.

app/controllers/admin/sessions_controller.rb
```diff
@@ -0,0 +1,11 @@
+module Admin
+  class SessionsController < BaseAdminController
+    include AdminSessionTimeout
```
tekin: This is already included in `BaseAdminController` so doesn't need to be included again.

app/controllers/concerns/admin_session_timeout.rb
```diff
@@ -15,7 +15,7 @@ def admin_signed_in?
   end
 
   def admin_session_timed_out?
-    admin_signed_in? && session[:last_seen_at] < admin_timeout_in_minutes.minutes.ago
+    admin_signed_in? && session[:admin_last_seen_at] < admin_timeout_in_minutes.minutes.ago
```
tekin: When we ship this, any already signed-in admin users will get a 500 error because `session[:admin_last_seen_at]` will be `nil` and `nil < something` will blow up. We should be aware of this and do what we can to minimise disruption and confusion for claim checkers.

spec/requests/admin_sessions_spec.rb
```diff
@@ -0,0 +1,30 @@
+require "rails_helper"
+
+RSpec.describe "Admin Sessions", type: :request do
+  describe "#refresh" do
+    it "updates the last_seen_at session timestamp and responds with OK" do
+      sign_in_to_admin_with_role(DfeSignIn::User::SERVICE_OPERATOR_DFE_SIGN_IN_ROLE_CODE, "_user_id", "_org_id")
+      get admin_root_path
```
tekin: I think this `get` is redundant.

spec/requests/admin_timeout_spec.rb
```diff
@@ -7,38 +7,49 @@
...
-      end
+      follow_redirect!
+      expect(response.body).to include("Your session has timed out due to inactivity")
     end
   end
 
-  context "no action performed just within the timeout period" do
-    let(:before_expiry) { timeout_length_in_minutes.minutes - 2.seconds }
+  it "does not extend the session when the user performs a claim action" do
+    travel before_expiry do
```
tekin: Do we need to travel to before expiry here? Would calling `start_student_loans_claim` inside the `travel after_expiry` block below would still provide a failing spec for the bug? If so that would make it easier to follow and also closer to the behaviour that triggered the bug in the first place.

app/controllers/concerns/admin_session_timeout.rb
pezholio: How about safe navigation here? 

```ruby
admin_signed_in? && session[:admin_last_seen_at].&< admin_timeout_in_minutes.minutes.ago
```

tekin: That looks v.icky, especially as it should only be a temporary issue until their sessions are updated. I think it might be enough to either do the deploy when server operators aren't around, or ask them to sign-out whilst we do the deploy.

spec/requests/admin_timeout_spec.rb
pezholio: Yeah, that's clearer 👍

app/controllers/concerns/admin_session_timeout.rb
pezholio: Yeah, hence my asking 😃. It's probably not worth the long-term ickyness for the small benefit we get

app/controllers/admin/sessions_controller.rb
tekin: This is still here.

pezholio: Ah, I removed it from the public `SessionsController` (which is also legit), but not here

PR #853, opened 2020-02-11
Remove `verify_response` from a Claim

pezholio:
This was introduced when we first started integrating with Verify, as we didn't have a huge amount of confidence in the data it would return, so kept the raw response against a claim too. It has come in handy in the past, but, as the Verify integration is more or less stable, it makes less sense (and is a big risk) for us to be storing this much PII, so we should remove it.

### DIFF
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -57,7 +57,6 @@ class Claim < ApplicationRecord
     submitted_at: false,
     updated_at: false,
     govuk_verify_fields: false,
-    verify_response: true,
     banking_name: true,
     building_society_roll_number: true,
     payment_id: false,
--- a/app/models/claim/verify_response_parameters_parser.rb
+++ b/app/models/claim/verify_response_parameters_parser.rb
@@ -8,18 +8,10 @@ class Claim
   #   => { first_name: "Margaret", last_name: "Hamilton", date_of_birth: "1936-08-17", ...}
   #
   # As well as keys for the personal information of the user (name, address,
-  # etc) the returned Hash includes two additional keys:
-  #
-  #   govuk_verify_fields – these are the keys for the attributes that have come
-  #                     back in the Verify response. Recording these allows us
-  #                     to determine which attributes came from Verify and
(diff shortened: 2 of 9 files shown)


pezholio:
[Drop the verify_response column on claims](https://trello.com/c/qLYzS6LK/1274-drop-the-verifyresponse-column-on-claims)

db/schema.rb
```diff
@@ -10,12 +10,18 @@
...
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
   enable_extension "plpgsql"
 
+  create_table "big_ass_table", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.string "name"
+    t.datetime "created_at", precision: 6, null: false
+    t.datetime "updated_at", precision: 6, null: false
+  end
+
```
robbiepaul: :)

pezholio: 🤣 I totally missed that!

app/models/claim/verify_response_parameters_parser.rb
```diff
@@ -29,7 +29,7 @@ def initialize(response_parameters)
     end
 
     def attributes
-      identity_attributes.merge(govuk_verify_fields: govuk_verify_fields, verify_response: @response_parameters)
+      identity_attributes.merge(govuk_verify_fields: govuk_verify_fields)
```
robbiepaul: I think we need to update the Tomdoc for this model as well: https://github.com/DFE-Digital/dfe-teachers-payment-service/blob/870efe931ffadf2f32450be696d3c2690b5cded0/app/models/claim/verify_response_parameters_parser.rb#L18-L23

app/models/claim.rb
```diff
@@ -57,7 +57,6 @@ class Claim < ApplicationRecord
     submitted_at: false,
     updated_at: false,
     govuk_verify_fields: false,
-    verify_response: true,
```
tekin: oops, this should almost certainly have been `false`...

PR #855, opened 2020-02-11
Implement usability improvements for the student loans questions

pezholio:
This adds a couple of small improvements to the flow in the student loans questions, moving the explanation to a details component, and adding subheadings to every question in the flow.

# Screenshots

image

image

image

image

image

image

image

### DIFF
--- a/app/views/claims/student_loan.html.erb
+++ b/app/views/claims/student_loan.html.erb
@@ -7,17 +7,13 @@
     <%= form_for current_claim, url: claim_path(current_policy_routing_name) do |form| %>
       <%= form.hidden_field :has_student_loan %>
         <%= form_group_tag current_claim do %>
-          <fieldset class="govuk-fieldset" aria-describedby="student_loans-hint" role="group">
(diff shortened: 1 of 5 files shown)


app/views/claims/student_loan.html.erb
```diff
@@ -10,14 +10,11 @@
           <fieldset class="govuk-fieldset" aria-describedby="student_loans-hint" role="group">
             <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
               <h1 class="govuk-fieldset__heading">
+                <span class="govuk-caption-xl">About your student loans</span>
```
robbiepaul: Do you think we should have the caption outside of the `h1`? The design system seems to suggest if it's a section name it should be outside and if it's part of the heading it should be inside. My guess is there's a semantic difference for assistive tech.

https://design-system.service.gov.uk/styles/typography/#headings-with-captions

pezholio: Ah, I misread that. Lemme sort that out 👍

PR #856, opened 2020-02-11
Introduce qualifications and employment checks

robbiepaul:
We chose to start with qualification and employment checks as they are common to each policy, with the exception that StudentLoans has some additional information but we can tackle that later.

Eventually we'll get to a point where a service operator can mark these checks as completed or not.

## Screenshots

Screenshot 2020-02-11 at 18 04 32
Screenshot 2020-02-12 at 16 04 57
Screenshot 2020-02-12 at 16 05 12

### DIFF
--- a/app/assets/stylesheets/components/task_list.scss
+++ b/app/assets/stylesheets/components/task_list.scss
@@ -1,4 +1,16 @@
-// Task list pattern
+/*
+GOV.UK Task list pattern (experimental)
+
+This is an experimental GOV.UK pattern, the styles and markup
+have been taken from the example given in the GOV.UK prototyping
+kit (https://github.com/alphagov/govuk-prototype-kit/blob/master/app/assets/sass/patterns/_task-list.scss)
+
+For more information on the pattern and when to use it take a
+look at the GOV.UK Design System.
+
+https://design-system.service.gov.uk/patterns/task-list-pages/
+
+*/
 
 .app-task-list {
   list-style-type: none;
--- a/app/assets/stylesheets/components/task_list.scss
+++ b/app/assets/stylesheets/components/task_list.scss
@@ -1,0 +1,65 @@
+// Task list pattern
+
+.app-task-list {
+  list-style-type: none;
+  padding-left: 0;
+  margin-top: 0;
+  margin-bottom: 0;
+  @include govuk-media-query($from: tablet) {
+    min-width: 550px;
+  }
+}
+
+.app-task-list__section {
+  display: table;
+  @include govuk-font($size: 24, $weight: bold);
+}
+
+.app-task-list__section-number {
+  display: table-cell;
+
+  @include govuk-media-query($from: tablet) {
+    min-width: govuk-spacing(6);
+    padding-right: 0;
+  }
+}
+
+.app-task-list__items {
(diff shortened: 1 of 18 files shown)


robbiepaul:
[Introduce claim checking task list](https://trello.com/c/flXOrTyv/1299-introduce-claim-checking-task-list)

app/assets/stylesheets/components/task_list.scss
```diff
@@ -0,0 +1,65 @@
+// Task list pattern
```
tekin: I think it would be good to reference the place this SCSS came from as we haven't written it ourselves. That way we give a hint to future developer that it might need updating or may become redundant if/when it stops being an experimental pattern and becomes actually part of `govuk-frontend`. We should be able to reference it in the GOV.UK Prototype Kit.

robbiepaul: Yea good idea 👍

robbiepaul:
I started writing the code to introduce the ITT specialism/specialism subject etc and found myself duplicating a lot of the conditional logic we wrote in `EligibilityAdminAnswersPresenter#answers`. 

I decided it wouldn't be too much extra work to refactor the code slightly so we could reuse the logic in these new checks and we get the added benefit of it working with student loans.

I'll see if I can get some time from a claims checker to make sure we haven't missed anything.

robbiepaul:
As per our conversation earlier, I've changed the code in the last commit to introduce a new presenter for the new claims checks.

app/views/admin/claims/_answer_section.html.erb
```diff
@@ -1,7 +1,15 @@
-<h2 class="govuk-heading-m">
+<% description ||= nil %>
+
+<h2 class="govuk-heading-l">
   <%= heading %>
 </h2>
 
+<% if description.present? %>
```
tekin: We can use the `local_assigns` helper here to avoid having to do the conditional assignment above.

app/helpers/admin/claims_helper.rb
```diff
@@ -1,5 +1,6 @@
 module Admin
   module ClaimsHelper
+    include Admin::PresenterMethods
```
tekin: Are these presenter methods still needed here?

spec/requests/admin_checks_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+        get admin_claim_checks_path(claim_id: claim.id)
+
+        expect(response.body).to include(claim.reference)
+        expect(response.body).to include("Qualifications")
+        expect(response.body).to include("Employment")
+      end
+    end
+
+    # Compatible with claims from each policy
+    Policies.all.each do |policy|
```
tekin: 👍🏽

app/helpers/admin/claims_helper.rb
robbiepaul: Good catch, this was a leftover from earlier work. I've removed it

app/views/admin/claims/_answer_section.html.erb
robbiepaul: Thanks, sorted!

PR #857, opened 2020-02-12
Reduce sensitivity of Azure alerts

jacksonj04:
Reduces the sensitivity of Azure alerts.

* CPU and Memory alerts now trigger at 80% average usage sustained over 15 minutes (was previously 5 minutes).
* HTTP 500 errors now trigger at three or more errors within a minute (was previously two or more)

### DIFF
--- a/azure/templates/app_alerts.json
+++ b/azure/templates/app_alerts.json
@@ -88,7 +88,7 @@
               "metricNamespace": "Microsoft.Web/sites",
               "metricName": "Http5xx",
               "operator": "GreaterThan",
-              "threshold": 1,
+              "threshold": 2,
               "timeAggregation": "Total"
             }
           ]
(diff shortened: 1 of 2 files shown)


azure/templates/app_alerts.json
```diff
@@ -110,8 +110,8 @@
         "enabled": "[parameters('enableAlerts')]",
         "description": "[concat('Alert when average CPU utilization for ', variables('appServicePlanName'), ' is greater than 80%.')]",
         "severity": 1,
-        "evaluationFrequency": "PT1M",
-        "windowSize": "PT5M",
+        "evaluationFrequency": "PT5M",
```
pezholio: It would be nice to have this configurable per environment. We probably care less about issues like this in dev than we do in prod

robbiepaul:
Looking good, can we also increase the tolerances for the the `AverageResponseTime` check? That one has been quite sensitive in the past

azure/templates/app_alerts.json
jacksonj04: I think for these type of alerts we probably want to know just as much on development as we do on production - something pegging the CPU at 100% or throwing a bunch of HTTP 500 errors is just as crucial.

jacksonj04:
The `AverageResponseTime` check requires a sustained average response time of over 2 seconds in a 5 minute window, so to me that feels like the alert is correct and something is happening which is causing a noticeable service degradation.

robbiepaul:
Ah I didn't realise it was that high already. 

We should definitely investigate what's triggering it to happen so often. Looking at my previous alert emails it only seems to happen on development, the last alert for it on production was back in November. The only difference I can think between the two is development isn't set to 'AlwaysOn' - maybe it's something to do with the app warming up from cold?

PR #858, opened 2020-02-12
Fix school search button text changing when submitting

pezholio:
Out of the box, the Rails `submit_tag` helper sets a `data-disable-with` value, which tells the Unobstructive Javascript Scripting Adapter what to set the input's value to when disabling the submit button after it is clicked. In our JS, we don't change this when changing the value, so whe a user clicks on the "Continue" button, it switches back to "Search". This updates the JS so the `data-disable-with` attribute is changed whenever the value of the submit button is changed.


pezholio:
[Search button text changes when you click on it](https://trello.com/c/eEIpX7VD/1248-search-button-text-changes-when-you-click-on-it)

PR #860, opened 2020-02-12
Hide unknown gender warning for decided claims

jacksonj04:
Where a claim has had a decision made, do not show the message about needing a payroll gender to make a decision.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- The admin page for a claim no longer warns about a missing payroll gender once
+  a decision has been made
+
 ## [Release 053] - 2020-02-12
 
 - Don't share the last seen at session variables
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- The admin page for a claim no longer warns about a missing payroll gender once
+  a decision has been made.
```
tekin: No full stop on these.

spec/features/admin_claim_check_spec.rb
```diff
@@ -90,6 +90,16 @@
...
         expect(page).to have_content("This claim cannot be approved, the payroll gender is missing")
       end
+
+      context "When a decision has been made" do
+        let!(:claim_missing_payroll_gender_with_decision) { create(:claim, :rejected, payroll_gender: :dont_know) }
+
+        scenario "User is not informed that the claim cannot be approved" do
+          visit admin_claim_path(claim_missing_payroll_gender_with_decision)
+
+          expect(page).to_not have_content("This claim cannot be approved, the payroll gender is missing")
```
tekin: Minor thing but for specs that check for the absence of something, making the thing you look for as generic and small as possible makes the spec more robust against content changes. So verifying the page does not contain the words `"payroll gender"` would reduce the chance of this spec becoming a false-positive. Or better still, extract the message to the locale file, that way we reduce the duplication here in the specs and in the views and can be confident that content changes won't mean these specs also need to change.

PR #865, opened 2020-02-12
Introduce academic year modelling on policies and claims

tekin:
This lays some of the groundwork for supporting future academic years in the service, specifically it prepares the way for adjusting the QTS question to reflect the qualifying years that are eligible. What this PR covers:

- Introduces a `PolicyConfiguration#current_academic_year` attribute to represent the year each service currently is accepting claims for
- Introduces a `Claim#academic_year` attribute so we know the academic year each claim was made in

There are also two data migrations to run when this ships to backfill the data:

- `rails runner db/data/20200212151005_set_current_academic_year_on_policy_configurations.rb` to set current_academic_year to "2019/2020" for the two policies
- `rails runner db/data/20200212154927_set_academic_year_on_claims.rb` to back-fill all existing claims with the "2019/2020" academic year

Once this PR has shipped and the data migrations run there will be a follow-up PR to add validations to both `PolicyConfiguration` and `Claim` to validate these values are always present. We have to wait until after this ships, otherwise existing in-progress claims will error as invalid. Other things that will follow on from this:

- An update to the admin interface so that the current academic year can be changed on each policy
- An update that will change the QTS questions based on the current academic year and the rules for each policy

## Policy background

This work is required because each policy has their own unique criteria for eligibility based on the year that teachers qualified. Maths and Physics eligibility is based on teachers that are in the first 5 years of their careers. Therefor each academic year, the QTS year threshold needs to go up by one (from 2014/2015 to 2015/2016 when the academic year 2020/2021 starts). TSLR has its own more complicated rules around QTS year (it won't change until 2025...), which will also need to be accounted for.

### DIFF
--- a/app/controllers/verify/authentications_controller.rb
+++ b/app/controllers/verify/authentications_controller.rb
@@ -59,12 +59,16 @@ module Verify
       Rollbar.debug("Verify::RedactedResponse", parameters: redacted_response.parameters)
     end
 
+    def current_policy_routing_name
+      claim_from_session&.policy&.routing_name
+    end
+
     # This controller is not namespaced to a policy so we can't redirect a user
     # to a policy-specific start page if a claim isn't in progress. Instead
     # redirect to the root URL and let the routing take care of sending the user
     # to the right place.
     def send_unstarted_claiments_to_the_start
-      redirect_to root_url unless current_claim.persisted?
+      redirect_to root_url unless current_policy_routing_name
     end
   end
 end
--- a/spec/requests/claims_spec.rb
+++ b/spec/requests/claims_spec.rb
@@ -25,18 +25,20 @@ RSpec.describe "Claims", type: :request do
   end
 
   describe "claims#create request" do
-    it "creates a new Claim for the policy and redirects to the QTS question" do
-      expect { start_student_loans_claim }.to change { Claim.count }.by(1)
+    Policies.all.each do |policy|
+      it "creates a new Claim for the policy and redirects to the next question in the sequence" do
+        expect { start_claim(policy) }.to change { Claim.count }.by(1)
 
-      claim = Claim.last
-      expect(claim.eligibility).to be_kind_of(StudentLoans::Eligibility)
+        claim = Claim.last
+        expect(claim.eligibility).to be_kind_of(policy::Eligibility)
 
-      expect(response).to redirect_to(claim_path(StudentLoans.routing_name, "claim-school"))
-    end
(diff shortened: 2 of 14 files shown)


db/data/20200212151551_foo_bar_bag.rb
```diff
@@ -0,0 +1,3 @@
+# Run me with `rails runner db/data/20200212151551_foo_bar_bag.rb`
```
pezholio: Should this have been committed?

spec/requests/claims_spec.rb
```diff
@@ -25,18 +25,20 @@
   end
 
   describe "claims#create request" do
-    it "creates a new Claim for the policy and redirects to the QTS question" do
-      expect { start_student_loans_claim }.to change { Claim.count }.by(1)
+    Policies.all.each do |policy|
+      it "creates a new Claim for the policy and redirects to the next question in the sequence" do
```
pezholio: It'd be good to have the policy referenced in the name of the spec here

app/controllers/verify/authentications_controller.rb
```diff
@@ -59,12 +59,16 @@ def report_redacted_response
       Rollbar.debug("Verify::RedactedResponse", parameters: redacted_response.parameters)
     end
 
+    def current_policy_routing_name
```
pezholio: Would you mind rewording this commit? (92d91cba47917c389d6d2fc4dacc626a8ab14dad). I'm finding it a bit tricky to follow when reading the commits in order. It might even be better to combine withe the subsequent commit, as I'm finding it difficult to understand why this commit was added in isolation

db/data/20200212151551_foo_bar_bag.rb
tekin: Nope, there's a [fixup commit](https://github.com/DFE-Digital/dfe-teachers-payment-service/pull/865/commits/5cd20dad33ad693ab1b9f7a21d5cf1759d96d122) that just needs squashing down.

spec/requests/claims_spec.rb
tekin: Good shout.

app/controllers/verify/authentications_controller.rb
tekin: Good shout, commits combined. Does the new commit make more sense now?

pezholio: Yeah, that's great 👍

PR #869, opened 2020-02-13
Update Heroku configuration to enable acceptance testing apps

pezholio:
We've had review apps set up for a while now, and they haven't worked properly. This fixes them up to use the correct buildpacks, as well as a script that correctly bootstraps the database, so we can use the app.


tekin:
Amazing, great work folks. Can we add some documentation to around these so that current and future developers know what the deal is?

PR #870, opened 2020-02-13
Allow service operators to change the current academic year on a policy

tekin:
This updates the admin interface so that service operators can change the academic year that each policy is accepting claims for:

Screenshot 2020-02-13 at 16 22 35

Screenshot 2020-02-13 at 16 22 39

This is the second part of the work to support future claim years. The final part is to update the claimant journey and eligibility so that they reflect the academic year for which claims are being accepted for.

### DIFF
--- a/app/views/admin/policy_configurations/edit.html.erb
+++ b/app/views/admin/policy_configurations/edit.html.erb
@@ -8,14 +8,24 @@
     <div class="govuk-panel govuk-panel--informational">
       <% if @policy_configuration.open_for_submissions? %>
         <h3 class="govuk-panel__title">Service open</h3>
-        <p class="govuk-panel__body">This service is currently open. Users can submit claims.</p>
+        <p class="govuk-panel__body">
+          This service is currently open and accepting claims for the
+          <%= @policy_configuration.current_academic_year%> academic year.
+        </p>
       <% else %>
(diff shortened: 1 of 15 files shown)


app/views/admin/policy_configurations/edit.html.erb
```diff
@@ -1,22 +1,32 @@
...
+        <p class="govuk-panel__body">
+          This service is currently closed. Users can not submit claims.
+        </p>
       <% end %>
     </div>
 
     <%= form_with model: [:admin, @policy_configuration] do |f| %>
 
+      <div class="govuk-form-group">
+        <%= f.label :current_academic_year, "Accepting claims for academic year", class: "govuk-label" %>
```
titlescreen: I wonder is `Accepting claims` could cause confusion? It would be possible to be accepting claims for 2020/2021 but have the service closed.

Perhaps something like claim window would be more appropriate?

tekin: We need to be careful here, there is a technical and semantic difference between the academic year and the claim window. The value here represents the academic year, which runs September to July, whilst the claim window runs September to end of Feb. I think we should go with how it is now and see if service operators find it confusing. When I showed it to Phil last week, he was happy with it.

PR #874, opened 2020-02-17
Run two concurrent worker processes

pezholio:
This updates the worker container to run a command that spawns two worker processes and then tails the logs, so we can see what's going on. I don't think specifying the delayed_job log file location will interfere with logit, but I'll keep an eye on it.

### DIFF
--- a/azure/resource_groups/app/parameters/development.template.json
+++ b/azure/resource_groups/app/parameters/development.template.json
@@ -181,6 +181,9 @@
     "ENVIRONMENT_NAME": {
       "value": "development"
     },
+    "WORKER_COUNT": {
+      "value": 2
+    },
     "VSP.VERIFY_ENVIRONMENT": {
       "value": "INTEGRATION"
     },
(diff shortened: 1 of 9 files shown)


pezholio:
[Run two concurrent worker processes](https://trello.com/c/jUOjxSGG/1331-run-two-concurrent-worker-processes)

azure/resource_groups/app/parameters/development.template.json
```diff
@@ -181,6 +181,9 @@
     "ENVIRONMENT_NAME": {
       "value": "development"
     },
+    "WORKER_COUNT": {
+      "value": 1
```
robbiepaul: Do you think we should go for parity with production here? It should help us pick up on any issues sooner

jacksonj04: Argh, my brain still treats `development` as "to develop in" not "staging". Fixed!

PR #878, opened 2020-02-18
Swap Net/HTTP for HTTPClient in file_download

jacksonj04:
This resolves growing memory usage during the school data import job.

### DIFF
--- a/lib/file_download.rb
+++ b/lib/file_download.rb
@@ -1,11 +1,8 @@
-require "net/http"
-
-# Used to download a remote file. Handles redirects up to a maximum of 5.
+# Used to download a remote file. Handles redirects up to the given maximum.
 # Returns the file as a `Tempfile`
 class FileDownload
   MAX_REDIRECTS = 5
 
-  class TooManyRedirects < StandardError; end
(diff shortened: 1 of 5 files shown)


pezholio:
Just a quick one, althought we don't currently need to, I think it's worth us adding HTTPClient to the Gemfile, just in case any upstream dependencies change the library they use.

lib/file_download.rb
```diff
@@ -22,17 +19,14 @@ def fetch
   private
 
   def download_file(url, redirect_limit = MAX_REDIRECTS)
```
pezholio: We probably don't need the `redirect_limit` parameter now, as we're only calling this method once

lib/file_download.rb
```diff
@@ -21,18 +18,15 @@ def fetch
 
   private
 
-  def download_file(url, redirect_limit = MAX_REDIRECTS)
```
pezholio: Actually, thinking about it, this doesn't _really_ need to be a method on its own at all now. The logic can just be moved into the `fetch` method above

PR #879, opened 2020-02-18
1025 accessibility testing admin area

titlescreen:
This is the first of several PRs to fix accessibility issues within the claim admin area. It is looking at the A level of accessibility and includes
- fixing the order of headers
- ensuring inputs have associated headers
- checking that table headers do not have empty <th>
- Adding in meaningful page titles to individual pages

### DIFF
--- a/app/views/admin/claims/index.html.erb
+++ b/app/views/admin/claims/index.html.erb
@@ -1,4 +1,4 @@
-<% content_for(:page_title) { "View claims - #{t("service_name")} – GOV.UK"} %>
+<% content_for(:page_title) { page_title("View claims") } %>
 
 <%= link_to "Back", admin_root_path, class: "govuk-back-link" %>
 <div class="govuk-grid-row">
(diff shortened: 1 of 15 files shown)


app/views/admin/claims/index.html.erb
```diff
@@ -31,7 +33,7 @@
               <th scope="col" class="govuk-table__header">Decision warning</th>
             <% end %>
             <th scope="col" class="govuk-table__header">Decision deadline</th>
-            <th scope="col" class="govuk-table__header"></th>
+            <td scope="col" class="govuk-table__header"></td>
```
robbiepaul: I don't think a `td` can have a `scope` attribute in HTML5. Maybe we should either remove the scope or add an `aria-label` that describes the column?

titlescreen: Thanks bud, I've added an aria role as suggested 👍

robbiepaul: I think `td[role=colheader]` is semantically equal to `th[scope=col]` - so I think the issue of having an empty column header still stands? I could be wrong, I'm not sure how screen readers will read tables. My guess would be that each column needs a header for when you tab through the cells?

PR #881, opened 2020-02-18
Remove the `tmp-app` from the deploy script

pezholio:
We've deprecated this in favour of the actual test environment in Azure, which gives us more flexibility.


pezholio:
[Remove the `--tmp-app` from the deploy script](https://trello.com/c/ammTu2Fa/1298-remove-the-tmp-app-from-the-deploy-script)

PR #885, opened 2020-02-19
Resolve bugs with scrubbed claims

jacksonj04:
Resolves a few issues with scrubbing PII from old claims:

* [x] Moves the generation of the two-month-ago date after which claims are considered 'old' from being a class constant into being set during instance initialisation.
* [x] Duplicate claim spotting behaviour with claims which have been scrubbed is not as expected.

### DIFF
--- a/app/models/claim/pii_scrubber.rb
+++ b/app/models/claim/pii_scrubber.rb
@@ -23,7 +23,7 @@ class Claim
       :building_society_roll_number,
     ]
 
-    MINIMUM_DATE_OF_LAST_ACTION = 2.months.ago
+    TIME_BEFORE_CLAIM_CONSIDERED_OLD = 2.months
 
     def scrub_completed_claims
       old_claims_rejected_or_paid.update_all(attribute_values_to_set)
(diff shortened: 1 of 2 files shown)


jacksonj04:
Regarding exceptions on claims with PII removed and the matched data spotter, this is actually an issue with trying to view an unsubmitted claim.

app/models/claim/pii_scrubber.rb
```diff
@@ -23,7 +23,11 @@ class PiiScrubber
       :building_society_roll_number,
     ]
 
-    MINIMUM_DATE_OF_LAST_ACTION = 2.months.ago
+    TIME_BEFORE_CLAIM_CONSIDERED_OLD = 2.months
+
+    def initialize
+      @minimum_time_of_last_action = TIME_BEFORE_CLAIM_CONSIDERED_OLD.ago
```
tekin: I reckon we can forgo this instance variable and just call `TIME_BEFORE_CLAIM_CONSIDERED_OLD.ago` in the scope method below.

PR #889, opened 2020-02-24
Vary the QTS question based on the academic year configured in the policy

tekin:
This updates the claim journey such that the QTS question will vary the options based on the academic year that the policy is currently configured to accept answers for. Further work is [in-progress](https://github.com/DFE-Digital/dfe-teachers-payment-service/compare/1223-vary-qts-eligibility-based-on-academic-year?expand=1) to update the admin area so that it reflects the actual answer the user gave, but right now it is still hard-coded to display the current years. That work will be delivered admin area will be updated in a separate PR.

### DIFF
--- a/app/models/maths_and_physics/eligibility_answers_presenter.rb
+++ b/app/models/maths_and_physics/eligibility_answers_presenter.rb
@@ -76,7 +76,7 @@ module MathsAndPhysics
     def qts_award_year
       [
         I18n.t("questions.qts_award_year"),
-        I18n.t("maths_and_physics.questions.qts_award_years.#{eligibility.qts_award_year}"),
+        I18n.t("answers.qts_award_years.#{eligibility.qts_award_year}", year: qts_answer_academic_year.to_s(:long)),
         "qts-year",
       ]
     end
@@ -126,6 +126,14 @@ module MathsAndPhysics
       when "yes" then "Yes"
       when "no" then "No"
       else I18n.t("maths_and_physics.answers.has_uk_maths_or_physics_degree.#{eligibility.has_uk_maths_or_physics_degree}")
+      end
+    end
+
+    def qts_answer_academic_year
+      if eligibility.awarded_qualified_status_on_or_after_cut_off_date?
+        MathsAndPhysics.first_eligible_qts_award_year
+      else
+        MathsAndPhysics.first_eligible_qts_award_year - 1
       end
     end
   end
--- a/app/models/maths_and_physics/eligibility_answers_presenter.rb
+++ b/app/models/maths_and_physics/eligibility_answers_presenter.rb
@@ -17,21 +17,109 @@ module MathsAndPhysics
     # [2]: slug for changing the answer.
     def answers
       [].tap do |a|
-        a << [I18n.t("maths_and_physics.questions.teaching_maths_or_physics"), (eligibility.teaching_maths_or_physics? ? "Yes" : "No"), "teaching-maths-or-physics"]
-        a << [I18n.t("questions.current_school"), eligibility.current_school_name, "current-school"]
-        a << [I18n.t("maths_and_physics.questions.initial_teacher_training_subject"), I18n.t("maths_and_physics.answers.initial_teacher_training_subject.#{eligibility.initial_teacher_training_subject}"), "initial-teacher-training-subject"]
-        a << [I18n.t("maths_and_physics.questions.initial_teacher_training_subject_specialism"), I18n.t("maths_and_physics.answers.initial_teacher_training_subject_specialism.#{eligibility.initial_teacher_training_subject_specialism}"), "initial-teacher-training-subject-specialism"] if eligibility.initial_teacher_training_subject_specialism.present?
-        a << [I18n.t("maths_and_physics.questions.has_uk_maths_or_physics_degree"), degree_answer, "has-uk-maths-or-physics-degree"] if eligibility.has_uk_maths_or_physics_degree.present?
-        a << [I18n.t("questions.qts_award_year"), I18n.t("maths_and_physics.questions.qts_award_years.#{eligibility.qts_award_year}"), "qts-year"]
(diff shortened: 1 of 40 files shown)


app/models/maths_and_physics/eligibility_answers_presenter.rb
```diff
@@ -128,5 +128,13 @@ def degree_answer
       else I18n.t("maths_and_physics.answers.has_uk_maths_or_physics_degree.#{eligibility.has_uk_maths_or_physics_degree}")
       end
     end
+
+    def qts_answer_academic_year
+      if eligibility.awarded_qualified_status_on_or_after_cut_off_date?
+        MathsAndPhysics.first_eligible_qts_award_year
+      else
+        MathsAndPhysics.first_eligible_qts_award_year - 1
```
robbiepaul: In theory a user shouldn't be able to see this answer on the check-your-answers page, they should see the "ineligible" page first. Is there a reason why you decided to include it?

app/models/student_loans/eligibility_answers_presenter.rb
```diff
@@ -87,5 +87,13 @@ def student_loan_amount
         "student-loan-amount",
       ]
     end
+
+    def qts_answer_academic_year
+      if eligibility.awarded_qualified_status_on_or_after_cut_off_date?
+        StudentLoans.first_eligible_qts_award_year
+      else
+        StudentLoans.first_eligible_qts_award_year - 1
```
robbiepaul: Same as above

app/models/student_loans.rb
```diff
@@ -1,5 +1,10 @@
-# frozen_string_literal: true
```
robbiepaul: I've noticed we're quite inconsistent with using frozen string literals (we didn't have it in the `MathsAndPhysics` module for example), I'm not sure if the benefits are enough to worry about but I thought I'd point it out incase you didn't mean to remove it.

app/models/student_loans.rb
```diff
@@ -35,4 +37,29 @@ def feedback_url
...
+  # the *financial year*. So for example if you qualify in 2021/2022, you are
+  # eligible to claim back student loan repayments you make in the 2021/2022
+  # "financial year", which ends April 2022, and the claim for that period can
+  # be made from the start of the 2022/2023 "academic year".
+  #
+  # So to give concrete examples, teachers qualifying in 2013/2014 can make
+  # claims up to 2024/2025, and a teacher qualifying in 2014/2015 can make
+  # claims up to 2025/2026 and so on.
+  def first_eligible_qts_award_year
+    eleven_years_prior = AcademicYear.new(configuration.current_academic_year) - 11
```
robbiepaul: At first I was thinking "why is it 11 here and not 10?" but the comment and examples explained it nicely 👍

config/locales/en.yml
```diff
@@ -137,8 +140,8 @@ en:
       initial_teacher_training_subject_specialism: "Which science subject did your initial teacher training specialise in?"
       has_uk_maths_or_physics_degree: "Do you have a UK undergraduate or postgraduate degree in maths or physics?"
       qts_award_years:
-        before_september_2014: "In or before the academic year 2013 to 2014"
-        on_or_after_september_2014: "In or after the academic year 2014 to 2015"
+        before_cut_off_date: "In or before the academic year 2013 to 2014"
+        on_or_after_cut_off_date: "In or after the academic year 2014 to 2015"
```
robbiepaul: Can we remove these policy specific translation strings now that we're using `t("answers.qts_award_years.on_or_after_cut_off_date", year: current_policy.first_eligible_qts_award_year.to_s(:long))` in the view?

config/locales/en.yml
```diff
@@ -187,8 +190,8 @@ en:
     support_email_address: "studentloanteacherpayment@digital.education.gov.uk"
     questions:
       qts_award_years:
-        before_september_2013: "In or before the academic year 2012 to 2013"
-        on_or_after_september_2013: "In or after the academic year 2013 to 2014"
+        before_cut_off_date: "In or before the academic year 2012 to 2013"
+        on_or_after_cut_off_date: "In or after the academic year 2013 to 2014"
```
robbiepaul: Same as above

app/models/maths_and_physics/eligibility_answers_presenter.rb
tekin: It's funny, I originally had it assuming the answer was the eligible one, and have [done the same for the admin presenter](https://github.com/DFE-Digital/dfe-teachers-payment-service/commit/83ab345ae22a80f1f8a1d1fb0c8dafd660833f05) in the other branch. There is a part of me that does feel a little uncomfortable making the assumption that this presenter will only be used to display eligible claims though...

app/models/student_loans.rb
tekin: Yeah, we're definitely not consistent. I think I removed this one here so that the shape of this file more closely the source for the `MathsAndPhysics` module. Maybe I should have gone the other way and added it to `MathsAndPhysics` instead.

app/models/student_loans.rb
tekin: In the follow-up work in the other branch it gets a name too: https://github.com/DFE-Digital/dfe-teachers-payment-service/compare/1223-vary-qts-eligibility-based-on-academic-year?expand=1#diff-3f76becab371c5075ab8d7f820fa49faR62

config/locales/en.yml
tekin: Not yet, at least not without breaking a bunch of specs. There are still a bunch of places the existing keys are being used, mostly in the specs. Part of the follow-up work to update admin so that it dynamically shows the correct academic year will also migrate to using the new keys everywhere. At that point we can ditch these old keys.

app/models/student_loans.rb
tekin: I've re-instated the magic quote and added it to the other policy namespace for consistency.

app/models/maths_and_physics/eligibility_answers_presenter.rb
robbiepaul: Yeah, I think you're right. It is possible in an edge-case that a user may be able to see this answer on the check your answers page.

PR #898, opened 2020-02-25
Maths & Physics claims use new tasks interface

robbiepaul:
This is a temporary step to encourage service operators to check Maths and Physics claims using the new "checks" or "tasks" interface.


robbiepaul:
[Maths & Physics claims use new "tasks" interface](https://trello.com/c/JPUkGSws/1362-maths-physics-claims-use-new-tasks-interface)

robbiepaul:
That's a good point, I've added a link to "View full claim" (from the prototype https://fast-shelf-85846.herokuapp.com/)

PR #899, opened 2020-02-25
A service operator can amend a claim

lawrence-forooghian:
# Description

This adds an "Amend claim" button which service operators can use to amend certain attributes of a claim. The commit messages give lots of information about this and why we're doing it.

# Opinions wanted

There are couple of technical things I was hoping for opinions on; I've left comments on the PR. Nothing massive.

# Screenshots

## Amend claim link on `admin/claim#show`
image

## Amend claim page
image

## Validation errors on amend claim page
image

## Amendments timeline
image

### DIFF
--- a/app/controllers/admin/amendments_controller.rb
+++ b/app/controllers/admin/amendments_controller.rb
@@ -1,0 +1,32 @@
+class Admin::AmendmentsController < Admin::BaseAdminController
+  before_action :ensure_service_operator
+
+  def new
+    @claim = Claim.find(params[:claim_id])
+    @amendment = @claim.amendments.build
+  end
+
+  def create
+    @claim = Claim.find(params[:claim_id])
+    @amendment = Amendment.amend_claim(@claim, claim_params, amendment_params)
+
+    if @amendment.persisted?
+      redirect_to admin_claim_url(@claim)
+    else
+      render "new"
+    end
+  end
+
+  private
+
+  def claim_params
+    params.require(:amendment).require(:claim).permit(*Claim::AMENDABLE_ATTRIBUTES)
+  end
+
+  def amendment_params
+    {
+      notes: params[:amendment][:notes],
+      created_by: admin_user
+    }
+  end
+end
--- a/app/controllers/admin/amendments_controller.rb
+++ b/app/controllers/admin/amendments_controller.rb
@@ -1,13 +1,13 @@
 class Admin::AmendmentsController < Admin::BaseAdminController
+  before_action :load_claim
   before_action :ensure_service_operator
+  before_action :ensure_claim_is_amendable
 
   def new
-    @claim = Claim.find(params[:claim_id])
     @amendment = @claim.amendments.build
   end
 
   def create
-    @claim = Claim.find(params[:claim_id])
     @amendment = Amendment.amend_claim(@claim, claim_params, amendment_params)
 
     if @amendment.persisted?
@@ -18,6 +18,16 @@ class Admin::AmendmentsController < Admin::BaseAdminController
   end
 
   private
+
+  def load_claim
+    @claim = Claim.find(params[:claim_id])
+  end
+
+  def ensure_claim_is_amendable
+    unless @claim.amendable?
+      render "not_amendable"
+    end
+  end
 
   def claim_params
     params.require(:amendment).require(:claim).permit(*Claim::AMENDABLE_ATTRIBUTES)
--- a/app/models/amendment.rb
+++ b/app/models/amendment.rb
@@ -1,0 +1,17 @@
+# A service operator can amend a claim. When they do so, an Amendment record is
+# created. The amendments of a claim provide an audit trail which explains why
+# the claim has its current attribute values. It stores the details of which
+# attributes were changed, and optionally their old and new values.
+#
+# The claim_changes attribute is a hash whose keys (String) are the names of
+# the attributes changed. The values are either
+# - an array [old_value, new_value]
+# - nil, meaning that this personal data has been removed from the amendment
+class Amendment < ApplicationRecord
+  belongs_to :claim
+  belongs_to :created_by, class_name: "DfeSignIn::User"
+  serialize :claim_changes, Hash
+
+  validates :claim_changes, presence: {message: "To amend the claim you must change at least one value"}
+  validates :notes, presence: {message: "Enter a message to explain why you are making this amendment"}
+end
--- a/app/models/amendment.rb
+++ b/app/models/amendment.rb
@@ -14,6 +14,7 @@ class Amendment < ApplicationRecord
 
   validates :claim_changes, presence: {message: "To amend the claim you must change at least one value"}
   validates :notes, presence: {message: "Enter a message to explain why you are making this amendment"}
+  validate :claim_must_be_amendable, on: :create
 
   # Updates the claim using the attributes given in claim_attributes, and uses
   # these changes to create an associated Amendment record with additional
@@ -47,5 +48,11 @@ class Amendment < ApplicationRecord
     end
 
     amendment
+  end
+
+  private
+
+  def claim_must_be_amendable
+    errors.add(:claim, "must be amendable") unless claim.amendable?
   end
 end
--- a/app/models/amendment.rb
+++ b/app/models/amendment.rb
@@ -14,4 +14,38 @@ class Amendment < ApplicationRecord
 
   validates :claim_changes, presence: {message: "To amend the claim you must change at least one value"}
   validates :notes, presence: {message: "Enter a message to explain why you are making this amendment"}
+
+  # Updates the claim using the attributes given in claim_attributes, and uses
+  # these changes to create an associated Amendment record with additional
+  # attributes given by amendment_attributes.
+  #
+  # If the operation fails, then the claim’s changes will not be persisted, and
+  # the Amendment record will not be saved.
+  #
+  # Returns an Amendment record.
+  # - If the operation was successful, this will return true for persisted?
+  # - If the operation failed, this will return false for persisted?, and the
+  # amendment’s errors will have been populated with the errors from the claim.
+  def self.amend_claim(claim, claim_attributes, amendment_attributes)
+    amendment = Amendment.new(claim: claim, **amendment_attributes)
+
+    Claim.transaction do
+      claim.assign_attributes(claim_attributes)
+
+      unless claim.save(context: :submit)
+        amendment.valid?
+        amendment.errors.merge!(claim.errors)
+        amendment.errors.delete(:claim_changes)
+        raise ActiveRecord::Rollback
+      end
+
+      amendment.claim_changes = claim.previous_changes
+        .slice(*Claim::AMENDABLE_ATTRIBUTES)
+        .reject { |_, values| values.all?(&:blank?) }
+        .to_h
+      raise ActiveRecord::Rollback unless amendment.save
+    end
+
+    amendment
+  end
 end
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -81,6 +81,7 @@ class Claim < ApplicationRecord
 
   has_one :decision
   has_many :checks
+  has_many :amendments
 
   belongs_to :eligibility, polymorphic: true, inverse_of: :claim, dependent: :destroy
   accepts_nested_attributes_for :eligibility, update_only: true
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -275,6 +275,10 @@ class Claim < ApplicationRecord
     eligibility&.current_school
   end
 
+  def amendable?
+    submitted? && !decision&.rejected? && !payment && !pii_removed?
+  end
+
   private
 
   def normalise_trn
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -28,6 +28,15 @@ class Claim < ApplicationRecord
     :banking_name,
     :building_society_roll_number
   ].freeze
+  AMENDABLE_ATTRIBUTES = %i[
+    teacher_reference_number
+    payroll_gender
+    date_of_birth
+    student_loan_plan
+    bank_sort_code
+    bank_account_number
+    building_society_roll_number
+  ].freeze
   FILTER_PARAMS = {
     address_line_1: true,
     address_line_2: true,
--- a/app/models/claim/pii_scrubber.rb
+++ b/app/models/claim/pii_scrubber.rb
@@ -26,7 +26,13 @@ class Claim
     TIME_BEFORE_CLAIM_CONSIDERED_OLD = 2.months
 
     def scrub_completed_claims
-      old_claims_rejected_or_paid.update_all(attribute_values_to_set)
+      Claim.transaction do
+        old_claims_rejected_or_paid.includes(:amendments).each do |claim|
+          scrub_amendments_personal_data(claim)
+        end
+
+        old_claims_rejected_or_paid.update_all(attribute_values_to_set)
+      end
     end
 
     private
@@ -46,6 +52,22 @@ class Claim
           minimum_time: TIME_BEFORE_CLAIM_CONSIDERED_OLD.ago,
           rejected: Decision.results.fetch(:rejected)
         )
+    end
+
+    def scrub_amendments_personal_data(claim)
+      claim.amendments.each do |amendment|
+        scrub_amendment_personal_data(amendment)
+      end
+    end
+
+    def scrub_amendment_personal_data(amendment)
+      attributes_to_scrub = PII_ATTRIBUTES_TO_DELETE.map(&:to_s) & amendment.claim_changes.keys
+      personal_data_mask = attributes_to_scrub.to_h { |attribute| [attribute, nil] }
+      amendment.claim_changes.merge!(personal_data_mask)
+
+      amendment.personal_data_removed_at = Time.zone.now
+
+      amendment.save!
     end
   end
 end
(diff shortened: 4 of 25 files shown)


spec/factories/amendments.rb
```diff
@@ -0,0 +1,13 @@
+FactoryBot.define do
+  factory :amendment do
+    association :claim, factory: [:claim, :submitted]
+    association :created_by, factory: :dfe_signin_user
+    claim_changes { ActiveSupport::HashWithIndifferentAccess.new(teacher_reference_number: [generate(:teacher_reference_number).to_s, claim.teacher_reference_number]) }
```
lawrence-dxw: I'd like to ask for opinions on this. I've made it match the class returned by `ActiveModel::Dirty#previous_changes`, which is what we use to populate this attribute in the app. But I'm wondering

1.  whether it's the right thing to be storing in the database
2. whether the specs need to match the app code so exactly

spec/requests/admin_claims_spec.rb
```diff
@@ -71,6 +71,37 @@
...
+        end
+
+        context "when the claim has been amended" do
+          let(:claim) { create(:claim, :submitted, teacher_reference_number: "1234567") }
+          let!(:amendment) { create(:amendment, claim: claim, claim_changes: {teacher_reference_number: ["7654321", "1234567"]}) }
+
+          it "renders a timeline of changes" do
+            get admin_claim_path(claim)
+
+            expect(response.body).to include("changed from 7654321 to 1234567")
```
lawrence-dxw: Request for opinions: I'm a bit uneasy about the lack of request specs for the claim timeline, but it's hard to test without parsing the HTML in these specs, because the text is split over various HTML tags. It's just about possible to make a useful-ish test for the case where personal data hasn't been removed, as I have here, but it's impossible for the case where personal data has been removed, where the only string I can search for is "changed"...

So the timeline is mainly tested via tests of the `admin/timeline_helper` spec, and I'm wondering whether that's enough.

spec/requests/admin_amendments_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+        expect(amendment.claim_changes).to be_a(ActiveSupport::HashWithIndifferentAccess)
+        expect(amendment.notes).to eq("Claimant made a typo")
+        expect(amendment.created_by).to eq(service_operator)
+
+        expect(claim.teacher_reference_number).to eq("7654321")
+        expect(claim.bank_sort_code).to eq("111213")
+        expect(claim.date_of_birth).to eq(new_date_of_birth)
+      end
+
+      it "saves the normalised value in the amendment when updating bank sort code" do
```
lawrence-dxw: I have some request specs here which I wrote before extracting some code from the controller into the model. So now there is some overlap between these specs and the model specs. I'm wondering whether this is a problem. In general, we don't seem to mind a bit of overlap between specs.

lawrence-dxw:
[A service operator can amend values on a claim](https://trello.com/c/510bZMyP/1279-a-service-operator-can-amend-values-on-a-claim)

app/models/claim.rb
```diff
@@ -262,6 +263,10 @@ def school
     eligibility&.current_school
   end
 
+  def amendable?
+    submitted? && !decision&.rejected? && !passed_scheduled_payment_date?
```
jacksonj04: I think this is actually 
```suggestion
    submitted? && !decision&.rejected? && !payment.present?
```
since once we have imported the payment confirmation file there's nothing we can do, even if we import the file on a Tuesday and it's not scheduled to pay until the Friday

spec/requests/admin_claims_spec.rb
pezholio: Could you use a feature test and check for the presence / absence of particular selectors on a page? Something like:

```ruby
expect(page).to have_selector("p.hmcts-timeline__by[text='changed']")
```

spec/factories/amendments.rb
lawrence-dxw: Spoke to Tekin about this and he said that the best thing would be to save a normal `Hash` instead of one with indifferent access, using the keys that come in `previous_changes`, which are strings. So I've made that change.

Aside: `serialize ..., Hash` actually [serializes using YAML](http://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Serialization/ClassMethods.html#method-i-serialize), so that's how you‘d find out what the serialization looks like, not `to_s`.

app/models/amendment.rb
```diff
@@ -0,0 +1,26 @@
...
+# the claim has its current attribute values. It stores the details of which
+# attributes were changed, and optionally their old and new values.
+#
+# The claim_changes attribute is a hash whose keys are the names of the attributes changed. The values are either
+# - an array [old_value, new_value]
+# - nil, meaning that this personal data has been removed from the amendment
+class Amendment < ApplicationRecord
+  belongs_to :claim
+  belongs_to :created_by, class_name: "DfeSignIn::User"
+  serialize :claim_changes, Hash
```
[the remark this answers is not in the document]
lawrence-dxw: This comment is outdated now that we're storing it as a `Hash` 😃

app/models/amendment.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  validates :claim_changes, presence: {message: "To amend the claim you must change at least one value"}
+  validates :notes, presence: {message: "Enter a message to explain why you are making this amendment"}
+
+  AMENDABLE_ATTRIBUTES = %i[
```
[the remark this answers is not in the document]
lawrence-dxw: Right you are. Gonna move it.

spec/factories/amendments.rb
pezholio: Aaah, righto, that makes sense

app/controllers/admin/amendments_controller.rb
```diff
@@ -19,6 +19,16 @@ def create
 
   private
 
+  def load_claim
+    @claim = Claim.find(params[:claim_id])
+  end
+
+  def ensure_claim_is_amendable
+    unless @claim.amendable?
+      render "not_amendable"
```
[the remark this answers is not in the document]
lawrence-dxw: Our error messages aren't actually that noticeable, and I worried that it might end up looking like the claim was amended successfully. I think it's best to either render the form that the user was trying to submit, with error messages (which doesn't work in the case where the claim has been anonymised), or otherwise to show a completely different view, which seemed like the simplest. I don't think this extra little template adds much maintenance overhead, what do you think?

app/models/amendment.rb
```diff
@@ -23,4 +23,37 @@ class Amendment < ApplicationRecord
...
+  # attributes given by amendment_attributes.
+  #
+  # If the operation fails, then the claim’s changes will not be persisted, and
+  # the Amendment record will not be saved.
+  #
+  # Returns an Amendment record.
+  # - If the operation was successful, this will return true for persisted?
+  # - If the operation failed, this will return false for persisted?, and the
+  # amendment’s errors will have been populated with the errors from the claim.
+  def self.amend(claim, claim_attributes, amendment_attributes)
```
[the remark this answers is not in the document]
lawrence-dxw: Sure thing, changed it 👍

app/models/amendment.rb
```diff
@@ -23,4 +23,37 @@ class Amendment < ApplicationRecord
...
+      unless claim.save(context: :submit)
+        amendment.valid?
+        amendment.errors.merge!(claim.errors)
+        amendment.errors.delete(:claim_changes)
+        raise ActiveRecord::Rollback
+      end
+
+      amendment.claim_changes = claim.previous_changes
+        .slice(*Amendment::AMENDABLE_ATTRIBUTES)
+        .reject { |_, values| values.all? { |value| value.blank? } }
```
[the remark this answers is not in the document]
lawrence-dxw: Nice catch, ta.

app/models/claim/pii_scrubber.rb
```diff
@@ -47,5 +53,23 @@ def old_claims_rejected_or_paid
...
+        scrub_amendment_personal_data(amendment)
+      end
+    end
+
+    def scrub_amendment_personal_data(amendment)
+      PII_ATTRIBUTES_TO_DELETE.each do |attribute|
+        if amendment.claim_changes.key?(attribute)
+          amendment.claim_changes[attribute] = nil
+        end
+      end
```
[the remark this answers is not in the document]
lawrence-dxw: Nice idea! Took the idea of finding the relevant keys and went with this:

```ruby
attributes_to_scrub = PII_ATTRIBUTES_TO_DELETE.map(&:to_s) & amendment.claim_changes.keys
attributes_to_scrub.each { |attribute| amendment.claim_changes[attribute] = nil }
```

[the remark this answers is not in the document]
pezholio: Ah, even better 👍

spec/models/amendment_spec.rb
```diff
@@ -16,4 +16,195 @@
...
+        expect(claim.reload.amendments).to be_empty
+
+        expect(amendment.errors.keys).to eq([:claim_changes])
+      end
+    end
+
+    context "when updating an attribute that gets normalised in a before_save hook" do
+      let(:claim_attributes) do
+        {
+          bank_sort_code: "010203",
```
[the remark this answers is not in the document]
lawrence-dxw: You make an excellent point, it's a rather useless test as it is! Changed it.

spec/models/amendment_spec.rb
```diff
@@ -16,4 +16,195 @@
     amendment.notes = "Claimant made a typo"
     expect(amendment).to be_valid
   end
+
+  describe ".amend" do
```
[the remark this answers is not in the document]
lawrence-dxw: Thank you!

app/views/admin/amendments/new.html.erb
```diff
@@ -0,0 +1,170 @@
...
+    <h1 class="govuk-heading-xl">
+      Amend claim <%= @claim.reference %>
+    </h1>
+
+    <%=
+      if @amendment.errors.any?
+        render(
+          "shared/error_summary",
+          instance: @amendment,
+          errored_field_id_overrides: Amendment::AMENDABLE_ATTRIBUTES
```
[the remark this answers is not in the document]
lawrence-dxw: I've pulled it into a helper, which means I can do a bit of naming and commenting, but I'm still not sure how clear it is. Could you take a look?

spec/models/amendment_spec.rb
```diff
@@ -16,4 +16,195 @@
...
+
+        expect(claim.reload.amendments).to be_empty
+        expect(claim.teacher_reference_number).to eq("1234567")
+        expect(claim.payroll_gender).to eq("male")
+
+        expect(amendment.errors.keys).to match_array([:teacher_reference_number])
+      end
+    end
+
+    context "given invalid claim attributes and invalid amendment attributes" do
```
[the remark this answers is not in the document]
lawrence-dxw: :+1:

spec/features/admin_amend_claim_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+      student_loan_plan: :plan_1,
+      bank_sort_code: "010203",
+      bank_account_number: "47274828",
+      building_society_roll_number: "RN 123456",)
+  end
+  let(:date_of_birth) { 25.years.ago.to_date }
+  let(:service_operator) { create(:dfe_signin_user, given_name: "Jo", family_name: "Bloggs") }
+
+  scenario "Service operator amends a claim" do
+    travel_to Time.current
```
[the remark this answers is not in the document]
lawrence-dxw: I have _absolutely_ no idea why that's there! Killed it.

spec/features/admin_amend_claim_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    sign_in_to_admin_with_role(DfeSignIn::User::SERVICE_OPERATOR_DFE_SIGN_IN_ROLE_CODE, service_operator.dfe_sign_in_id)
+
+    visit admin_claim_url(claim)
+
+    click_on "Amend claim"
+
+    fill_in "Teacher reference number", with: "7654321"
+    select "Male", from: "Payroll gender"
+    new_date_of_birth = 30.years.ago.to_date
```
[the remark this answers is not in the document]
lawrence-dxw: 👍

spec/requests/admin_amendments_spec.rb
lawrence-dxw: 👍

app/controllers/admin/amendments_controller.rb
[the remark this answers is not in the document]
pezholio: Ah, OK, that's fair enough 👍

app/views/admin/amendments/new.html.erb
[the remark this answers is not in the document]
pezholio: I think it's a bit clearer, and takes the logic out of the view, so I'd say job done 👍

app/models/claim.rb
lawrence-forooghian: Makes sense. Changed it, and also added an explicit check that the claim hasn't had its personal data removed.

app/models/claim.rb
```diff
@@ -344,4 +348,8 @@ def determine_student_loan_plan
       Claim::NO_STUDENT_LOAN
     end
   end
+
+  def passed_scheduled_payment_date?
+    scheduled_payment_date < Time.zone.now if scheduled_payment_date
```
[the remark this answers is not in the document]
lawrence-forooghian: :+1: changed

spec/requests/admin_claims_spec.rb
lawrence-forooghian: Yeah, I've added a feature spec for it. Didn't actually need to go to the level of checking for individual tags; the `have_content` matcher does a pretty good job of smooshing everything together so I can just do

```ruby
expect(page).to have_content("Teacher reference number\nchanged from 7654321 to 1234567")
expect(page).to have_content("Bank account number\nchanged")
expect(page).not_to have_content("Bank account number\nchanged from")
```

app/models/claim/pii_scrubber.rb
```diff
@@ -47,5 +53,21 @@ def old_claims_rejected_or_paid
...
+    def scrub_amendments_personal_data(claim)
+      claim.amendments.each do |amendment|
+        scrub_amendment_personal_data(amendment)
+      end
+    end
+
+    def scrub_amendment_personal_data(amendment)
+      attributes_to_scrub = PII_ATTRIBUTES_TO_DELETE.map(&:to_s) & amendment.claim_changes.keys
+      personal_data_mask = attributes_to_scrub.to_h { |attribute| [attribute, nil] }
+      amendment.claim_changes.merge!(personal_data_mask)
```
pezholio: 👍

app/models/claim.rb
```diff
@@ -343,4 +357,8 @@ def determine_student_loan_plan
...
+  end
```
jacksonj04: I don't think this method is used any more

spec/requests/admin_claims_spec.rb
pezholio: Sweet 👍

app/models/claim.rb
```diff
@@ -343,4 +357,8 @@ def determine_student_loan_plan
       Claim::NO_STUDENT_LOAN
     end
   end
+
+  def passed_scheduled_payment_date?
```
tekin: Is this method still being used?

app/models/claim.rb
lawrence-forooghian: Thanks!

app/models/claim.rb
lawrence-forooghian: Nope, thanks for catching it.

app/models/claim.rb
```diff
@@ -262,6 +272,10 @@ def school
     eligibility&.current_school
   end
 
+  def amendable?
+    submitted? && !decision&.rejected? && !payment && !pii_removed?
```
jacksonj04: I think that `!payment` is checking that it hasn't been _payrolled_, not that it hasn't been _paid_. I'm pretty sure this is actually correct behaviour (ie to make an amendment someone must first remove the claim from a payroll run), but just so we're all clear.

lawrence-forooghian: Yep, thanks for clarifying - realised that's what you meant. The commit message for this change makes it clear.

lawrence-forooghian:
I did some user testing with a claims checker for this feature. We tested the case of changing a TRN because the claimant contacted us after realising they got it wrong. This is a case that they saw in real life recently.

- When filling in the change notes, they said “what would be good would be to have the name”. They then realised that it was personal data so that it shouldn’t go in the notes. I wonder if we can do better on the guidance there.

- It’s not clear to them much they should write in the change notes. For example, would we expect “details didn’t match DQT, so we contacted claimant, who told us they’d got their TRN wrong and gave us the right one”

- They pointed out that it’s quite easy to accidentally amend two things. They didn’t think it was a big risk. I asked whether he thought people would take a look at the timeline after making an amendment, to check the change they’d just made. They said yes, but I’m not sure if that’s only once I’d pointed this out as an option,

- They said it would be useful for the landing page for a claim to mention whether the claim has been amended. So we might want to add it to the tasks page for maths and physics.

@titlescreen, I think we should go through these and decide whether we want to try and address any of them before an initial release of the feature.

They also asked a good question - if a claimant gets their TRN wrong and hence their data isn't in the DQT export, how do we check their claim? I think this is outside the scope of this feature, but it's one I'll raise with Phil and Kristi.

PR #902, opened 2020-02-26
Update Admin to show QTS award year answer based on when claim was made

tekin:
This is the second part of the work to support future claim windows. It updates the back-office screens so that the QTS award year displayed to claim checkers will reflect the year the claim was made. For example, an eligible Maths & Physics claim made this year (2019/2020) will show the claimant's answer as "In or after the academic year 2014 to 2015", whilst a Maths & Physics claim made next year (2020/2021) will show "In or after the academic year 2015 to 2016".

### DIFF
--- a/app/models/maths_and_physics/admin_checks_presenter.rb
+++ b/app/models/maths_and_physics/admin_checks_presenter.rb
@@ -1,4 +1,11 @@
 module MathsAndPhysics
+  # Used to display the information a claim checker needs to check to either
+  # approve or reject a claim
+  #
+  # Note this presenter is only intented for use with eligible claims and
+  # therefor makes certain assumptions about the claim and eligibility.
+  # Specifically it assumes the QTS question was answered with
+  # :on_or_after_cut_off_date.
   class AdminChecksPresenter
     include Admin::PresenterMethods
 
@@ -10,7 +17,7 @@ module MathsAndPhysics
 
     def qualifications
       [].tap do |a|
-        a << ["Award year", I18n.t("maths_and_physics.questions.qts_award_years.#{eligibility.qts_award_year}")]
+        a << ["Award year", qts_award_year_answer]
         a << ["ITT subject", (eligibility.initial_teacher_training_subject_specialism || eligibility.initial_teacher_training_subject).humanize]
         a << ["Maths or Physics degree", maths_or_physics_degree] if eligibility.has_uk_maths_or_physics_degree.present?
       end
@@ -34,6 +41,11 @@ module MathsAndPhysics
       elsif eligibility.has_uk_maths_or_physics_degree == "has_non_uk"
         "Non-UK Maths or Physics degree"
       end
+    end
+
+    def qts_award_year_answer
+      qts_cut_off_for_claim = MathsAndPhysics.first_eligible_qts_award_year(claim.academic_year)
+      I18n.t("answers.qts_award_years.on_or_after_cut_off_date", year: qts_cut_off_for_claim.to_s(:long))
     end
   end
 end
--- a/app/models/maths_and_physics/admin_checks_presenter.rb
+++ b/app/models/maths_and_physics/admin_checks_presenter.rb
@@ -1,11 +1,6 @@
 module MathsAndPhysics
   # Used to display the information a claim checker needs to check to either
-  # approve or reject a claim
-  #
-  # Note this presenter is only intented for use with eligible claims and
-  # therefor makes certain assumptions about the claim and eligibility.
-  # Specifically it assumes the QTS question was answered with
-  # :on_or_after_cut_off_date.
+  # approve or reject a claim.
   class AdminChecksPresenter
     include Admin::PresenterMethods
 
@@ -17,7 +12,7 @@ module MathsAndPhysics
 
     def qualifications
       [].tap do |a|
-        a << ["Award year", qts_award_year_answer]
+        a << ["Award year", eligibility.qts_award_year_answer]
         a << ["ITT subject", itt_subject]
         a << ["Maths or Physics degree", maths_or_physics_degree] if eligibility.has_uk_maths_or_physics_degree.present?
       end
@@ -45,11 +40,6 @@ module MathsAndPhysics
       elsif eligibility.has_uk_maths_or_physics_degree == "has_non_uk"
         "Non-UK Maths or Physics degree"
       end
-    end
-
-    def qts_award_year_answer
-      qts_cut_off_for_claim = MathsAndPhysics.first_eligible_qts_award_year(claim.academic_year)
-      I18n.t("answers.qts_award_years.on_or_after_cut_off_date", year: qts_cut_off_for_claim.to_s(:long))
     end
   end
 end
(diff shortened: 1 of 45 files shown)


app/models/maths_and_physics/admin_checks_presenter.rb
```diff
@@ -1,4 +1,11 @@
 module MathsAndPhysics
+  # Used to display the information a claim checker needs to check to either
+  # approve or reject a claim
+  #
+  # Note this presenter is only intented for use with eligible claims and
+  # therefor makes certain assumptions about the claim and eligibility.
```
lawrence-forooghian: "therefore"

tekin: This documentation gets removed in a later refactor so maybe we can let it slide 😬

app/models/maths_and_physics/eligibility.rb
```diff
@@ -45,6 +45,7 @@ class Eligibility < ApplicationRecord
       on_or_after_cut_off_date: 1,
     }, _prefix: :awarded_qualified_status
 
+    has_one :claim, as: :eligibility, inverse_of: :eligibility
```
lawrence-forooghian: It would be more consistent with your earlier commit "Switch AdminChecksPresenters to accept claim" if you changed the `…::EligibilityAdminAnswersPresenter` classes to accept a claim instead of an eligibility, instead of adding this new association so that you can get to a claim from an eligibility. What do you think?

app/models/maths_and_physics/eligibility.rb
```diff
@@ -106,6 +107,15 @@ def initial_teacher_training_specialised_in_maths_or_physics?
       itt_subject_maths? || itt_subject_physics? || itt_specialism_physics?
     end
 
+    # Returns a String representing the answer given for the QTS question when
+    # the claim was made.
+    def qts_award_year_answer
```
lawrence-forooghian: I agree that this is definitely a presentation method. I don‘t feel very comfortable with it being in the model (and off the top of my head I can‘t think of a precedent for this in the codebase).

If the aim of this is to have presentation logic that can be shared between multiple presenters, you could consider adding a `PresenterMethods` module which is shared by all presenters (or put it in a helper and include it in the presenters). We already have an `Admin::PresenterMethods` module so there’s precedent for it.

spec/models/claim/pii_scrubber_spec.rb
```diff
@@ -6,26 +6,26 @@
   it "does not delete details from a submitted claim" do
     claim = create(:claim, :submitted, updated_at: over_two_months_ago)
 
-    expect { Claim::PiiScrubber.new.scrub_completed_claims }.not_to change { claim.reload.attributes }
+    expect { Claim::PiiScrubber.new.scrub_completed_claims }.not_to change { claim.reload.attributes_before_type_cast }
```
lawrence-forooghian: Looking at this it’s not clear to me why we‘ve had to switch from `attributes` to `attributes_before_type_cast`. I’m assuming that it’s because whatever notion of equality that the `change` matcher uses is struggling with the custom type. But if that’s the case, I think it would be good to validate that assumption and explain it in the commit message at least.

But, are we sure that we can’t get this working just with `attributes`? Because future developers will, I think, expect to be able to use `attributes`, and it would be unfortunate if they have to figure out each time they try to write a test like this that they need to use `attributes_before_type_cast` to make it work.

app/models/maths_and_physics/admin_checks_presenter.rb
lawrence-forooghian: :+1:

app/models/maths_and_physics/eligibility.rb
tekin: Consistent in one way (all the presenters would accept a claim), but at the expense of encapsulation. The non-"checks" presenters are all "eligibility" presenters. They are responsible for presenting an eligibility to a user. It would make for a strange API for them to accept a claim to do so. Besides which, in the final version of the code these presenters do not actually call any method on the claim itself, and instead only talk to the eligibility. This relationship primarily exists so that the eligibility can see which year academic claim was made and therefor can use that to determine the QTS cut-off that was used for the QTS question. More on this in the comment below...

app/models/maths_and_physics/eligibility.rb
tekin: Yeah, I prevaricated a lot about introducing this method but felt on balance the system is safer and more easily worked with having it here and not elsewhere.

I don't believe the aim is to _"have presentation logic that can be shared between multiple presenters"_. Yes, that's a benefit, but for me the primary aim is encapsulate the behaviour such that we can ask an eligibility: "what was the user response to the QTS question?" and for it to give us an answer _based on the year the claim was made_, without us having to jump through hoops using presenters, and/or helper methods in shared modules. I get that this doesn't follow existing precedence (e.g. we convert boolean values to "yes"/"no" responses in the presenters), but this feels different to those as this encapsulates some complex and important behaviour that brings together: the year the claim was made; the rules for eligibility for the specific policy; the actual stored value representing the users response. My worry is that having this behaviour live in some shared "presentation" code puts it somewhere that isn't immediately obvious, makes it harder to use, and also makes the system more difficult to understand and reason about. 

As a bar minimum the eligibility should be able to report the before/after state for the question, plus the resulting year value that is needed to rehydrate the sentence the user chose (i.e. the combination of `qts_award_year` and `year_for_answer` here in this method). I guess we could have a method that returns those two values to represent the answer given, and then have the actual translation into a human-readable string happen in the presenters. That would make for a bit of a weird method though as it would be returning a pretty abstract data structure that will only make sense and ever be use in a very specific context: turning it into the human-readable string.

spec/models/claim/pii_scrubber_spec.rb
tekin: So, yes, the specs fail when we do a straight comparison of the attributes, which I assumed was down to the use of our custom type as that was the only thing that changed, but to test the assumption more fully I've dug into the code for the matcher and it looks like it _should_ pass with the typecasting as the `RSpec::Support::FuzzyMatcher` that gets used under the hood doesn't complain about attributes with the type-cast attribute. Will dig further...

app/models/maths_and_physics/eligibility.rb
lawrence-forooghian: Yeah, I can't think of a better way of representing the answer - as you say, a single-use data structure, which would probably just have an API of "give me a human-readable answer" wouldn't be any more useful really.

It's interesting that this is the first time that we've had to solve a problem like this - reconstructing an answer. Until now we've been able to store the user's answer directly in the database. But it's the first dynamic question, so makes sense.

One suggestion: update the comment to make it clear that the string returned is not just a string "representing" the answer - which could mean something machine-readable - but is in fact a human-readable description of the answer.

app/models/maths_and_physics/eligibility.rb
lawrence-forooghian: Makes sense - I hadn‘t noticed that in the end we‘re only calling methods on the eligibility anyway.

spec/models/claim/pii_scrubber_spec.rb
tekin: Re the attribute type-casting, figured it out. It looks like RSpec uses `Object#hash` to do object comparison with nested objects – https://github.com/rspec/rspec-expectations/pull/1034. Defining `AcademicYear#hash` and have it return a consistent value the same academic year removes the need for this change.

app/models/maths_and_physics/eligibility.rb
tekin: 👍

PR #903, opened 2020-02-26
Use a shared partial for QTS ineligibility

tekin:
The actual QTS question template is already shared between the policies so it makes sense to also share the ineligibility reason. This reduces the service area of changes for future policies, which are all likely to include a QTS question.


app/mailers/claim_mailer.rb
```diff
@@ -13,7 +13,7 @@ def approved(claim)
 
   def rejected(claim)
     @claim_description = claim_description(claim)
-    @possible_rejection_reasons = I18n.t("#{claim.policy.routing_name.underscore}.possible_rejection_reasons", qts_year: ineligible_qts_year(claim))
+    @possible_rejection_reasons = I18n.t("#{claim.policy.locale_name}.possible_rejection_reasons", qts_year: ineligible_qts_year(claim))
```
lawrence-forooghian: "locale name" makes me think that it will be the name of a locale e.g. `en`, `es`, `pt`. Something that makes it clear it's a key in the locales file might be better, e.g. `i18n_key`.

tekin: How about `locale_key`?

lawrence-forooghian: Works for me 👍

PR #904, opened 2020-02-27
Remove `linuxFxVersion` from app service

pezholio:
The presence of `linuxFxVersion` (which is the Docker image) is telling the app service to deploy straight away. If we remove this, it keeps the previous version running in the background and only pulls down the latest version and starts the container once the slots have been swapped.


pezholio:
Good call - added that too 👍

PR #914, opened 2020-03-02
Simplify log formatting

pezholio:
While attempting to diagnose what's happening with our logs, I've noticed other teams that are using `LogStashLogger` are appending extra info into their logs using LogStashLogger, rather than via SemanticLogger. I'm also piping the logs to stdout as well, which will make it easier to see when we have discrepancies between what's being logged and what it appearing in Logit.


lawrence-forooghian:
Oh nice, didn't know you could do this 👍

PR #917, opened 2020-03-02
Allow claims to have multiple decisions

jacksonj04:
This lays the groundwork for decisions being undoable (setting the claim back to an undecided status) and allowing a subsequent change of decision whilst still retaining a complete history of decision changes.

The new `decision` method returns the latest decision on a `Claim`, and is functionally identical to current behaviour.

There is one notable pair of changes to specs, which is the removal of the spec ensuring payroll runs cannot be generated when there would be two claims containing the same personal details but different bank numbers. Reaching this point should not be possible, as claims with matching details but different bank numbers cannot be approved. This behaviour is now explicitly tested in `decision_spec.rb` to prevent regressions.

The original test now fails where it previously passed since `:approved` and `:rejected` traits in the claims factory now use `create` instead of `build`. This change ensures tests pass where previously they would fail due to reliance on the `decision` relation within a `Claim` returning values even when unsaved.

There should be no user-facing changes or changes to behaviour as a result of this pull request.

### DIFF
--- a/app/controllers/admin/decisions_controller.rb
+++ b/app/controllers/admin/decisions_controller.rb
@@ -11,7 +11,7 @@ class Admin::DecisionsController < Admin::BaseAdminController
   end
 
   def create
-    @decision = @claim.build_decision(decision_params.merge(created_by: admin_user))
+    @decision = @claim.decisions.build(decision_params.merge(created_by: admin_user))
     if @decision.save
       send_claim_result_email
       RecordOrUpdateGeckoboardDatasetJob.perform_later([@claim.id])
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -88,7 +88,7 @@ class Claim < ApplicationRecord
   enum student_loan_courses: {one_course: 0, two_or_more_courses: 1}
   enum student_loan_plan: STUDENT_LOAN_PLAN_OPTIONS
 
-  has_one :decision
+  has_many :decisions
   has_many :checks
   has_many :amendments
 
@@ -161,9 +161,9 @@ class Claim < ApplicationRecord
 
   scope :unsubmitted, -> { where(submitted_at: nil) }
   scope :submitted, -> { where.not(submitted_at: nil) }
-  scope :awaiting_decision, -> { submitted.left_outer_joins(:decision).where(decisions: {claim_id: nil}) }
-  scope :approved, -> { joins(:decision).where("decisions.result" => :approved) }
-  scope :rejected, -> { joins(:decision).where("decisions.result" => :rejected) }
+  scope :awaiting_decision, -> { submitted.left_outer_joins(:decisions).where(decisions: {claim_id: nil}) }
+  scope :approved, -> { joins(:decisions).where("decisions.result" => :approved) }
+  scope :rejected, -> { joins(:decisions).where("decisions.result" => :rejected) }
   scope :approaching_decision_deadline, -> { awaiting_decision.where("submitted_at < ? AND submitted_at > ?", DECISION_DEADLINE.ago + DECISION_DEADLINE_WARNING_POINT, DECISION_DEADLINE.ago) }
   scope :passed_decision_deadline, -> { awaiting_decision.where("submitted_at < ?", DECISION_DEADLINE.ago) }
   scope :payrollable, -> { approved.where(payment: nil) }
@@ -192,6 +192,10 @@ class Claim < ApplicationRecord
 
   def approvable?
     submitted? && !payroll_gender_missing? && !decision_made? && !payment_prevented_by_other_claims?
+  end
+
+  def decision
+    decisions.last
   end
 
   def decision_made?
--- a/spec/factories/claims.rb
+++ b/spec/factories/claims.rb
@@ -56,12 +56,16 @@ FactoryBot.define do
 
     trait :approved do
       submitted
-      association(:decision, factory: [:decision, :approved], strategy: :build)
+      after(:build) do |claim|
+        claim.decisions.append(build(:decision, :approved))
+      end
     end
 
     trait :rejected do
       submitted
-      association(:decision, factory: [:decision, :rejected], strategy: :build)
+      after(:build) do |claim|
+        claim.decisions.append(build(:decision, :rejected))
+      end
     end
 
     trait :ineligible do
--- a/spec/models/claim_spec.rb
+++ b/spec/models/claim_spec.rb
@@ -426,6 +426,28 @@ RSpec.describe Claim, type: :model do
     end
   end
 
+  describe "approved" do
+    let!(:submitted_claims) { create_list(:claim, 5, :submitted) }
+    let!(:unsubmitted_claims) { create_list(:claim, 2, :submittable) }
+    let!(:approved_claims) { create_list(:claim, 5, :approved) }
+    let!(:rejected_claims) { create_list(:claim, 5, :rejected) }
+
+    it "returns approved claims" do
+      expect(subject.class.approved).to match_array(approved_claims)
+    end
+  end
+
+  describe "rejected" do
+    let!(:submitted_claims) { create_list(:claim, 5, :submitted) }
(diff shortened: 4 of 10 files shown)


spec/models/claim_spec.rb
```diff
@@ -654,34 +664,34 @@
 
   describe "#amendable?" do
     it "returns false for a claim that hasn’t been submitted" do
-      claim = build(:claim, :submittable)
+      claim = create(:claim, :submittable)
```
robbiepaul: Does this one need to be changed to `create`?

spec/models/claim_spec.rb
```diff
@@ -654,34 +664,34 @@
...
       expect(claim.amendable?).to eq(false)
     end
 
     it "returns true for a submitted claim" do
-      claim = build(:claim, :submitted)
+      claim = create(:claim, :submitted)
```
robbiepaul: Same question as above

spec/requests/admin_payroll_runs_spec.rb
```diff
@@ -32,23 +32,6 @@
 
         expect(response).to redirect_to(admin_payroll_run_path(payroll_run))
       end
-
-      context "when a payroll run contains claims from the same person with different bank details" do
-        it "doesn’t create a payroll run and shows an error" do
```
robbiepaul: I feel a little uneasy about losing this spec. We rely on other parts of the system preventing us from getting in this state, but those parts could change. However it's not likely to change so if leaving this spec in causes complication I'm happy for it to be removed.

app/controllers/admin/decisions_controller.rb
```diff
@@ -11,7 +11,7 @@ def new
   end
 
   def create
-    @decision = @claim.build_decision(decision_params.merge(created_by: admin_user))
+    @decision = Decision.new(decision_params.merge(claim: @claim, created_by: admin_user))
```
pezholio:
```suggestion
    @decision = @claim.decisions.build(decision_params.merge(created_by: admin_user))
```

spec/models/claim_spec.rb
```diff
@@ -426,6 +426,28 @@
     end
   end
 
+  describe "approved" do
+    let!(:submitted_claims) { create_list(:claim, 5, :submitted) }
+    let!(:unsubmitted_claims) { create_list(:claim, 2, :submittable) }
+    let!(:approved_claims) { create_list(:claim, 5, :approved) }
+    let!(:rejected_claims) { create_list(:claim, 5, :rejected) }
```
robbiepaul: Do we need to create this many claims to test this? I think we should aim to keep the number of database writes down in our test suite

spec/models/claim_spec.rb
```diff
@@ -426,6 +426,28 @@
...
+    it "returns approved claims" do
+      expect(subject.class.approved).to match_array(approved_claims)
+    end
+  end
+
+  describe "rejected" do
+    let!(:submitted_claims) { create_list(:claim, 5, :submitted) }
+    let!(:unsubmitted_claims) { create_list(:claim, 2, :submittable) }
+    let!(:approved_claims) { create_list(:claim, 5, :approved) }
+    let!(:rejected_claims) { create_list(:claim, 5, :rejected) }
```
robbiepaul: Same as above

spec/models/claim_spec.rb
```diff
@@ -426,6 +426,28 @@
...
   end
 
+  describe "approved" do
+    let!(:submitted_claims) { create_list(:claim, 5, :submitted) }
+    let!(:unsubmitted_claims) { create_list(:claim, 2, :submittable) }
+    let!(:approved_claims) { create_list(:claim, 5, :approved) }
+    let!(:rejected_claims) { create_list(:claim, 5, :rejected) }
+
+    it "returns approved claims" do
+      expect(subject.class.approved).to match_array(approved_claims)
```
robbiepaul: I think using `described_class.approved` or `Claim.approved` would be more explicit here, using subject feels like an unnecessary indirection

```suggestion
      expect(described_class.approved).to match_array(approved_claims)
```

app/models/claim.rb
```diff
@@ -161,9 +161,9 @@ class Claim < ApplicationRecord
 
   scope :unsubmitted, -> { where(submitted_at: nil) }
   scope :submitted, -> { where.not(submitted_at: nil) }
-  scope :awaiting_decision, -> { submitted.left_outer_joins(:decision).where(decisions: {claim_id: nil}) }
-  scope :approved, -> { joins(:decision).where("decisions.result" => :approved) }
-  scope :rejected, -> { joins(:decision).where("decisions.result" => :rejected) }
+  scope :awaiting_decision, -> { submitted.left_outer_joins(:decisions).where(decisions: {claim_id: nil}) }
+  scope :approved, -> { joins(:decisions).where("decisions.result" => :approved) }
+  scope :rejected, -> { joins(:decisions).where("decisions.result" => :rejected) }
```
pezholio: With these two scopes, it's likely that claims that have been approved then subsequently rejected will show in both scopes. Is this something that will come later?

jacksonj04: Yes - the next PR in the chain introduces some far more robust testing around these when it introduces the idea of something being able to move back to undecided.

app/models/claim.rb
```diff
@@ -194,6 +194,10 @@ def approvable?
     submitted? && !payroll_gender_missing? && !decision_made? && !payment_prevented_by_other_claims?
   end
 
+  def decision
+    decisions.order(created_at: :desc).first
```
tekin: This is the same as:
```suggestion
    decisions.last
```

spec/factories/claims.rb
```diff
@@ -56,12 +56,16 @@
 
     trait :approved do
       submitted
-      association(:decision, factory: [:decision, :approved], strategy: :build)
+      after(:create) do |claim|
+        create(:decision, :approved, claim: claim)
```
tekin: We should avoid calling `create` in factories, otherwise it becomes impossible to instantiate an instance without it being written to the database:

```suggestion
      after(:build) do |claim|
        build(:decision, :approved, claim: claim)
```

tekin: Actually, you may need to assign it to the claim if the record isn't being saved... Something like this may work:

```suggestion
      after(:build) do |claim|
        claim.decisions.build(attributes_for(:decision, :approved))
```

spec/factories/claims.rb
```diff
@@ -56,12 +56,16 @@
...
+      end
     end
 
     trait :rejected do
       submitted
-      association(:decision, factory: [:decision, :rejected], strategy: :build)
+      after(:create) do |claim|
+        create(:decision, :rejected, claim: claim)
```
tekin: as above, let's avoid calling `create` if we can.

spec/models/amendment_spec.rb
```diff
@@ -18,7 +18,7 @@
   end
 
   it "is invalid if its claim is not amendable" do
-    amendment = build(:amendment, claim: build(:claim, :rejected))
+    amendment = build(:amendment, claim: create(:claim, :rejected))
```
tekin: I think this change becomes redundant if we move the factory to using `build` instead of `create`.

spec/models/claim_spec.rb
tekin: I'm with @robbiepaul. Can we write these as `Claim.approved` and avoid the indirection? It doesn't help comprehension and I can't see what the benefit is.

spec/models/claim_spec.rb
tekin: as above, if we change the factory we can hopefully leave these as they are.

spec/factories/claims.rb
jacksonj04: I've gone with `claim.decisions.append(build(:decision, :rejected, claim: claim))`. `claim.decisions.build` complains that `decisions is invalid`, which as far as I can tell is because it is returning a single `decision` rather than a list. There is no `build_list` available.

PR #918, opened 2020-03-02
Add CPU and memory alerts for worker container instance

robbiepaul:
Unfortunately `CpuPercentage` and `MemoryPercentage` metrics aren't
available for container instance monitoring. We have to use `CpuUsage` and
`MemoryUsage` instead, this means the threshold has to be a concrete
value rather than a proportion.

- CpuUsage is measured in millicores (1,000 per core)
- MemoryUsage is measured in bytes (1,000,000 per MB)

I've set the values as 800 millicores (80% of our 1 core) and
1200MB (80% of our 1500MB available memory).

We could parameterise these units in the container instance deploy and
pass them in to the app_alerts.json then use `[mul(value, 0.8)]` however
these values aren't likely to change so it feels like overkill.

https://docs.microsoft.com/en-gb/azure/azure-monitor/platform/metrics-supported#microsoftcontainerinstancecontainergroups

### DIFF
--- a/azure/resource_groups/alerts/template.json
+++ b/azure/resource_groups/alerts/template.json
@@ -31,6 +31,9 @@
     "appServicePlanId": {
       "type": "string"
     },
+    "workerContainerInstanceId": {
+      "type": "string"
+    },
     "vspAppServiceId": {
       "type": "string"
     },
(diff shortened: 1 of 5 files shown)


robbiepaul:
Opening as draft because I made a couple of changes that need verifying on the test environment before merging

robbiepaul:
This is working and ready for review

PR #925, opened 2020-03-03
Expose Git commit hash in Rollbar

lawrence-forooghian:
When investigating a previous error in Rollbar, I wanted to know which
version of the code was running when the error occurred. We can achieve
this by setting Rollbar’s `config.code_version`.

I think it makes sense to bake the Git commit hash into the Docker
image, since it provides metadata about which code is contained inside
the image and the context in which the image was built. I’ve put the
ENV instruction below the COPY instructions for caching reasons, since
the commit hash changes more frequently than the copied files.

I haven’t found a way to test this end to end. I’ve confirmed that the
`web` image built by CI, which gets pushed to Docker Hub, contains the
GIT_COMMIT_HASH environment variable, but I don't know if there's an
easy way to actually test that the exceptions triggered end up tagged
properly in Rollbar.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -109,6 +109,9 @@ RUN if [ ${RAILS_ENV} = "production" ]; then \
 
 EXPOSE 3000
 
+ARG GIT_COMMIT_HASH
+ENV GIT_COMMIT_HASH ${GIT_COMMIT_HASH}
+
 CMD [ "bundle", "exec", "rails", "server" ]
 
 # ------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


lawrence-forooghian:
I'd welcome opinions on whether we _need_ to test this end-to-end.

pezholio:
One way we could test this is in test - tag and push a Docker image and then deploy to test (we'd need to change the `DOCKER_IMAGE_TAG` in the deploy script). We can then raise an error manually and confirm it shows up in Rollbar.

lawrence-forooghian:
Ah, cool, I'll give that a go. And I think I can trigger an error manually by running `Rollbar.error` in a console.

lawrence-forooghian:
Tested it and updated the commit message.

PR #928, opened 2020-03-04
Accessibility testing admin area part 2

titlescreen:
This is the second PR for updating the admin area. It contains a series of commits mostly focused on screen reader improvements such as adding in visually hidden text within tables or links.

### DIFF
--- a/app/views/admin/policy_configurations/index.html.erb
+++ b/app/views/admin/policy_configurations/index.html.erb
@@ -30,7 +30,11 @@
             <th scope="row" class="govuk-table__header"><%= policy_service_name(policy_configuration.policy.routing_name) %></th>
             <td class="govuk-table__cell"><%= policy_configuration.current_academic_year %></td>
             <td class="govuk-table__cell"><%= policy_configuration.open_for_submissions? ? "Open" : "Closed" %></td>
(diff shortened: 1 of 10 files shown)


app/views/admin/policy_configurations/index.html.erb
```diff
@@ -30,7 +30,11 @@
             <th scope="row" class="govuk-table__header"><%= policy_service_name(policy_configuration.policy.routing_name) %></th>
             <td class="govuk-table__cell"><%= policy_configuration.current_academic_year %></td>
             <td class="govuk-table__cell"><%= policy_configuration.open_for_submissions? ? "Open" : "Closed" %></td>
-            <td class="govuk-table__cell"><%= link_to "Change", edit_admin_policy_configuration_path(policy_configuration), class: "govuk-link" %></td>
+            <td class="govuk-table__cell">
+              <%= link_to edit_admin_policy_configuration_path(policy_configuration), class: "govuk-link" do %>
+                Change <span class="govuk-visually-hidden"> <%= policy_service_name(policy_configuration.policy.routing_name) %></span>
```
pezholio: Probably doesn't make a _huge_ amount of difference, but there's two spaces here (one outside the span and one inside the span)

titlescreen: No worries dude I will get that fixed 👍

PR #929, opened 2020-03-04
Get the Azure deployment slots feature working

pezholio:
This updates our deployment script to deploy the Docker images that have been tagged with build numbers directly. 

Previously we'd pull down the image that has been tagged with a specific build number, and then tag it with `development` or `production` as part of our deployment process. This caused problems because once an infrastructure deploy had happened, this triggered a restart on our app services, which then pulled down the latest version of the code, rather than waiting for the slots to be swapped over. 

Doing it this way ensures that each version of the app service is tightly coupled to the version of the code, so changes don't sneak in too early.

**Once this is approved, I'll need to make changes to the deployment script before merging.**

### DIFF
--- a/bin/azure-deploy
+++ b/bin/azure-deploy
@@ -43,19 +43,16 @@ case $ENVIRONMENT_NAME in
   "development")
     SUBSCRIPTION_ID="8655985a-2f87-44d7-a541-0be9a8c2779d"
     RESOURCE_GROUP_PREFIX="s118d01"
-    DOCKER_IMAGE_TAG="development"
     VSP_DOCKER_IMAGE_TAG="20191030.1"
     ;;
   "test")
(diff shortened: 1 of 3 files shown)


bin/azure-deploy
```diff
@@ -105,6 +105,12 @@ while [ "$2" ]; do
   shift
 done
 
+if [ -z "$DOCKER_TAG" ]; then
+  echo "Usage: $0 ENVIRONMENT_NAME --docker-tag=DOCKER_TAG"
```
robbiepaul: I wonder if this is going to add an extra burden on developers to find the build number when running the script outside of the pipeline. Should we keep `development` / `production` / `test` tags for this purpose? Or perhaps introduce a release pipeline for test?

pezholio: Yeah, that's a good shout 👍

PR #932, opened 2020-03-05
Run deployment as part of the CI pipeline

pezholio:
This commit refactors the CI pipeline so it has three distinct stages:

- The first stage is as before - build the Docker container and run tests
- The second stage then only runs on a push to the master branch, and deploys the app to development
- The third stage only runs when a tag has been pushed, and this deploys the app to production

Both deploy stages use the `azure-deploy-template.yml` file, which DRYsthings up quite a bit.

### DIFF
--- a/azure-deploy-template.yml
+++ b/azure-deploy-template.yml
@@ -1,0 +1,66 @@
+parameters:
+  azureSubscription:
+  resourceGroupPrefix:
+  environmentName:
+  slackWebhookUrl:
+
+jobs:
+  - job: run_deploy_script
+    displayName: Run deploy script
+    steps:
+      - task: UseRubyVersion@0
+        displayName: 'Use Ruby >= 2.4'
+      - task: AzureCLI@1
+        displayName: 'Deploy Azure'
+        inputs:
+          azureSubscription: ${{parameters.azureSubscription}}
+          scriptPath: '$(System.DefaultWorkingDirectory)/Build/repository/bin/azure-deploy'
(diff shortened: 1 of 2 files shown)


pezholio:
[Define release pipeline using YAML](https://trello.com/c/oY6HDyg0/1409-define-release-pipeline-using-yaml)

azure-deploy-template.yml
```diff
@@ -0,0 +1,66 @@
...
+  - job: send_slack_notification
+    dependsOn: clean_up
+    displayName: Send Slack notification
+    steps:
+      - task: Bash@3
+        displayName: 'Send Slack Notification'
+        inputs:
+          targetType: filePath
+          filePath: './$(System.DefaultWorkingDirectory)/Build/repository/bin/slack-alert'
+          arguments: 'test $(Build.BuildNumber) $(Release.ReleaseId) ${{parameters.slackWebhookUrl}} SUCCESS'
```
robbiepaul: Do we consider the `slackWebhookUrl` a secret? I'm not sure, I haven't done much with them before. 

If we do we'd probably want to make it a secret variable in the same way we do the Docker password (https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#secret-variables)

pezholio: Ah, good catch. I think it is _relatively_ sensitive in that people could spam the channel if they had it, so it's best to keep it a secret.

robbiepaul: Ah I see, cool. The risk is relatively low at the moment because with out current setup only people with access can see the output of the pipeline

PR #934, opened 2020-03-05
Student Loans Financial year varies based on the currently configured academic year

tekin:
Each academic year when the claim window re-opens, claimants will be claiming for student loans paid in a new financial year. This updates the service so that the content (questions and ineligibility reasons) the user sees reflects the year that they are currently claiming against.

We're now in a position where _almost_ all the questions for student loans have defined helper methods. It may make sense in future work to make it so all questions have defined helpers, rather than a mixture of helper and direct access of locale values, but I think if we decide to do that it should be as a separate and smaller PR.

### DIFF
--- a/app/mailers/claim_mailer.rb
+++ b/app/mailers/claim_mailer.rb
@@ -2,40 +2,43 @@ class ClaimMailer < ApplicationMailer
   helper :application
 
   def submitted(claim)
-    @claim_description = claim_description(claim)
-    view_mail_with_claim_and_subject(claim, "Your claim #{@claim_description} has been received, reference number: #{claim.reference}")
+    @claim = claim
+    @claim_description = claim_description
+    view_mail_with_claim_and_subject("Your claim #{@claim_description} has been received, reference number: #{claim.reference}")
   end
 
   def approved(claim)
-    @claim_description = claim_description(claim)
-    view_mail_with_claim_and_subject(claim, "Your claim #{@claim_description} has been approved, reference number: #{claim.reference}")
+    @claim = claim
+    @claim_description = claim_description
+    view_mail_with_claim_and_subject("Your claim #{@claim_description} has been approved, reference number: #{claim.reference}")
   end
 
   def rejected(claim)
-    @claim_description = claim_description(claim)
-    @possible_rejection_reasons = I18n.t("#{claim.policy.locale_key}.possible_rejection_reasons", qts_year: ineligible_qts_year(claim))
-    view_mail_with_claim_and_subject(claim, "Your claim #{@claim_description} has been rejected, reference number: #{claim.reference}")
+    @claim = claim
+    @claim_description = claim_description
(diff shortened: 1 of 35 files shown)


app/mailers/claim_mailer.rb
```diff
@@ -2,40 +2,43 @@ class ClaimMailer < ApplicationMailer
   helper :application
 
   def submitted(claim)
-    @claim_description = claim_description(claim)
-    view_mail_with_claim_and_subject(claim, "Your claim #{@claim_description} has been received, reference number: #{claim.reference}")
+    @claim = claim
+    @claim_description = claim_description
+    view_mail_with_claim_and_subject("Your claim #{@claim_description} has been received, reference number: #{claim.reference}")
```
pezholio: It doesn't make much difference, but I think it might be neater to refer to the `@claim` instance variable here (and in other calls to `view_mail_with_claim_and_subject`) now we have the reference.

spec/models/student_loans_spec.rb
```diff
@@ -22,4 +22,14 @@
       expect(StudentLoans.first_eligible_qts_award_year(AcademicYear.new(2030))).to eq AcademicYear.new(2019)
     end
   end
+
+  describe ".current_financial_year" do
+    it "returns a human-friendly string for the financial year the policy is currently accepting claims for" do
+      expect(policy_configuration.current_academic_year).to eq AcademicYear.new("2025/2026")
```
pezholio: I'm a bit confused as to _why_ this returns that particular Academic Year, as there doesn't seem to be any specific teardown/setup in this test. Can we make this a bit clearer?

spec/mailers/claim_mailer_spec.rb
```diff
@@ -107,7 +107,15 @@
       let(:mail) { ClaimMailer.rejected(claim) }
 
       it "mentions “the eligible-school during the financial year” reason" do
-        expect(mail.body.encoded).to include("you did not teach at an eligible school between 6 April 2018 and 5 April 2019")
+        expect(policy_configurations(:student_loans).current_academic_year).to eq AcademicYear.new("2025/2026")
```
pezholio: Again, I'm not sure why the academic year is set to be this year. Can we have something that specifically sets it first?

spec/models/student_loans_spec.rb
tekin: It's the currently-configured academic year [as defined the fixtures for policy configurations](https://github.com/DFE-Digital/dfe-teachers-payment-service/blob/master/spec/fixtures/policy_configurations.yml#L3).

So it's less an expectation on behaviour and more an assertion on the current state of the system/data and also tells the reader what the currently configured year is so they understand how the expected value for financial year relates to it. I can see though that it's not particularly clear where it's coming from though. I've replaced this and the other expectation with a comment instead.

pezholio: Cool 👍

PR #938, opened 2020-03-05
Fix Slack deployment failure notifications

pezholio:
Previously, both Slack notification steps were wrapped in one job, which had a `dependsOn` of the previous job. This meant that the job would only run if `clean_up` had run successfully. Moving the failure message to its own job with a `condition` of `failed()` means this job only runs if any of the previous jobs have failed.


pezholio:
[Deployment failures should notify the team](https://trello.com/c/S9NtZwHs/1413-deployment-failures-should-notify-the-team)

robbiepaul:
I think we should close this while we work on the release pipeline

pezholio:
Read my mind 😄

PR #943, opened 2020-03-09
Exclude claims for the other policy from claims with matching details

pezholio:
We've had reports of a number of false positives when searching for matching attributes, because a number of applicants are eligible for both policies.

This adds the source claim's policy to the `claims_to_compare` query, ensuring that only claims for the source claim's policy are returned when searching for potential duplicate claims.

I've also removed the optional `claims_to_compare` parameter as it's not used anywhere, and complicates the class somewhat without any real benefit.

This also comes with some slight copy tweaks, which should make things easier to understand:

image

image

### DIFF
--- a/app/models/claim/matching_attribute_finder.rb
+++ b/app/models/claim/matching_attribute_finder.rb
@@ -7,13 +7,19 @@ class Claim
       ["bank_account_number", "bank_sort_code", "building_society_roll_number"]
     ].freeze
 
-    def initialize(source_claim, claims_to_compare = Claim.submitted)
+    def initialize(source_claim)
       @source_claim = source_claim
-      @claims_to_compare = claims_to_compare
     end
 
+    # Returns a list of claims that could potentially be from the same applicant
+    # because they either share a same single attribute with the source claim,
+    # (for example, the same TRN, or email), or they share a group of attributes
+    # with the source claim (for example bank sort code, account number and roll
+    # number).
+    #
+    # This may not necessarily mean the claim cannot be approved, but means it
+    # warrants a degree of caution and further investigation.
     def matching_claims
-      claims = @claims_to_compare.where.not(id: @source_claim.id)
       match_queries = nil
 
       ATTRIBUTE_GROUPS_TO_MATCH.each do |attributes|
@@ -27,10 +33,16 @@ class Claim
         match_queries = match_queries.nil? ? query : match_queries.or(query)
       end
 
-      claims.merge(match_queries)
+      claims_to_compare.merge(match_queries)
     end
 
     private
+
+    def claims_to_compare
+      Claim.submitted
+        .by_policy(@source_claim.policy)
+        .where.not(id: @source_claim.id)
+    end
 
     def values_for_attributes(attributes)
       attributes.map { |attribute|
--- a/app/views/admin/claims/_claims_with_matching_details.html.erb
+++ b/app/views/admin/claims/_claims_with_matching_details.html.erb
@@ -1,5 +1,7 @@
 <table id="claims-with-matches" class="govuk-table govuk-!-margin-bottom-9">
-  <caption class="govuk-table__caption govuk-heading-l">Claims with matching details</caption>
+  <caption class="govuk-table__caption govuk-heading-l">
+    <%= @claim.policy.short_name %> <%= "claim".pluralize(@matching_claims.count) %> with matching details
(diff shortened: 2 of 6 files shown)


pezholio:
[Exclude claims for the other policy from "claims with matching details"](https://trello.com/c/yVAX5u9m/1408-exclude-claims-for-the-other-policy-from-claims-with-matching-details)

spec/models/claim/matching_attribute_finder_spec.rb
```diff
@@ -2,24 +2,27 @@
...
+    let(:source_claim) { create(:claim, claim_attributes) }
+
+    let!(:matching_claim_with_different_policy) { create(:claim, :submitted, claim_attributes.merge(policy: MathsAndPhysics)) }
     let!(:claim_with_no_matching_attributes) { create(:claim, :submitted) }
     let!(:unsubmitted_claim_with_matching_teacher_reference_number) { create(:claim, :submittable, teacher_reference_number: source_claim.teacher_reference_number) }
 
     subject(:matching_claims) { Claim::MatchingAttributeFinder.new(source_claim).matching_claims }
 
-    it "does not include the source claim, or claims that do not match, or claims that are not submitted" do
+    it "does not include the source claim, or claims that do not match, or claims that are not submitted, or claims that match, but have a different policy" do
```
tekin: This spec has got a bit smelly – it's trying to communicate too many things in a single spec/line. How about we break out three focused specs for each? The three records that are being created in the `let!` calls could be created inline on each of the specs instead of being created for every single spec in this file.

app/views/admin/claims/_claims_with_matching_details.html.erb
```diff
@@ -1,5 +1,7 @@
 <table id="claims-with-matches" class="govuk-table govuk-!-margin-bottom-9">
-  <caption class="govuk-table__caption govuk-heading-l">Claims with matching details</caption>
+  <caption class="govuk-table__caption govuk-heading-l">
+    <%= t("admin.duplicate_attributes_title", count: @matching_claims.count, policy: @claim.policy.short_name) %>
```
tekin: This one is simple enough that maybe we should avoid the indirection of the locale file and inline it instead?

```suggestion
    <%= @claim.policy.short_name %> <%= "claim".pluralize(@matching_claims.count) %> with matching details
```

spec/models/claim/matching_attribute_finder_spec.rb
pezholio: 👍 Sounds like a plan

spec/models/claim/matching_attribute_finder_spec.rb
```diff
@@ -2,24 +2,41 @@
...
+    end
+
+    it "does not include unsubmitted claims with matching attributes" do
+      create(:claim, :submittable, teacher_reference_number: source_claim.teacher_reference_number)
+
+      expect(matching_claims).to be_empty
+    end
+
+    it "does not include claims that match, but have a different policy" do
+      create(:claim, :submitted, claim_attributes.merge(policy: MathsAndPhysics))
```
tekin: All the other specs grab an attribute from `source_claim` so this appears to be the only place where `claim_attributes` is used aside from when `source_claim` is created. Which is making me wonder if it's worth it to extract out the attributes. Could this spec not do the same as the others and grab the TRN from source_claim? It would mean one less instance variable to understand and trace through the specs.

app/models/claim/matching_attribute_finder.rb
```diff
@@ -7,13 +7,19 @@ class MatchingAttributeFinder
...
     end
 
+    # Returns a list of claims that could potentially be from the same applicant
+    # because they either share a same single attribute with the source claim,
+    # (for example, the same TRN, or email), or they share a group of attributes
+    # with the source claim (for example bank sort code, account number and roll
+    # number).
+    #
+    # This may not necessarily mean the claim cannot be approved, but means it
+    # warrants a degree of caution and further investigation.
```
tekin: 👍🏽

spec/models/claim/matching_attribute_finder_spec.rb
pezholio: Yeah, that makes sense, I guess I was trying to be super-clear about the fact that they were from the same person, but it'll show up in the matcher regardless of the number of attributes that match.

PR #948, opened 2020-03-09
Show incomplete checks warning on decision screen

lawrence-forooghian:
# Summary

This adds a list of incomplete checks to the `admin/decisions#new` screen. If a service operator goes to the decision screen before completing all the checks, they will see this warning.

More details in commit messages.

# Screenshot

image

### DIFF
--- a/spec/models/claim_spec.rb
+++ b/spec/models/claim_spec.rb
@@ -717,4 +717,18 @@ RSpec.describe Claim, type: :model do
       expect(claim.amendable?).to eq(false)
     end
   end
+
+  describe "#incomplete_check_names" do
+    Policies.all.each do |policy|
+      it "returns an array of the checks that haven’t been completed on the claim" do
+        claim = build(:claim, :submitted, checks: [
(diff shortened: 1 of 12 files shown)


spec/models/claim_spec.rb
```diff
@@ -717,4 +717,33 @@
       expect(claim.amendable?).to eq(false)
     end
   end
+
+  describe "#incomplete_check_names" do
```
tekin: I find specs where there are multiple contexts that only have a single example in them hard to read, and the fact that there's only single `it` blocks in each context makes me think they are redundant because the context isn't being shared.

Could we simplify the entire thing into a single `it "returns an array of the checks that haven't been completed on the claim"`? In that example we'd verify the return value includes all the checks on a claim with no checks, and then add a check and verify the array now excludes the check that has been performed. It would communicate the expected behaviour more clearly and succinctly.

lawrence-forooghian: Yep, makes sense - I've changed that.

PR #950, opened 2020-03-10
Rename checks to tasks

tekin:
The verb "check" wasn't quite correct as not all of the steps taken to check a claim are an actual "check", for example, providing a missing payroll gender. This renames the model and also updates the UI to be more consistent with it's use of the word "task".

We'll want to ask claim checks to pause checking whilst we roll out this release as it's not been written in a way that would cope with requests being made whilst the database is being migrated and the code reloaded.


config/routes.rb
```diff
@@ -88,7 +88,7 @@ def matches?(request)
     get "/auth/failure", to: "auth#failure"
 
     resources :claims, only: [:index, :show] do
-      resources :checks, only: [:index, :show, :create], param: :check, constraints: {check: %r{#{Admin::ChecksController::CHECKS_SEQUENCE.join("|")}}}
+      resources :tasks, only: [:index, :show, :create], param: :task, constraints: {task: %r{#{Admin::TasksController::TASKS_SEQUENCE.join("|")}}}
```
pezholio:
```suggestion
      resources :tasks, only: [:index, :show, :create], param: :name, constraints: {name: %r{#{Admin::TasksController::TASKS_SEQUENCE.join("|")}}}
```

I'm wondering if it's better to call the `task` param `name`, as this better reflects the mapping in the database? It also means, when we get to the yes/no work, we can have a `task` param hash, which is more in line with Rails' defaults.

tekin: Makes sense, I'll update it.

tekin: Renamed, as requested: https://github.com/DFE-Digital/dfe-teachers-payment-service/compare/98e21787586d404cd621af73f55ca394039a2354..fe9802e2d27814c178aacfa63f5c241463fe235b

PR #951, opened 2020-03-10
1367 rename pii to personal data

LornaHarwood:
**Changes made:**

Replaced the term PII with 'personal data' to reflect the language of GDPR

### DIFF
--- a/app/jobs/delete_pii_from_old_claims_job.rb
+++ b/app/jobs/delete_pii_from_old_claims_job.rb
@@ -1,10 +1,10 @@
-# Runs weekly at 00:30 on a Sunday morning, and instructs the PiiScrubber class
+# Runs weekly at 00:30 on a Sunday morning, and instructs the PersonalDataScrubber class
 # to remove PII from eligible claims.
 class DeletePiiFromOldClaimsJob < CronJob
   self.cron_expression = "30 0 * * 0"
 
   def perform
     Rails.logger.info "Deleting PII from old claims which have been rejected or paid"
-    Claim::PiiScrubber.new.scrub_completed_claims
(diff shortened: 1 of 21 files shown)


lawrence-forooghian:
I've noticed a couple of issues with the changes to the `Claim` model when I tried to run the migrations. I think we should sit down and look at the whole PR again more carefully.

lawrence-forooghian:
This is ready for somebody to review now.

app/jobs/delete_pii_from_old_claims_job.rb
```diff
@@ -1,10 +0,0 @@
-# Runs weekly at 00:30 on a Sunday morning, and instructs the PiiScrubber class
-# to remove PII from eligible claims.
-class DeletePiiFromOldClaimsJob < CronJob
```
tekin: Because we have renamed this scheduled job we'll need to manually unschedule it by removing the `Delayed::Job` record from the database to prevent the worker trying to load a class that doesn't exist and raising errors. You'll be able to identify it by examining the `handler` attribute on the job records in the database, it'll contain the a string like `"job_class: DeletePiiFromOldClaimsJob"` in it.

lawrence-forooghian: Would you prefer that we do this in the console, or as a data migration in `db/data`?

tekin: Ooh, good question! Maybe a good opportunity to experience writing a data migration?

lawrence-forooghian: Added a data migration, could you take a look please?

PR #952, opened 2020-03-10
Add a task for checking the student loan amount on student loans claims

tekin:
This introduces a task to check the student loan amount on student loans claims. Now that we have this task, we can switch student loans claims to default to the new task-based claim checking interface.

This PR introduces a new `ClaimCheckingTasks` class that we can use to model the various checks that may or may not be possible on any given claim.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- Add a task for confirming the student loan amount on Student Loans claims
+
 ## [Release 060] - 2020-03-10
 
 - Fix question bug on school search results page
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog]
 ## [Unreleased]
 
 - Add a task for confirming the student loan amount on Student Loans claims
+- Student Loans claims default to new task-based claim checking interface
 
 ## [Release 060] - 2020-03-10
 
--- a/app/models/claim_checking_tasks.rb
+++ b/app/models/claim_checking_tasks.rb
@@ -16,4 +16,9 @@ class ClaimCheckingTasks
   def applicable_task_names
     TASK_NAMES
   end
+
+  # Returns an Array of tasks names that have not been completed on the claim.
+  def incomplete_task_names
+    applicable_task_names - claim.tasks.map(&:name)
+  end
 end
--- a/app/models/claim_checking_tasks.rb
+++ b/app/models/claim_checking_tasks.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+# This models the tasks that need to be performed on a claim as part of the
+# claim checking process.
+class ClaimCheckingTasks
+  TASK_NAMES = %w[qualifications employment].freeze
+
+  attr_reader :claim
+
+  def initialize(claim)
+    @claim = claim
+  end
+
+  # Returns an Array task names that need to be performed on the claim during
+  # claim checking before it should be approved.
+  def applicable_task_names
+    TASK_NAMES
+  end
+end
--- a/app/models/claim_checking_tasks.rb
+++ b/app/models/claim_checking_tasks.rb
@@ -3,7 +3,7 @@
 # This models the tasks that need to be performed on a claim as part of the
 # claim checking process.
 class ClaimCheckingTasks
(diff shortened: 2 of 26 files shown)


app/models/claim_checking_tasks.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  attr_reader :claim
+
+  def initialize(claim)
+    @claim = claim
+  end
+
+  # Returns an Array of all the tasks that need to be performed on the claim
+  # during claim checking before it should be approved.
+  def applicable_tasks
```
lawrence-forooghian: Reading the documentation and method name, I'd expect this method to return an array of `Task` instances. Perhaps we can be more precise with our words and distinguish between a task (`Task` instance) and a task name?

This is a point that applies more broadly across this PR but just going to mention it in one place until hearing your thoughts.

spec/models/claim_checking_tasks_spec.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe ClaimCheckingTasks do
+  let(:claim) { build(:claim) }
+  let(:checking_tasks) { ClaimCheckingTasks.new(claim) }
+
+  describe "#applicable_tasks" do
+    it "returns the checks that apply to the claim" do
```
lawrence-forooghian:
```suggestion
    it "returns the tasks that apply to the claim" do
```

app/views/admin/tasks/student_loan_amount.html.erb
```diff
@@ -0,0 +1,23 @@
+<% content_for(:page_title) { page_title("Claim #{@claim.reference} student loan amount check for #{policy_service_name(@claim.policy.routing_name)}") } %>
```
lawrence-forooghian: Looks like this gets added as a blank file in the commit "ClaimCheckingTask to model checking tasks". I think you probably meant to add it in a later commit instead.

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- Add a task for confirming the student loan amount on Student Loans claims
```
lawrence-forooghian: I think this line belongs in the previous commit.

config/locales/en.yml
```diff
@@ -211,3 +211,6 @@ en:
         employment:
           summary: "Check employment information"
           description: "Check the claimant's previous and current schools match the below information from their claim."
+        student_loan_amount:
+          summary: "Check student loan amount"
+          description: "Check  the claimant’s student loan amount and plan type match the information we hold about their loan."
```
lawrence-forooghian:
```suggestion
          description: "Check the claimant’s student loan amount and plan type match the information we hold about their loan."
```

spec/models/claim_checking_tasks_spec.rb
```diff
@@ -10,6 +10,13 @@
     it "returns the checks that apply to the claim" do
       expect(checking_tasks.applicable_tasks).to eq %w[qualifications employment]
     end
+
+    it "includes the a task for student loan amount for a StudentLoans claim" do
```
lawrence-forooghian:
```suggestion
    it "includes a task for student loan amount for a StudentLoans claim" do
```

config/routes.rb
```diff
@@ -88,7 +88,7 @@ def matches?(request)
     get "/auth/failure", to: "auth#failure"
 
     resources :claims, only: [:index, :show] do
-      resources :tasks, only: [:index, :show, :create], param: :name, constraints: {name: %r{#{Admin::TasksController::TASKS_SEQUENCE.join("|")}}}
+      resources :tasks, only: [:index, :show, :create], param: :name, constraints: {name: %r{#{ClaimCheckingTasks::TASK_NAMES.join("|")}}}
```
lawrence-forooghian: Another thought: currently visiting `admin/claims/:claim_id/tasks/student_loan_amount` for a maths and physics claim raises an exception. Should we be returning a 404, same way as we restrict claim journey slugs to those which are relevant to the policy?

tekin: I don't feel strongly that we need to do this. It feels like an edge case we could ignore because these pages are not public-facing and it would require a service operator to hack the URL to get to the page as it shouldn't be linked to from anywhere.

lawrence-forooghian: 👍

PR #954, opened 2020-03-11
Deploy to development in pipelines

pezholio:
This uses the Azure Pipeline to deploy as well as run the tests if the branch is master. This keeps the two things tightly coupled, and also allows us to keep our deployment steps tightly coupled to our code, rather than having them defined in Azure DevOps via a UI. This can also be adapted to deploy to production when we're ready to do so.

I've also put all our Docker-related tasks in a seperate bin script, so it's easier to follow, rather than having to parse a massive bunch of YAML.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -28,23 +28,30 @@ stages:
               git reset --hard
               git clean -xdf
             displayName: Clean repository
-      - job: build_and_push_web_dependencies
+      - job: docker_tasks
         dependsOn: run_tests
-        displayName: Build and push web dependencies
+        displayName: Docker tasks
         steps:
           - script: docker login -u $(dockerId) -p $pass
             env:
               pass: $(dockerPassword)
             displayName: Login to DockerHub
           - script: |
-              ./bin/build-docker-images "ref/heads/master" \
-                $(Build.SourceVersion) \
+              ./bin/build-docker-images $(Build.SourceVersion) \
                 $(dockerRegistry)/$(dockerImageName) \
                 $(Build.BuildNumber)
             displayName: Prepare and push Docker images
+            condition: eq(variables['Build.SourceBranch'], 'refs/heads/master')
+          - script: |
+              ./bin/cache-docker-images $(Build.SourceVersion) \
+                $(dockerRegistry)/$(dockerImageName) \
+                $(Build.BuildNumber)
+            displayName: Prepare Docker images for caching
+            condition: ne(variables['Build.SourceBranch'], 'refs/heads/master')
       - job: publish_repository
-        dependsOn: build_and_push_web_dependencies
+        dependsOn: docker_tasks
         displayName: Publish repository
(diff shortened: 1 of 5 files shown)


azure-pipelines.yml
```diff
@@ -6,106 +6,58 @@ pool:
...
+            env:
+              pass: $(dockerPassword)
+            displayName: Login to DockerHub
+          - script: |
+              ./bin/build-docker-images "ref/heads/master" \
+                $(Build.SourceVersion) \
+                $(dockerRegistry)/$(dockerImageName) \
+                $(Build.BuildNumber)
+            displayName: Prepare and push Docker images
+      - job: publish_repository
```
robbiepaul: Do we only want to publish the repository when merging to master?

bin/slack-alert
```diff
@@ -2,11 +2,11 @@
 
 ENVIRONMENT_NAME=$1
 BUILD_NUMBER=$2
-RELEASE_ID=$3
+BUILD_ID=$3
 WEBHOOK_URL=$4
 RESULT=$5
 
-RELEASE_URL="https://dev.azure.com/dfe-ssp/S118-Teacher-Payments-Service/_releaseProgress?_a=release-pipeline-progress&releaseId=$RELEASE_ID"
+RELEASE_URL="https://dfe-ssp.visualstudio.com/S118-Teacher-Payments-Service/_build/results?buildId=$BUILD_ID&view=results"
```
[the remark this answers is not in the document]
pezholio: The release happens as part of the build now (see https://dfe-ssp.visualstudio.com/S118-Teacher-Payments-Service/_build/results?buildId=50614&view=results for an example)

azure-pipelines.yml
pezholio: Yeah, this just publishes the artifact, which we need for deploy.

azure-pipelines.yml
```diff
@@ -6,106 +6,58 @@ pool:
...
+      - job: build_and_push_web_dependencies
+        dependsOn: run_tests
+        displayName: Build and push web dependencies
+        steps:
+          - script: docker login -u $(dockerId) -p $pass
+            env:
+              pass: $(dockerPassword)
+            displayName: Login to DockerHub
+          - script: |
+              ./bin/build-docker-images "ref/heads/master" \
```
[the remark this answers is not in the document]
pezholio: Good call 👍

bin/slack-alert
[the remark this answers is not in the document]
robbiepaul: Ahh ok gotcha

azure-pipelines.yml
robbiepaul: It doesn't have a `condition` though, so won't it publish the artifact on non-master builds?

pezholio: Oh gotcha. You're right 😄

pezholio:
I'm gonna take a look at seperating out the caching and building of Docker images, they do seem like separate things

PR #958, opened 2020-03-11
Add Yes/No option to claim checking tasks

pezholio:
This updates the claim checking tasks to have a boolean value of `passed` and allows users to set this via a Yes/No interface on each task. It also shows the state of each task on the task list page and has a data migration to update older tasks that have already been created to set an explicit `passed` state.

# Screenshots

image

image

### DIFF
--- a/app/controllers/admin/tasks_controller.rb
+++ b/app/controllers/admin/tasks_controller.rb
@@ -1,6 +1,7 @@
 class Admin::TasksController < Admin::BaseAdminController
   before_action :ensure_service_operator
   before_action :load_claim
+  before_action :ensure_task_has_not_already_been_completed, only: [:create]
 
   def index
     @claim_checking_tasks = ClaimCheckingTasks.new(@claim)
@@ -14,16 +15,25 @@ class Admin::TasksController < Admin::BaseAdminController
 
   def create
     @claim_checking_tasks = ClaimCheckingTasks.new(@claim)
-    @claim.tasks.create!(check_params)
-    redirect_to next_task_path
-  rescue ActiveRecord::RecordInvalid
-    redirect_to admin_claim_task_path(@claim, name: current_task_name), alert: "This task has already been completed"
+    @task = @claim.tasks.build(check_params)
+    if @task.save
+      redirect_to next_task_path
+    else
+      @tasks_presenter = @claim.policy::AdminTasksPresenter.new(@claim)
+      render current_task_name
+    end
   end
 
   private
 
   def load_claim
     @claim = Claim.includes(:tasks).find(params[:claim_id])
+  end
+
+  def ensure_task_has_not_already_been_completed
+    if @claim.tasks.find_by(name: current_task_name)
+      redirect_to admin_claim_task_path(@claim, name: current_task_name), alert: "This task has already been completed"
+    end
   end
 
   def current_task_name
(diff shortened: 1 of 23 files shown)


pezholio:
[Add yes/no option to claim checking tasks](https://trello.com/c/9akLKVFg/1407-add-yes-no-option-to-claim-checking-tasks)

spec/requests/admin_tasks_spec.rb
```diff
@@ -81,6 +81,16 @@
             end
           end
 
+          context "when a task's passed flag is not set" do
+            it "doesn't create a task and shows an error" do
+              expect {
+                post admin_claim_tasks_path(claim, name: "qualifications")
+              }.not_to change { Task.count }
```
tekin: I know all the other specs use `Task.count`, but I think we can be more specific here:
```suggestion
              }.not_to change { claim.tasks.count }
```

app/controllers/admin/tasks_controller.rb
```diff
@@ -42,4 +52,11 @@ def next_task_path
       new_admin_claim_decision_path(@claim)
     end
   end
+
+  def check_params
+    params.fetch(:task, {})
```
tekin: Because we always expect the params to arrive via a scoped `task` form we would be better off using the more standard `params.require(:task)` line here. We may need to add an `f.hidden_field :passed` line to the form to ensure we always get something posted, but that's a standard pattern that we're already employing in the yes/no forms in the public-facing site.

pezholio: Ah yeah, I put that in initally to keep the tests green, but forgot to take it out

app/helpers/admin/claims_helper.rb
```diff
@@ -69,6 +69,20 @@ def matching_attributes(first_claim, second_claim)
...
+
+      if task.present?
+        status = task_status(task)
+        tag_classes = "govuk-tag app-task-list__task-completed"
+      else
+        status = "Incomplete"
+        tag_classes = "govuk-tag app-task-list__task-completed govuk-tag--inactive"
+      end
+
+      content_tag("strong", status, class: tag_classes, id: "#{task_name}-status")
```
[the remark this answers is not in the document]
pezholio: Ah no, I initially intended to use it in tests, then decided against it

app/helpers/admin/claims_helper.rb
```diff
@@ -69,6 +69,20 @@ def matching_attributes(first_claim, second_claim)
       matching_attributes.to_h.compact.keys.map(&:humanize).sort
     end
 
+    def task_status_tag(claim, task_name)
+      task = claim.tasks.find_by(name: task_name)
```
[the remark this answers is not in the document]
pezholio: Ah, good call. I've updated the spec too, so we don't have to create the tasks/claim in the DB too.

app/controllers/admin/tasks_controller.rb
tekin: This is still using `fetch`, were you going to change it to `require`?

pezholio: Arrrgh, dammit. Yep

pezholio: Sorted, also added another spec to make sure it works with the hidden field in play

PR #963, opened 2020-03-12
Add the claims with matching details task

pezholio:
This adds the a new task that allows service operators to review any claims with matching details before approving a claim.

# Screenshots

image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- Add a task to check claims with matching attributes
+
 ## [Release 062] - 2020-03-12
 
 - Add Yes/No option to claim checking tasks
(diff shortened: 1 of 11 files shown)


pezholio:
[Add the "Claims with matching details" task](https://trello.com/c/jloVMsuc/1385-add-the-claims-with-matching-details-task)

spec/features/admin_claim_check_spec.rb
```diff
@@ -130,6 +130,8 @@
         let!(:claim) { create(:claim, :submitted, personal_details.merge(bank_sort_code: "582939", bank_account_number: "74727752")) }
 
         scenario "User is informed that the claim cannot be approved" do
+          pending("This needs a new step to be added")
```
lawrence-forooghian: I'd lean towards squashing these two commits into a single commit which adds this new claim checking task. Reduces churn of adding and removing stuff in the specs, and makes sure that we don't have any commits where the claim checking flow is broken.

(and then the CHANGELOG commit can be squashed into here, since the whole feature is implemented in a single commit)

pezholio: Yeah, fair enough. It worked for me while I was working on it, but no reason why is can't be squashed down

PR #965, opened 2020-03-13
Add the identity confirmation task for claims that didn't go through GOV.UK Verify

lawrence-forooghian:
This replaces the existing banner which is displayed on the full claim page and on the claim decision page, which tells the service operator to contact the claimant at their school to verify their identity.

# Open question

Was it safe to remove the existing banners? Do we think that there's a chance that a claims checker might still use the full claim page to approve a claim?

# Screenshots

image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog]
 ## [Unreleased]
 
 - Add a task to check claims with matching attributes
+- Add a claim checking task to verify identity of claimants who didn’t use
+  GOV.UK Verify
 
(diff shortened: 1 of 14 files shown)


lawrence-forooghian:
I felt that the previous question of "Did the claimant submit the claim" wasn't very clear (the claimant is, at least to me, by definition the person who submitted the claim so the answer is yes always). After chatting with Mark and Tekin have changed it to include the claimant’s name:

image

lawrence-forooghian:
Agreed, think it’s safest. I’ve also updated the feature specs to match your PR #963, to jump straight to the extra question.

PR #970, opened 2020-03-16
Remove TRN from payment discrepancy attribute list

tekin:
When creating payments, claims are grouped by their TRN, which means it doesn't make sense to check them for discrepancies when combining multiple claims into a single payment. TRN appears in the list as a hangover from when we were grouping claims using the national_insurance_number instead.


tekin:
Closing this as on reflection and following discussion with @lawrence-forooghian, it's worth keeping the safety of the validation that would prevent a payment being valid with claims that have different TRNs, even though that shouldn't happen through the standard mechanism for creating payments.

PR #972, opened 2020-03-18
Make decisions on a claim undoable

jacksonj04:
All decisions now have a boolean `undone` column which defaults to `false`. Setting this to `true` indicates that a decision has been undone.

When a decision is marked as `undone` it is no longer included in any of the scopes which return `accepted`/`rejected`/`awaiting decision` claims, nor in a `Claim`'s `decision` method (which returns the latest valid decisions).

Changes to specs are to reflect the fact that these scopes now rely on an actual database query, which doesn't apply to objects brought into existence with `build` instead of `create`.

### DIFF
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -200,7 +200,7 @@ class Claim < ApplicationRecord
   end
 
   def decision_made?
-    decision&.persisted?
+    decision.present? && decision.persisted?
   end
 
   def payroll_gender_missing?
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -257,6 +257,14 @@ class Claim < ApplicationRecord
     personal_data_removed_at.present?
   end
 
+  def payrolled?
+    payment.present?
+  end
+
+  def scheduled_for_payment?
+    scheduled_payment_date.present?
+  end
+
   def full_name
     [first_name, middle_name, surname].compact.join(" ")
   end
@@ -281,7 +289,7 @@ class Claim < ApplicationRecord
   end
 
   def amendable?
-    submitted? && !decision&.rejected? && !payment && !personal_data_removed?
+    submitted? && !decision&.rejected? && !payrolled? && !personal_data_removed?
   end
 
   private
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -162,9 +162,9 @@ class Claim < ApplicationRecord
 
   scope :unsubmitted, -> { where(submitted_at: nil) }
   scope :submitted, -> { where.not(submitted_at: nil) }
-  scope :awaiting_decision, -> { submitted.left_outer_joins(:decisions).where(decisions: {claim_id: nil}) }
-  scope :approved, -> { joins(:decisions).where("decisions.result" => :approved) }
-  scope :rejected, -> { joins(:decisions).where("decisions.result" => :rejected) }
+  scope :awaiting_decision, -> { submitted.joins("LEFT OUTER JOIN decisions ON decisions.claim_id = claims.id AND decisions.undone = false").where(decisions: {claim_id: nil}) }
+  scope :approved, -> { joins(:decisions).merge(Decision.active.approved) }
+  scope :rejected, -> { joins(:decisions).merge(Decision.active.rejected) }
   scope :approaching_decision_deadline, -> { awaiting_decision.where("submitted_at < ? AND submitted_at > ?", DECISION_DEADLINE.ago + DECISION_DEADLINE_WARNING_POINT, DECISION_DEADLINE.ago) }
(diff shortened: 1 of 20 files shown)


app/controllers/admin/undo_decision_controller.rb
```diff
@@ -0,0 +1,42 @@
+class Admin::AmendmentsController < Admin::BaseAdminController
```
robbiepaul: Did you mean to commit this file?

jacksonj04: Nope, spotted it just before you commented!

robbiepaul: Ah 🙌

spec/models/claim_spec.rb
```diff
@@ -484,14 +502,23 @@
...
-      claim.decisions.append(build(:decision, result: "rejected", created_at: DateTime.now))
+      claim = create(:claim, :submitted)
+      create(:decision, result: "approved", claim: claim, created_at: 7.days.ago)
+      create(:decision, result: "rejected", claim: claim, created_at: DateTime.now)
 
       expect(claim.decision.result).to eq "rejected"
     end
   end
 
+  describe "#decision" do
```
robbiepaul: We already describe `#decision` above this, could we move this example to that block?

app/models/decision.rb
```diff
@@ -4,14 +4,17 @@ class Decision < ApplicationRecord
 
   validates :result, :created_by, presence: {message: "Make a decision to approve or reject the claim"}
   validate :claim_must_be_approvable, if: :approved?, on: :create
+  validate :claim_must_have_undoable_decision, if: :undone?, on: :update
+
+  scope :not_undone, -> { where(undone: false) }
```
robbiepaul: I wrestled with the naming of this scope a little, could we avoid the double negative somehow? Perhaps something like `:active` would be better?

app/models/claim.rb
```diff
@@ -257,6 +257,14 @@ def personal_data_removed?
     personal_data_removed_at.present?
   end
 
+  def payrolled?
+    payment.present?
+  end
+
+  def paid?
```
pezholio: The naming of this could be a bit misleading. If the payment date is in the future, it might be _considered_ paid, but not actually paid. I don't think it really matters for our case, but maybe something like `marked_as_paid?` might be better

spec/support/geckoboard_helpers.rb
```diff
@@ -92,9 +92,9 @@ def request_body_matches_geckoboard_data_for_claims?(request, claims)
         "check" => claim.decision.present? ? claim.decision.result : "unchecked",
         "checked_at" => (claim.decision.present? ? claim.decision.created_at : DateTime.parse("1970-01-01")).strftime("%Y-%m-%dT%H:%M:%S%:z"),
         "number_of_days_to_check" => claim.decision&.number_of_days_since_claim_submitted,
-        "paid" => claim.payment.present?.to_s,
-        "paid_at" => (claim.payment.present? ? claim.scheduled_payment_date : DateTime.parse("1970-01-01")).strftime("%Y-%m-%dT%H:%M:%S%:z"),
-        "award_amount" => claim.eligibility.present? ? (claim.award_amount * 100).to_i : nil
+        "award_amount" => claim.eligibility.present? ? (claim.award_amount * 100).to_i : nil,
+        "paid" => claim.paid?.to_s,
+        "paid_at" => (claim.paid? ? claim.scheduled_payment_date : DateTime.parse("1970-01-01")).strftime("%Y-%m-%dT%H:%M:%S%:z")
```
pezholio: Is there any reason the order of these has changed? No big deal, it just made the diff a bit harder to parse.

spec/models/claim_spec.rb
```diff
@@ -555,6 +555,18 @@
     end
   end
 
+  describe "#personal_data_removed?" do
```
pezholio: 🌟

app/models/claim.rb
```diff
@@ -200,7 +200,7 @@ def decision
   end
 
   def decision_made?
-    decision&.persisted?
+    decision.present? && decision.persisted?
```
pezholio: Yeah, this is nicer. I know `nil` is falsey in Ruby, but I prefer `*?` methods to return `true` or `false`

app/models/claim.rb
```diff
@@ -195,12 +195,12 @@ def approvable?
     submitted? && !payroll_gender_missing? && !decision_made? && !payment_prevented_by_other_claims?
   end
 
-  def decision
-    decisions.last
+  def latest_decision
+    decisions.where(undone: false).last
```
tekin: we can use the scope here:
```suggestion
    decisions.active.last
```

PR #974, opened 2020-03-18
Fix deployments

pezholio:
We're having issues deploying due to a change in the way ARM templates are handled. For some reason, running `string` against a created array now appears to flatten out the array into a comma seperated list, so we need to change how we build the `SERVICE_ENTITY_IDS` JSON array. This builds up the array using a combination of a `[`, an (escaped) single quote `''`, the VSP entity ID and a `]`. The means if the value of `VSP.VERIFY_ENTITY_ID` was `http://example.com`, the resulting JSON string would be `['http://example.com']`.

I've also updated the command we use to create deployments, as we were getting a deprecation warning.

### DIFF
--- a/azure/resource_groups/app/template.json
+++ b/azure/resource_groups/app/template.json
@@ -615,7 +615,7 @@
               },
               {
                 "name": "SERVICE_ENTITY_IDS",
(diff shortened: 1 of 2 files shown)


pezholio:
Yep, works like a charm 👍

PR #979, opened 2020-03-19
Support amending the student loan amount

robbiepaul:
It's not a particularly elegant solution, but this change will allow service operators to amend the student loan amount. I've also added a upper limit of £5000 to the amendment.

### Screenshots

Screenshot 2020-03-19 at 14 29 06

Screenshot 2020-03-19 at 12 26 16

### DIFF
--- a/app/models/amendment.rb
+++ b/app/models/amendment.rb
@@ -41,7 +41,7 @@ class Amendment < ApplicationRecord
       end
 
       changes_hash = claim.previous_changes.merge(claim.eligibility.previous_changes)
-        .slice(*Claim::AMENDABLE_ATTRIBUTES + Claim::AMENDABLE_ELIGIBILITY_ATTRIBUTES)
+        .slice(*Claim::AMENDABLE_ATTRIBUTES + Policies::AMENDABLE_ELIGIBILITY_ATTRIBUTES)
         .reject { |_, values| values.all?(&:blank?) }
         .to_h
 
--- a/app/models/amendment.rb
+++ b/app/models/amendment.rb
@@ -40,10 +40,13 @@ class Amendment < ApplicationRecord
         raise ActiveRecord::Rollback
       end
 
-      amendment.claim_changes = claim.previous_changes
-        .slice(*Claim::AMENDABLE_ATTRIBUTES)
+      changes_hash = claim.previous_changes.merge(claim.eligibility.previous_changes)
+        .slice(*Claim::AMENDABLE_ATTRIBUTES + Claim::AMENDABLE_ELIGIBILITY_ATTRIBUTES)
         .reject { |_, values| values.all?(&:blank?) }
         .to_h
+
+      amendment.claim_changes = changes_hash
+
       raise ActiveRecord::Rollback unless amendment.save
     end
 
--- a/app/models/amendment.rb
+++ b/app/models/amendment.rb
@@ -33,7 +33,7 @@ class Amendment < ApplicationRecord
     Claim.transaction do
       claim.assign_attributes(claim_attributes)
 
-      unless claim.save(context: :submit)
+      unless claim.save(context: [:submit, :amendment])
         amendment.valid?
         amendment.errors.merge!(claim.errors)
         amendment.errors.delete(:claim_changes)
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -36,6 +36,7 @@ class Claim < ApplicationRecord
     bank_account_number
     building_society_roll_number
   ].freeze
+  AMENDABLE_ELIGIBILITY_ATTRIBUTES = %i[student_loan_repayment_amount].freeze
   FILTER_PARAMS = {
     address_line_1: true,
     address_line_2: true,
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -36,7 +36,6 @@ class Claim < ApplicationRecord
     bank_account_number
     building_society_roll_number
   ].freeze
-  AMENDABLE_ELIGIBILITY_ATTRIBUTES = %i[student_loan_repayment_amount].freeze
   FILTER_PARAMS = {
     address_line_1: true,
     address_line_2: true,
--- a/app/models/policies.rb
+++ b/app/models/policies.rb
@@ -5,6 +5,8 @@ module Policies
     MathsAndPhysics
   ].freeze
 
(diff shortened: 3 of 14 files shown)


robbiepaul:
[Support amending the student loan amount](https://trello.com/c/DUezjzsV/1434-support-amending-the-student-loan-amount)

app/views/admin/amendments/new.html.erb
```diff
@@ -126,6 +126,27 @@
...
+            <% end %>
+          </div>
+          <div class="govuk-grid-column-two-thirds">
+            <div class="govuk-form-group">
+              <%= eligibility_form.text_field(
+                    :student_loan_repayment_amount,
+                    value: currency_value_for_number_field(eligibility_form.object.student_loan_repayment_amount),
+                    class: "govuk-input govuk-input--width-10") %>
+            </div>
+          </div>
```
titlescreen: How would you feel about using the below so that it is visually consistent with the loan amount question that the claimants see?

I think the narrower input would discourage users from entering big numbers as anything over `1000.00` gets pushed so service operators will hopefully see that they have perhaps entered in an extra 0 when they did not mean to. 

image

```
<div class="govuk-currency-input">
                <span class="govuk-currency-input__unit">&pound;</span>
                <%= eligibility_form.text_field(
                      :student_loan_repayment_amount,
                      value: currency_value_for_number_field(eligibility_form.object.student_loan_repayment_amount),
                      class: "govuk-input govuk-currency-input__input govuk-input--width-10") %>
              </div>
```

If we do this then the current_input.SCSS will need updating.

```
.govuk-currency-input {
  position: relative;
  &__unit {
    position: absolute;
    bottom: 7px;
    left: 7px;
    @include govuk-font(19, bold);
  }

  &__input.govuk-input {
    padding-left: 20px;
  }
}
```

app/models/student_loans/eligibility.rb
```diff
@@ -51,6 +51,7 @@ class Eligibility < ApplicationRecord
     validates :mostly_performed_leadership_duties, on: [:"mostly-performed-leadership-duties", :submit], inclusion: {in: [true, false], message: "Select yes if you spent more than half your working hours on leadership duties"}, if: :had_leadership_position?
     validates :student_loan_repayment_amount, on: [:"student-loan-amount", :submit], presence: {message: "Enter your student loan repayment amount"}
     validates_numericality_of :student_loan_repayment_amount, message: "Enter a valid monetary amount", allow_nil: true, greater_than: 0, less_than_or_equal_to: 99999
+    validates :student_loan_repayment_amount, on: :amendment, numericality: {less_than: 5000, message: "Enter an amount below £5,000"}
```
titlescreen: I think it may be useful to have the error highlighted against the student loan input as well. I've included an example of how this is done for teacher reference numbers 

image

app/models/student_loans/eligibility.rb
```diff
@@ -51,6 +51,7 @@ class Eligibility < ApplicationRecord
     validates :mostly_performed_leadership_duties, on: [:"mostly-performed-leadership-duties", :submit], inclusion: {in: [true, false], message: "Select yes if you spent more than half your working hours on leadership duties"}, if: :had_leadership_position?
     validates :student_loan_repayment_amount, on: [:"student-loan-amount", :submit], presence: {message: "Enter your student loan repayment amount"}
     validates_numericality_of :student_loan_repayment_amount, message: "Enter a valid monetary amount", allow_nil: true, greater_than: 0, less_than_or_equal_to: 99999
```
titlescreen: Do we need something a bit more explicit for when a SO tries to enter a value of £0, or is this too edge-casey. Now that I have written it I'm probably a 2/10 on this. 🤷‍♀️

app/models/student_loans/eligibility.rb
robbiepaul: Yeah good catch `errors_tag` wasn't displaying the error correctly because it's a nested attribute. I'll push up a fix

app/models/student_loans/eligibility.rb
robbiepaul: A value of £0 will be caught by the validation rule above this one

app/views/admin/amendments/new.html.erb
robbiepaul: Sorted, thanks Rich.

This is what it looks like now:
Screenshot 2020-03-19 at 14 27 23

titlescreen:
Cool looks good to me 👍

spec/models/amendment_spec.rb
```diff
@@ -194,6 +194,32 @@
       end
     end
 
+    context "when updating a nested attribute" do
```
tekin:
```suggestion
    context "when amending the claim’s eligibility attributes" do
```

app/models/amendment.rb
```diff
@@ -33,15 +33,15 @@ def self.amend_claim(claim, claim_attributes, amendment_attributes)
...
+      unless claim.save(context: [:submit, :amendment])
         amendment.valid?
         amendment.errors.merge!(claim.errors)
         amendment.errors.delete(:claim_changes)
         raise ActiveRecord::Rollback
       end
 
-      amendment.claim_changes = claim.previous_changes
-        .slice(*Claim::AMENDABLE_ATTRIBUTES)
+      amendment.claim_changes = claim.previous_changes.merge(claim.eligibility.previous_changes)
```
tekin: I feel like this chain of hash munging has got long enough to warrant extracting it to a method to try and make things more easily understood.

app/models/claim.rb
```diff
@@ -36,6 +36,7 @@ class Claim < ApplicationRecord
     bank_account_number
     building_society_roll_number
   ].freeze
+  AMENDABLE_ELIGIBILITY_ATTRIBUTES = %i[student_loan_repayment_amount].freeze
```
tekin: Should this live on the eligibility class?

app/models/policies.rb
```diff
@@ -5,6 +5,8 @@ module Policies
     MathsAndPhysics
   ].freeze
 
+  AMENDABLE_ELIGIBILITY_ATTRIBUTES = POLICIES.map { |policy| policy::Eligibility::AMENDABLE_ATTRIBUTES }.flatten.freeze
```
robbiepaul: Nice! I did wonder how best to achieve this 👍

PR #983, opened 2020-03-23
Support searching for claims with more than just the reference

pezholio:
This adds a class to search claims for a number of different attributes, and updates the search 
page to list the search results, rather than just redirecting to a single claim.

image

image

### DIFF
--- a/app/controllers/admin/claims_controller.rb
+++ b/app/controllers/admin/claims_controller.rb
@@ -22,14 +22,14 @@ class Admin::ClaimsController < Admin::BaseAdminController
   end
 
   def search
-    return unless params[:reference].present?
+    return unless params[:query].present?
 
-    claim = Claim.find_by(reference: params[:reference].upcase)
+    @claims = Claim::Search.new(params[:query]).claims.includes(:eligibility)
 
-    if claim
-      redirect_to(admin_claim_url(claim))
-    else
-      flash.now[:notice] = "Cannot find a claim with reference \"#{params[:reference]}\""
+    if @claims.none?
+      flash.now[:notice] = "Cannot find a claim for query \"#{params[:query]}\""
+    elsif @claims.one?
+      redirect_to(admin_claim_url(@claims.first))
     end
   end
 
--- a/app/models/claim/search.rb
+++ b/app/models/claim/search.rb
@@ -1,0 +1,31 @@
+class Claim
+  # Accepts a search term and returns all Claims that match against any of the
+  # attributes defined in the `SEARCHABLE_ATTRIBUTES` constant. Both subject
+  # and attribute are downcased, so the search is case-insensitive.
+  class Search
+    attr_accessor :search_term
+
+    SEARCHABLE_ATTRIBUTES = %w[
+      reference
+      email_address
+      surname
+      teacher_reference_number
+    ]
+
+    def initialize(search_term)
+      @search_term = search_term
+    end
+
+    def claims
+      SEARCHABLE_ATTRIBUTES.inject(Claim.none) do |relation, attribute|
+        relation.or(search_by(attribute))
+      end
+    end
+
+    private
+
+    def search_by(attribute)
+      Claim.where("LOWER(#{attribute}) = LOWER(?)", search_term)
+    end
+  end
+end
--- a/app/views/admin/claims/search.html.erb
+++ b/app/views/admin/claims/search.html.erb
@@ -10,7 +10,7 @@
     <%= form_with url: search_admin_claims_path, method: :get do |form| %>
       <div class="govuk-form-group">
         <%= form.label :query, "Enter a reference, email address, Teacher Reference Number, or surname", class: "govuk-label" %>
-        <%= form.text_field :query, class: "govuk-input" %>
+        <%= form.text_field :query, class: "govuk-input govuk-input--width-20" %>
       </div>
 
       <div class="govuk-form-group">
--- a/app/views/admin/claims/search.html.erb
+++ b/app/views/admin/claims/search.html.erb
@@ -9,8 +9,8 @@
 
     <%= form_with url: search_admin_claims_path, method: :get do |form| %>
       <div class="govuk-form-group">
-        <%= form.label :reference, "Enter a reference", class: "govuk-label" %>
-        <%= form.text_field :reference, class: "govuk-input" %>
+        <%= form.label :query, "Enter a reference, email address, Teacher Reference Number, or surname", class: "govuk-label" %>
+        <%= form.text_field :query, class: "govuk-input" %>
       </div>
 
       <div class="govuk-form-group">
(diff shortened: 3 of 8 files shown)


pezholio:
[Support searching for claims with more than just the reference](https://trello.com/c/N6q303Yb/1437-support-searching-for-claims-with-more-than-just-the-reference)

app/models/claim/search.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+
+    def claims
+      claims = search_by(:reference)
+
+      SEARCHABLE_ATTRIBUTES.each do |attribute|
+        claims = claims.or(search_by(attribute))
+      end
+
+      claims
```
tekin: It made me uncomfortable that `reference` wasn't in `SEARCHABLE_ATTRIBUTES`. I think we can rewrite this method such that we can put it there:

```suggestion
      SEARCHABLE_ATTRIBUTES.inject(Claim.none) do |relation, attribute|
        relation.or(search_by(attribute))
      end
```

spec/models/claim/search_spec.rb
```diff
@@ -0,0 +1,67 @@
+require "rails_helper"
+
+RSpec.describe Claim::Search do
+  subject(:search) { Claim::Search.new(query) }
+
+  let!(:other_claims) { create_list(:claim, 7) }
```
tekin: Maybe reduce this down a tad, we only really need a single other claim for these specs to give us confidence that non-matching claims aren't returned.

app/models/claim/search.rb
```diff
@@ -0,0 +1,31 @@
+class Claim
+  class Search
+    attr_accessor :query
```
tekin: "query" is quite a loaded term. Can we rename this to something less ambiguous? Maybe "search_term" or "search_string"?

app/controllers/admin/claims_controller.rb
```diff
@@ -22,14 +22,12 @@ def show
   end
 
   def search
-    return unless params[:reference].present?
+    return unless params[:query].present?
 
-    claim = Claim.find_by(reference: params[:reference].upcase)
+    @claims = Claim::Search.new(params[:query]).claims
```
tekin: Let's avoid an n+1 and preload eligibilities (used by the `Claim#policy` method):
```suggestion
    @claims = Claim::Search.new(params[:query]).claims.includes(:eligibility)
```

app/views/admin/claims/search.html.erb
```diff
@@ -9,14 +9,42 @@
...
+        </thead>
+        <tbody class="govuk-table__body">
+          <% @claims.each do |claim| %>
+            <tr class="govuk-table__row">
+              <th scope="row" class="govuk-table__header"><%= claim.reference %></th>
+              <td class="govuk-table__cell"><%= claim.full_name %></td>
+              <td class="govuk-table__cell"><%= claim.policy.short_name %></td>
+              <td class="govuk-table__cell"><%= claim.academic_year %></td>
+              <td class="govuk-table__cell">
+                <%= link_to 'View tasks', admin_claim_tasks_path(claim), class: "govuk-link" %>
```
tekin: I wonder if it makes sense to use different words here because we're no longer listing just claims that are awaiting checking:

```suggestion
                <%= link_to "View details"', admin_claim_tasks_path(claim), class: "govuk-link" %>
```

spec/features/admin_search_spec.rb
```diff
@@ -5,7 +5,7 @@
     sign_in_to_admin_with_role(DfeSignIn::User::SERVICE_OPERATOR_DFE_SIGN_IN_ROLE_CODE)
   end
 
-  it "redirects to the claim" do
+  it "allows a service operator to view the claim" do
```
tekin:
```suggestion
  scenario "a service operator can search for claims" do
```

spec/requests/admin_claims_spec.rb
```diff
@@ -107,19 +107,31 @@
   end
 
   describe "claims#search" do
-    let(:claim) { create(:claim, :submitted) }
+    let(:claim1) { create(:claim, :submitted, surname: "Wayne") }
+    let(:claim2) { create(:claim, :submitted, surname: "Wayne") }
 
     it "redirects to a claim when one exists" do
```
tekin: The code isn't redirecting so these words are no longer quite right. Also this spec appears to replicate the newly added spec below almost identically. Do we actually need both?

app/models/claim/search.rb
```diff
@@ -0,0 +1,31 @@
+class Claim
+  class Search
```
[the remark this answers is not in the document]
pezholio: 👍

app/models/claim/search.rb
pezholio: Yes! I forgot about `.none`. Good shout

pezholio:
Cool, nice one. I did think about keeping the redirection functionality in, but I came to the opposite conclusion. Happy to put it back in 👍

spec/requests/admin_claims_spec.rb
pezholio: I've tweaked this now, so we do redirect, so both specs are now legit

spec/models/claim/search_spec.rb
tekin: This has been marked as resolved but it's still creating seven claims.

app/controllers/admin/claims_controller.rb
```diff
@@ -22,14 +22,16 @@ def show
...
+    return unless params[:query].present?
 
-    claim = Claim.find_by(reference: params[:reference].upcase)
+    @claims = Claim::Search.new(params[:query]).claims.includes(:eligibility)
 
-    if claim
-      redirect_to(admin_claim_url(claim))
+    if @claims.blank?
+      flash.now[:notice] = "Cannot find a claim for query \"#{params[:query]}\""
+    elsif @claims.count == 1
```
tekin: Ruby's Enumerable has a `one?` method:

```suggestion
    elsif @claims.one?
```

app/controllers/admin/claims_controller.rb
```diff
@@ -22,14 +22,16 @@ def show
...
+      redirect_to(admin_claim_url(@claims.first))
     else
```
tekin: Reckon we can get rid of this last `else` as it's redundant: if the redirect doesn't happen above, the controller will render the `search` template anyway.

spec/requests/admin_claims_spec.rb
```diff
@@ -107,19 +107,30 @@
   end
 
   describe "claims#search" do
-    let(:claim) { create(:claim, :submitted) }
+    let(:claim1) { create(:claim, :submitted, surname: "Wayne") }
+    let(:claim2) { create(:claim, :submitted, surname: "Wayne") }
```
tekin: `claim2` is not being used anywhere in this file as far as I can see.

app/controllers/admin/claims_controller.rb
```diff
@@ -22,14 +22,16 @@ def show
...
   def search
-    return unless params[:reference].present?
+    return unless params[:query].present?
 
-    claim = Claim.find_by(reference: params[:reference].upcase)
+    @claims = Claim::Search.new(params[:query]).claims.includes(:eligibility)
 
-    if claim
-      redirect_to(admin_claim_url(claim))
+    if @claims.blank?
```
tekin: Don't forget; `blank?` is an ActiveSupport extension primarily intended for dealing with things that might be strings. Either `none?` or `empty?` is more appropriate here:

```suggestion
    if @claims.empty?
```

spec/models/claim/search_spec.rb
pezholio: Arrgh. Sorry!

app/controllers/admin/claims_controller.rb
pezholio: Ooh, nice

app/controllers/admin/claims_controller.rb
pezholio: Yeah, I just thought it might be a bit clearer to spell it out, but, on reflection, that statement does look a bit icky, and as it's default behaviour, shouldn't be too surprising

PR #988, opened 2020-03-24
Set healthcheck endpoint to monitor

pezholio:
This tells Azure to monitor the healthcheck endpoint every 2 minutes. If an instance does not respond within 10 minutes, the instance is taken out of the pool. It's a manual process to put the app back in the pool, and we're not (currently) able to be alerted if something is taken out of the pool (this is a Windows-only thing at the moment). That said, it's better than us accidentally routing people to a broken instance.

Ideally, we'll want to have autohealing, but this is (again) a Windows-only thing.

More on the Health Check functionality is [here](https://github.com/projectkudu/kudu/wiki/Health-Check-%28Preview%29)

### DIFF
--- a/azure/templates/app.json
+++ b/azure/templates/app.json
@@ -183,7 +183,8 @@
         "httpsOnly": true,
         "siteConfig": {
           "alwaysOn": "[parameters('appServiceAlwaysOn')]",
-          "appSettings": "[parameters('appServiceEnvironmentVariables')]"
+          "appSettings": "[parameters('appServiceEnvironmentVariables')]",
+          "healthCheckPath": "/healthcheck"
         }
       }
     },


pezholio:
[Investigate if Azure can take out unhealthy instances from an App Service](https://trello.com/c/qc2CMLF5/886-investigate-if-azure-can-take-out-unhealthy-instances-from-an-app-service)

pezholio:
I've tested this on test and it deploys fine. Tricky to test if it _works_, as I can't manually make a single instance break reliably.

pezholio:
I think autohealing is the thing, but from my research, seems it's a Windows-only thing at the mo. We could _maybe_ jury-rig our own autohealing, with an app that responds to a ping when an instance is down and restarts the app, but this would take time and effort (and I don't even know if it's possible!)

PR #989, opened 2020-03-24
Exclude claims from previous academic years on matching claims screen

pezholio:
This adds a new scope to filter by academic year, and then uses that scope in the matching attribute finder to only find claims that are for the same academic year as the source claim.

### DIFF
--- a/spec/models/claim/matching_attribute_finder_spec.rb
+++ b/spec/models/claim/matching_attribute_finder_spec.rb
@@ -9,6 +9,7 @@ RSpec.describe Claim::MatchingAttributeFinder do
         email_address: "genghis.khan@mongol-empire.com",
         bank_account_number: "34682151",
         bank_sort_code: "972654",
+        academic_year: AcademicYear.new("2019"),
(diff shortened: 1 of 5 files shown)


pezholio:
[Exclude claims from previous academic years on "matching claims" screen](https://trello.com/c/Qj2tIt9B/1426-exclude-claims-from-previous-academic-years-on-matching-claims-screen)

spec/models/claim/matching_attribute_finder_spec.rb
```diff
@@ -37,6 +38,12 @@
       expect(matching_claims).to be_empty
     end
 
+    it "does not include claims that match, but have a different academic year" do
+      create(:claim, :submitted, teacher_reference_number: source_claim.teacher_reference_number, academic_year: AcademicYear.new("2020"))
```
lawrence-forooghian: I think we need to lock the policy to `StudentLoans` here.

pezholio: Good call 👍

PR #992, opened 2020-03-25
Transform DQT report rows to unique records

robbiepaul:
When we built the automated DQT checks we were working on the assumption
that there would be a single row for each record. It turns out there
can be multiple rows for each claimant.

We need to munge these rows down to unique records so we can iterate
through them and safely create checks for each claim.

The claim reference, date of birth and QTS award year are all consistent
between grouped rows so we can use the first row's value. For the ITT
subject and degree JAC codes we need to collate and combine these values
from the multiple grouped rows.

As this new class is responsible for transforming the data in to a shape
we expect I used the opportunity to move the logic to do with
parsing the dates and transforming columns to arrays.

### DIFF
--- a/app/models/automated_checks/dqt_report_consumer.rb
+++ b/app/models/automated_checks/dqt_report_consumer.rb
@@ -19,17 +19,16 @@ module AutomatedChecks
       return if errors.any?
 
       claims = Claim.awaiting_task("qualifications")
-      csv.rows.each do |row|
-        reference = row.fetch("dfeta text2")
-        qts_date = row.fetch("dfeta qtsdate")
-        claim = claims.detect { |c| c.reference == reference }
-        next if qts_date.blank? || claim.nil?
-        if claim.policy::DQTRecord.new(row.to_h).eligible? && row_matches_claim?(row, claim)
+      records = DQTReportCsvToRecords.new(csv.rows).transform
+      records.each do |record|
+        claim = claims.detect { |c| c.reference == record.fetch(:claim_reference) }
+        next if record.fetch(:qts_date).blank? || claim.nil?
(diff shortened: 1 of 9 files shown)


robbiepaul:
[BUG: Automated DQT checks need to support a record having multiple rows](https://trello.com/c/Syhdb5YN/1450-bug-automated-dqt-checks-need-to-support-a-record-having-multiple-rows)

app/models/automated_checks/dqt_report_consumer.rb
```diff
@@ -19,17 +19,16 @@ def ingest
...
-      csv.rows.each do |row|
-        reference = row.fetch("dfeta text2")
-        qts_date = row.fetch("dfeta qtsdate")
-        claim = claims.detect { |c| c.reference == reference }
-        next if qts_date.blank? || claim.nil?
-        if claim.policy::DQTRecord.new(row.to_h).eligible? && row_matches_claim?(row, claim)
+      records = DQTReportCsvToRecords.new(csv.rows).transform
+      records.each do |record|
+        claim = claims.detect { |c| c.reference == record.fetch(:claim_reference) }
+        next if record.fetch(:qts_date).blank? || claim.nil?
```
pezholio: Should we add `next if record.fetch(:qts_date).blank?` to top of the loop? If we're not going to do anything with this record, there's no point trying to find the claim.

robbiepaul: I wanted to avoid too many `next if` conditionals. I don't think it's too computationally intensive to find the claim in an array of objects, the query fetching all the claims happens outside the loop now. I'll do some quick research to see if this should be optimised though

pezholio: 👍

PR #993, opened 2020-03-26
User interface for undoing a decision

pezholio:
This builds on #972 to add a UI to allow a decision to be undone. It works in a similar way to how other amendments are made, by adding creating an Amendment record with notes and details on what changes have been made.

# Screenshots

image

image

image

### DIFF
--- a/app/models/amendment.rb
+++ b/app/models/amendment.rb
@@ -53,6 +53,35 @@ class Amendment < ApplicationRecord
     amendment
   end
 
+  # Undoes a decision and makes a new Amendment record with notes and details
+  # on what the decison was previously.
+  #
+  # If the operation fails, then the claim’s changes will not be persisted, and
+  # the Amendment record will not be saved.
+  #
+  # Returns an Amendment record.
+  # - If the operation was successful, this will return true for persisted?
+  # - If the operation failed, this will return false for persisted?, and the
+  # amendment’s errors will have been populated with the errors from the
+  # decision (this will usually only happen if the decision cannot be reversed).
+  def self.undo_decision(decision, amendment_params)
+    amendment = Amendment.new(claim: decision.claim, **amendment_params)
+
+    ActiveRecord::Base.transaction do
+      amendment.claim_changes = {
+        decision: [decision.result, "undecided"]
+      }
+
+      decision.update!(undone: true)
+      amendment.save!
+    end
+
+    amendment
+  rescue ActiveRecord::RecordInvalid
+    amendment.errors.merge!(decision.errors)
+    amendment
+  end
+
   private
 
   def claim_must_be_amendable
--- a/app/views/admin/amendments/new.html.erb
+++ b/app/views/admin/amendments/new.html.erb
@@ -150,6 +150,20 @@
         </div>
       <% end %>
     <% end %>
+
+    <% if @claim.latest_decision %>
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-one-third">
+          <p class="govuk-body">Claim decision</p>
+        </div>
+        <div class="govuk-grid-column-two-thirds">
+          <p class="govuk-body">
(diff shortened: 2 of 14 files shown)


pezholio:
[UI for reversing the decision made on a claim](https://trello.com/c/hvrBiIK0/1291-ui-for-reversing-the-decision-made-on-a-claim)

app/models/amendment.rb
```diff
@@ -53,6 +53,35 @@ def self.amend_claim(claim, claim_attributes, amendment_attributes)
...
+  # decision (this will usually only happen if the decision cannot be reversed).
+  def self.undo_decision(decision, amendment_params)
+    amendment = Amendment.new(claim: decision.claim, **amendment_params)
+
+    ActiveRecord::Base.transaction do
+      amendment.claim_changes = {
+        decision: [decision.result, "undecided"]
+      }
+
+      decision.update!(result: "undecided", undone: true)
```
robbiepaul: Do we want to update the result old decision to `undecided`? I feel like it goes against having the `undone` flag. Now the previous decisions will be both undecided and undone. Should we preserve the old decision result and just mark it undone?

app/views/admin/amendments/new.html.erb
```diff
@@ -150,6 +150,19 @@
...
     <% end %>
+
+    <% if @claim.decisions.any? %>
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-one-third">
+          <p class="govuk-body">Claim decision</p>
+        </div>
+        <div class="govuk-grid-column-two-thirds">
+          <p class="govuk-body">
+            <%= link_to "Amend decision", new_admin_claim_decision_undo_path(@claim, @claim.latest_decision), class: "govuk-link" %>
```
robbiepaul: I think we should show the current decision here as well, maybe before the link?

app/views/admin/decisions_undo/new.html.erb
```diff
@@ -0,0 +1,32 @@
...
+      the decision and mark it as undecided?
+    </span>
+  </div>
+</div>
+
+<%= form_with url: admin_claim_decision_undos_path(@claim, @decision), scope: :amendment, model: @amendment do |f| %>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <div class="govuk-form-group">
+        <%= f.label :notes, "Amendment Notes", class: "govuk-label" %>
```
robbiepaul:
```suggestion
        <%= f.label :notes, "Change notes", class: "govuk-label" %>
```

Just to be consistent with the claim amendment screen

app/views/admin/amendments/new.html.erb
robbiepaul: Also, perhaps we should be explicit about the result of the action here? Maybe `Undo decision`?

app/views/admin/decisions_undo/new.html.erb
```diff
@@ -0,0 +1,32 @@
...
+    <h1 class="govuk-heading-l">
+      Amend claim decision for Claim <%= @claim.reference %>?
+    </h1>
+
+    <%= render("shared/error_summary", instance: @amendment) if @amendment.errors.any? %>
+
+    <span class="govuk-hint">
+      Claim ABC1234 is currently approved. Do you want to undo
+      the decision and mark it as undecided?
+    </span>
```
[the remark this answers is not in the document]
pezholio: Oops!

app/models/amendment.rb
pezholio: Yeah, I did wonder that. I think you're right. Undone is undone, then we can also keep the history a bit better too.

app/views/admin/amendments/new.html.erb
pezholio: Yeah, nice. I've tweaked it to look like this now:

image

pezholio:
Cool, just running the copy by Faye and I'll get it merged in 👍

PR #994, opened 2020-03-26
Make the new task view primary

LornaHarwood:
We have now switched both policies to using the new "task"-based UI as the primary interface for viewing and checking claims, but there are still a few things that either only exist on the old "full view", or bits of interaction that assume the old "full view" is the main screen for viewing a claim.

- Remove the "decision" form from the "full claim" view. There should only be one place where a user decides the outcome of a claim.
- Searching for a claim should take the user to the "tasks" view, not the "full claim" view.
- Amending a claim should take the user back to the "tasks" view, not the "full claim" view.

### DIFF
--- a/app/controllers/admin/amendments_controller.rb
+++ b/app/controllers/admin/amendments_controller.rb
@@ -11,7 +11,7 @@ class Admin::AmendmentsController < Admin::BaseAdminController
     @amendment = Amendment.amend_claim(@claim, claim_params, amendment_params)
 
     if @amendment.persisted?
-      redirect_to admin_claim_url(@claim)
+      redirect_to admin_claim_tasks_url(@claim), notice: "Claim has been amended successfully"
     else
       render "new"
     end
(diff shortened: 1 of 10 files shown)


LornaHarwood:
[Make the new "task list" view of a claim the primary screen for working with claims](https://trello.com/c/DthVLP1a/1435-make-the-new-task-list-view-of-a-claim-the-primary-screen-for-working-with-claims)

robbiepaul:
Looks good to me. Do you think the "Cancel" button on the amendment screen should go to the tasks view as well?

https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/b1bca30ca20e2b3f2951a537b28d44f1502d63bb/spec/features/admin_amend_claim_spec.rb#L86

tekin:
> Looks good to me. Do you think the "Cancel" button on the amendment screen should go to the tasks view as well?
> 
> https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/b1bca30ca20e2b3f2951a537b28d44f1502d63bb/spec/features/admin_amend_claim_spec.rb#L86

Good shout @robbiepaul, we missed that. We're about to update the claim tasks view so it has a separate amendments tab so I reckon we can pick that up as part of that work.

PR #1002, opened 2020-03-30
Run the automated DQT checks in a transaction

robbiepaul:
There's a potential race condition where a user could manually complete
a qualification task at the same time (after `Claim.awaiting_task`) and
cause an exception to occur. By wrapping the process in a transaction
we can make the script fail gracefully.

### DIFF
--- a/spec/requests/admin_qualification_report_upload_spec.rb
+++ b/spec/requests/admin_qualification_report_upload_spec.rb
@@ -15,6 +15,12 @@ RSpec.describe "Admin qualification report upload" do
 
   describe "qualification_report_upload#create" do
     let(:file) { Rack::Test::UploadedFile.new(StringIO.new(csv), "text/csv", original_filename: "dqt_data.csv") }
+    let(:csv) do
+      <<~CSV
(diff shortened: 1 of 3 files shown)


spec/requests/admin_qualification_report_upload_spec.rb
```diff
@@ -62,6 +61,19 @@
       end
     end
 
+    context "when there's an error creating the tasks" do
+      it "displays an error and prompts the user to try again" do
```
tekin: I think this spec will pass even without the transaction wrapping the work because the exception would get raised when the first task is being created so the rollback behaviour doesn't have coverage.

robbiepaul: Ahh. I had originally wanted to write a test where the second record caused the exception and we could check no records were created, but I couldn't figure out how to set that up. I'll have a think

tekin: You know what, let's not stress this too much given how much time we have left. It's probably not worth the extra effort vs the other things we want to get done by the middle of next week.

PR #1006, opened 2020-03-30
Ensure the surname in DQT matches the one in the claim

robbiepaul:
Currently claims checkers do an extra informal identity verification
on the record in DQT to make sure it matches the claimant. They check
the name and National Insurance number (NiNo). The NiNo doesn't always
exist in DQT so instead we can add the extra verification step as a
match against the surname.

We recognise that names are subject to change and can legitimately not
match, but these claims will still be checked just manually by a person
instead of automatically.

Surnames that come back from GOV.UK Verify can be all upper case. To
allow for this we need to make the comparison case insensitive.

### DIFF
--- a/app/models/automated_checks/dqt_report_csv_to_records.rb
+++ b/app/models/automated_checks/dqt_report_csv_to_records.rb
@@ -12,6 +12,7 @@ module AutomatedChecks
         {
           claim_reference: grouped_rows.first.fetch(CLAIM_REFERENCE_COLUMN),
           qts_date: parse_date(grouped_rows.first.fetch("dfeta qtsdate")),
+          surname: extract_surname(grouped_rows.first.fetch("fullname")),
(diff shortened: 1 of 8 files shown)


app/models/automated_checks/dqt_report_csv_to_records.rb
```diff
@@ -30,5 +31,9 @@ def parse_date(date)
     rescue
       nil
     end
+
+    def extract_surname(fullname)
+      fullname&.split(" ")&.last
```
robbiepaul: I tried to do `fullname && fullname.split(" ").last` to avoid all the lonely operators but Standard complained. Any suggestions to make the neater are welcome :)

tekin: Maybe this?
```suggestion
      (fullname || "").split(" ").last
```

robbiepaul: Thanks!

PR #1012, opened 2020-03-30
Service operators can add notes to claims

tekin:
This adds a simple notes feature that allows service operators to add notes against claims:

Screenshot 2020-03-30 at 17 08 31

### DIFF
--- a/spec/models/note_spec.rb
+++ b/spec/models/note_spec.rb
@@ -1,0 +1,4 @@
+require "rails_helper"
+
+RSpec.describe Note, type: :model do
+end
--- a/spec/requests/admin_notes_spec.rb
+++ b/spec/requests/admin_notes_spec.rb
@@ -4,9 +4,8 @@ RSpec.describe "admin/notes controller" do
   let(:claim) { create(:claim, :submitted) }
 
   describe "admin/notes#index" do
-    before { @signed_in_user = sign_in_as_service_operator }
-
-    it "list the notes on a claim" do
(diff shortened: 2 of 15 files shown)


spec/models/note_spec.rb
```diff
@@ -0,0 +1,4 @@
+require "rails_helper"
+
+RSpec.describe Note, type: :model do
```
robbiepaul: Do we need this file if there are no specs for the model?

spec/requests/admin_notes_spec.rb
```diff
@@ -0,0 +1,65 @@
+require "rails_helper"
+
+RSpec.describe "admin/notes controller" do
+  let(:claim) { create(:claim, :submitted) }
+
+  describe "admin/notes#index" do
+    it "list the notes on a claim when the user is a service operators" do
```
robbiepaul:
```suggestion
    it "list the notes on a claim when the user is a service operator" do
```

tekin:
@robbiepaul good point re: the form. I imagine most claims will have very few notes, but I'll check in with Rich and Mark, seee what they say.

spec/models/note_spec.rb
tekin: Not really, but don't think it costs anything either. It also means it's there ready for someone if/when they need to add specs to the model.

tekin:
Cheers @robbiepaul, word is that service operators should see the most important info first, so form stays at the bottom, plus I've reversed the order of the claims so that the most recent one comes at the top.

PR #1021, opened 2020-03-31
Added Cantium info

FayeGoodworth:
Added information that Cantium will appear on bank statements as the sender.


FayeGoodworth:
[Tell users their payment comes from Cantium](https://trello.com/c/7VBLawL3/1444-tell-users-their-payment-comes-from-cantium)

PR #1024, opened 2020-04-01
Document how to use the development website

lawrence-forooghian:
This explains:

- where to find the development website
- where to find the basic auth credentials
- how to get through GOV.UK Verify in development

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,12 @@ by [running a live console](#running-a-live-console).
 Development is automatically built and deployed when commits are pushed to
 `master`.
 
+You can check the progress of the build and release in
+[Azure DevOps](https://dev.azure.com/dfe-ssp/S118-Teacher-Payments-Service).
+
+The development website is at
+https://development.additional-teaching-payment.education.gov.uk.
+
 ### Production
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -207,6 +207,16 @@ by [running a live console](#running-a-live-console).
 Development is automatically built and deployed when commits are pushed to
 `master`.
 
+You can check the progress of the build and release in
+[Azure DevOps](https://dev.azure.com/dfe-ssp/S118-Teacher-Payments-Service).
+
+The development website:
+
+- is at https://development.additional-teaching-payment.education.gov.uk
+- is protected by basic HTTP authentication – the credentials are in the
```
tekin, line 216: So right now I think the basic auth credentials are blank, which effectively turns it off. Maybe we should turn it back on now?

lawrence-forooghian: Sure, I didn't realise that we'd turned it off. I'll restore the credentials to the key vault, so they'll be active from the next deploy.

lawrence-forooghian: As discussed, we fully removed basic auth in 0ebe811, so just removing the documentation.

PR #1027, opened 2020-04-02
Move shipping logs to logstash to Filebeat

robbiepaul:
We currently send log entries to Logstash as inline TCP messages.
We've found this to be quite inconsistent and several log entries have
gone missing.

Filebeat is considered a more reliable way of sending log entries to
Logstash. It will track and retry any entries that have failed to be sent.

I attempted to run Filebeat in a container instance and mount the logs
(from Azure file storage) to the container. However Filebeat wasn't
picking up any changes to the file. On further investigation, they
recommend that you don't use network mounted files and instead install
Filebeat on the same client as the logs[1].

Unfortunately Filebeat don't have an official package available for
Alpine so we have to manually download and install it. To do this we
have to install the `libc6-compat` dependency[2].

Because the Filebeat command doesn't exit and outputs logs to std out we
have to use `&` to fork the process so the `bundle exec rails server`
command executes as well. It feels a little messy and we may want to run
it in the background/as a service.

[1] https://www.elastic.co/guide/en/beats/filebeat/master/filebeat-network-volumes.html
[2] https://discuss.elastic.co/t/filebeat-6-x-could-not-support-running-under-os-alpine/116195/11


Dockerfile
```diff
@@ -74,6 +75,18 @@ RUN mkdir -p ${APP_HOME}
 WORKDIR ${APP_HOME}
 # End
 
+# Download and install filebeat for sending logs to logstash
+ENV FILEBEAT_VERSION=7.6.2
+RUN apk add --update-cache curl && \
```
erbridge, line 80: `curl` should already have been installed above

Dockerfile
```diff
@@ -74,6 +75,18 @@ RUN mkdir -p ${APP_HOME}
...
+    curl https://artifacts.elastic.co/downloads/beats/filebeat/filebeat-${FILEBEAT_VERSION}-linux-x86_64.tar.gz -o filebeat.tar.gz && \
```
erbridge, line 81: Would be good to compare a checksum if possible. See what we did for the Verify image.

robbiepaul: Yea that's a good point, I've added in the checksum. Thanks F

azure/resource_groups/app/parameters/development.template.json
```diff
@@ -149,7 +149,7 @@
       }
     },
     "LOGSTASH_PORT": {
-      "value": 17002
```
lawrence-forooghian, line 152: Looks like we're changing from TCP-SSL. Why did we make this change? I think it would be good to put a bit of explanation in the commit message.

lawrence-forooghian: Also, we should get rid of the `logstash-logger` gem.

robbiepaul: Thanks Lawrence. I've updated the commit message and added a commit that removes the `logstash-logger` gem.

PR #1029, opened 2020-04-03
Add secondary SAML encryption key for VSP

robbiepaul:
In preparation for rotating the certifications we need to configure the
VSP to accept the new SAML encryption key while still using the old
one. The rotation will happen in a separate change.


robbiepaul:
[Renew VSP production certificates before April 25, 2020](https://trello.com/c/ekWDQj1S/844-renew-vsp-production-certificates-before-april-25-2020)

robbiepaul:
Yeah I double checked it was there 👍

PR #1035, opened 2020-04-03
Make the new SAML encryption key primary and remove old

robbiepaul:
We're now successfully using the new certificates so we can remove the old key and make the new key the primary one.


robbiepaul:
[Renew VSP production certificates before April 25, 2020](https://trello.com/c/ekWDQj1S/844-renew-vsp-production-certificates-before-april-25-2020)

PR #1038, opened 2020-04-03
Display decision result on the task view page for a claim

LornaHarwood:
When a decision is made on a claim, the claim's task view will now show the result within the header.
For claims which haven't had a decision made, the task view page will just display the claim reference (as shown in the before screenshot)

**Before:**
Screenshot 2020-04-03 at 17 43 29

**After:**
Screenshot 2020-04-03 at 17 42 53

This ticket has taken longer than expected to complete, so I have broken it up into separate Pull Requests as I believe that this can be merged alone. 
The next parts to this ticket are:
- to show the payroll status
- to show the claim amount
- to remove other information that will be scrubbed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ The format is based on [Keep a Changelog]
 
 - Improve message shown when inconsistent payroll information is found
 - Service operators can add a support ticket to claims
+- Display the claim's decision within the heading on the task view page
 
 ## [Release 068] - 2020-03-31
 
--- a/app/helpers/admin/claims_helper.rb
+++ b/app/helpers/admin/claims_helper.rb
@@ -95,6 +95,14 @@ module Admin
       content_tag("strong", status, class: tag_classes)
     end
 
+    def claim_summary_heading(claim)
+      if claim.decision_made?
+        claim.reference + " – " + claim.latest_decision.result.capitalize
+      else
+        claim.reference
+      end
+    end
+
     private
 
     def matching_attributes_for(claim)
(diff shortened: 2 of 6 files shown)


LornaHarwood:
[Improve the experience of viewing an approved/rejected claim](https://trello.com/c/ceFnKu1Z/1436-improve-the-experience-of-viewing-an-approved-rejected-claim)

app/helpers/claims_helper.rb
```diff
@@ -58,4 +58,12 @@ def payment_answers(claim)
   def date_of_birth_string(claim)
     claim.date_of_birth && l(claim.date_of_birth)
   end
+
+  def format_claim_reference(claim)
```
tekin, line 62: There is actually two `claim_helper.rb` files: one in the root of the `helpers` folder and another namespaced under `admin`. Because this method is used in the admin area of the application I reckon it belongs in the `admin/claim_helper.rb` file.

Also wondering if we can give this method a more appropriate (intention-revealing) name. "format claim reference" sounds like a method that would format a reference. It feels like this method's responsibility is actually to give us a _heading for the claim summary_, which happens to be a claim reference for undecided claims, and a claim reference with an outcome for decided claims. Maybe something like `claim_summary_heading` would better describe what this method gives us?

spec/helpers/claims_helper_spec.rb
```diff
@@ -219,4 +219,22 @@
       expect(helper.student_loan_answers(claim)).to eq expected_answers
     end
   end
+
+  describe "#format_claim_reference" do
```
tekin, line 223: When we move the helper to the other file we'll also want to move these specs to the corresponding spec file: `spec/helpers/admin/claims_helper_spec.rb`

tekin:
Oh, and a changelog entry would be good.

app/helpers/claims_helper.rb
LornaHarwood: I agree, `claim_summary_heading` is more descriptive. I have made this change.

LornaHarwood:
I think my changes have removed your comments about using the correct `claims_helper.rb` file and the relating spec file. (I potentially did the amendments incorrectly for a PR). I have moved the code into `app/helpers/admin/claims_helper.rb` and `spec/helpers/admin/claims_helper_spec.rb`

tekin:
You'll need to rebase with master as there is a conflict in the CHANGELOG.

PR #1044, opened 2020-04-06
Add instructions for onboarding a support developer

lawrence-forooghian:
I have not yet tested this end to end with a developer, so it might need more tweaks.


lawrence-forooghian:
I've improved the documentation about how to set up Google 2FA, explaining how to do it without a phone.

PR #1050, opened 2020-04-07
Task view shows the claims payroll status and claim amount

LornaHarwood:
The task view now displays the `Payroll status` and the `Claim amount` 

**Before:** 
Screenshot 2020-04-07 at 13 07 13

**After:**
Screenshot 2020-04-07 at 13 06 37

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog]
 ## [Unreleased]
 
 - Display payroll status of claim within task view
+- Display claim amount within task view
 
 ## [Release 073] - 2020-04-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
+- Display payroll status of claim within task view
+
 ## [Release 073] - 2020-04-07
 
 - Move shipping logs to logstash to Filebeat
--- a/app/views/admin/tasks/_claim_summary.html.erb
+++ b/app/views/admin/tasks/_claim_summary.html.erb
@@ -49,6 +49,17 @@
         <%= l(claim.submitted_at) %>
       </dd>
     </div>
+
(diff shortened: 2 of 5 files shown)


LornaHarwood:
[Improve the experience of viewing an approved/rejected claim](https://trello.com/c/ceFnKu1Z/1436-improve-the-experience-of-viewing-an-approved-rejected-claim)

app/views/admin/tasks/_claim_summary.html.erb
```diff
@@ -49,6 +49,17 @@
         <%= l(claim.submitted_at) %>
       </dd>
     </div>
+
+    <div class="govuk-summary-list__row">
+      <dt class="govuk-summary-list__key">
+        Payroll status
```
jacksonj04, line 55: I'm not sure "Payroll status" is the clearest indicator of what this field will tell us - "Awaiting payroll" is certainly a status, but "April 2020" is more of an identifier of which payroll the claim was paid out in.

Perhaps changing the output when a payroll has been processed to something like "Included in April 2020 payroll run"?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog]
 ## [Unreleased
 
 - Display payroll status of claim within task view
+- Display claim amount of claim within task view
```
jacksonj04, line 10: Redundant `claim` in this sentence

app/views/admin/tasks/_claim_summary.html.erb
LornaHarwood: This is a good point. I can see value in making it clearer and more descriptive.
I had used "April 2020", or rather "Month Year", as this was what Rich had stated within the ticket description points. 
@tekin - do you feel this would be helpful?

LornaHarwood: Based on the limited time we have, I think it will need to be left as it is and could be something to be looked at for future devs

Issue #1063, opened 2020-04-29
Broken links in documentation


CristinaRO:
All the links in the first paragraph are currently broken:

https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/master/docs/infrastructure-deployment.md#deployment-process

lawrence-forooghian:
Weird – the same technique of putting the links at the bottom of the document seems to work in `CHANGELOG.md`. I'll take a look, thanks!

lawrence-forooghian:
Opened #1071 to fix this.

PR #1071, opened 2020-05-11
Fix broken links in documentation

lawrence-forooghian:
These are “reference-style links” according to
https://www.markdownguide.org/basic-syntax#reference-style-links.

See https://github.com/DFE-Digital/claim-additional-payments-for-teaching/issues/1063


CristinaRO:
Would it be too out of scope to fix the "you'll recieve" typo as well? >.>

PR #1158, opened 2020-09-24
Update data report request csv, corresponding tests and changelong

xswoodsx:

### DIFF
--- a/app/models/claim/data_report_request.rb
+++ b/app/models/claim/data_report_request.rb
@@ -8,7 +8,9 @@ class Claim
     HEADERS = [
       "Claim reference",
       "Teacher reference number",
-      "Full name"
+      "Full name",
+      "Email",
+      "Date of birth"
     ].freeze
 
     def initialize(claims)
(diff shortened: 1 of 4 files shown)


app/models/claim/data_report_request.rb
```diff
@@ -21,7 +23,9 @@ def to_csv
           csv << [
             claim.reference,
             claim.teacher_reference_number,
-            claim.full_name
+            claim.full_name,
+            claim.email_address,
+            claim.date_of_birth
```
jacksonj04, line 28: A heads up from past experience which I'm not certain if it applies here or not - you may want to double check that Excel will happily understand this date without ambiguity (eg "04/08/2020" will upset it as it won't know if we're talking April or August). You may want to be explicit about the output format.

jacksonj04 (changes requested):
I'd definitely recommend squashing these two commits into one, so you just have a single commit which passes tests.

app/models/claim/data_report_request.rb
xswoodsx: Checked in excel and it worked fine.

jacksonj04 (approved):

PR #1184, opened 2020-10-30
Fix incorrect display of claims in payroll run

lawrence-forooghian:
We were using two different ways to gather the list of claims to be
displayed for a payment. Each of these imposes a different ordering on
the claims in the SQL that it generates:

1. payment.claims.first: this does ORDER BY claims.created_at
2. payment.claims: this does no ORDER BY and hence returns Postgres's
   default order, which is undefined [1].

This means that when the Postgres default order doesn’t match the
created_at order, these two sources of data won’t agree and hence we
might see duplicate and missing claims against a payment.

This commit changes it so that we only use one source of data -
payment.claims.

I haven’t added a regression test here because I’m not sure how I’d set
up example data where the Postgres order doesn’t match the created_at
order, given that the Postgres order is unpredictable.

[1] https://www.postgresql.org/docs/current/queries-order.html

Zendesk: https://dxw.zendesk.com/agent/tickets/12942

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog]
 
 - Update the DQT automation to include HECOS codes as well as JACS codes
 - Update email copy for claim submitted and claim approved
+- Fix incorrect display of claims in payroll run
 
 ## [Release 086] - 2020-10-08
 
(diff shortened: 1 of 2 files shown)


jacksonj04 (approved):
LGTM

PR #1185, opened 2020-11-02
Remove some extra detail from claim submitted email

lawrence-forooghian:
The client requested that we remove the additional information about
phoning the claimant to verify their identity. The email is now the same
regardless of whether the claimant’s identity is already verified or
not. This is because they have had to “change (their) processes”.

This tweaks the original change made in e5a4543.

(No change to the `CHANGELOG` since it's already covered in the "Unreleased" section there.)


pezholio (approved):
LGTM 👍

PR #1241, opened 2021-02-01
Disable alerts on development

robbiepaul:
There's noone actively working on the project, the alerts for
development are mostly ignored and generate noise.

This can easily be reversed when people start sprinting on the project
again.

The production alerts will continue to notify us but they may need some
tweaking as they are also generating a lot of noise.


jacksonj04 (approved):
Thank You

PR #1242, opened 2021-02-01
Reduce sensitivity of high response time alert

robbiepaul:
The high response time alert is triggering a lot. Looking at the history
it seems to get back to normal after ~5 minutes. The impact on the user
would be minimal so I don't think it's worth investigating the
underlying cause just yet.

We previously reduced the sensitivity of the high CPU usage alert (in
f1e949da55a5c9b5e7c56d763801ce4d30b6d68b). I've opted to use the same
thresholds as that change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog]
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


WorkSutton (approved):
LGTM

PR #1249, opened 2021-02-11
Reduce frequency of high response time alert

robbiepaul:


robbiepaul:
Closing in favour of #1242

PR #1254, opened 2021-02-23
[ECP-194]- update documentation as the process has changed

WorkSutton:
[ECP-194](https://dfedigital.atlassian.net/jira/software/projects/ECP/boards/84?selectedIssue=ECP-194) As part of reviewing access via DFE Sign-in it was identified that the process on the DfE side had changed and this is the resulting update to documentation

### DIFF
--- a/docs/dfe-sign-in.md
+++ b/docs/dfe-sign-in.md
@@ -7,28 +7,60 @@ The project’s Confluence wiki contains
 [instructions](https://dfedigital.atlassian.net/wiki/spaces/TP/pages/1102053404/Granting+and+revoking+access+to+back-office)
 on how to add new service operators to the production environment.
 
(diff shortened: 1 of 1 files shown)


mattonfoot (approved):
LGTM

mattonfoot (approved):
Thanks for Clarifying these

WorkSutton:
@mattonfoot Just double checking as Kristi has put a comment in the JIRA. Could you confirm please, so if I need to I can rework the ticket.

```
Hi Mark, I think the “obtaining security access” step is superfluous. This was what i did first which didn’t seem to lead anywhere, so you can remove that section. Otherwise, that all looks good to me   

Regarding your other questions Service Portal is the correct name.

The question about DSI approval is definitely one for Matt Smith
```

PR #1283, opened 2021-03-29
[feature/ECP-317] - Update ADR records

mahsanshakir:
- reflects the decision to deprecate ADR-0008
- reflects new ADR 0012 to replace GOVUK Verify with DqT integration
This pull request is only to update the document so that we could record all architecture decisions that could make an impact on the service.

Co-authored-by: Muhammad Shakir <muhammad.shakir@and.digital>


martikat (approved):
LGTM

PR #1295, opened 2021-04-19
[story/ECP-464] - User Journey Page 4 - 'Are you currently employed as a Supply Teacher'

martikat:
Add this page for the Early Career Payments claimant journey

### DIFF
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -30,7 +30,7 @@ class ClaimsController < BasePublicController
   def update
     current_claim.attributes = claim_params
     current_claim.reset_dependent_answers
-    current_claim.eligibility.reset_dependent_answers
+    current_claim.eligibility.reset_dependent_answers unless current_claim.has_ecp_policy?
 
     if current_claim.save(context: page_sequence.current_slug.to_sym)
       redirect_to claim_path(current_policy_routing_name, next_slug)
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -106,50 +106,50 @@ class Claim < ApplicationRecord
 
   validates :academic_year_before_type_cast, format: {with: PolicyConfiguration::ACADEMIC_YEAR_REGEXP}
 
-  validates :payroll_gender, on: [:gender, :submit], presence: {message: "Choose the option for the gender your school’s payroll system associates with you"}
+  validates :payroll_gender, on: [:gender, :submit], presence: {message: "Choose the option for the gender your school’s payroll system associates with you"}, unless: :has_ecp_policy?
 
-  validates :first_name, on: [:name, :submit], presence: {message: "Enter your first name"}
+  validates :first_name, on: [:name, :submit], presence: {message: "Enter your first name"}, unless: :has_ecp_policy?
   validates :first_name, length: {maximum: 100, message: "First name must be 100 characters or less"}
 
   validates :middle_name, length: {maximum: 100, message: "Middle name must be 100 characters or less"}
 
-  validates :surname, on: [:name, :submit], presence: {message: "Enter your last name"}
+  validates :surname, on: [:name, :submit], presence: {message: "Enter your last name"}, unless: :has_ecp_policy?
   validates :surname, length: {maximum: 100, message: "Last name must be 100 characters or less"}
 
-  validates :address_line_1, on: [:address, :submit], presence: {message: "Enter your building and street address"}
+  validates :address_line_1, on: [:address, :submit], presence: {message: "Enter your building and street address"}, unless: :has_ecp_policy?
   validates :address_line_1, length: {maximum: 100, message: "Address lines must be 100 characters or less"}
   validates :address_line_2, length: {maximum: 100, message: "Address lines must be 100 characters or less"}
   validates :address_line_3, length: {maximum: 100, message: "Address lines must be 100 characters or less"}
   validates :address_line_4, length: {maximum: 100, message: "Address lines must be 100 characters or less"}
 
-  validates :postcode, on: [:address, :submit], presence: {message: "Enter your postcode"}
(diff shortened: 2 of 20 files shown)


dbrunnellAD (approved):

dbrunnellAD (commented):
I've approved the PR but will we be updating the changelog for each new page?

nomasprime (approved):

app/models/claim.rb
```diff
@@ -304,6 +302,10 @@ def decision_undoable?
 
   private
 
+  def ecp_skip_validation
```
nomasprime, line 305: Methods that return boolean should probably end in `?`. Better name would be something like `is_ecp_policy?`.

app/controllers/claims_controller.rb
```diff
@@ -30,7 +30,7 @@ def show
   def update
     current_claim.attributes = claim_params
     current_claim.reset_dependent_answers
-    current_claim.eligibility.reset_dependent_answers
+    current_claim.eligibility.reset_dependent_answers unless current_claim.policy == EarlyCareerPayments
```
nomasprime, line 33: Why are we doing this?

spec/models/claim_spec.rb
```diff
@@ -905,14 +905,15 @@
   end
 
   describe "Early Career Payments claim" do
```
nomasprime, line 907: These specs made it in because the previous PR was rubber stamped but actually they don't look right and should be somewhere else.

I'm all for leveraging debt effectively but that's not what we're doing here.

Let's pair for a few minutes and get this sorted now rather than let it get out of hand and slow us down even more later on.

nomasprime (changes requested):
Generally looks good, just a few quick changes.

app/controllers/claims_controller.rb
WorkSutton: @nomasprime This was what I talked about in the walkthrough yesterday, I could have opted to put the method into the `EarlyCareerPayments::Eligibility` class. I opted for this instead

spec/models/claim_spec.rb
WorkSutton: Sure thing

spec/models/claim_spec.rb
```diff
@@ -905,14 +905,15 @@
...
-    let(:eligibility) { build(:early_career_payments_eligibility) }
+    let(:eligibility) { build(:early_career_payments_eligibility, :eligible) }
 
     describe "#submittable?" do
```
nomasprime, line 910: Should be in context under existing describe for `#submittable?`.

spec/models/claim_spec.rb
```diff
@@ -921,7 +922,7 @@
     end
 
     it "triggers validations on the eligibility appropriate to the context" do
```
nomasprime, line 924: Again, should be in context under existing validations.

app/controllers/claims_controller.rb
nomasprime: Understand this now.

Personally would have opted for the latter because conditional logic to decide whether to call something is usually a smell. Can't remember technical term off the top of my head but see null object for something similar.

nomasprime:
One last thing, there are still lots of pending specs from the rubber stamped PR. Reiterate, normally wouldn't merge those at all.

Why are they still here?

spec/models/claim_spec.rb
WorkSutton: Done as suggested. Thanks for pointing out the error in my logic and positioning @nomasprime

spec/models/claim_spec.rb
WorkSutton: @nomasprime As above, moved as we discussed

WorkSutton:
@nomasprime Have removed all tracks of Pending specs by changing from using `Policies.all` to just having [MathAndPhysics, StudentLoans]` as the items to be iterated over for driving those previously pending specs where I conditionally made them pending as fixing all the Admin issues is not part of the Claimant journey.

We have tickets for the Admin phases later in the project

Hope that makes sense

spec/models/claim_spec.rb
```diff
@@ -148,11 +148,24 @@
     expect(claim).to be_valid(:submit)
   end
 
-  it "triggers validations on the eligibility appropriate to the context" do
-    claim = build(:claim)
+  describe "Policy eligiblitiy validations appropriate to the context" do
```
nomasprime, line 151: What context?

nomasprime (changes requested):

spec/models/claim_spec.rb
```diff
@@ -148,11 +148,24 @@
...
+      claim = build(:claim)
 
-    expect(claim).not_to be_valid(:"qts-year")
-    expect(claim.errors.values).to include(["Select when you completed your initial teacher training"])
+      expect(claim).not_to be_valid(:"qts-year")
+      expect(claim.errors.values).to include(["Select when you completed your initial teacher training"])
+    end
+
+    context "Early Career Payments claim" do
+      let(:eligibility) { build(:early_career_payments_eligibility, :eligible) }
```
nomasprime, line 160: Don't think this is used.

app/models/claim.rb
WorkSutton: Fixed as suggested. Its temporary anyway until such time as we have a valid claim. Then this can be removed

spec/models/claim_spec.rb
WorkSutton: This is what we have just spent time on fixing so hopefully this is all good now from our joint pairing session

app/controllers/claims_controller.rb
nomasprime, line 33: How come not calling `is_ecp_policy?` here?

nomasprime (approved):
Yay! Got there 🎉

PR #1299, opened 2021-04-20
[story/ECP-467] - User Journey Page 7 - 'Are you currently subject to action for poor performance?'

dbrunnellAD:
View build for the poor performance question.


config/locales/en.yml
```diff
@@ -210,3 +210,4 @@ en:
     landing_page: "Apply for an early-career payment"
     questions:
       nqt_in_academic_year_after_itt: "Did you do your NQT the academic year after your ITT?"
+      poor_performance: "Are you currently subject to action for poor performance?"
```
WorkSutton, line 213: @dbrunnellAD this needs a new line after the last entry. Thats why it has that no smoking sign at the end in red. Could you fix and push up that chnage please.

Otherwise the view Looks Good to Me (my LGTM version) and I will pick-up to do the wiring-up. Thanks

martikat (approved):
LGTM

app/models/early_career_payments/slug_sequence.rb
```diff
@@ -14,6 +14,7 @@ class SlugSequence
       "supply-teacher",
       "entire-term-contract",
       "employed-directly",
+      "formal-performance-action",
```
martikat, line 17: Just had a check against the v7 prototype and it looks like the slug is called "poor-performance", so I think this will need to be changed to match up to the prototype

config/locales/en.yml
```diff
@@ -234,3 +235,4 @@ en:
         modern_foreign_languages: "Modern foreign languages"
         physics: "Physics"
         none_of_the_above: "None of the above"
+
```
martikat, line 238: There is some whitespace here that needs to be removed

app/models/early_career_payments/slug_sequence.rb
dbrunnellAD: Hi we changed it from poor performance to stick to what its called on the previous journeys to stay consistent with the codebase

config/locales/en.yml
dbrunnellAD: removed the line

martikat (approved):

martikat (approved):

PR #1304, opened 2021-04-22
[story/ECP-479] - User Journey Page 16 - 'You are eligible for an early careers payment'

martikat:
View build for the eligible page for ECP


app/models/early_career_payments.rb
```diff
@@ -49,4 +49,8 @@ def first_eligible_qts_award_year(claim_year = nil)
   def last_ineligible_qts_award_year
     first_eligible_qts_award_year - 1
   end
+
+  def student_loan_balance_url
```
WorkSutton, line 53: @martikat There should be a spec to cover this method in the corresponding model spec. Could you add please?

app/views/early_career_payments/claims/eligibility_confirmed.html.erb
```diff
@@ -0,0 +1,61 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">You are eligible <%= I18n.t("early_career_payments.claim_description") %></h1>
+    <p class="govuk-body">
+      Based on the answers you have provided you will be able to claim £5,000.
+      This could increase to £7,500 if the school you teach at is considered
+      to be in an uplift area.
+    </p>
+
+    <p>
+      <%= link_to "Uplift areas", "#{EarlyCareerPayments.eligibility_page_url}#local-authorities-with-uplift-payments", class: "govuk-link" %>
```
WorkSutton, line 13: Not 100% sure this is the same eligibility page url as referenced elsewhere in the application. Could you have a look at what the M+P & TSLR ones are and reach out to Laura/Muhammad about this please. Not a blocker, but something to be aware of.

WorkSutton: Looking at the code in the Prototype V7 on the link there is are `target` and `title` attributes. Can you get these added to please. Please check with Brian to confirm, I think we use `title` for assistive technologies.

app/views/early_career_payments/claims/eligibility_confirmed.html.erb
```diff
@@ -0,0 +1,61 @@
...
+    </p>
+
+    <h2 class="govuk-heading-l">What happens next?</h2>
+
+    <p class="govuk-body govuk-!-font-weight-bold">
```
WorkSutton, line 18: Where did the `govuk-!-font-weight-bold` class come from?

app/views/early_career_payments/claims/eligibility_confirmed.html.erb
```diff
@@ -0,0 +1,61 @@
...
+          your school, the certificate you got when you qualified, or
+          by contacting the teacher qualifications helpdesk on
+          <%= link_to "0207 593 5394", "#", class: "govuk-link" %> or
+          <%= link_to "qts.enquiries@education.gov.uk", "#", class: "govuk-link" %>
+        </p>
+      </li>
+      <li>
+        <p class="govuk-body">
+          the academic year in which you completed your ITT - you
+          can find this on your QTS certificate or through teach
```
WorkSutton, line 45: The prototype V7 has `the` not `teach` as the last word before the link to `teacher self-service`

app/views/early_career_payments/claims/eligibility_confirmed.html.erb
```diff
@@ -0,0 +1,61 @@
...
+          <%= link_to "teacher self-service", "#", class: "govuk-link" %> website
+        </p>
+      </li>
+      <li>
+        <p class="govuk-body">
+          when you took out your student loan (if you have one) - if
+          you don't know when this was, you can find out by logging into your
+          <%= link_to "student loan account", EarlyCareerPayments.student_loan_balance_url, class: "govuk-link" %>
```
WorkSutton, line 53: Again should have a `target` and `title` attribute.

Please see the documentation https://design-system.service.gov.uk/styles/typography/#links which I believe gives good guidance on this. Might also be worth checking with Brian.
See `Opening links in a new tab` section

WorkSutton (changes requested):
Just a few minor niggles around some links. Happy to assist if my notes are not clear, or unsure.

app/views/early_career_payments/claims/eligibility_confirmed.html.erb
martikat: The font weight class can be found here: https://design-system.service.gov.uk/styles/typography/#font-weight

app/models/early_career_payments.rb
martikat: Added spec in spec/models/early_career_payments_spec.rb

WorkSutton:
LGTM for the wiring-up phase

PR #1307, opened 2021-04-23
[story/ECP-471] - User Journey Page 10 - 'Which subject did you do your undergraduate ITT in?'

martikat:
View build for the undergraduate ITT subject page for ECP

### DIFF
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -306,10 +306,6 @@ class Claim < ApplicationRecord
 
   private
 
-  def is_ecp_policy?
-    policy == EarlyCareerPayments
-  end
-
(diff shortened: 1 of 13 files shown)


dbrunnellAD (approved):

app/models/claim.rb
```diff
@@ -306,10 +306,6 @@ def has_ecp_policy?
 
   private
 
-  def is_ecp_policy?
```
nomasprime, line 309: Curious why this is still here? Thought we renamed before.

WorkSutton: Yes picked this up and have removed. Was a c/p error. Apologies

PR #1312, opened 2021-04-28
Fix GOV.UK Verify removal [ECP-593]

nomasprime:
- Remove verified trait from submittable claim factory
- Fix automated checks update admin claim tasks with DQT
- Fix automated checks DQT report consumer spec
- Fix admin checks claim specs

Bugs are fixed but UX needs some work, especially admin tasks identity confirmation:

- The question for outstanding claims that went through Verify don't mention it
- The question for claims that haven't gone through Verify say they haven't which is technically true but, apart from the above, we're not actually using Verify anymore.


WorkSutton (approved):
Nice catch on this bug @nomasprime

PR #1321, opened 2021-05-04
[story/ECP-504] - User Journey Page 36 - 'Student Loans: Did you take out a Postgraduate Doctoral Loan on or after 1 August 2016?'

martikat:


app/views/early_career_payments/claims/doctoral_loan.html.erb
```diff
@@ -0,0 +1,48 @@
...
+    <%= render("shared/error_summary", instance: current_claim, errored_field_id_overrides: { "eligibility.postgraduate_doctoral_loan": "claim_eligibility_attributes_postgraduate_doctoral_loan_true" }) if current_claim.errors.any? %>
+    <%= form_for current_claim, url: path_for_form do |form| %>
+      <%= form_group_tag current_claim do %>
+        <%= form.fields_for :eligibility, include_id: false do |fields| %>
+
+          <%= fields.hidden_field :postgraduate_doctoral_loan %>
+
+          <fieldset class="govuk-fieldset">
+
+            <span class="govuk-caption-xl">About your student loans</span>
```
WorkSutton, line 15: @martikat Just looking at the layouts in the prototype. I think this needs to be outside of the <fieldset> tag? Can you double check please?

app/views/early_career_payments/claims/masters_loan.html.erb
```diff
@@ -13,27 +13,28 @@
           <fieldset class="govuk-fieldset">
 
             <span class="govuk-caption-xl">About your student loans</span>
```
WorkSutton, line 15: Same as comment above. Just need to double check please

WorkSutton (changes requested):
@martikat Nothing major in the code, but would like us to check the html file for correct positioning of the span containing `About your student loans`

Apologies I did not catch this in the previous view which is near identical

WorkSutton (approved):
@martikat Thanks for checking and making the changes. Approving

PR #1322, opened 2021-05-04
[story/ECP-494] - Personal Details: What gender does your school's payroll system associate with you?

dbrunnellAD:
- Dynamic title sizing added to pre existing file
- Tests added to ECP claim journey

### DIFF
--- a/app/models/early_career_payments/slug_sequence.rb
+++ b/app/models/early_career_payments/slug_sequence.rb
@@ -26,6 +26,7 @@ module EarlyCareerPayments
       # personal details phase of claim journey
       "how_we_will_use_information_provided",
       "personal-details",
+      "gender",
       # student loans phase of claim journey
       "masters-loan",
       "doctoral-loan",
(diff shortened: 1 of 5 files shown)


app/models/early_career_payments/slug_sequence.rb
```diff
@@ -26,6 +26,7 @@ class SlugSequence
       # personal details phase of claim journey
       "how_we_will_use_information_provided",
       "personal-details",
+      "gender",
```
WorkSutton, line 29: I find our use of `gender` and `payroll gender` confusing. Is this the name of the slug for the URL that our UCD colleagues gave us? Or an existing one for the current process in TSLR / M+P?

dbrunnellAD: The slug in the M&P journey is just labeled as "gender" - I went with sticking to the consistency of the codebase

WorkSutton: ok lets be pragmatic and we can alter once we have confirmation from UT / AC feedback if indeed it does need to change. I'm all in favour of consistency

WorkSutton (approved):
LGTM

PR #1274, opened 2021-03-24
[task/ECP-319] - remove references to GOVUK Verify from Azure scripts

WorkSutton:
[JIRA ECP-319](https://dfedigital.atlassian.net/browse/ECP-319)
Remove the from the Azure resource groups and templates references to:

- Signing Key
- SAML_PRIMARY_ENCRYPTION_KEY
- SAML_SECONDARY_ENCRYPTION_KEY
- vspDockerImageTag
- vspAppServiceDockerImage
- vspAppServiceRuntimeStack
- vspAppServicePlanSkuTier
- vspAppServicePlanSkuSize
- vspAppServicePlanSkuFamily
- vspAppServicePlanSkuCapacity
- VSP.VERIFY_ENVIRONMENT
- VSP.EUROPEAN_IDENTITY_ENABLED
- VSP.VERIFY_ENTITY_ID
- GOVUK_VERIFY_VSP_HOST
- vspAppServicePlanId
- vspAppServicePlanName
- vspAppServiceName
- vspAppServiceId
- verifyHubPossibleOutboundIpAddresses
- vspDeploymentName
- vspAppServiceWebTestDeploymentName
- vspAppServiceWebTestName

Remove the vsp specific json file
Remove from Alerts

PR #1344, opened 2021-05-10
Revert "[task/ECP-319] - remove references to GOVUK Verify from Azure scripts"

WorkSutton:
Reverts DFE-Digital/claim-additional-payments-for-teaching#1274

This is being reverted because on investigation into why Releases are failing on Azure Development (and would be too on Production) are caused by the incorrect sequencing of resource deletion.
It turns out that the last Microsoft component to be deleted is the `AppServicePlan` so that needs to be decoupled from this removal.

As per this article from Microsoft - https://docs.microsoft.com/en-us/answers/questions/169500/unable-to-delete-virtual-network-and-a-resource-gr.html

Also reverting as about to go into Private Beta. Prior to this we want to be have the code deployed in Azure development for our Business colleagues to perform exploratory testing against.

Approach discussed and investigation by @WorkSutton and @and-lucas-kelly

### DIFF
--- a/azure/resource_groups/alerts/template.json
+++ b/azure/resource_groups/alerts/template.json
@@ -34,6 +34,12 @@
     "workerContainerInstanceId": {
       "type": "string"
     },
+    "vspAppServiceId": {
+      "type": "string"
+    },
+    "vspAppServicePlanId": {
+      "type": "string"
+    },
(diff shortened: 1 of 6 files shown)


and-lucas-kelly (approved):
reviewed with MS

PR #1383, opened 2021-05-24
reverted back to dockerhub

and-lucas-kelly:


nomasprime (commented):
Not really in a position to review DevOps changes but I would say the commented out code should probably be deleted rather than merged into master. Can always pop those changes uncommented onto another branch if the motivation is to implement those changes later.

and-lucas-kelly:
The code that has been commented is the code that should go live rather than the code I’m having to put into master. Rather than loose it I’ve left it commented out for easy revert

nomasprime (approved):

PR #1384, opened 2021-05-24
Add DQT client vars to deployment

nomasprime:


and-lucas-kelly (approved):

nomasprime:
@and-lucas-kelly any idea re build error?

nomasprime:
Thought the build failure was due to the `RUN` statement on `Dockerfile:117` but turns out the actual error's further up the output and related to DQT initializer.

Not sure why it's been done like this but I'll add a guard clause to the `json_to_hash` method that should prevent the error.

PR #1386, opened 2021-05-24
[story/ECP-274] - Implement One Time Password (OTP) Email Component in ECP User Journey

WorkSutton:

### DIFF
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -1,5 +1,6 @@
 class ClaimsController < BasePublicController
   include PartOfClaimJourney
+  include OneTimePassword
 
   skip_before_action :send_unstarted_claiments_to_the_start, only: [:new, :create, :timeout]
   before_action :check_page_is_in_sequence, only: [:show, :update]
@@ -23,8 +24,6 @@ class ClaimsController < BasePublicController
   end
 
   def show
-    logger.debug("ClaimsController#show: #{}")
-    logger.debug("  - OTP section: #{}") if params[:slug] == "email-verification"
     search_schools if params[:school_search]
     render current_template
   end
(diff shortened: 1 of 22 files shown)


app/controllers/claims_controller.rb
```diff
@@ -113,4 +117,18 @@ def claim_slug_sequence
...
+
+  def generate_one_time_password
+    if params[:slug] == "email-address" && current_claim.has_ecp_policy?
+      one_time_password = generate_otp
+      ClaimMailer.ecp_email_verification(current_claim, one_time_password).deliver_now
+      session[:sent_one_time_password_at] = Time.now
+    end
+  end
+
+  def store_in_session_one_time_password
```
nomasprime, line 129: There's obviously a lot of model logic in this controller but, that aside, I thought we concluded there's no need to store generated OTPs. Would be good to know why we need to now?

app/views/early_career_payments/claims/_one_time_password.html.erb
```diff
@@ -0,0 +1,38 @@
+<div class="govuk-grid-row">
```
nomasprime, line 1: We currently don't send HTML emails, presumably because of the inherent complications.

Why are we doing this now and does this need testing across email clients?

nomasprime (approved):
Given the situation, approved but I hope you've taken onboard comments re email, this will probably break for many users without proper testing.

app/controllers/claims_controller.rb
WorkSutton: @nomasprime We need to store the OTP so we can determine if the reason we have nil is because it has timed out vs its a wrong challenge code

app/views/early_career_payments/claims/_one_time_password.html.erb
WorkSutton: @nomasprime I assumed that the emails not being HTML was because that was the requirement when the product was originally delivered. It might be to do with how Notify is used too. I have just been consistent as we only have `.text.erb` templates in the existing repo

nomasprime: Sorry, I realise what the actually change is now. All good 👍

PR #1400, opened 2021-05-28
removed dqt_client_port from container build

and-lucas-kelly:


nomasprime:
Think there might have been a misunderstanding on Friday. Don't think we need to remove this, doesn't it work as expected with `nil`?

and-lucas-kelly:
You can't have a null vale in Key Vault and i'm not sure how it would behave if we added nil as the value. Will the port number not always be set in the code?

nomasprime:
Didn't we already demonstrate it as working with no value?

nomasprime (approved):
Doesn't look like the right fix but, given limited support during release week, I'll approve on faith that it's a fix, at least for now.

PR #1403, opened 2021-05-31
[story/ECP-908] - change to using correct award amounts in eligible later 2022 screen

WorkSutton:
Show the base amount and the uplifted amount based on subject and academic ITT year
Add details of criteria in an initializer file

### DIFF
--- a/app/helpers/claims/show_helper.rb
+++ b/app/helpers/claims/show_helper.rb
@@ -11,5 +11,21 @@ module Claims
     def shared_view_css_class_size(claim)
       claim.has_ecp_policy? ? "l" : "xl"
     end
+
+    def base_and_uplift_award_amounts(claim, year)
+      return nil unless claim.has_ecp_policy?
+
+      base_amount = MATRIX.dig(claim.eligibility.eligible_itt_subject, claim.eligibility.itt_academic_year, year)
+      uplifted = case base_amount
+      when 0
+        0
+      when 2000
+        3000
+      when 5000
+        7500
+      end
+
+      {base: BigDecimal(base_amount), uplifted: BigDecimal(uplifted)}
+    end
   end
 end
--- a/app/models/early_career_payments/eligibility_matrix_calculator.rb
+++ b/app/models/early_career_payments/eligibility_matrix_calculator.rb
@@ -1,12 +1,5 @@
 module EarlyCareerPayments
   class EligibilityMatrixCalculator
-    MATRIX = {
-      mathematics: ["2018_2019", "2019_2020", "2020_2021"],
-      physics: ["2020_2021"],
-      chemistry: ["2020_2021"],
-      foreign_languages: ["2020_2021"]
-    }.freeze
-
     attr_reader :eligibility
 
     def initialize(eligibility)
@@ -16,7 +9,7 @@ module EarlyCareerPayments
     def eligible_later?
       return false if subject == :mathematics && itt_academic_year == "2018_2019"
 
-      MATRIX.dig(subject)&.include?(itt_academic_year) ? true : false
+      MATRIX.dig(subject.to_s, itt_academic_year).nil? ? false : true
     end
 
     private
--- a/config/initializers/early_career_payments/eligibility_matrix_calculator.rb
+++ b/config/initializers/early_career_payments/eligibility_matrix_calculator.rb
@@ -1,0 +1,1 @@
+MATRIX = YAML.safe_load(File.read(File.expand_path("../eligibility_matrix.yml", __FILE__))).freeze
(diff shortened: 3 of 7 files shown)


martikat (approved):

app/helpers/claims/show_helper.rb
```diff
@@ -11,5 +11,21 @@ def claim_submitted_title(claim)
     def shared_view_css_class_size(claim)
       claim.has_ecp_policy? ? "l" : "xl"
     end
+
+    def base_and_uplift_award_amounts(claim, year)
```
nomasprime, line 15: Seems like this should be named after an object that it returns, ie, missing object.

app/helpers/claims/show_helper.rb
```diff
@@ -11,5 +11,21 @@ def claim_submitted_title(claim)
...
+      return nil unless claim.has_ecp_policy?
+
+      base_amount = MATRIX.dig(claim.eligibility.eligible_itt_subject, claim.eligibility.itt_academic_year, year)
```
nomasprime, line 18: Not obvious what this does, missing method?

app/models/early_career_payments/eligibility_matrix_calculator.rb
```diff
@@ -1,12 +1,5 @@
 module EarlyCareerPayments
   class EligibilityMatrixCalculator
```
nomasprime, line 2: Didn't mention on previous PR but I think in future we could do better on consolidating and naming eligibility related code.

app/models/early_career_payments/eligibility_matrix_calculator.rb
```diff
@@ -16,7 +9,7 @@ def initialize(eligibility)
     def eligible_later?
       return false if subject == :mathematics && itt_academic_year == "2018_2019"
```
nomasprime, line 10: Also, didn't mention feels odd having mixture of symbols and strings here. These could be missing objects.

app/models/early_career_payments/eligibility_matrix_calculator.rb
```diff
@@ -16,7 +9,7 @@ def initialize(eligibility)
...
 
-      MATRIX.dig(subject)&.include?(itt_academic_year) ? true : false
+      MATRIX.dig(subject.to_s, itt_academic_year).nil? ? false : true
```
nomasprime, line 12: Didn't say anything before given current situation but glad you changed this 👍

config/initializers/early_career_payments/eligibility_matrix_calculator.rb
```diff
@@ -0,0 +1 @@
+MATRIX = YAML.safe_load(File.read(File.expand_path("../eligibility_matrix.yml", __FILE__))).freeze
```
nomasprime, line 1: Why YAML not inline Ruby?

spec/helpers/claims/show_helper_spec.rb
```diff
@@ -38,4 +38,76 @@
...
+
+      it "returns nil" do
+        expect(helper.base_and_uplift_award_amounts(claim, nil)).to be_nil
+      end
+    end
+
+    context "with a EarlyCareerPayments policy" do
+      let(:policy) { EarlyCareerPayments }
+
+      describe "Chemistry/Physics/Foreign Languages" do
```
nomasprime, line 54: Again, this kind of thing should be context.

nomasprime (approved):
Few things but, again, given current direction I'll approve.

Quite a few things we could improve on the specs but didn't comment because we'll probably have to revisit later anyway.

app/models/early_career_payments/eligibility_matrix_calculator.rb
WorkSutton: Yes I agree, eligibility is everywhere and its been a challenge as to what to call some methods/classes. I went with the MatrixCalculator as it calcuates if a ECP claims eligibility is actually one of 3 statuses after all of the eligibility questions have been successfully answered. These are:

- Eligible Now (in this claim window/AcademicYear)
- Eligible Later (say in 2022 if in 2021 and meets all criteria)
- No longer eligible

app/models/early_career_payments/eligibility_matrix_calculator.rb
WorkSutton: @nomasprime Yes I changed it as I changed the data structure to better reflect additional information. Its definitely better just using the `dig` method.

spec/helpers/claims/show_helper_spec.rb
WorkSutton: Is it ok to have context, context, context? I thought the linters got fussy and liked 2 levels of describes, then multiple contexts if more than 3 levels deep.
Maybe thats rubocop which we are overriding with `standard` which I have to say personally I am not a fan of

PR #1409, opened 2021-06-03
[story/ECP-918] - V10 Content Change: 'You might be eligible to claim in the future when you’ve completed your ITT'

martikat:


app/views/early_career_payments/claims/_ineligibility_reason_ineligible_nqt_in_academic_year_after_itt.html.erb
```diff
@@ -0,0 +1,15 @@
+<p class="govuk-body">
+  If you’re a trainee teacher and plan to complete your studies in this academic
+  year you might be able to apply in the next academic year.
+</p>
+
+<h3 class="govuk-heading-m">
```
WorkSutton, line 6: @martikat In the prototype this is an `h1` element. Where did the `h3` get specified. I have looked in both V10 and V11 documents and both have this as a `h1`.

app/views/early_career_payments/claims/ineligible.html.erb
```diff
@@ -6,9 +6,11 @@
 
     <%= render "ineligibility_reason_#{current_claim.eligibility.ineligibility_reason}" %>
 
-    <p class="govuk-body">
-      If you want to know more about who can apply, check our
-      <%= link_to "eligibility page", EarlyCareerPayments.eligibility_page_url, class: "govuk-link" %>.
-    </p>
+    <% if current_claim.eligibility.ineligibility_reason != :ineligible_nqt_in_academic_year_after_itt %>
```
WorkSutton, line 9: @martikat I think you might have mentioned this to me. Did we check with UCD team if this is required? I am looking at the V10 prototype and cannot see this on the error page. Could you double check please (if you have not already done so)

WorkSutton (changes requested):
@martikat Nice work on splitting out this new error page. My reason for request changes is to simply clarify some of the questions I have raised.
Technically looks good. Thank you

app/views/early_career_payments/claims/_ineligibility_reason_ineligible_nqt_in_academic_year_after_itt.html.erb
martikat: Updated this to be `h1`

app/views/early_career_payments/claims/ineligible.html.erb
martikat: This is to hide this text as it is not used on the page. The prototype does not show this text for this ineligibility reason page

app/helpers/early_career_payments_helper.rb
```diff
@@ -4,6 +4,8 @@ def ineligible_heading(claim)
       content_tag(:h1, I18n.t("early_career_payments.ineligible.poor_performance_heading"), class: "govuk-heading-xl")
     elsif claim.eligibility.ineligibility_reason == :ineligible_current_school
       content_tag(:h1, I18n.t("early_career_payments.ineligible.school_heading"), class: "govuk-heading-xl")
+    elsif claim.eligibility.ineligibility_reason == :ineligible_nqt_in_academic_year_after_itt
+      content_tag(:h1, I18n.t("early_career_payments.ineligible.reason.nqt_after_itt"), class: "govuk-heading-l")
```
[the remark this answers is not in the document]
martikat: Have added the inset text and updated the heading size to be xl to match with the other ineligible pages

WorkSutton (approved):

PR #1412, opened 2021-06-07
Update README.md

and-lucas-kelly:
updated the rails call to include bin

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ The migrations are stored in the `db/data` folder.
 
 - To generate a migration: `rails g data_migration add_this_to_that`
 - To run the data migration:
-  `rails runner db/data/$FILENAME_OF_THE_GENERATED_MIGRATION`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -191,7 +191,7 @@ The migrations are stored in the `db/data` folder.
 
 - To generate a migration: `rails g data_migration add_this_to_that`
```
WorkSutton, line 192: @and-lucas-kelly being really pedantic here, but I think for extra clarity it should say `data migration`.
I know its in the topic area, but I have added this as on the line directly after we call it a `data migration`

and-lucas-kelly: makes sense to me.

nomasprime (changes requested):
Been thinking about this and don't think it's really worth our time:

1. Can be run in multiple ways
2. Not sure runtime paths need to be in scope of this doc
3. If it is, should be updated with multiple ways and throughout, eg, bullet above is missing bin

and-lucas-kelly:
after review from RJ and MS decided changes not needed or needs an overall review and update.

PR #1413, opened 2021-06-07
[bug/ECP-907] - Edit National Insurance Number from the ECP Personal Details page is broken

martikat:
Fixed all edit links that go to the personal details page (Full name, Date of birth and National Insurance number)


WorkSutton (changes requested):
Could we have some specs written please @martikat. 
If uncertain on how to write those then please reach out. Thank you.

WorkSutton (approved):

PR #1459, opened 2021-06-28
[story/ECP-865] - Update 'Which subject did you do your qualification  in?' screen for Overseas Recognition

WorkSutton:

### DIFF
--- a/app/models/early_career_payments/eligibility.rb
+++ b/app/models/early_career_payments/eligibility.rb
@@ -73,7 +73,9 @@ module EarlyCareerPayments
     def qualification_name
       return qualification.gsub("_itt", " ITT") if qualification.split("_").last == "itt"
 
-      qualification.humanize.downcase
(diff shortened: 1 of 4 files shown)


app/models/early_career_payments/eligibility.rb
```diff
@@ -73,7 +73,9 @@ def policy
     def qualification_name
       return qualification.gsub("_itt", " ITT") if qualification.split("_").last == "itt"
 
-      qualification.humanize.downcase
+      qualification_attained = qualification.humanize.downcase
+
+      qualification_attained == "assessment only" ? qualification_attained : qualification_attained + " qualification"
```
nomasprime, line 78: Seems odd to add ` qualification` to the qualification name on the  business side of things, would be like `Mark Sutton name`. Sounds like something that should be handled on the view?

nomasprime (approved):

PR #1461, opened 2021-06-28
[story/ECP-866] - Update question for ITT Year to 'Which academic year did you earn your qualified teacher status (QTS)?' for AO and OR routes into teaching

WorkSutton:

### DIFF
--- a/app/views/early_career_payments/claims/itt_year.html.erb
+++ b/app/views/early_career_payments/claims/itt_year.html.erb
@@ -1,6 +1,4 @@
-<% qualification = current_claim.eligibility.qualification %>
-<% start_or_complete = current_claim.eligibility.qualification == "undergraduate_itt" ? "complete" : "start" %>
-<% content_for(:page_title, page_title(t("early_career_payments.questions.itt_academic_year", start_or_complete: start_or_complete, qualification: qualification), policy: current_policy_routing_name, show_error: current_claim.errors.any?)) %>
+<% content_for(:page_title, page_title(t("early_career_payments.questions.itt_academic_year.qualification.#{current_claim.eligibility.qualification}"), policy: current_policy_routing_name, show_error: current_claim.errors.any?)) %>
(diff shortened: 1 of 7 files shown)


config/locales/en.yml
```diff
@@ -257,7 +257,13 @@ en:
           overseas_recognition: "This is the subject you qualified with."
       eligible_itt_subject: "Which subject did you do your %{qualification} in?"
       teaching_subject_now: "Do you teach %{eligible_itt_subject} now?"
-      itt_academic_year: "In what academic year did you %{start_or_complete} your %{qualification}?"
+      itt_academic_year:
```
nomasprime, line 260: Tying translations to the way you've broken down a grammatical structure probably isn't a good idea because that structure might not be compatible with other languages.

I know we only have English but then you could say why do we have translations at all. I'd change this to make sure it's compatible with other languages, see example directly below your change.

nomasprime (changes requested):

config/locales/en.yml
WorkSutton: @nomasprime so you just want 4 entries?

We already do this elsewhere in the file, so not sure why being called out now?

I think we are using translations more for the simplicity of we only have to change them in one place

app/views/early_career_payments/claims/itt_year.html.erb
```diff
@@ -52,20 +50,22 @@
...
-          </summary>
-          <div class="govuk-details__text">
-            <p class="govuk-body">
-              You might still be eligible to claim if your ITT coincided with one of the
-              academic years stated, even if it didn’t start or complete in one of those
-              years.
-            </p>
-          </div>
-        </details>
+        <% if current_claim.eligibility.qualification == "postgraduate_itt" || current_claim.eligibility.qualification == "undergraduate_itt" %>
```
nomasprime, line 53: To avoid duplication couldn't you do something like `%w[foo bar].include?(something)`?

nomasprime (approved):
Pre-approved, up to you if you want to clean up a bit.

PR #1476, opened 2021-07-05
Update DQT client qualified teacher status to handle null data response

nomasprime:


WorkSutton (approved):

WorkSutton:
@nomasprime I would still like to see the ECP JIRA ticket number on these so they can be traced in the scenario of an audit.

PR #1477, opened 2021-07-05
Fix admin claim tasks update with DQT API without matching DQT identity

nomasprime:
We were returning the wrong response when stubbing the request, ie, worked locally but not when deployed.

ECP-1056

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ The format is based on [Keep a Changelog]
     later journey
   - Add 'Mobile number' screen
   - Update DQT client qualified teacher status to handle 404 response
+  - Fix admin claim tasks update with DQT API without matching DQT identity
 
 ## [Release 091] - 2021-06-24
 
--- a/app/models/automated_checks/claim_verifiers/identity.rb
+++ b/app/models/automated_checks/claim_verifiers/identity.rb
@@ -14,9 +14,10 @@ module AutomatedChecks
       def perform
         return unless awaiting_task?("identity_confirmation")
 
-        complete_match ||
-          partial_match ||
-          no_match
+        # Order matters so that subsequent conditions in methods fall through to execute the right thing
+        no_match ||
+          complete_match ||
+          partial_match
       end
 
       private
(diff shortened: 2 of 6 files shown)


WorkSutton:
@nomasprime I would still like to see the ECP JIRA ticket number on these so they can be traced in the scenario of an audit.

WorkSutton:
@nomasprime please can we be consistent with our commit messages using what we agreed with the style in front of the message. Thanks

CHANGELOG.md
```diff
@@ -27,7 +27,8 @@ The format is based on [Keep a Changelog]
   - Add 'Set up a Reminder - Enter Personal Details' screen to ECP eligible
     later journey
   - Add 'Mobile number' screen
-  - Update DQT client qualified teacher status to handle null data response
+  - Update DQT client qualified teacher status to handle 404 response
```
WorkSutton, line 30: This should have been in the previous PR I believe @nomasprime

app/models/automated_checks/claim_verifiers/identity.rb
```diff
@@ -14,9 +14,10 @@ def initialize(
       def perform
         return unless awaiting_task?("identity_confirmation")
 
-        complete_match ||
-          partial_match ||
-          no_match
+        # Order matters and seems to create unnecessary conditionals
```
WorkSutton, line 17: @nomasprime could you maybe elaborate on this somewhere? So when we come back to look at code in future we have some background? It feels like this might be some kind of tech debt?

app/models/automated_checks/claim_verifiers/qualifications.rb
```diff
@@ -14,9 +14,10 @@ def initialize(
       def perform
         return unless awaiting_task?("qualifications")
 
-        complete_match ||
-          partial_match ||
-          no_match
+        # Order matters and seems to create unnecessary conditionals
```
WorkSutton, line 17: Same comment as in previous file

WorkSutton (changes requested):

app/models/automated_checks/claim_verifiers/identity.rb
nomasprime: Not sure how to elaborate because I don't know without actually doing the refactor? It's just the ORs and subsequent conditionals aren't very manageable.

Suggested when this work was started it followed a similar design to Rails validations, not sure why it didn't take that direction but imagine it would end up something like that.

WorkSutton (approved):

PR #1482, opened 2021-07-07
[story/ECP-886] - Add 'OTP verification for mobile number' screen

WorkSutton:

### DIFF
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -122,6 +122,9 @@ class ClaimsController < BasePublicController
       one_time_password = generate_otp
       ClaimMailer.email_verification(current_claim, one_time_password).deliver_now
       session[:sent_one_time_password_at] = Time.now
+    elsif params[:slug] == "mobile-number"
+      one_time_password = generate_otp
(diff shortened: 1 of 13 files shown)


app/controllers/claims_controller.rb
```diff
@@ -122,6 +122,9 @@ def generate_one_time_password
       one_time_password = generate_otp
       ClaimMailer.email_verification(current_claim, one_time_password).deliver_now
       session[:sent_one_time_password_at] = Time.now
+    elsif params[:slug] == "mobile-number"
+      one_time_password = generate_otp
+      Rails.logger.debug "\n\n  ** =================== **\nSMS one_time_password: \n#{one_time_password}\n  ** =================== **\n"
```
nomasprime, line 127: Debug?

WorkSutton: Yes, so that we see the OTP when the SMS is sent. Not doing yet. Was doing some reading on how to use Notify last night. There is a response object, but not sure how we will integrate that into the feature spec

nomasprime (approved):

PR #1494, opened 2021-07-15
[story/ECP-632] Change admin manage services accepting claims for academic year select to display four years instead of two

nomasprime:
Also, removed redundant JS spec: nothing actually changes in the spec depending on whether JS in enabled or not.


nomasprime:
Maintenance and no longer tied to issue, hence no `[type/ID]` in commit message.

WorkSutton (approved):

PR #1498, opened 2021-07-19
[story/ECP-611] - Add backlinks

d-a-v-e:
_https://dfedigital.atlassian.net/browse/ECP-611_
________

Adds backlinks to all screen on all pages, barring: claims, complete, existing-session. 

TODO

- [x] specs

### DIFF
--- a/app/assets/javascripts/components/backlink.js
+++ b/app/assets/javascripts/components/backlink.js
@@ -6,7 +6,7 @@
 
   // unhide backlink and add listener
   backlink.classList.remove("govuk-visually-hidden");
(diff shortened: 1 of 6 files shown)


d-a-v-e:
@WorkSutton I've left out a test for each non-backlink page. "/claim" is tested and shown to correclty not display the back link. Unless you can suggest a workaround, I'd have to step through to each of the other non-backlink pages or disable the redirect perhaps...which feels like overkill to me (plus adds a good few seconds to the testing suite!).

WorkSutton (approved):

PR #1510, opened 2021-07-27
[story/ECP-641] Show admin claim identity confirmation task questions when claim verifier identity not fully matched

nomasprime:


app/views/admin/tasks/_form.html.erb
```diff
@@ -1,12 +1,29 @@
...
       <legend class="govuk-fieldset__legend govuk-fieldset__legend--l ">
         <h3 class="govuk-heading-l">
-          <%= I18n.t("#{claim.policy.to_s.underscore}.admin.task_questions.#{task_name}") %>
+          <%= I18n.t("#{translation_base}.title", claim_full_name: @claim.full_name) %>
         </h3>
       </legend>
 
+      <%
+          I18n.t(
+            "#{claim.policy.to_s.underscore}.admin.task_questions.#{task_name}.body",
```
WorkSutton, line 13: @nomasprime Could this not use the translation_base variable that is on line 1?

eg
```
I18n.t("#{translation_base}.body"
```

app/views/admin/tasks/_form.html.erb
```diff
@@ -1,12 +1,29 @@
...
+            default: nil,
+            link: link_to(
+              I18n.t("#{translation_base}.notes"),
+              admin_claim_notes_path(@claim)
+            )
+          )&.html_safe.tap do |translation| %>
+        <% break unless translation %>
```
WorkSutton, line 20: @nomasprime Is this to cater for the fact there are no translations for `early_career_payments.task_questions`?

Is there are reason there aren't any entries for this key?

app/views/admin/tasks/identity_confirmation.html.erb
```diff
@@ -17,37 +17,7 @@
...
-              <div class="govuk-radios__item">
-                <%= f.radio_button(:passed, false, class: "govuk-radios__input") %>
-                <%= f.label "passed_false", "No", class: "govuk-label govuk-radios__label" %>
-              </div>
-            </div>
-          </fieldset>
-          <%= f.submit "Save and continue", class: "govuk-button", data: {module: "govuk-button"} %>
-        </div>
-      <% end %>
+      <%= render "form", task_name: "identity_confirmation", claim: @claim %>
```
WorkSutton, line 20: @nomasprime Any reason for not being consistent with the use of `render :partial: <partial_name>, locals: { <key>: <value> }` style used elsewhere in the application?

WorkSutton (changes requested):
A couple of small changes. Also I believe for future developers who come to work with the codebase, that it would be good in the PR and in the en.yml to explain why there are no entries for the ECP Policy task questions

app/views/admin/tasks/_form.html.erb
nomasprime: There's no presenter for ECP.

app/views/admin/tasks/identity_confirmation.html.erb
nomasprime: This code was copied from another view so it's consistent with that 😄

app/views/admin/tasks/_form.html.erb
nomasprime: 😆 Pretty much the reason I did it, must have been distracted 👍

app/views/admin/tasks/_form.html.erb
WorkSutton: @nomasprime out of curiosity purely, why is there no presenter for ECP?

WorkSutton (approved):

PR #1516, opened 2021-08-02
Bug/ecp 929 otp not working on e069 journey

d-a-v-e:
https://dfedigital.atlassian.net/browse/ECP-929

### DIFF
--- a/lib/one_time_password.rb
+++ b/lib/one_time_password.rb
@@ -10,11 +10,11 @@ module OneTimePassword
   private
 
   def generate_otp
-    ROTP::TOTP.new(OTP_SECRET, issuer: "Claim Additional Payments for Teaching", interval: OTP_PASSWORD_INTERVAL).now
+    ROTP::TOTP.new(OTP_SECRET, issuer: "Claim Additional Payments for Teaching").now
   end
 
   def verify_otp(challenge_code)
-    totp = ROTP::TOTP.new(OTP_SECRET, issuer: "Claim Additional Payments for Teaching", interval: OTP_PASSWORD_INTERVAL)
-    totp.verify(challenge_code)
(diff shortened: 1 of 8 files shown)


lib/one_time_password.rb
```diff
@@ -10,11 +10,11 @@ module OneTimePassword
...
   def generate_otp
-    ROTP::TOTP.new(OTP_SECRET, issuer: "Claim Additional Payments for Teaching", interval: OTP_PASSWORD_INTERVAL).now
+    ROTP::TOTP.new(OTP_SECRET, issuer: "Claim Additional Payments for Teaching").now
   end
 
   def verify_otp(challenge_code)
-    totp = ROTP::TOTP.new(OTP_SECRET, issuer: "Claim Additional Payments for Teaching", interval: OTP_PASSWORD_INTERVAL)
-    totp.verify(challenge_code)
+    totp = ROTP::TOTP.new(OTP_SECRET, issuer: "Claim Additional Payments for Teaching")
+    totp.verify(challenge_code, drift_behind: OTP_PASSWORD_INTERVAL)
```
WorkSutton, line 18: @d-a-v-e So if I understand correctly this will be our 900 seconds (15 minutes) + whatever seconds remaining on the default of 30 seconds? So min would be 901 seconds (15 mins 1 second) and maximum would be 929 seconds (15 mins 29 seconds)?

WorkSutton (commented):
Just a small question around naming and maximum duration of validity.

lib/one_time_password.rb
d-a-v-e: sounds about right!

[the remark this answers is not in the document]
d-a-v-e: I'll rename it 👍

WorkSutton (approved):

PR #1521, opened 2021-08-04
Bug/ecp1019 nokogiri upgrade

and-lucas-kelly:


and-lucas-kelly:
separate way to do the upgrade explored

PR #1523, opened 2021-08-05
[story/ECP-1042] - Create admin claim note when identity confirmation mailer is sent

nomasprime:


WorkSutton (approved):
Nice work @martikat and @nomasprime :+1:

PR #1524, opened 2021-08-06
[story/ECP-487] - User Journey Page 21 - 'Eligible Later: One Time Password'

d-a-v-e:
https://dfedigital.atlassian.net/browse/ECP-487


app/controllers/reminders_controller.rb
```diff
@@ -1,26 +1,87 @@
...
+  def current_slug
+    Reminder::SLUGS[current_slug_index]
+  end
+
+  def current_slug_index
+    Reminder::SLUGS.index(params[:slug]) || 0
+  end
+
+  def current_policy_routing_name
+    "early-career-payments"
```
WorkSutton, line 59: @d-a-v-e Just a question, if we want to extend reminders at a point in the future I assume we could alter this to be policy determination similar to how its done for each Policy?

WorkSutton (approved):
Really nice work on this @d-a-v-e

PR #1540, opened 2021-08-12
[story/ECP-1135] - DAC Report Issue 3, Error links missing for Bank or Building Society account page

WorkSutton:


app/models/claim.rb
```diff
@@ -394,12 +395,12 @@ def building_society_roll_number_must_be_in_a_valid_format
   end
 
   def bank_account_number_must_be_eight_digits
-    errors.add(:bank_account_number, "Bank account number must be 8 digits  – check you've entered the correct number or check with your bank for an 8 digit version") \
+    errors.add(:bank_account_number, "Account number must be 8 digits") \
       if bank_account_number.present? && normalised_bank_detail(bank_account_number) !~ /\A\d{8}\z/
   end
```
d-a-v-e, line 400: my pref is to use a guard clause instead, but if this is used elsewhere leave as is

```ruby
def bank_account_number_must_be_eight_digits
    return unless bank_account_number.present? && normalised_bank_detail(bank_account_number) !~ /\A\d{8}\z/

    errors.add(:bank_account_number, "Account number must be 8 digits")
end
```

app/models/claim.rb
```diff
@@ -394,12 +395,12 @@ def building_society_roll_number_must_be_in_a_valid_format
...
 
   def bank_sort_code_must_be_six_digits
-    errors.add(:bank_sort_code, "Sort code must contain six digits") \
+    errors.add(:bank_sort_code, "Sort code must be 6 digits") \
       if bank_sort_code.present? && normalised_bank_detail(bank_sort_code) !~ /\A\d{6}\z/
   end
```
d-a-v-e, line 405: same comment as above.

d-a-v-e (approved):

PR #1542, opened 2021-08-12
[story/ECP-642] Add ECP admin claim qualifications task screen

nomasprime:


app/models/academic_year.rb
```diff
@@ -47,19 +53,28 @@ def for(date)
...
+  def initialize(start_year = nil)
+    self.years = start_year
+  end
+
+  def eql?(other)
+    to_s == other.to_s
   end
 
   def to_s(format = :default)
+    return "None" if [start_year, end_year].include? nil
```
nomasprime, line 65: - [ ] Specs

app/models/academic_year.rb
```diff
@@ -47,19 +53,28 @@ def for(date)
...
+
     if format == :long
       "#{start_year} to #{end_year}"
     else
       "#{start_year}/#{end_year}"
     end
   end
 
+  def ==(other)
```
nomasprime, line 74: - [ ] Specs

db/schema.rb
```diff
@@ -129,7 +129,9 @@
     t.boolean "subject_to_formal_performance_action"
     t.integer "eligible_itt_subject"
     t.boolean "teaching_subject_now"
-    t.integer "itt_academic_year"
+    t.string "itt_academic_year", limit: 9
+    t.boolean "postgraduate_masters_loan"
```
nomasprime, line 133: Weird, where did loans comes from?

spec/factories/claims.rb
```diff
@@ -11,7 +11,8 @@
 
     after(:build) do |claim, evaluator|
       claim.eligibility = build(*evaluator.eligibility_factory) unless claim.eligibility
-      claim.academic_year = "2019/2020" unless claim.academic_year_before_type_cast
+      claim_academic_year = claim.policy == EarlyCareerPayments ? AcademicYear.new([*Date.today.year..Date.today.year + 4].sample) : "2019/2020"
```
nomasprime, line 14: Actually thought this should be converted to string for factories. Need to test this cause I've used strings elsewhere but the actual object would be better.

spec/factories/early_career_payments/eligibilities.rb
```diff
@@ -9,7 +9,9 @@
       qualification { :postgraduate_itt }
       eligible_itt_subject { :mathematics }
       teaching_subject_now { true }
-      itt_academic_year { "2018_2019" }
+      itt_academic_year { "2018/2019" }
```
nomasprime, line 12: Same as previous.

spec/features/eligible_later_early_career_payments_claim_spec.rb
```diff
@@ -18,15 +18,17 @@
...
+        {itt_subject: "mathematics", itt_academic_year: AcademicYear.new(2019), award_amount: number_to_currency(7_500, precision: 0)},
+        {itt_subject: "mathematics", itt_academic_year: AcademicYear.new(2020), award_amount: number_to_currency(3_000, precision: 0)},
+        {itt_subject: "physics", itt_academic_year: AcademicYear.new(2020), award_amount: number_to_currency(3_000, precision: 0)},
+        {itt_subject: "chemistry", itt_academic_year: AcademicYear.new(2020), award_amount: number_to_currency(3_000, precision: 0)},
+        {itt_subject: "foreign_languages", itt_academic_year: AcademicYear.new(2020), award_amount: number_to_currency(3_000, precision: 0)}
       ].each do |scenario|
         scenario "with ITT subject #{scenario[:itt_subject]} in ITT academic year #{scenario[:itt_academic_year]} it displays award amount of #{scenario[:award_amount]}" do
-          claim.eligibility.public_send("itt_subject_#{scenario[:itt_subject].gsub(/\s/, "_").downcase}!")
-          claim.eligibility.public_send("itt_academic_year_#{scenario[:itt_academic_year].gsub(/\s-\s/, "_")}!")
+          claim.eligibility.update(
```
nomasprime, line 28: Move this back to enum methods or keep like this? Think I find this clearer.

spec/models/early_career_payments/admin_tasks_presenter_spec.rb
```diff
@@ -5,14 +5,16 @@
...
+    build(
+      :claim,
+      academic_year: AcademicYear.new(2021),
+      eligibility: build(
+        :early_career_payments_eligibility,
         current_school: school,
-        initial_teacher_training_subject: :maths,
-        initial_teacher_training_subject_specialism: nil,
-        qts_award_year: "on_or_after_cut_off_date"))
+        eligible_itt_subject: :mathematics,
```
nomasprime, line 14: Not sure about these types.

spec/models/early_career_payments/eligibility_answers_presenter_spec.rb
```diff
@@ -11,7 +11,7 @@
       qualification: :postgraduate_itt,
       eligible_itt_subject: :chemistry,
       teaching_subject_now: true,
-      itt_academic_year: "2019_2020"
+      itt_academic_year: "2019/2020"
```
nomasprime, line 14: - [ ] Check type

spec/models/early_career_payments/eligibility_spec.rb
```diff
@@ -78,17 +78,17 @@
       end
 
       [
-        {itt_subject: "Mathematics", itt_academic_year: "2019 - 2020"},
-        {itt_subject: "Mathematics", itt_academic_year: "2020 - 2021"},
-        {itt_subject: "Physics", itt_academic_year: "2020 - 2021"},
-        {itt_subject: "Chemistry", itt_academic_year: "2020 - 2021"},
-        {itt_subject: "Foreign languages", itt_academic_year: "2020 - 2021"}
+        {itt_subject: "mathematics", itt_academic_year: "2019/2020"},
```
nomasprime, line 81: Maybe I can use `AcademicYear` throughout this file too?

spec/models/early_career_payments/slug_sequence_spec.rb
```diff
@@ -74,17 +74,17 @@
       end
 
       [
-        {itt_subject: "Mathematics", itt_academic_year: "2019 - 2020"},
-        {itt_subject: "Mathematics", itt_academic_year: "2020 - 2021"},
-        {itt_subject: "Physics", itt_academic_year: "2020 - 2021"},
-        {itt_subject: "Chemistry", itt_academic_year: "2020 - 2021"},
-        {itt_subject: "Foreign languages", itt_academic_year: "2020 - 2021"}
+        {itt_subject: "mathematics", itt_academic_year: "2019/2020"},
```
nomasprime, line 77: Maybe `AcademicYear` again?

db/schema.rb
WorkSutton: @nomasprime This came from the ticket to Add Postgraduate Loans to both M+P and TSLR journeys. Originally the attributes were on the ECP eligibility model, having followed how `loan_amount` works where its on an eligibility model. It always felt wrong, and needed moving. ECP-1130 as an enabled for ECP-446.
Does it not show up in `git blame`?

spec/factories/early_career_payments/eligibilities.rb
```diff
@@ -9,7 +9,9 @@
...
+      postgraduate_masters_loan { true }
+      postgraduate_doctoral_loan { true }
```
WorkSutton, line 14: @nomasprime these should no longer exist on this factory, they are now in the claims factory as parts of the following traits:

- `:with_student_loans`
- `:with_no_student_loan`

Rebasing should fix it, but might be painful as a lot has changed in the last week

spec/features/eligible_later_early_career_payments_claim_spec.rb
WorkSutton: @nomasprime Do you mean line 28 vs previous lines 28 & 29 with the `public_send` methods?
I think I prefer the new change if that helps

spec/models/early_career_payments/admin_tasks_presenter_spec.rb
WorkSutton: @nomasprime Are you asking for an opinion or help with this? 
I am unsure if the comment is here for you as a reminder to review and then decide, or if you are asking for help.

I can see that lines 13-15 that are deleted were not appropriate or associated to an ECP, as they are attributes for M+P.

spec/models/early_career_payments/eligibility_spec.rb
WorkSutton: @nomasprime If it is working elsewhere then I think it would be good to be consistent.

If you think it needs some explanation please add appropriate comments. The code base already has good comments where something needs explaining. So I think it would be good to have comments.

Possibly need and update to an ADR?

spec/models/early_career_payments/slug_sequence_spec.rb
WorkSutton: @nomasprime As per previous comment, Yes I think it would be good to be consistent

WorkSutton (changes requested):
@nomasprime Looks like it is taking shape. Are the tick boxes marked `Specs` for specs still waiting to be written?

Looks like needs a rebase due to errors around Postgraduate Loans and where they are sitting at present in this PR

app/models/early_career_payments/eligibility.rb
```diff
@@ -38,11 +38,11 @@ class Eligibility < ApplicationRecord
     }, _prefix: :itt_subject
 
     enum itt_academic_year: {
-      "2018_2019": 0,
-      "2019_2020": 1,
-      "2020_2021": 2,
-      none_of_the_above: 3
-    }, _prefix: :itt_academic_year
+      AcademicYear.new(2018) => AcademicYear::Type.new.serialize(AcademicYear.new(2018)),
```
WorkSutton, line 41: @nomasprime Does this create an integer to work with AR enums?

app/models/early_career_payments/eligibility.rb
```diff
@@ -174,6 +174,138 @@ def award_amounts
...
+    def first_eligible_itt_academic_year
+      award_amounts = [
+        AwardAmount.new(
+          :mathematics,
+          AcademicYear.new(2018),
+          AcademicYear.new(2021),
+          5_000,
+          7_500
+        ),
+        AwardAmount.new(
```
WorkSutton, line 194: @nomasprime You may remember that back in the following PR https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/1403 I had introduced a `yml file` to hold all this data. I think I prefer it in that format over this, as it is easier to read (IMHO), and can easily be updated and then loaded through something like the following:

```ruby
MATRIX = YAML.safe_load(File.read(File.expand_path("../eligibility_matrix.yml", __FILE__))).freeze
```

app/models/early_career_payments/eligibility.rb
nomasprime: No, string enum.

app/models/early_career_payments/eligibility.rb
```diff
@@ -174,6 +174,138 @@ def award_amounts
...
 
+    AwardAmount = Struct.new(
+      :itt_subject,
+      :itt_academic_year,
+      :claim_academic_year,
+      :base_amount,
+      :uplift_amount
+    )
+
+    def first_eligible_itt_academic_year
```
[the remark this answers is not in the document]
nomasprime: Discussed on call, should be part of eligibility.

app/models/early_career_payments/eligibility.rb
nomasprime: Think we can achieve similar legibility without overhead of loading from YAML.

db/schema.rb
nomasprime: Removed.

spec/factories/early_career_payments/eligibilities.rb
nomasprime: Removed.

spec/models/early_career_payments/admin_tasks_presenter_spec.rb
nomasprime: Thought I needed serialized types at DB level and cast otherwise but not sure this is the case, mabye Rails is doing some magic.

spec/factories/claims.rb
nomasprime: Did a quit test and these should be serialized values at DB level, ie, strings.

spec/models/early_career_payments/admin_tasks_presenter_spec.rb
nomasprime: No magic, should be serialized value. If it works otherwise, it's coincidental.

spec/models/early_career_payments/eligibility_spec.rb
nomasprime: Turns out these need to be serialized DB values too.

app/models/early_career_payments/eligibility.rb
WorkSutton: @nomasprime Ah, yes, a while since I have used Enums, completely forgot about being able to also use strings. Makes sense

app/models/academic_year.rb
```diff
@@ -80,4 +95,15 @@ def -(other)
...
     AcademicYear.new(start_year + other)
   end
+
+  private
+
+  def years=(start_year)
+    if start_year.nil?
+      @start_year = @end_year = nil
+    else
+      @start_year = start_year.nil? ? nil : start_year.to_s.split("/").first.to_i
```
WorkSutton, line 105: @nomasprime My sense is that start_year.nil? has been covered in the if branch. Have I missed understood something in your proposed change?

spec/factories/claims.rb
```diff
@@ -11,7 +11,15 @@
 
     after(:build) do |claim, evaluator|
       claim.eligibility = build(*evaluator.eligibility_factory) unless claim.eligibility
-      claim.academic_year = "2019/2020" unless claim.academic_year_before_type_cast
+
+      claim_academic_year =
+        if claim.policy == EarlyCareerPayments
+          AcademicYear::Type.new.serialize(AcademicYear.new([*Date.today.year..Date.today.year + 4].sample))
```
WorkSutton, line 17: @nomasprime why is 2025 included in this array to sample from? I thought the policy for ECP ran for the Academic Years between 2021 thru 2024

app/models/academic_year.rb
nomasprime: Good spot. Think this got used in different context and hasn't been refactored how it should have been.

spec/factories/claims.rb
nomasprime: Offset error and realised I already wrote that part of the line in a nicer way elsewhere, refactoring.

app/helpers/admin/policy_configurations_helper.rb
```diff
@@ -1,8 +1,7 @@
 module Admin
   module PolicyConfigurationsHelper
     def options_for_academic_year
-      years = 4
-      (0..(years - 1)).map { |relative_year| AcademicYear.current + relative_year }
+      (0..3).map { |relative_year| AcademicYear.current + relative_year }
```
WorkSutton, line 4: @nomasprime So when the current `AcademicYear` is `2022`, from this method we will still get 4 new `AcademicYears`?

```
(0..3).map { |relative_year| AcademicYear.current + relative_year }
=> [#<AcademicYear:0x000055c3ddba6430 @start_year=2022, @end_year=2023>, #<AcademicYear:0x000055c3ddba6070 @start_year=2023, @end_year=2024>, #<AcademicYear:0x000055c3ddba5d28 @start_year=2024, @end_year=2025>, #<AcademicYear:0x000055c3ddba59e0 @start_year=2025, @end_year=2026>]
```

I am not sure this is what we are trying to achieve. For ECP it runs until the final `AcademicYear` of 2024/2025.

So the number in the array returned should be decreasing by 1 each year, should it not?

app/models/early_career_payments/admin_tasks_presenter.rb
```diff
@@ -17,6 +17,20 @@ def identity_confirmation
...
+    def qualifications
+      [].tap do |a|
+        a << [
+          "ITT start/end year",
+          I18n.t("answers.qts_award_years.on_date", year: eligibility.first_eligible_itt_academic_year.to_s(:long))
+        ]
+
+        a << [
+          "ITT subject",
+          eligibility.eligible_itt_subject.humanize
```
WorkSutton, line 29: @nomasprime have you applied `humanize` to all instances of `eligibility.eligible_itt_subject`? Or should it only be for specific instances? Looking at some of the specs they now seem inconsistent. 

e.g. https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/0bd4306f056e9dc773960f5b425e91bbac34e564/spec/features/early_career_payments_claim_spec.rb

WorkSutton (changes requested):

app/models/early_career_payments/admin_tasks_presenter.rb
nomasprime: Not sure what you mean. Could you elaborate, maybe give line numbers?

app/helpers/admin/policy_configurations_helper.rb
nomasprime: Technically this should be tied to those years but business was fine with the next four years. Should raise a ticket though.

app/models/early_career_payments/admin_tasks_presenter.rb
WorkSutton: As discussed in zoom, what you have said in response makes sense. Happy to close

WorkSutton (approved):
A tough piece of work. Nicely handled @nomasprime

PR #1544, opened 2021-08-16
[story/ECP-1139] - Hide any Mobile/SMS related OTP screens

martikat:


app/views/claims/mobile_number.html.erb
```diff
@@ -12,10 +12,10 @@
           <%= form.label :mobile_number, t("questions.mobile_number"), class: "govuk-label govuk-label--#{shared_view_css_size}" %>
         </h1>
 
-        <div id="mobile-number-hint" class="govuk-hint">
+        <%# <div id="mobile-number-hint" class="govuk-hint">
           We will send you a text message with a 6-digit password, this will be
           used on the next screen.
-        </div>
+        </div> %>
```
WorkSutton, line 18: @martikat Did this work to hide the relevant content? I would have expected that you would have used the HTML ignore syntax of `<!-- <div class="govuk-form-group"> -->`

If it works great, and I have learnt something about .html.erb files and another way of ignoring content.

timallen-and: It's an ERB comment, which means that the block will get removed server side rather than be delivered as a commented out block to the user. An elegant solution!

timallen-and (approved):

app/views/claims/mobile_number.html.erb
WorkSutton: @timallen-and @martikat Thats really interesting, sublime if I comment uses the style I said, and then what one would have to do is to comment out the specific ERB pieces. So have learnt something genuinely useful with this. :+1:

d-a-v-e (changes requested):
@martikat I would suggest using `xit` and equivalent instead of commenting out. just keeps them in the logs.

d-a-v-e (approved):

PR #1545, opened 2021-08-16
[story/ECP-582] - send reminder email on ECP year open

d-a-v-e:


app/jobs/send_reminder_emails_job.rb
```diff
@@ -0,0 +1,13 @@
+class SendReminderEmailsJob < ApplicationJob
+  def perform(year)
+    reminders.each do |reminder|
```
WorkSutton, line 3: @d-a-v-e Do we want to batch these up into groups of 100/500/1000? Thinking 500 might be the sweet spot, if there are several thousand, but don't fully have a handle on numbers we are expecting. So 100 might be a safer batching number

spec/factories/reminders.rb
```diff
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :reminder do
-    full_name { "Miss Sophia Xu" }
-    email_address { "s.xu@hotmail.com" }
+    full_name { Faker::Name.name }
```
WorkSutton, line 3: @d-a-v-e I think if we are going to introduce Faker for data then we should be moving to using it all over the RSpec suite. Personally I feel we should probably not introduce that at this point. I think the original team made a conscious decision not to use this gem in order to keep the number of gem dependencies to a minimum. Was there a particular reason for deciding to add the gem, that mocked data that existed before did not fill?

app/jobs/send_reminder_emails_job.rb
d-a-v-e: good shout, seemed like theres only going to be a few but no reason to not batch them.

spec/factories/reminders.rb
d-a-v-e: I created a few sets/list of reminder emails and I didn't want duplicate records in the DB so faker seemed like a good choice. I'm happy to leave it for now.

WorkSutton: @d-a-v-e ok makes sense

WorkSutton (approved):
Nice work @d-a-v-e

PR #1558, opened 2021-08-18
[story/ECP-920] - add reminder flow to eligible now journey

d-a-v-e:
https://dfedigital.atlassian.net/browse/ECP-920

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -15,6 +15,6 @@ class SubmissionsController < BasePublicController
 
   def show
     render :show
-    clear_claim_session
+    clear_claim_session unless current_claim.has_ecp_policy?
   end
 end
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -13,6 +13,8 @@ class SubmissionsController < BasePublicController
     end
   end
 
+  # Clear session unless this was an ECP policy. If ECP policy, user has
+  # oppotunity to set a reminder up where name and email is re-used from claim.
   def show
     render :show
     clear_claim_session unless current_claim.has_ecp_policy?
--- a/spec/support/capybara.rb
+++ b/spec/support/capybara.rb
@@ -5,19 +5,23 @@ rescue LoadError
   nil
 end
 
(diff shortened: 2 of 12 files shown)


spec/support/capybara.rb
```diff
@@ -5,19 +5,23 @@
   nil
 end
 
-Capybara.register_driver :headless_chrome do |app|
+view_mode = ENV.fetch("VIEW", false) == "true" ? :chrome : :headless_chrome
+args = %w[disable-dev-shm-usage no-sandbox window-size=1280x1280]
+args = args << "headless" unless view_mode == :chrome
```
d-a-v-e, line 10: I've added this so it's easier to view feature specs live in a browser, just prefix the ENV variable VIEW when running a spec, e.g.: 

```terminal
VIEW=true be rspec spec/features/ecp_eligible_now_reminder_set_spec.rb
```
_be = bundle exec_

d-a-v-e: to disable, just set `VIEW=false` (or anything but `true`)

app/controllers/submissions_controller.rb
```diff
@@ -13,8 +13,10 @@ def create
     end
   end
 
+  # Clear session unless this was an ECP policy. If ECP policy, user has
+  # oppotunity to set a reminder up where name and email is re-used from claim.
```
WorkSutton, line 17: @d-a-v-e Should we be clearing down the `current_claim_session` for ECP at some point later in the journey, so that it can be used, but is actually cleared eventually?

WorkSutton (commented):
Nice work @d-a-v-e. Just the one question, otherwise looks good from my perspective

app/controllers/submissions_controller.rb
d-a-v-e: not sure i understand the Q, but the session is cleared later on in the reminders flow: https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/1558/files#diff-b8e469ddb63fc4efe9e712d5b2f68490f9e3b4bc9d055bc312dd42e90ddb8402R21

WorkSutton (approved):

PR #1570, opened 2021-08-20
[story/ECP-1140] Add important notes to claim (DQT active alert)

nomasprime:


WorkSutton:
@nomasprime I noticed when you demo'd this the other day that we had a red panel.

I think we should be asking an opinion from our UCD team - Andy Davies now.

app/assets/stylesheets/components/banner.scss
```diff
@@ -1,5 +1,5 @@
-.banner-temporary-message-without-action {
-  background: govuk-colour("blue");
+.banner {
+  background: govuk-colour("red");
```
WorkSutton, line 2: Please see main comment, I believe we should not be making design decisions ourselves given we have a UCD (Ux/Content/Design) team member

WorkSutton (approved):
Approving after the meeting this morning the visuals.

PR #1576, opened 2021-08-23
[bug/ECP-1158] - Fix admin qualifications task for ECP

martikat:


nomasprime:
Why remove `:long` when it's supposed to be a valid arg? The real issue appears to be the `first_eligible_itt_academic_year` is returning `nil` which has a method `.to_s` that doesn't accept any args.

Suspect the real issue's that with an existing deployment the old enum values for the `itt_academic_year` field haven't been migrated.

PR #1589, opened 2021-08-27
[bug/ECP-1158] - One off Data Migration to fix qualifications link throwing error

WorkSutton:
- PolicyConfiguration for ECP - update current_academic_year from 2020/2021 to 2021/2022
 - Find all ECP Claims with attribute 'academic_year' set to 2020/2021 - update to 2021/2022


nomasprime (dismissed):
Can be drastically simplified and, appreciate the sentiment but, all the output seems unnecessary.

Also, don't see anything here updating enums we talked about which means they'll be ineligible, ie, cohort needs to be eligible too.

timallen-and (approved):

PR #1596, opened 2021-08-28
[story/ECP-1063] Update claim verifiers to handle multiple partial matches

nomasprime:


db/schema.rb
```diff
@@ -131,6 +131,8 @@
     t.integer "eligible_itt_subject"
     t.boolean "teaching_subject_now"
     t.string "itt_academic_year", limit: 9
+    t.boolean "postgraduate_masters_loan"
+    t.boolean "postgraduate_doctoral_loan"
```
WorkSutton, line 135: @nomasprime These two attributes have been moved to the claim model. Could you please fix

WorkSutton (changes requested):
Please fix the schema migration issue called out. Have not reviewed rest yet

db/schema.rb
nomasprime: Missed this. Don't know why it keeps coming back.

I'll remove again.

WorkSutton (approved):

PR #1602, opened 2021-08-31
[story/ECP-1072] - removes DQT identity check mailer

d-a-v-e:
https://dfedigital.atlassian.net/browse/ECP-1072


WorkSutton (approved):
Looks good @d-a-v-e, thanks for jumping on this request from business so quickly :+1:

PR #1607, opened 2021-08-31
[bug/ECP-1180] Stop claim full name adding redundant spaces when middle name is an empty or blank string

nomasprime:


WorkSutton (approved):
Nice :+1:

PR #1610, opened 2021-09-01
[story/ECP-1182] Fix ECP admin claim qualifications task question to first claim year eligibility (Mathematics 2018)

nomasprime:


WorkSutton (changes requested):
@nomasprime I know you don't like FIXME's or TODO's but I feel strongly we need some in code documentation on this one. No ideas if we will be fixing before we leave the project. So I believe it is essential to call this out here. Could you add something plesae?

nomasprime:
We will definitely be doing this and even if we weren't, would you propose we move all incomplete issues to the code when we hand over?

Code's pretty self-explanatory and this pretty much next thing I'm working on, comment would duplicate issue, so no need for extraneous comment.

timallen-and (commented):
Please document it as if you were to be hit by a bus while nipping out for lunch and another Dev has to pick up the work! I mean, I hope you don't get hit by bus but you never know.

WorkSutton (approved):
Being pragmatic and approving mainly due to time constraints

PR #1613, opened 2021-09-01
[story/ECP-1183] Make admin claim verifier qualifications matching comparable to previous CSV matching

nomasprime:


app/models/automated_checks/claim_verifiers/qualifications.rb
```diff
@@ -35,10 +32,23 @@ def complete_match
...
+          "Not eligible"
+        else
+          <<~HTML
+            #{match == :none ? "Ine" : "E"}ligible:
+            <pre>
+              ITT subject codes: #{dqt_teacher_status.itt_subject_codes}
+              Degree codes:      #{dqt_teacher_status.degree_codes}
+              QTS award date:    #{dqt_teacher_status.qts_award_date}
+            </pre>
+          HTML
```
WorkSutton, line 46: @nomasprime Assume when we do the work you describe in the ticket this will be removed/change?
As said in a previous review this feels presentational and therefore in the wrong place.
Given time constraints and urgency will accept

WorkSutton (approved):

PR #1616, opened 2021-09-01
[story/ECP-1195] Add new DQT fields to admin claim verifier qualifications notes

nomasprime:


nomasprime:
Flaky spec constantly failing. If still failing tomorrow, will have to ignore until it can be fixed.

WorkSutton (approved):
@nomasprime same comments on some other PR's feels like we have presentational details  in the model layer. Assume this will change when you do the upcoming work?
Given time constraints and urgency approving

PR #1624, opened 2021-09-03
[bug/ECP-1201] - Update schools fixture data that is used in seeding

WorkSutton:


spec/fixtures/schools.yml
```diff
@@ -62,3 +65,4 @@ bradford_grammar_school:
   town: Bradford
   postcode: BD9 4JP
   phone_number: "01274542492"
+  establishment_number: 6103
```
martikat, line 68: A new line needs to be added here

martikat (changes requested):

martikat (approved):

PR #1625, opened 2021-09-03
[story/ECP-1070] - create decisions csv

d-a-v-e:
https://dfedigital.atlassian.net/browse/ECP-1070


WorkSutton (approved):
@d-a-v-e Nice work on this. Great that you could refactor the original version and remove those methods we discussed.

Only comment is on the `read-only`. Not sure if we might have to change that down the road for the Rails upgrade. Have a sneaking suspicion that when I looked at the upgrade to 6.1 before that `read-only` was throwing some warnings. Lets cross that bridge when we get to it

PR #1629, opened 2021-09-05
[bug/ECP-1202] - Update link_to with correct path

WorkSutton:
- allows a Trainee Teacher answering No to NQT question to set a reminder

### DIFF
--- a/app/controllers/reminders_controller.rb
+++ b/app/controllers/reminders_controller.rb
@@ -76,6 +76,7 @@ class RemindersController < BasePublicController
 
   def build_reminder_from_claim
     return unless current_claim
+    return if current_claim.eligibility.nqt_in_academic_year_after_itt == false && current_claim.has_ecp_policy?
 
     Reminder.new(
       full_name: current_claim.full_name,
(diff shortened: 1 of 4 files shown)


app/controllers/reminders_controller.rb
```diff
@@ -76,6 +76,7 @@ def reminder_from_session
 
   def build_reminder_from_claim
     return unless current_claim
+    return if current_claim.eligibility.nqt_in_academic_year_after_itt == false && current_claim.has_ecp_policy?
```
d-a-v-e, line 79: so this means they will use the "default" reminder defined below if these conditions are met? will the claim object not have an academic year or subject if those conditions are like that?

WorkSutton: @d-a-v-e It was failing looking for in the `eligibile_later_year` method at this line - https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/master/app/models/early_career_payments/eligibility.rb#L261. 

Given the claimant - a Trainee Teacher - has answered 'No - I am a trainee Teacher` on the very first question, there was never going to be an itt_academic_year, nor a eligible_itt_subject.

So I felt given we had a `default` that this could be used. Hence adding the guard clause based on `nqt_in_academic_year_after_itt == false` and scoping it to only ECP policy.

If you have concerns then lets have a chat. :smiley:

d-a-v-e: ok great, that makes sense!

d-a-v-e (approved):

PR #1633, opened 2021-09-07
removed ARM template code

and-lucas-kelly:


WorkSutton (approved):
@and-lucas-kelly Really nice to finally remove the old ARM template stuff

PR #1638, opened 2021-09-07
[Story/ECP-1090] upload decisions CSV to GCS

d-a-v-e:


.env.development
```diff
@@ -9,3 +9,6 @@ DQT_CLIENT_HOST=https://s118d01funcappfapp.azurewebsites.net/
 
 ADMIN_ALLOWED_IPS=::1,127.0.0.1
 ENVIRONMENT_NAME=local
+
+# google cloud storage
+STORAGE_CREDENTIALS=./google_cloud_storage.json
```
WorkSutton, line 14: @d-a-v-e wonder if this should be optional on DEV similar to how it is for the old GA as per the README.md

d-a-v-e: moved to README

WorkSutton (approved):
:+1:

PR #1650, opened 2021-09-13
[story/ECP-1086] - analytics daily stats

d-a-v-e:
https://dfedigital.atlassian.net/browse/ECP-1086


spec/models/claim_stats/daily.rb
```diff
@@ -0,0 +1,27 @@
...
+      end
+      ClaimStats.refresh
+    end
+
+    let(:expected_csv) do
+      <<~CSV
+        extract_date,policy,average_claim_submission_length,average_claim_decision_length,applications_started_total,applications_submitted_total,applications_rejected_total,applications_accepted_total,applications_started_daily,applications_submitted_daily,applications_rejected_daily,applications_accepted_daily
+        #{Date.yesterday},early career payments,-0.0,0.0,1,1,0,1,0,0,0,0
+        #{Date.yesterday},maths and physics,-0.0,0.0,1,1,0,1,0,0,0,0
+        #{Date.yesterday},student loans,-0.0,0.0,1,1,0,1,0,0,0,0
```
d-a-v-e, line 19: note to self: the final stat `applications_accepted_daily` should not be 0. Investigating...

app/models/claim_stats.rb
```diff
@@ -0,0 +1,11 @@
+class ClaimStats < ApplicationRecord
+  self.primary_key = "claim_id"
+
+  def readonly?
+    true
+  end
+
+  def self.refresh
+    Scenic.database.refresh_materialized_view(table_name, concurrently: false, cascade: false)
```
d-a-v-e, line 9: The materialised view will need refreshing. However, I will do this in the upload portion (covered in the next ticket) and refresh the view as part of the background job.

db/views/claim_stats_v01.sql
```diff
@@ -0,0 +1,24 @@
...
+  CASE c.submitted_at
+    WHEN null THEN null
+    ELSE extract(epoch from c.submitted_at - c.created_at)
+  END AS submission_length,
+  CASE d.created_at
+    WHEN null THEN null
+    ELSE extract(epoch from d.created_at - c.submitted_at)
+  END AS decision_length
+FROM decisions d
+RIGHT JOIN claims c ON c.id = d.claim_id
```
d-a-v-e, line 24: This view isn't used directly, instead it's used in ClaimStats::Daily to build the aggregated daily stats. This is because we need to pass in the day to aggregate on into the SQL.

spec/models/claim_stats/daily.rb
d-a-v-e: ...fixed in the new specs.

spec/models/claim_stats/daily.rb
```diff
@@ -0,0 +1,80 @@
...
+        started_previously.times { create(:claim, policy: policy, created_at: previously) }
+
+        approved_yesterday.times { create(:claim, :approved, policy: policy, created_at: yesterday, submitted_at: yesterday + 60.seconds) }
+        approved_previously.times { create(:claim, :approved, policy: policy, created_at: previously, submitted_at: previously + 60.seconds) }
+
+        rejected_yesterday.times { create(:claim, :rejected, policy: policy, created_at: yesterday, submitted_at: yesterday + 60.seconds) }
+        rejected_previously.times { create(:claim, :rejected, policy: policy, created_at: previously, submitted_at: previously + 60.seconds) }
+      end
+
+      allow_any_instance_of(Decision).to receive(:readonly?).and_return(false)
```
WorkSutton, line 56: @d-a-v-e Can we do this without the old style of `allow_any_instance_of`?
https://relishapp.com/rspec/rspec-mocks/docs/working-with-legacy-code/any-instance

d-a-v-e: In that instance I’m just using this so I can get the data in the correct state. decisions don’t allow us to write to them once written and I'm not using this for testing `Decision`.

WorkSutton: Ah, ok, thanks for the explaination :+1:

WorkSutton (approved):

PR #1655, opened 2021-09-14
[story/ECP-1087] - Upload daily stats

d-a-v-e:
https://dfedigital.atlassian.net/browse/ECP-1087

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog]
 
 ## [Unreleased]
+- Google Analytics
+  - Adds CSV view for claim stats
 
 ## [Release 095] - 2021-09-09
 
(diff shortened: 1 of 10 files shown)


app/jobs/send_decisions_analytics_job.rb
```diff
@@ -18,10 +18,10 @@ def query
   end
 
   def file_name
-    @file_name ||= "decisions-analytics_#{date}.csv"
+    @file_name ||= "decisions-data/decisions-analytics_#{date}.csv"
```
WorkSutton, line 21: @d-a-v-e is this directory (bucket) then filename?

d-a-v-e: its `folder/file_name.ext` (the bucket is static and defined as an ENV separately for each environment..just adding them now actually!)

WorkSutton (approved):

PR #1657, opened 2021-09-14
[story/ECP-1236] Dockerize local production environment

nomasprime:


docker-compose.production.yml
```diff
@@ -0,0 +1,62 @@
+version: "3.7"
+
+x-env_files:
```
WorkSutton, line 3: @nomasprime why `x-env_files`? What does tis represent?

.env.production
```diff
@@ -0,0 +1,18 @@
+DFE_SIGN_IN_ISSUER=https://pp-oidc.signin.education.gov.uk:443
+DFE_SIGN_IN_REDIRECT_BASE_URL=https://localhost:3000
+DFE_SIGN_IN_IDENTIFIER=teacherpayments
+
+DFE_SIGN_IN_API_CLIENT_ID=teacherpayments
+DFE_SIGN_IN_API_ENDPOINT=https://pp-api.signin.education.gov.uk
+
+DQT_CLIENT_HOST=https://s118d01funcappfapp.azurewebsites.net/
+
+ENVIRONMENT_NAME=review
```
WorkSutton, line 10: @nomasprime Why is the environment called `review`?

WorkSutton:
@nomasprime @and-lucas-kelly how do we ensure that this is not run in true PRODUCTION?

.env.production
nomasprime: Seeds.

docker-compose.production.yml
nomasprime: [Extension fields](https://docs.docker.com/compose/compose-file/compose-file-v3/#extension-fields).

nomasprime:
> @nomasprime @and-lucas-kelly how do we ensure that this is not run in true PRODUCTION?
> As I believe it is not meant to be. That rather this is just setup so we can debug locally.

We don't run dotenv in production.

WorkSutton (approved):

PR #1659, opened 2021-09-15
[story/ECP-1233-1234] - Backfill analytics

d-a-v-e:
https://dfedigital.atlassian.net/browse/ECP-1233
https://dfedigital.atlassian.net/browse/ECP-1234


lib/tasks/backfill_daily_stats_analytics.rake
```diff
@@ -0,0 +1,7 @@
+desc 'For backfilling analytics CSVs'
+task backfill_daily_stats: :environment do
+  (Date.new(2021,9,6)..Date.new(2021,9,14)).each do |date|
+    SendAnalytics::DailyStatsJob.perform_later(date: date)
+    SendAnalytics::DecisionsJob.perform_later(date: date)
+  end
+end
```
d-a-v-e, line 7: Have tested with my GCS instance and this correctly creates the CSVs:

Screenshot 2021-09-15 at 16 17 08
Screenshot 2021-09-15 at 16 16 59

WorkSutton: @d-a-v-e Great documentation, really appreciate this style. It resonates with how I like to work and make PR's easily understandable :+1:

app/jobs/send_analytics/daily_stats_job.rb
```diff
@@ -0,0 +1,13 @@
+module SendAnalytics
+  class DailyStatsJob < CronJob
+    self.cron_expression = "0 3 * * *" # every day at 3am
+
+    queue_as :analytics
+
+    def perform(date: Date.yesterday)
+      Rails.logger.info "Sending daily stats CSV for #{date}..."
```
WorkSutton, line 8: @d-a-v-e Are you using `info` so that we can set this level in TEST/DEV (or a level that picks it up)? So it won't be adding to the log in PROD

WorkSutton (approved):
@d-a-v-e Will we remove the rake task `lib/tasks/backfill_daily_stats_analytics.rake` after all the back population has successfully completed? It feels to me like this is a one-off task that doesn't need to say in the repo.

app/jobs/send_analytics/daily_stats_job.rb
d-a-v-e: this won't go to prod logs?

WorkSutton (approved):
Approving after latest changes

PR #1677, opened 2021-09-22
[bug/ECP-1192] - Start page - 'check your eligibility today' link directs to wrong page

martikat:


WorkSutton:
@martikat Can we check this as this is now inconsistent with what we do for M+P and TSLR. I am happy to jump on a call with Laura if its required. But I don't believe this is correct from a consistency standpoint. Not what you have done, rather what the business have asked

WorkSutton (approved):

PR #1679, opened 2021-09-23
[story/ECP-887] - Add GOVUK Notify SMS functionality

WorkSutton:


WorkSutton:
To test locally you will need to add to your .env file the key `NOTIFY_API_KEY` and its value can be found in the KV store on Azure development

spec/lib/notify_sms_message_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+        err = instance_double("Net::HTTPClientError")
+        allow(err).to receive(:code).and_return(400)
+        allow(err).to receive(:body).and_return("boom")
+
+        message.send(:instance_variable_set, :@sms_client, sms_client)
+        allow(sms_client).to receive(:send_sms).and_raise(NotifySmsMessage::NotifySmsError, err)
+      end
+
+      it "catches the error and raises a NotifySMSError" do
+        expect { message.deliver! }.to raise_error(NotifySmsMessage::NotifySmsError)
```
d-a-v-e, line 38: seems like this just tests the code in the test, as you have this just above:
```ruby
allow(sms_client).to receive(:send_sms).and_raise(NotifySmsMessage::NotifySmsError, err)
```

spec/lib/notify_sms_message_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    it "sends a SMS message" do
+      message.deliver!
+      expect(sms_client).to have_received(:send_sms).once
+    end
+
+    context "when the Nofify client raises an error" do
+      before do
+        err = instance_double("Net::HTTPClientError")
+        allow(err).to receive(:code).and_return(400)
+        allow(err).to receive(:body).and_return("boom")
```
d-a-v-e, line 31: both of these `allow` look unused. If they are necessary, I think they can just go in the `instance_double()` definition like so:
```ruby
err = instance_double("Net::HTTPClientError", code: 400, body: "boom")
```

app/models/concerns/one_time_password_checkable.rb
```diff
@@ -1,31 +1,35 @@
...
+    return unless CATEGORIES.include?(one_time_password_category)
+    return write_attribute(:email_verified, true) if otp.valid? && %w[claim_email reminder_email].include?(one_time_password_category)
+    return write_attribute(:mobile_verified, true) if otp.valid? && one_time_password_category == "claim_mobile"
 
     errors.add(:one_time_password, otp.warning)
   end
 
   def otp
-    @otp ||= OneTimePassword::Validator.new(one_time_password, set_sent_one_time_password_at)
+    @otp ||= OneTimePassword::Validator.new(one_time_password, sent_one_time_password_at, nil, one_time_password_category)
```
d-a-v-e, line 33: rather than pass in `nil` to `OneTimePassword::Validator` I'd say get rid of the `ISSUER` setting in `Validator` and `Generator` it doesn't look like it is needed. I copied this over from the original code.I thought it was used in validation but if you can just pass in `nil` and it still works, then I guess not!

d-a-v-e (changes requested):
Some suggestions and some questions. Jump on a call to go through if you like 👍

lib/one_time_password/validator.rb
```diff
@@ -14,12 +15,13 @@ def warning
       return "Enter your one time password" if code.blank?
       return "Your one time password must be #{LENGTH}-digits" if wrong_length?
       return "Your one time password has expired, request a new one" if expired?
-      "Enter the correct one time password that we emailed to you" if incorrect?
+      return "Enter the correct one time password that we sent to you in a text message" if incorrect? && otp_category == "claim_mobile"
+      "Enter the correct one time password that we emailed to you" if incorrect? && (otp_category == "claim_email" || otp_category.nil?)
     end
```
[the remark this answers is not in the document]
WorkSutton: Goto suggestion @d-a-v-e , will do this.

d-a-v-e (approved):

PR #1686, opened 2021-09-29
[story/ECP-624] - DQT switchover

d-a-v-e:
https://dfedigital.atlassian.net/browse/ECP-624


config/application.rb
```diff
@@ -19,6 +19,7 @@
 require_relative "../lib/one_time_password"
 require_relative "../lib/send_analytics"
 require_relative "../lib/upload"
+require_relative "../lib/fwy"
```
d-a-v-e, line 22: (this name is temporary)

app/models/student_loans/dqt_record.rb
```diff
@@ -23,11 +23,11 @@ class DqtRecord
...
     def initialize(record, _ = nil)
-      @qts_award_date = record.fetch(:qts_date)
-      @degree_codes = record.fetch(:degree_codes, ["ANY"])
-      @itt_start_date = record.fetch(:itt_date, nil)
-      @itt_subject_codes = record.fetch(:itt_subject_codes, ["ANY"])
-      @qualification_name = record.fetch(:qualification_name, nil)
+      @qts_award_date = record.qts_date)
+      @degree_codes = record.degree_codes || ["ANY"]
+      @itt_start_date = record.itt_date
+      @itt_subject_codes = record.itt_subject_codes || ["ANY"]
```
d-a-v-e, line 29: @WorkSutton who's best to ask about the behaviour here? at the moment, the new API client would return: `[nil, nil, nil]` if there were no values for the subject codes. I'm wondering if I should compress to get rid of the nils, then in this portion of the code run `.presence` and supply `["ANY"]` if it returns `nil`

PR #1724, opened 2021-10-20
[story/ECP-1284] - Use subject instead of code in DQTRecord for ECP

d-a-v-e:


WorkSutton:
Closing as will be replaced by work in ECP-1288 which targets both `subject_code` and `subject`

PR #1743, opened 2021-11-13
[story/ECP-1288] - Add ITT Subject and ITT Subject Codes to task notes for DqT automation

WorkSutton:


martikat (approved):

spec/features/admin_claim_tasks_update_with_dqt_api_spec.rb
```diff
@@ -1486,6 +1496,7 @@ def task_outcome
               9,
               1
             ),
+            itt_subjects: ["applied mathematics", "maths and statistics", :maths],
```
d-a-v-e, line 1499: I don't think we can/would receive a symbol

d-a-v-e (approved):

spec/features/admin_claim_tasks_update_with_dqt_api_spec.rb
WorkSutton: @d-a-v-e thanks, have updated.

martikat (approved):

PR #1757, opened 2021-11-24
[bug/ECP-1298] - Exclude data prior to 2021/2022 Academic Year in stats extracts

WorkSutton:


martikat (approved):

WorkSutton:
@martikat going to re-work this as it looks like the senic gem supports versioning. Will ask for a re-review when ready

martikat (approved):

PR #1845, opened 2022-02-07
[bug/ECP-1388] - fix load order for claims associated eligibilities to match csv dqt data

WorkSutton:
- Refactor importer and extract common methods
 - Add DqT CSV Data Import validator


WorkSutton:
This fixes the issue with data being associated incorrectly between DqT CSV test data and the eligibilities that are generated for a claim. Previously they were in `:desc order` but this bug fix corrects the ordering.

Additionally CSV import checking is now performed post test data seeding

martikat (approved):

PR #1867, opened 2022-02-22
[story/ECP-1361] - Upload TPS data

martikat:


WorkSutton:
@martikat Cannot remember if we decided to split this into two and make it like some of the other loaders, or to keep it similar to the schools loader. For now given time constraints will approve this, but later if we have time and its necessary we might want to move to the two object/file style employed elsewhere.

Based on that looking at approving

PR #1868, opened 2022-02-22
[bug/ECP-1390] - Decisions stats dashboard page showing incorrect data

martikat:


lib/tasks/backfill_daily_stats_analytics.rake
```diff
@@ -1,6 +1,6 @@
 desc "For backfilling analytics CSVs"
 task backfill_daily_stats: :environment do
-  (Date.new(2021, 9, 6)..Date.new(2021, 11, 24)).each do |date|
+  (Date.new(2021, 9, 6)..Date.new(2022, 2, 22)).each do |date|
```
WorkSutton, line 3: @martikat will want to change this date to be the date it gets merged in

WorkSutton (changes requested):
@martikat I think it would be best to split the backfill task into two more tasks, and to update the existing task as follows:

Copy the existing task twice, naming the new ones as follows:

- `backfill_decision_stats` and removing from it the DailyStatsJob
- `backfill_daily_stats` and removing from it the DecisionsJob
- update the exisiting task name from `backfill_daily_stats` to `backfill_all_stats`

That way when we come to run this as we know the current issue is only the DecisionStats we can simply run the correct version

Also I would setup each task to accept two parameters for start and end dates. And if they are not passed in default them as follows:

- start_date (2021, 9, 6) - the date these service went live in 2021
- end_date - Date.today (which will mean it runs yesterdays data in)

WorkSutton (approved):

WorkSutton (changes requested):
@martikat Woops, I made an error in approving. Have just realised that from the command line you will get errors without formatting the entry in zsh. Best to go down the route of putting a comment in about format of dates. And then instead of using Date.new simply pass the date as a string (don't have to put as string in rake).
Then convert string to date using Date.parse(<date of interest>, "%Y-%m-%d") where `string_of_interest` is say 2022-01-31

Any questions, please ask

WorkSutton (approved):
Nice work @martikat :+1:

PR #1898, opened 2022-03-31
[story/ECP-939] - Automate Employment TPS checks

WorkSutton:
TODO:
 
 Fix other specs - mainly feature specs
 Demo with business against A/C
 Create job to run the claim verifier for the TPS employment checks. This should be when a new upload of data is made which is approximately every 2 weeks.
 Will need to add additional specs for that

### DIFF
--- a/app/models/automated_checks/claim_verifiers/census_subjects_taught.rb
+++ b/app/models/automated_checks/claim_verifiers/census_subjects_taught.rb
@@ -41,7 +41,7 @@ module AutomatedChecks
       end
 
       def no_match
-        return if school_workforce_census.nil? || eligible?
+        return unless school_workforce_census.nil? || !eligible?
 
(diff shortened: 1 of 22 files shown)


app/models/automated_checks/claim_verifiers/census_subjects_taught.rb
```diff
@@ -41,7 +41,7 @@ def no_data
       end
 
       def no_match
-        return if school_workforce_census.nil? || eligible?
+        return unless school_workforce_census.nil? || !eligible?
```
kenfodder, line 44: Just double-checking with you this is as expected, at a glance `school_workforce_census.nil?` isn't negated or the condition isn't `&&`.

vassyz:
I'll try to close this for now and create a new PR as it seems to have some issues with CI.

PR #1914, opened 2022-04-29
Move `current-school` step to beginning of ECP journey

nigel-lowry:
Draft PR to check sanity of approach:

* Right now the PR does not actually move the current-school step to the beginning! It's how it is at the moment to make sure it doesn't break the existing wizards (assuming the specs we've inherited are sufficient). However, I've changed the current-school to be the first step locally and via manual verification it works (including avoiding some weird JavaScript problems I found manually on a previous attempt)
* The problem with the current-school step is because it appeared later in the journey it made the assumption that the Claim record was already persisted. (It seems to be the only step which makes this assumption but there could be others). Other wizard steps actually create the record in the view if it's not already persisted which is a smell
* The approach here is to create the record before the first step is fully loaded. This means we can make the assumption that the record is persisted before we enter the wizard (I'm not worrying about it now but there's the potential to delete a lot of dead code given this precondition)
* A spec for the back links was using a loophole to test the presence of back links. This approach has closed that loophole so I'd need to devise another way of testing the same thing. Back links aren't working on Safari anyway so we need to a ticket to sort them out anyway
* If we're happy with the approach, I'll still have to make the change to the EarlyCareerPayments::SlugSequence and fix the ~150 specs which rely on the current ordering

### DIFF
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -8,6 +8,7 @@ class ClaimsController < BasePublicController
   before_action :check_claim_not_in_progress, only: [:new]
   before_action :clear_claim_session, only: [:new]
   before_action :prepend_view_path_for_policy
+  before_action :persist, only: [:new]
 
   def new
     render first_template_in_sequence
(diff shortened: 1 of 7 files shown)


app/controllers/claims_controller.rb
```diff
@@ -8,6 +8,7 @@ class ClaimsController < BasePublicController
   before_action :check_claim_not_in_progress, only: [:new]
   before_action :clear_claim_session, only: [:new]
   before_action :prepend_view_path_for_policy
+  before_action :persist, only: [:new]
```
slorek, line 11: Suggestion: move the `persist` call inside the `new` action for slightly improved clarity.

app/controllers/claims_controller.rb
```diff
@@ -8,6 +8,7 @@ class ClaimsController < BasePublicController
...
 
   def new
     render first_template_in_sequence
```
slorek, line 14: Will this line now be redundant given that L94 now redirects (I'm assuming to the `show` action)?

slorek (commented):
👏  Looks like the start of a good approach to resolve some issues. One thing to be aware of is the potential for an increase in orphaned records from abandoned forms - we may want to consider how we will clean these up.

PR #1915, opened 2022-04-29
Move `choose-school` step to beginning of ECP journey

nigel-lowry:
JIRA: https://dfedigital.atlassian.net/browse/CAPT-167

Moves `choose-school` step to beginning of ECP wizard in preparation for combined ECP/LUP journey.

Two things to note:
* This exposes a problem with the back links which must be rethought in a future ticket. As an aside, they currently use JavaScript because they're calling the Web browser's back button functionality which is wrong and they don't work on Safari; and
* You'll find you be asked about an existing claim more often during development. This is because claims are created on viewing the first page of the wizard rather than when submitting the answer to the first step. I think this makes sense though.

### DIFF
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -10,18 +10,11 @@ class ClaimsController < BasePublicController
   before_action :prepend_view_path_for_policy
 
   def new
-    render first_template_in_sequence
+    persist
   end
 
   def create
-    current_claim.attributes = claim_params
-
-    if current_claim.save(context: page_sequence.slugs.first.to_sym)
-      session[:claim_id] = current_claim.to_param
-      redirect_to claim_path(current_policy_routing_name, next_slug)
-    else
-      render first_template_in_sequence
-    end
+    persist
   end
 
   def show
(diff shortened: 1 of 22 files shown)


app/controllers/claims_controller.rb
```diff
@@ -10,18 +10,11 @@ class ClaimsController < BasePublicController
   before_action :prepend_view_path_for_policy
 
   def new
-    render first_template_in_sequence
+    persist
```
vassyz, line 13: Isn't it an issue that in the `new` call you are creating a claim? I'm assuming we'll have a lot of empty claims just form people who open this page and close it straight away.

vassyz:
You also have to update the order in `eligibility_answers_presenter.rb` for the `Check your answers` page.

app/controllers/claims_controller.rb
```diff
@@ -83,6 +76,14 @@ def next_slug
     page_sequence.next_slug
   end
 
+  def persist
+    current_claim.attributes = claim_params
+
+    current_claim.save!
+    session[:claim_id] = current_claim.to_param
+    redirect_to claim_path(current_policy_routing_name, page_sequence.slugs.first.to_sym)
```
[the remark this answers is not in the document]
nigel-lowry: Good spot, I'll delete the method

app/controllers/claims_controller.rb
kenfodder: @vassyz This should clean things up, so shouldn't be an issue.

This means however you rearrange the pages, it should just work then because they all expect a claim and it's present. A little less fragile.

kenfodder (approved):

vassyz (approved):

PR #1917, opened 2022-05-04
Fix misc typos

vassyz:
I was working on my current task and noticed a few typos. I think it's a bit tidier to have a separate PR.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -66,7 +66,6 @@ en:
         no_address_found: "No addresses found matching those details"
     date_of_birth: "What is your date of birth?"
     teacher_reference_number: "What is your teacher reference number (TRN)?"
-    national_insurance_number: "What's your National Insurance number?"
     has_student_loan: "Are you currently paying off your student loan?"
     has_masters_and_or_doctoral_loan: "Did you take out a Postgraduate Masters or a Postgraduate Doctoral loan?"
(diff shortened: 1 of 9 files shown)


slorek (approved):

config/locales/en.yml
```diff
@@ -66,7 +66,6 @@ en:
         no_address_found: "No addresses found matching those details"
     date_of_birth: "What is your date of birth?"
     teacher_reference_number: "What is your teacher reference number (TRN)?"
-    national_insurance_number: "What's your National Insurance number?"
```
nigel-lowry, line 69: Is this unused then?

config/locales/en.yml
```diff
@@ -371,7 +370,7 @@ en:
                 qualification:
                   undergraduate_itt: "Select which subject you did your undergraduate ITT in"
                   postgraduate_itt: "Select which subject you did your postgraduate ITT in"
-                  assessment_only: "Select which subject you did your assessment only in"
+                  assessment_only: "Select which subject you did your assessment in"
```
nigel-lowry, line 373: If that's on the screen, I probably wouldn't change that

config/locales/en.yml
vassyz: It's a duplicate, there's a 2nd value `national_insurance_number: "What is your National Insurance number?"`

PR #1920, opened 2022-05-09
Rework policies

kenfodder:


nigel-lowry (approved):

app/controllers/base_public_controller.rb
```diff
@@ -2,13 +2,14 @@ class BasePublicController < ApplicationController
   include ClaimSessionTimeout
 
   helper_method :current_policy, :current_policy_routing_name, :claim_timeout_in_minutes
+  before_action :add_view_paths
   before_action :end_expired_claim_sessions
   after_action :update_last_seen_at
 
   private
 
   def current_policy
```
slorek, line 11: I think we would not want there to be a single `current_policy` but rather multiple `current_policies` - and, given the introduction of `Journey` we probably want this to be `current_journey` instead.

app/controllers/base_public_controller.rb
```diff
@@ -2,13 +2,14 @@ class BasePublicController < ApplicationController
...
-    Policies[current_policy_routing_name]
+    Journeys.policy_for_routing_name(current_policy_routing_name)
```
slorek, line 12:
```suggestion
    Journeys.policies_for_routing_name(journey_name)
```

app/controllers/base_public_controller.rb
```diff
@@ -2,13 +2,14 @@ class BasePublicController < ApplicationController
...
   end
 
   def current_policy_routing_name
```
slorek, line 15:
```suggestion
  def journey_name
```

app/models/journeys.rb
```diff
@@ -0,0 +1,48 @@
+# Policies should not control the front-end journey as one journey can be an application
+# for more than one policy
+class Journeys
```
slorek, line 3: Prefer singular naming.
```suggestion
class Journey
```

app/controllers/concerns/part_of_claim_journey.rb
```diff
@@ -25,23 +25,41 @@ def send_unstarted_claimants_to_the_start
...
 
+  # Placeholder needs session work
+  def current_ecp_claim
+    @current_ecp_claim
+  end
+
+  # Placeholder needs session work
+  def current_lup_claim
+    @current_lup_claim
+  end
```
slorek, line 42: Suggestion: something more generic like `current_claims` being either an array, or hash of `Policy` => `Claim` objects

app/controllers/concerns/part_of_claim_journey.rb
```diff
@@ -25,23 +25,41 @@ def send_unstarted_claimants_to_the_start
...
+
+  # This will need to manage multiple claims
   def claim_from_session
     Claim.find(session[:claim_id]) if session.key?(:claim_id)
   end
 
   def build_new_claim
```
slorek, line 49:
```suggestion
  def build_new_claims
```

app/controllers/concerns/part_of_claim_journey.rb
```diff
@@ -25,23 +25,41 @@ def send_unstarted_claimants_to_the_start
...
-    )
+    Journeys.policies_for_routing_name(current_policy_routing_name).map do |policy|
+      claim = Claim.new(
+        eligibility: policy::Eligibility.new,
+        academic_year: policy_configuration.current_academic_year
+      )
+      @current_ecp_claim = claim if policy == EarlyCareerPayments
+      @current_lup_claim = claim if policy == LevellingUpPayments
+      claim
+    end.first
```
slorek, line 58: Again making the `claims` more generic and an Array so we're not hard-coding references to the two policies.

```suggestion
    @claims = Journeys.policies_for_routing_name(current_policy_routing_name).map do |policy|
      claim = Claim.new(
        eligibility: policy::Eligibility.new,
        academic_year: policy_configuration.current_academic_year
      )
      claim
    end.first
```

app/models/journeys.rb
```diff
@@ -0,0 +1,48 @@
...
+  ALL = [
+    {
+      routing_name: "student-loans",
```
slorek, line 6: I think there's probably a way we can move this routing name out of the model and into the routes.rb where it belongs (possibly by mounting or with a module), but this is a good step.

app/controllers/concerns/part_of_claim_journey.rb
kenfodder: This would make sense when doing the work which addresses storing the claim_ids in the session and adapting the  `claims_controller` to deal with multiple claims

app/models/journeys.rb
kenfodder: Agreed, sensible place to put it for now.

app/controllers/base_public_controller.rb
kenfodder: Will do this when we incorporate `current_journey`

PR #1922, opened 2022-05-10
Reimplement back links

nigel-lowry:
https://dfedigital.atlassian.net/browse/CAPT-200

Reimplements back links without recourse to JavaScript so back links now work in the Apple Safari browser

Removes breaking change created by https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/1915

This stops another problem where a page is re-rendered as part of the wizard and allows a whole step to be skipped over to change previous answers (this isn't always possible with the JavaScript links in Chrome)


app/models/page_sequence.rb
```diff
@@ -25,6 +25,17 @@ def next_slug
     slugs[current_slug_index + 1]
   end
 
+  def previous_slug
+    slug_index = current_slug_index
+    deadend_slugs = %w[complete existing-session eligible-now eligibility-confirmed eligible-later ineligible]
+
+    if slug_index.zero? || current_slug.in?(deadend_slugs)
```
vassyz, line 32: It's a personal preference but I usually use a guard clause on a conditional that ends a method.

```ruby
return nil if slug_index.zero? || current_slug.in?(deadend_slugs)

slugs[slug_index - 1]
```

vassyz (approved):

app/models/page_sequence.rb
kenfodder: Yeah I prefer that as well, short circuit first

nigel-lowry: Done, but it's one of the things I like about Ruby: you don't need to explicitly `return`

kenfodder: Just a style, I'm not too fussed just convention and rubocop does pick these up on certain scenarios

kenfodder (approved):

kenfodder (approved):

PR #1927, opened 2022-05-11
CAPT-221 Admin search for LUP claims

nigel-lowry:
https://dfedigital.atlassian.net/browse/CAPT-221

Adds factory to create LUP claims for testing Admin-side of application.

Adds a copy of the ECP fields to database and an extra field needed for a concurrent ticket. These fields may need to change to better suit LUP but allows things to work for now.

### DIFF
--- a/app/models/levelling_up_payments.rb
+++ b/app/models/levelling_up_payments.rb
@@ -1,3 +1,11 @@
 module LevellingUpPayments
   extend self
+
+  def short_name
+    I18n.t("levelling_up_payments.policy_short_name")
+  end
+
+  def routing_name
+    Journey.routing_name_for_policy(self)
+  end
 end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -378,3 +378,5 @@ en:
                   mathematics: "Select yes if you currently teach mathematics"
                   foreign_languages: "Select yes if you currently teach foreign languages"
                   physics: "Select yes if you currently teach physics"
+  levelling_up_payments:
+    policy_short_name: "Levelling Up Payments"
--- a/spec/support/admin_searchable_claim_shared_examples.rb
+++ b/spec/support/admin_searchable_claim_shared_examples.rb
@@ -1,0 +1,116 @@
+RSpec.shared_examples "Admin Searchable Claim" do |policy|
+  subject(:search) { Claim::Search.new(query) }
+
+  let(:reference) { "ABC123" }
+  let(:email) { "foo@example.com" }
+  let(:surname) { "Wayne" }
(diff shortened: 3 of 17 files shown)


app/models/levelling_up_payments.rb
```diff
@@ -1,3 +1,11 @@
 module LevellingUpPayments
   extend self
+
+  def short_name
+    I18n.t("levelling_up_payments.policy_short_name")
+  end
+
+  def routing_name
```
slorek, line 8: I think this goes in `Journey` now.

nigel-lowry: I've pasted something into all the policies to look into the `Journey` instead. Unfortunately `routing_name` is called all over the place so there's not a quick way of tidying up.

spec/support/admin_searchable_claim_shared_examples.rb
```diff
@@ -0,0 +1,115 @@
+RSpec.shared_examples "Admin Searchable Claim" do |policy|
+  subject(:search) { Claim::Search.new(query) }
+
+  let!(:other_claim) { create(:claim, :submitted) }
```
vassyz, line 4: This is not needed I think.

nigel-lowry: I tried to follow the pattern of the existing tests. `other_claim` is there to show that the search isn't returning an unmatching claim. I could put it in a before block instead so it doesn't need a variable name?

vassyz: Ok, a `before` block should work then, it doesn't need a name.

nigel-lowry: That's done

config/locales/en.yml
```diff
@@ -378,3 +378,6 @@ en:
                   mathematics: "Select yes if you currently teach mathematics"
                   foreign_languages: "Select yes if you currently teach foreign languages"
                   physics: "Select yes if you currently teach physics"
+  levelling_up_payments:
+    policy_short_name: "Levelling Up Payments"
+    policy_name: "Claim a levelling up payment"
```
kenfodder, line 383: This won't get used, just need the short one, shouldn't break anything 🤞 

`policy_name` is actually the name of the "service" under "Manage Services" and in turn is the text that appears in the black head banner and we're renaming the ECP service to the ECP&LUP service.

kenfodder (approved):

config/locales/en.yml
nigel-lowry: `policy_name`'s gone now, thanks

PR #1936, opened 2022-05-18
[CAPT-165] Add eligible degree subject page

vassyz:
Adds the following page when the answer to "Which subject did you do your postgraduate initial teacher training (ITT) in?" is "None of the above".

Screenshot 2022-05-31 at 09 06 22

Issues:
* The next page after clicking continue reads "Do you teach none of the above now?".
* The ineligible page when selecting "No" shows the ineligible reason for ECP claim.

### DIFF
--- a/app/models/early_career_payments/eligibility.rb
+++ b/app/models/early_career_payments/eligibility.rb
@@ -145,6 +145,10 @@ module EarlyCareerPayments
       "itt_academic_year" => ["eligible_itt_subject"]
     }.freeze
 
+    IGNORED_ATTRIBUTES = [
+      "eligible_degree_subject"
+    ]
+
     self.table_name = "early_career_payments_eligibilities"
 
     enum qualification: {
@@ -201,6 +205,13 @@ module EarlyCareerPayments
 
     def policy
       EarlyCareerPayments
+    end
+
+    def assign_attributes(*args)
+      super
+    rescue ActiveRecord::UnknownAttributeError
+      all_attributes_ignored = (args.first.keys - IGNORED_ATTRIBUTES).empty?
+      raise unless all_attributes_ignored
     end
 
     def trainee_teacher_in_2021?
--- a/app/models/early_career_payments/eligibility.rb
+++ b/app/models/early_career_payments/eligibility.rb
@@ -207,6 +207,8 @@ module EarlyCareerPayments
       EarlyCareerPayments
     end
 
+    # Rescues from errors for assignments coming from LUP-only fields
+    # eg. `claim.eligibility.eligible_degree_subject = true` will get ignored
     def assign_attributes(*args)
       super
     rescue ActiveRecord::UnknownAttributeError
--- a/app/models/early_career_payments/eligibility.rb
+++ b/app/models/early_career_payments/eligibility.rb
@@ -297,6 +297,10 @@ module EarlyCareerPayments
       end
     end
 
+    def eligible_none_of_the_above?
+      false
+    end
+
     private
 
     def find_cohorts(match_criteria:)
--- a/app/models/early_career_payments/slug_sequence.rb
+++ b/app/models/early_career_payments/slug_sequence.rb
@@ -80,7 +80,7 @@ module EarlyCareerPayments
         remove_student_loan_country_slugs(sequence)
         trainee_teacher_slugs(sequence) if claim.eligibility.trainee_teacher_in_2021?
         sequence.delete("future-eligibility") unless claim.eligibility.trainee_teacher_in_2021?
-        sequence.delete("eligible-degree-subject") unless claim.for_policy(LevellingUpPremiumPayments).eligibility.eligible_none_of_the_above?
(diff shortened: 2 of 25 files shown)


kenfodder (approved):

app/models/early_career_payments/eligibility.rb
```diff
@@ -203,6 +207,13 @@ def policy
       EarlyCareerPayments
     end
 
+    def assign_attributes(*args)
```
kenfodder, line 210: Might be worth adding a comment explaining why this is being done,

nigel-lowry (approved):

nigel-lowry (approved):

app/models/early_career_payments/slug_sequence.rb
```diff
@@ -79,6 +80,7 @@ def slugs
         remove_student_loan_country_slugs(sequence)
         trainee_teacher_slugs(sequence) if claim.eligibility.trainee_teacher_in_2021?
         sequence.delete("future-eligibility") unless claim.eligibility.trainee_teacher_in_2021?
+        sequence.delete("eligible-degree-subject") unless claim.for_policy(LevellingUpPremiumPayments)&.eligibility&.eligible_none_of_the_above?
```
kenfodder, line 83: Nitpicking - do we need the `&.` on those calls?

app/models/levelling_up_premium_payments/eligibility.rb
```diff
@@ -75,6 +84,15 @@ def award_amount
 
     # TODO - this need implementing later
```
kenfodder, line 85: Can this `TODO:` be removed now?

app/models/levelling_up_premium_payments/eligibility.rb
```diff
@@ -92,5 +110,52 @@ def calculate_award_amount
...
+    def poor_performance?
+      subject_to_formal_performance_action? ||
+        subject_to_disciplinary_action?
+    end
+
+    def ineligible_cohort?
+      return true if itt_academic_year == AcademicYear.new
+
+      false
+    end
```
kenfodder, line 159: Can this just be...?
```
def ineligible_cohort?
    itt_academic_year == AcademicYear.new
end
```

app/views/early_career_payments/claims/eligible_degree_subject.html.erb
```diff
@@ -0,0 +1,49 @@
+<% claim = current_claim.for_policy(LevellingUpPremiumPayments) %>
+
+<% content_for(:page_title, page_title(t("early_career_payments.questions.eligible_degree_subject"), policy: current_policy_routing_name, show_error: claim.errors.any?)) %>
+
+<% path_for_form = claim.persisted? ? claim_path(current_policy_routing_name) : claims_path(current_policy_routing_name) %>
```
kenfodder, line 5: Maybe for another PR, but isn't this always persisted? Check with @nigel-lowry maybe?

spec/features/ineligible_levelling_up_premium_payments_spec.rb
```diff
@@ -0,0 +1,74 @@
+require "rails_helper"
+
+RSpec.feature "Ineligible Levelling up premium payments claims" do
+  let(:eligibility) { LevellingUpPremiumPayments::Eligibility.order(:created_at).last }
+
+  scenario "When the school selected is LUP ineligible" do
+    start_levelling_up_premium_payments_claim
```
kenfodder, line 7: Again maybe another PR, given it's a combined journey, should we make different helper that returns both the claims, because visiting the url should create 2 claims.

spec/support/feature_helpers.rb
```diff
@@ -115,6 +115,11 @@ def start_early_career_payments_claim
     Claim.by_policy(EarlyCareerPayments).order(:created_at).last
   end
 
+  def start_levelling_up_premium_payments_claim
```
kenfodder, line 118: As mentioned earlier, this maybe a combined journey start vs start_(ecp|lup)_claim

app/views/early_career_payments/claims/eligible_degree_subject.html.erb
vassyz: I blindly copied this across, it seems to be in every view. @nigel-lowry Any input on this?

nigel-lowry: Before I moved the `current-school` slug to the start of the wizard (and created the record manually in the claims controller) this was necessary just in case a certain slug was ever moved to the start of the wizard (this wouldn't have worked with the `current-school` which is why I had to do a lot of work for that ticket).

I'm pretty sure you can get rid of all of those now (but probably only for the ECP/LUP journey) but I never actually tried it.

app/models/early_career_payments/slug_sequence.rb
vassyz: No, it was mostly a hack to get some tests passing. I'll update the specs.

PR #1937, opened 2022-05-18
WIP - Capt 271 delegate current claim - tests pass, reminders commented out

kenfodder:
DO NOT MERGE

This is just a progress PR for visibility...

- Tests all pass, excluding reminders which are pending
- Ugly code, needs tidying up
- Needs tests to check LUP claim is the "same" as the ECP claim
- Check any instances of `current_claim = ...` so the right model is returned

Getting there...

### DIFF
--- a/app/controllers/concerns/part_of_claim_journey.rb
+++ b/app/controllers/concerns/part_of_claim_journey.rb
@@ -25,22 +25,22 @@ module PartOfClaimJourney
     redirect_to current_policy.start_page_url unless current_claim.persisted?
   end
 
-  # Keep this for StudentLoans and MathsAndPhysics journeys
-  # Keep ECP journey working for now with current_claim as we build out combined journey
   def current_claim
     @current_claim ||= claim_from_session || build_new_claim
   end
 
-  # TASK: does NOT support multiple claims yet!
   def claim_from_session
-    Claim.find(session[:claim_id]) if session.key?(:claim_id)
+    if session.key?(:claim_id)
+      claim_ids = session[:claim_id].split(",")
+      claims = claim_ids.map { |cid| Claim.find(cid) }
+      CurrentClaim.new(claims: claims)
+    end
   end
 
   def build_new_claim
     CurrentClaim.new(claims: build_new_claims)
   end
 
-  # Beginning of setting up multiple claims in a combined journey
   def build_new_claims
     Journey.policies_for_routing_name(current_policy_routing_name).map do |policy|
       Claim.new(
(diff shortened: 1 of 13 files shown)


spec/models/current_claim_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    let(:lup_claim) { build(:claim, academic_year: "2022/2023", policy: lup_policy) }
+    let(:school) { create(:school) }
+
+    describe "#attributes=" do
+      it "sets the attributes on both claims" do
+        cc = CurrentClaim.new(claims: [ecp_claim, lup_claim])
+        cc.attributes = {"eligibility_attributes" => {"current_school_id" => school.id}}
+
+        expect(cc.claims.first.school.id).to eq(school.id)
+        expect(cc.claims.last.school.id).to eq(school.id)
```
vassyz, line 17: For tests like these I usually like to be more explicit, making sure it does the right thing:

```ruby
expect { cc.attributes = {"eligibility_attributes" => {"current_school_id" => school.id}} }
  .to change { cc.claims.first.school&.id }.from(nil).to(school.id)
  .and change { cc.claims.last.school&.id }.from(nil).to(school.id)
```

I think `CurrentClaim.new(claims: [ecp_claim, lup_claim])` and `{"eligibility_attributes" => {"current_school_id" => school.id}}` could also be memoized with a `let`.

app/controllers/concerns/part_of_claim_journey.rb
```diff
@@ -25,25 +25,24 @@ def send_unstarted_claimants_to_the_start
...
-  # Keep ECP journey working for now with current_claim as we build out combined journey
   def current_claim
     @current_claim ||= claim_from_session || build_new_claim
   end
 
-  # TASK: does NOT support multiple claims yet!
   def claim_from_session
-    Claim.find(session[:claim_id]) if session.key?(:claim_id)
+    if session.key?(:claim_id)
+      claim_ids = session[:claim_id].split(",")
```
vassyz, line 34: Can't we have an array in the sesssion? It should just work if you don't use join/split by `,` and have claim_ids as `claims.pluck(:id)`.

vassyz: Actually, maybe not. Because then you'd have to handle an Array everywhere. Your take has the same results for one claim in the session.

kenfodder: I think it works my new PR has it in, good feedback though!

PR #1939, opened 2022-05-19
Allow LUP claims to be edited by an admin

nigel-lowry:
https://dfedigital.atlassian.net/browse/CAPT-208

Allows LUP claims to be edited by an admin in a way consistent with the other policies.

While doing this, I found that some of the validation for monetary values was inconsistent between policies so addressed that.

Also agreed with Chris Williamson to remove the string parsing on the forms for monetary amounts.

### DIFF
--- a/app/models/early_career_payments/eligibility.rb
+++ b/app/models/early_career_payments/eligibility.rb
@@ -257,10 +257,6 @@ module EarlyCareerPayments
       super || calculate_award_amount
     end
 
-    def award_amount=(value)
-      super(value.to_s.gsub(/[£,\s]/, ""))
-    end
-
     def calculate_award_amount
       return BigDecimal("0.00") if current_school.nil?
 
--- a/app/models/early_career_payments/eligibility.rb
+++ b/app/models/early_career_payments/eligibility.rb
@@ -169,6 +169,12 @@ module EarlyCareerPayments
       AcademicYear.new => AcademicYear::Type.new.serialize(AcademicYear.new)
     }
 
+    @@max_award_amount_in_pounds = AWARD_AMOUNTS.collect(&:uplift_amount).max
+
+    def self.max_award_amount_in_pounds
+      @@max_award_amount_in_pounds
+    end
+
     has_one :claim, as: :eligibility, inverse_of: :eligibility
     belongs_to :current_school, optional: true, class_name: "School"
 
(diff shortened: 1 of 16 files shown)


app/models/early_career_payments/eligibility.rb
```diff
@@ -169,6 +169,12 @@ class Eligibility < ApplicationRecord
       AcademicYear.new => AcademicYear::Type.new.serialize(AcademicYear.new)
     }
 
+    @@max_award_amount_in_pounds = AWARD_AMOUNTS.collect(&:uplift_amount).max
```
vassyz, line 172: Are you ok with these class valiables? What about:

```ruby
@max_award_amount_in_pounds = AWARD_AMOUNTS.collect(&:uplift_amount).max

class << self
  attr_accessor :max_award_amount_in_pounds
end
```

vassyz: After I wrote this comment I saw the `award.rb` file and it seems you're a fan of using `@@`. If these classes don't get inherited I guess it's fine.

nigel-lowry: I totally know what you mean. I cracked open a Ruby textbook about the whole `@@` and class inheritance thing. This is somewhat related to the massive structure which will eventually hold the mapping from URN (the school identifier) to award amount. I definitely won't want multiple copies of that hanging about. I was hoping to kick the `@@` decision down the road until I work on the next story.

app/controllers/admin/amendments_controller.rb
```diff
@@ -32,6 +34,10 @@ def ensure_claim_is_amendable
     end
   end
 
+  def editable_award_amount_policy?(policy)
```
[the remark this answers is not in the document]
vassyz: It's used in a few places, search for `helper_method`. You're making the controller method accessible in the view. 

You can remove the `@is_editable_award_amount = ` and change the method to 

```ruby
helper_method :editable_award_amount?
def editable_award_amount?(policy)
  policy.in? [EarlyCareerPayments, LevellingUpPremiumPayments]
end
```

and update the view to use `<% if editable_award_amount?(@amendment.claim.policy) %>`

It should work.

Having said that, you don't need anything from the controller anyway, so why not move the method to a helper. `Admin::ClaimsHelper` would be appropriate I think.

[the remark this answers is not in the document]
nigel-lowry: Thanks I've made the change. That was a good explanation of helpers: it's been years since I've used one.

vassyz (approved):
Approved, but please wait for @slorek's review before merging.

PR #1942, opened 2022-05-20
Add hash of URNs to LUP award amounts

nigel-lowry:
First part of https://dfedigital.atlassian.net/browse/CAPT-227

Adds real values to the hash. Schools with zero award are not included in the hash to save space.

During review, *please load the large diff* for `app/models/levelling_up_premium_payments/award.rb` because there has been some refactoring.


app/models/levelling_up_premium_payments/award.rb
```diff
@@ -1,42 +1,2694 @@
...
 
       @school = school
     end
 
     def self.max
-      @@maximum_award_amount_in_pounds
+      @maximum_award_amount_in_pounds ||= urn_to_award_amount_in_pounds.values.max
+    end
+
+    def self.urn_to_award_amount_in_pounds
```
slorek, line 16: I think we would need to be able to retrieve this by academic year of the claim, e.g.

```suggestion
    def self.urn_to_award_amount_in_pounds(academic_year)
```

app/models/levelling_up_premium_payments/award.rb
```diff
@@ -1,42 +1,2710 @@
...
+      check_for_positive_value_in_case_zero_award_has_crept_into_hash
     end
 
     def amount_in_pounds
-      @@urn_to_award_amount_in_pounds.fetch(@school.urn, 0)
+      self.class.urn_to_award_amount_in_pounds(@year).fetch(@school.urn, 0)
+    end
+
+    private_class_method def self.urn_to_award_amount_in_pounds_for_2022_to_2023
+      @urn_to_award_amount_in_pounds_for_2022_to_2023 ||= {
```
vassyz, line 45: Why memoize here?

nigel-lowry: My thinking was because the hash is huge and I thought it would avoid repeated hash creation but you might be right. Also it's a class method so there's not multiple copies of this massive hash.

app/models/levelling_up_premium_payments/award.rb
```diff
@@ -1,42 +1,2710 @@
...
-    @@urn_to_award_amount_in_pounds = {
-      150000 => 1_000,
-      150001 => 3_000,
-      160000 => 0,
-      160001 => 0
-    }
+    private
 
-    @@maximum_award_amount_in_pounds = @@urn_to_award_amount_in_pounds.values.max
+    def check_for_positive_value_in_case_zero_award_has_crept_into_hash
```
vassyz, line 2706: I think I'd prefer something shorter, maybe `positive_value_in_case_zero_award_has_crept_into_hash?`

app/models/levelling_up_premium_payments/award.rb
```diff
@@ -1,42 +1,2704 @@
...
+      @year_to_max.fetch(year)
     end
 
-    def self.max
-      @@maximum_award_amount_in_pounds
+    def self.urn_to_award_amount_in_pounds(year)
+      year_to_awards.fetch(year)
+    end
+
+    def self.year_to_awards
```
vassyz, line 32: YOu are hardcoding 2022, but passing it into the `max` method.

app/models/levelling_up_premium_payments/award.rb
```diff
@@ -1,42 +1,2704 @@
...
+      raise "nil year" if year.nil?
+
+      raise "no LUP award mapping for #{year}" unless year.in? self.class.year_to_awards
 
       @school = school
+      @year = year
+    end
+
+    def self.max(year)
+      if @year_to_max.nil?
```
vassyz, line 17: Wouldn't `@year_to_max ||= {}` work?

vassyz (approved):

PR #1948, opened 2022-05-24
DQT API client for Levelling Up Premium Payments placeholder

slorek:
https://dfedigital.atlassian.net/browse/CAPT-211

When a claim is submitted it expects a `DqtRecord` class to exist within the policy namespace as per https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/master/app/models/automated_checks/claim_verifiers/qualifications.rb#L87.

I've added this class which requires implementation to ensure this doesn't break when an LUP claim is submitted.


kenfodder (approved):

spec/models/levelling_up_premium_payments/dqt_record_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+        itt_subject_codes: ["G1100"],
+        itt_start_date: Date.parse("1/9/2018"),
+        qts_award_date: Date.parse("31/8/2019"),
+        qualification_name: "BA"
+      }
+    )
+  end
+
+  describe "#eligible?" do
+    subject(:eligible?) { dqt_record.eligible? }
```
nigel-lowry, line 32: Could replace this whole `describe` body with `specify { expect(dot_record).to be_eligible }

nigel-lowry (approved):

spec/models/levelling_up_premium_payments/dqt_record_spec.rb
slorek: Cheers, I'm expecting this to be filled in with implementation so will leave it for now.

PR #1957, opened 2022-05-27
Move Journey into PolicyConfiguration

kenfodder:


app/models/policy_configuration.rb
```diff
@@ -12,6 +12,26 @@
 class PolicyConfiguration < ApplicationRecord
   ACADEMIC_YEAR_REGEXP = /\A20\d{2}\/20\d{2}\z/.freeze
 
+  SERVICES = [
```
slorek, line 15: Arguably these properties should probably be moved to columns on the database records as a next step. We already have `policy_types` being equivalent to the `policies` key here.

kenfodder: PolicyConfiguration only has a list of `policy_types` otherwise currently there is no other way of associating it to what routing_name they belong to. But we can't store the routing_name in the db because then we can't boot the routes.rb file without querying the DB, which is the noob error I hit earlier on CI.

app/models/policy_configuration.rb
```diff
@@ -12,6 +12,26 @@
...
+      policies: [StudentLoans]
+    },
+    {
+      routing_name: "maths-and-physics",
+      slugs: MathsAndPhysics::SlugSequence::SLUGS,
+      policies: [MathsAndPhysics]
+    },
+    {
+      routing_name: "early-career-payments",
+      slugs: EarlyCareerPayments::SlugSequence::SLUGS,
```
nigel-lowry, line 28: Not now but eventually we need to rename that slug sequence to cover both ECP and LUP

nigel-lowry (approved):

app/models/policy_configuration.rb
kenfodder: Good point, I've added a few more TODOs to where that might be affected.

PR #1973, opened 2022-06-08
Stop asking trainee teachers their ITT year

nigel-lowry:
Caveat: not sure what becomes of reminders when a trainee teacher now asks for one. The page says they'll get it in September but doesn't say which year (because we never asked them those details). Could be this September or never. I don't know anything about the reminder service so can't say.

### DIFF
--- a/spec/features/early_career_payments_trainee_teacher_spec.rb
+++ b/spec/features/early_career_payments_trainee_teacher_spec.rb
@@ -78,22 +78,17 @@ RSpec.feature "Trainee Teacher - Early Career Payments - journey" do
       expect(eligibility.nqt_in_academic_year_after_itt).to eql false
       expect(claim.eligibility.reload.qualification).to eq "postgraduate_itt"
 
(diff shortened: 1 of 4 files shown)


spec/features/early_career_payments_trainee_teacher_spec.rb
```diff
@@ -78,23 +78,18 @@
...
-
-      choose "2018 to 2019"
-      click_on "Continue"
-
       # - Which subject did you do your postgraduate ITT in
       expect(page).to have_text(
         I18n.t("early_career_payments.questions.eligible_itt_subject_trainee_teacher_in_2021")
       )
+
+      # for CAPT-350 will want to make sure `expect(page).to have_no_text("Foreign languages")` passes
```
vassyz, line 86: Should this be a TODO?

kenfodder: Yeah set to TODO, I'll tackle that.

kenfodder (approved):

PR #1979, opened 2022-06-14
Add subject logic for different ITT years

nigel-lowry:
Covers the five academic year variants (and one @vassyz already did the work for): 

* https://dfedigital.atlassian.net/browse/CAPT-355 (@vassyz 's ticket)
* https://dfedigital.atlassian.net/browse/CAPT-357
* https://dfedigital.atlassian.net/browse/CAPT-358
* https://dfedigital.atlassian.net/browse/CAPT-359
* https://dfedigital.atlassian.net/browse/CAPT-360

Really there are two parts to each ticket:

* the set of subjects; and
* where the flow goes next

This PR *only* deals with the set of subjects. The flow is too big a change so will go into a different ticket.


app/models/itt_subject_set.rb
```diff
@@ -0,0 +1,50 @@
...
+      if @ineligible_for_lup
+        %i(mathematics)
+      else
+        %i(chemistry computing mathematics physics)
+      end
+    when AcademicYear.new(2020)
+      if @ineligible_for_lup
+        %i(chemistry foreign_languages mathematics physics)
+      else
+        %i(chemistry computing foreign_languages mathematics physics)
```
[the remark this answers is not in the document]
nigel-lowry: Hopefully it won't!

app/models/itt_subject_set.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def subjects
+    return %i(chemistry computing mathematics physics) if @trainee_teacher
+
+    case @itt_academic_year
+    when AcademicYear.new(2017)
+      if @ineligible_for_lup
+        []
+      else
+        %i(chemistry computing mathematics physics)
```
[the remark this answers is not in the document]
nigel-lowry: Done

app/models/itt_subject_set.rb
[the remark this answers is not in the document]
vassyz: Check `CAPT-359` it uses `Languages`.

app/models/itt_subject_set.rb
```diff
@@ -0,0 +1,50 @@
...
+      else
+        %i[chemistry computing mathematics physics]
+      end
+    when AcademicYear.new(2020)
+      if @ineligible_for_lup
+        %i[chemistry foreign_languages mathematics physics]
+      else
+        %i[chemistry computing foreign_languages mathematics physics]
+      end
+    when AcademicYear.new(2021)
```
vassyz, line 42: This is the same as `AcademicYear.new(2017)`

nigel-lowry: @vassyz Good spot. I can put the two years in the same `when` branch, or do you think they're better in chronological order as above? Let me know which you prefer to get that approval

app/views/early_career_payments/claims/eligible_itt_subject.html.erb
```diff
@@ -21,7 +22,7 @@
             <%= errors_tag current_claim.eligibility, :eligible_itt_subject %>
 
             <div class="govuk-radios">
-              <% eligible_itt_subjects.each do |option| %>
```
kenfodder, line 24: L7 in this file, there is a reference to `_eligible_itt_subject_#{eligible_itt_subjects.first}" }) if`

app/models/itt_subject_set.rb
vassyz: Chronologically works best I think.

app/models/itt_subject_set.rb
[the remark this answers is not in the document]
nigel-lowry: I think the i18n key will have to remain as `foreign_languages` for now (there's a PR for this I'll work on next) but the i18n value will change to `Languages` (and where it's hardcoded in the views)

app/models/itt_subject_set.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+class IttSubjectSet
```
slorek, line 3: Suggestion: Should this go in the `LevellingUpPremiumPayments` namespace if it is only concerned with the LUP claim and no others, or the `EarlyCareerPayments` namespace alongide the `SlugSequence`?

Also it seems to be more of a view helper, since it's not used for any business logic e.g. validation or eligibility calculation. Or should we be validating the inputs against this set as well in the controller?

vassyz (approved):

app/models/itt_subject_set.rb
nigel-lowry: Annoyingly it's not unique to either namespace. It's *mostly* about LUP but where you see foreign languages make a reappearance one year that's because of ECP.

vassyz: I don't think it needs a namespace, but it does make sense to me to be put in a helper.

nigel-lowry: Also, you are right in that a controller somewhere should be validating that the subject is valid for a particular year but I've not done that now (I was going to kick that down the road) but I can work on that here if you like?

slorek: If this is intended to be used in follow up PRs then I don't mind too much.

slorek (approved):

kenfodder (approved):

PR #1982, opened 2022-06-14
CAPT-386 Implement new result page design

slorek:
https://dfedigital.atlassian.net/browse/CAPT-386

Implements the new result page design:

* The presentation changes depending on whether the user is eligible to claim for for one or two policies:
  * If the user qualifies for one of the two policies, they will see a button to continue their application
  * If the user is eligible for both policies, they will see two radio options allowing them to choose which option they want to apply for.

To do:

* Implement forms and back end selection (currently the ECP claim is always the one submitted)
* Ensure that an option has been selected on the radio version
* Delete the claim which has not been selected
* There are some slight spacing variances versus the screenshots supplied despite the same HTML used


app/models/current_claim.rb
```diff
@@ -61,7 +71,30 @@ def ineligible?
...
   def editable_attributes
     claims.flat_map { |c| c.eligibility.class::EDITABLE_ATTRIBUTES }.uniq
   end
+
+  def eligible_now
+    claims.select { |c| c.eligibility.eligible_now? }
+  end
+
+  def eligible_now_and_sorted
+    eligible_now.group_by { |c| c.award_amount.to_i }.map { |k, claims| claims.sort_by { |c| c.policy.short_name } }.reverse.flatten
```
nigel-lowry, line 91: Isn't it meant to be sorted by decreasing monetary amount, so the highest award appears first?

app/models/levelling_up_premium_payments/eligibility.rb
```diff
@@ -84,7 +83,7 @@ def eligible_later?
     end
 
     def award_amount
-      calculate_award_amount
+      super || calculate_award_amount
```
nigel-lowry, line 86: I had a weird bug where if it called `super` first then showed the award amount, if I revisited the page, the award amount would increase. I didn't really understand it, but getting rid of the call to `super` fixed it. You'll want to try coming back to the page and making sure the monetary amounts don't change.

app/models/page_sequence.rb
```diff
@@ -27,7 +27,7 @@ def next_slug
 
   def previous_slug
     slug_index = current_slug_index
-    dead_end_slugs = %w[complete existing-session eligible-now eligibility-confirmed eligible-later ineligible]
+    dead_end_slugs = %w[complete existing-session eligible-now eligible-later ineligible]
```
nigel-lowry, line 30: Currently these slugs are dead-ends in that you shouldn't see a back button. If we're going to allow people to change their answers, this might not be the right thing to do any longer.

app/views/early_career_payments/claims/eligibility_confirmed.html.erb
```diff
@@ -1,48 +1,25 @@
...
+<% content_for(:page_title, page_title("You’re eligible for an additional payment", policy: current_policy_routing_name, show_error: current_claim.errors.any?)) %>
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <% current_claim.claims.each do |claim| %>
-      <% if claim.eligibility.eligible_now? %>
-        <h1 class="govuk-heading-xl">You are eligible to apply for a <%= number_to_currency(claim.eligibility.award_amount, precision: 0) %> <%= claim.eligibility.policy.short_name.downcase.singularize %></h1>
-      <% end %>
+    <div class="govuk-panel govuk-panel--informational">
+      <h1 class="govuk-panel__title">You’re eligible for an additional payment</h1>
```
nigel-lowry, line 6: Another curly quote

app/views/early_career_payments/claims/eligibility_confirmed.html.erb
```diff
@@ -1,48 +1,25 @@
...
+    </div>
+
+    <% if current_claim.eligible_now.size == 1 %>
```
nigel-lowry, line 9: Can use `<% if current_claim.eligible_now.one? %>`

spec/features/early_career_payments_claim_spec.rb
```diff
@@ -628,16 +621,11 @@
     click_on("Continue")
 
     # - You are eligible for an early career payment
-    expect(page).to have_text("You are eligible")
-    expect(page).to have_text("for a £5,000 early-career payment")
+    expect(page).to have_text("You’re eligible for an additional payment")
```
nigel-lowry, line 624: Another curly quote

app/models/current_claim.rb
slorek: That's correct, this is asserted in the specs.

spec/models/current_claim_spec.rb
```diff
@@ -89,7 +89,7 @@
       let(:ecp_eligibility) { build(:early_career_payments_eligibility, :eligible) }
       let(:lup_eligibility) { build(:levelling_up_premium_payments_eligibility, :eligible) }
 
-      let(:ecp_claim) { build(:claim, academic_year: "2022/2023", eligibility: ecp_eligibility) }
+      let(:ecp_claim) { build(:claim, academic_year: "2021/2022", eligibility: ecp_eligibility) }
```
nigel-lowry, line 92: Is that year right?

app/views/early_career_payments/claims/eligibility_confirmed.html.erb
```diff
@@ -1,48 +1,25 @@
-<% content_for(:page_title, page_title("You are eligible to claim for early-career payment", policy: current_policy_routing_name)) %>
+<% content_for(:page_title, page_title("You’re eligible for an additional payment", policy: current_policy_routing_name, show_error: current_claim.errors.any?)) %>
```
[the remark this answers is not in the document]
slorek: It's not necessary with UTF-8 encoding: see https://www.gov.uk/government/publications/open-standards-for-government/cross-platform-character-encoding-profile.

spec/models/current_claim_spec.rb
slorek: Yes, I added a comment to the factory as it doesn't work with the 2022/23 year.

app/models/levelling_up_premium_payments/eligibility.rb
nigel-lowry: Please try it locally! I only saw this when I hit the back button and came back to the page and the award amount doubled. I never really understood why that was happening.

app/views/early_career_payments/claims/eligibility_confirmed.html.erb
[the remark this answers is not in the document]
nigel-lowry: We put in the HTML entities in GHWT because curly quotes can be so difficult to spot. Curly quotes are a pain generally if you try to search for text like "you're" but you've typed a straight quote (like you would naturally) then you don't get a match.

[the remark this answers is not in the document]
nigel-lowry: Basically if we can get away with straight quotes, it's less of a headache all round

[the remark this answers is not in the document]
slorek: Agreed (I had to re-type it because of the quote) and I'm not sure what the DfE convention is, but the content designer is responsible for this. The original content I replaced had curly quotes as well.

app/models/levelling_up_premium_payments/eligibility.rb
slorek: I've been back and forward through this page hundreds of times during the making of the PR, trust me ;)

app/views/early_career_payments/claims/eligibility_confirmed.html.erb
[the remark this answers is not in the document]
nigel-lowry: It's worth checking with them, because they might have copy-and-pasted themselves from a "helpful" app like Microsoft Word or Apple Mac Notes which automatically converts to curly quotes. If the designers haven't intentionally asked for curly quotes I'd change them to straight quotes. Also, if they don't understand what you're even asking, I'd just change them to straight quotes.

app/models/levelling_up_premium_payments/eligibility.rb
nigel-lowry: Ha! Fair enough!

app/models/levelling_up_premium_payments/eligibility.rb
```diff
@@ -6,7 +6,7 @@ class Eligibility < ApplicationRecord
 
     # use first year of LUP for now but this must come from a PolicyConfiguration
     validates :award_amount, on: :amendment, award_range: {max: LevellingUpPremiumPayments::Award.max(AcademicYear.new(2022))}
-    validates :eligible_degree_subject, on: [:"eligible-degree-subject", :submit], inclusion: {in: [true, false], message: "Select yes if you have a degree in an eligible subject"}
+    validates :eligible_degree_subject, on: [:"eligible-degree-subject"], inclusion: {in: [true, false], message: "Select yes if you have a degree in an eligible subject"}
```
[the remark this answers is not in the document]
kenfodder: @vassyz you added this original line https://github.com/DFE-Digital/claim-additional-payments-for-teaching/commit/039e810ffff66c328a9adb34511e618c3192efd9, would this cause an issue?

kenfodder (approved):

app/models/levelling_up_premium_payments/eligibility.rb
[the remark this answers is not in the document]
vassyz: Most field validation have the `:submit` option set. I think it needs to be there for when the question is being asked in the journey

app/models/current_claim.rb
vassyz: I feel like this line does too much, maybe assigning some variables would mke it a bit clearer.

app/views/early_career_payments/claims/eligibility_confirmed.html.erb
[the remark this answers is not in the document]
vassyz: VS Code highlights these quotes, so they are probably not the recommended way.

spec/factories/levelling_up_premium_payments/eligibilities.rb
```diff
@@ -9,6 +9,7 @@
       qualification { :postgraduate_itt }
       eligible_itt_subject { :mathematics }
       teaching_subject_now { true }
+      eligible_degree_subject { true }
```
vassyz, line 12: Is this required? It shouldn't need `eligible_degree_subject` if `eligible_itt_subject` is `:mathematics`.

spec/helpers/claims/show_helper_spec.rb
```diff
@@ -38,4 +38,50 @@
...
     end
   end
+
+  describe "#policy_name" do
+    subject(:name) { helper.policy_name(claim) }
+
+    context "with a StudentLoans policy" do
+      let(:policy) { StudentLoans }
+
+      it "returns a text string" do
```
vassyz, line 48: You can use the shorthand version for these specs eg. `it { is_expected.to eq "student loan" }`

spec/models/claim_spec.rb
```diff
@@ -610,6 +611,10 @@
       it "generates a reference" do
         expect(claim.reference).to_not eq nil
       end
+
+      it "sets the eligibility award amount" do
+        expect(claim.eligibility.attributes["award_amount"]).to be_a(BigDecimal)
```
vassyz, line 616: Why not `.to eq BigDecimal("2000.00")`

app/models/levelling_up_premium_payments/eligibility.rb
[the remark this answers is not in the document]
slorek: I think it needs to allow `nil` values if it is validating on `submit`. That's why I removed it as it doesn't appear to be needed on `submit` as this question isn't always asked depending on the journey taken.

spec/factories/levelling_up_premium_payments/eligibilities.rb
slorek: I was getting spec failures before this change but that seems to no longer be an issue after removing this so I will remove it now, thanks for flagging.

spec/models/claim_spec.rb
```diff
@@ -642,21 +653,37 @@
...
 
-      it "returns false" do
-        expect(claim.submit!).to eq false
+      it "raises an exception" do
+        expect { claim.submit! }.to raise_error(Claim::NotSubmittable)
       end
 
       it "doesn't change the reference number" do
-        expect { claim.submit! }.not_to(change { claim.reference })
+        expect {
```
vassyz, line 670: I'd personally add `RSpec::Matchers.define_negated_matcher :not_change, :change` to `spec_helper.rb` and combine these three specs:

```ruby
expect { claim.submit! }
  .to raise_error(Claim::NotSubmittable)
  .and not_change { claim.reference }
  .and not_change { claim.submitted_at }
```

spec/models/current_claim_spec.rb
```diff
@@ -187,12 +255,117 @@
...
+        let(:lup_amount) { 2000.0 }
+
+        it "orders the claims by highest award amount" do
+          expect(result).to eq([lup_claim, ecp_claim])
+        end
+      end
+    end
+
+    describe "#submit!" do
+      subject(:result) { cc.submit!(policy) }
```
vassyz, line 327: You're only using this as a subject in the last spec.

spec/factories/levelling_up_premium_payments/eligibilities.rb
vassyz: Could it be related to removing `:submit` from the validation?

spec/helpers/claims/show_helper_spec.rb
slorek: I'm not sure the shorthand version is better 🤷  Happy to go with it if preferred.

spec/models/claim_spec.rb
slorek: I'm not sure I care what the exact amount is here as that is dependent on the inputs, which we're not specific about and we're in danger of mixing testing cocnerns. What I'm testing is that _an amount_ is set.

spec/helpers/claims/show_helper_spec.rb
vassyz: It's not better, it's shorter. There are a lot of `it "returns a text string" do`.

spec/models/claim_spec.rb
nigel-lowry: Could go with:

```
expect(claim.eligibility.attributes["award_amount"]).to be_a(BigDecimal).and be_positive
```

vassyz: I'd go with just `be_positive` if what you want to test is that it's being set.

app/views/early_career_payments/claims/eligibility_confirmed.html.erb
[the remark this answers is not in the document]
slorek: Confirmed we should use curly apostrophes.

vassyz (approved):

PR #1995, opened 2022-06-24
Change url path from early-career-payments to additional-payments

kenfodder:
https://dfedigital.atlassian.net/browse/CAPT-395

This changes the URL from early-career-payments to additional-payments.

Note: This PR doesn't handle the redirect part of the ticket.

### DIFF
--- a/app/models/policy_configuration.rb
+++ b/app/models/policy_configuration.rb
@@ -16,20 +16,22 @@ class PolicyConfiguration < ApplicationRecord
     {
       routing_name: "student-loans",
       slugs: StudentLoans::SlugSequence::SLUGS,
-      policies: [StudentLoans]
+      policies: [StudentLoans],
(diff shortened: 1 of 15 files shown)


nigel-lowry (approved):

app/models/policy_configuration.rb
```diff
@@ -16,20 +16,22 @@ class PolicyConfiguration < ApplicationRecord
     {
       routing_name: "student-loans",
       slugs: StudentLoans::SlugSequence::SLUGS,
-      policies: [StudentLoans]
+      policies: [StudentLoans],
+      i18n_namespace: "student_loans"
```
slorek, line 20: This might be redundant/duplication as each policy model already specifies a `locale_key` e.g. https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/master/app/models/early_career_payments.rb#L36

PR #2001, opened 2022-06-28
[CAPT-211] Add degree codes from DQT response

vassyz:
Before
Screenshot 2022-06-29 at 11 14 17

After
Screenshot 2022-06-29 at 11 14 30

### DIFF
--- a/spec/lib/dqt/objects/teacher_spec.rb
+++ b/spec/lib/dqt/objects/teacher_spec.rb
@@ -385,9 +385,7 @@ RSpec.describe Dqt::Teacher do
   describe "#degree_codes" do
     subject(:degree_codes) { qualified_teaching_status.degree_codes }
 
-    it "returns an Array" do
-      expect(degree_codes).to eq ["100403", "100105"]
-    end
+    specify { expect(degree_codes).to contain_exactly("100403", "100105") }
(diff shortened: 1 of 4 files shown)


spec/lib/dqt/objects/teacher_spec.rb
```diff
@@ -359,8 +385,8 @@ def response(response)
   describe "#degree_codes" do
     subject(:degree_codes) { qualified_teaching_status.degree_codes }
 
-    it "returns Array" do
-      expect(degree_codes).to eq []
+    it "returns an Array" do
+      expect(degree_codes).to eq ["100403", "100105"]
```
nigel-lowry, line 389: That isn't what you're asserting though. I'd just go with `specify { expect(degree_codes).to contain_exactly("100403", "100105")}` and let `specify` generate you a failing message if it fails (rather than relying on what you typed)

nigel-lowry (approved):

spec/lib/dqt/objects/teacher_spec.rb
vassyz: Done in 8c6078e31a2a1d16a4e1074f3c7f68168eb68acd

PR #2002, opened 2022-06-28
Consolidate ECP and LUP logic

nigel-lowry:
The purpose of this PR is to get the slug sequence to check against the eligibility status of its claims rather than create many special cases which look at individual attributes (which are effectively separate eligibility checks outside the encapsulated `Eligibility` classes).

Notes:
* `!ineligible?` calls have now disappeared from the codebase. There's a status method we should use instead;
* The difficult cohort code has been deleted;
* The common elements of ECP and LUP have been brought into line. Literally 99% of the checks each policy makes are shared.


app/models/current_claim.rb
```diff
@@ -75,24 +75,52 @@ def respond_to_missing?(method_name, *args)
...
+    claims.any? { |c| c.eligibility.definitely_eligible_later? }
+  end
+
   def editable_attributes
     claims.flat_map { |c| c.eligibility.class::EDITABLE_ATTRIBUTES }.uniq
   end
 
   def eligible_now
-    claims.select { |c| c.eligibility.eligible_now? }
+    claims.select { |c| c.eligibility.status == :eligible_now }
```
kenfodder, line 123: Is this useful or used anywhere, it's really just select the first one that is `:eligible_now`

slorek (changes requested):
Really good work, some major improvements. Only some nitpicky changes from my side, especially around the use of operators. Also Rails provides some 'magic' methods for boolean properties which would make things easier I think.

app/models/concerns/eligibility_checkable.rb
```diff
@@ -0,0 +1,122 @@
...
+  def has_indicated_ineligible_school?
+    current_school.present? and !policy::SchoolEligibility.new(current_school).eligible?
+  end
+
+  def trainee_teacher?
+    nqt_in_academic_year_after_itt == false
+  end
+
+  def supply_teacher_lacking_either_long_contract_or_direct_employment?
+    employed_as_supply_teacher? and (has_entire_term_contract == false or employed_directly == false)
```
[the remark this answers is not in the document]
nigel-lowry: I'll do the operator changes but `!var?` is different from `x == false`

app/models/concerns/eligibility_checkable.rb
```diff
@@ -0,0 +1,122 @@
...
+  def newly_qualified_teacher?
+    nqt_in_academic_year_after_itt
+  end
+
+  def has_indicated_eligible_school?
+    current_school.present? and policy::SchoolEligibility.new(current_school).eligible?
+  end
+
+  def non_supply_teacher_or_supply_teacher_with_long_contract_and_direct_employment?
+    (employed_as_supply_teacher == false) or (employed_as_supply_teacher and has_entire_term_contract and employed_directly)
```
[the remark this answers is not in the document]
nigel-lowry: Breaks but I'll take the operators

app/models/current_claim.rb
```diff
@@ -101,6 +129,18 @@ def eligible_now_and_sorted
 
   private
 
+  def anything_eligible_now
```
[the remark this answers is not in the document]
nigel-lowry: There's a few methods there I'll rename similarly

app/models/page_sequence.rb
```diff
@@ -39,6 +43,27 @@ def in_sequence?(slug)
 
   private
 
+  def trainee_teacher_at_lup_school?
+    LevellingUpPremiumPayments.in?(claim.policies) and claim.eligibility.nqt_in_academic_year_after_itt == false and LevellingUpPremiumPayments::SchoolEligibility.new(claim.eligibility.current_school).eligible?
```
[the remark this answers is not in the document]
nigel-lowry: I'll take the operator changes only

app/models/concerns/eligibility_checkable.rb
```diff
@@ -0,0 +1,122 @@
...
+    common_eligible_later_attributes? && specific_eligible_later_attributes?
+  end
+
+  private
+
+  def common_ineligible_attributes?
+    [has_indicated_ineligible_school?, trainee_teacher?, supply_teacher_lacking_either_long_contract_or_direct_employment?, poor_performance?, no_selectable_subjects?, ineligible_cohort?, insufficient_teaching?].any?
+  end
+
+  def has_indicated_ineligible_school?
```
vassyz, line 46: The convention is that you don't usually add `has_` if the method ends in `?`. Same for `has_indicated_eligible_school?`.

app/models/concerns/eligibility_checkable.rb
```diff
@@ -0,0 +1,122 @@
...
+    employed_as_supply_teacher? and (has_entire_term_contract == false or employed_directly == false)
+  end
+
+  def poor_performance?
+    subject_to_formal_performance_action? || subject_to_disciplinary_action?
+  end
+
+  def no_selectable_subjects?
+    JourneySubjectEligibilityChecker.new(claim_year: claim_year, itt_year: itt_academic_year).current_and_future_subject_symbols(policy).empty?
+  rescue
```
vassyz, line 64: Can't we avoid this rescue? Or specifying an error class?

app/models/early_career_payments/award_amount_calculator.rb
```diff
@@ -0,0 +1,138 @@
...
+          case @subject_symbol
+          when :chemistry, :foreign_languages, :mathematics, :physics
+            if @school.eligible_for_early_career_payments_as_uplift?
+              3_000
+            else
+              2_000
+            end
+          end
+        end
+      when AcademicYear.new(2024)
```
vassyz, line 83: This is a bit hard to follow, but isn't this branch the same as 2022?

app/models/concerns/eligibility_checkable.rb
nigel-lowry: I'll do a bit of renaming in that class

spec/models/early_career_payments/award_amount_calculator_spec.rb
```diff
@@ -0,0 +1,354 @@
+require "rails_helper"
+
+describe EarlyCareerPayments::AwardAmountCalculator do
+  let(:base_school) { instance_double("School", eligible_for_early_career_payments?: true, eligible_for_early_career_payments_as_uplift?: false) }
```
vassyz, line 4: Why not the class? `instance_double(School)`?

nigel-lowry: I didn't like the factories

vassyz (approved):

PR #2010, opened 2022-06-30
Copy changes

slorek:
Implements various minor copy changes described in the tickets shown in the relevant commit messages.

Note: in order to avoid duplicating a template, there is a minor change to the slug sequence for ECP/LUP claims. This should not have any further impact beyond the slug changing on that page.

Added a feature spec which allows us to test the journey when there is no valid LUPP claim.


app/views/claims/information_provided.html.erb
```diff
@@ -13,56 +13,79 @@
 
     <ul class="govuk-list govuk-list--bullet">
       <li>the school you are currently employed to teach at</li>
-      <li>the Teaching Regulation Agency</li>
+      <li>the Department for Education's (DfE) database of qualified teachers</li>
```
nigel-lowry, line 16: That looks like a straight quote but you've been using curly quotes elsewhere

app/views/early_career_payments/claims/eligible_itt_subject.html.erb
```diff
@@ -40,24 +40,26 @@
...
-              If you qualified with science
-            </span>
-          </summary>
-          <div class="govuk-details__text">
-            <p class="govuk-body">
-              You will need to have specialised in either chemistry or physics to be eligible. If you specialised in biology then you will not be eligible. If you’re unsure, you can get this from the certificate you received when you qualified as a teacher.
-            </p>
-          </div>
-        </details>
+        <% if subject_symbols.any?{ |subject| [:chemistry, :physics].include?(subject) } %>
```
nigel-lowry, line 46: `if subject_symbols.in? [:chemistry, :physics]`

slorek: This won't work as `subject_symbols` is also an `Array`.

nigel-lowry (approved):

PR #2024, opened 2022-07-08
[CAPT-329] Include computing in school census checks

vassyz:
Unmatched:
Screenshot 2022-07-08 at 12 12 00

Matched:
Screenshot 2022-07-08 at 12 12 18

### DIFF
--- a/app/models/automated_checks/claim_verifiers/census_subjects_taught.rb
+++ b/app/models/automated_checks/claim_verifiers/census_subjects_taught.rb
@@ -62,9 +62,9 @@ module AutomatedChecks
           levelling_up_premium_payments_policy_subjects
         when StudentLoans
           student_loans_policy_subjects
+        else
+          return false
         end
-
-        return false unless match_against
 
         match_against.intersection(school_workforce_census_subjects).any?
       end
--- a/app/models/automated_checks/claim_verifiers/census_subjects_taught.rb
+++ b/app/models/automated_checks/claim_verifiers/census_subjects_taught.rb
@@ -35,7 +35,7 @@ module AutomatedChecks
       end
 
       def no_data
-        return unless school_workforce_census.nil?
+        return if school_workforce_census.present?
 
         create_task(match: nil)
       end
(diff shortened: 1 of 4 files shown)


app/models/automated_checks/claim_verifiers/census_subjects_taught.rb
```diff
@@ -58,10 +58,14 @@ def eligible?
         match_against = case claim.policy
         when EarlyCareerPayments
           early_career_payments_policy_subjects
+        when LevellingUpPremiumPayments
+          levelling_up_premium_payments_policy_subjects
         when StudentLoans
           student_loans_policy_subjects
         end
 
+        return false unless match_against
```
nigel-lowry, line 67: I'd have used an `else` in the case statement above for that

spec/models/automated_checks/claim_verifiers/census_subjects_taught_spec.rb
```diff
@@ -113,6 +124,10 @@ module ClaimVerifiers
                     expect(subject).to eq("[School Workforce Census] - Eligible:\n<pre>Subject 1: Mathematics / Mathematical Development (Early Years)\nSubject 2: Statistics\n</pre>\n")
                   end
 
+                  it "returns 'Eligible' with the school workforce census subjects", if: policy == LevellingUpPremiumPayments do
```
nigel-lowry, line 127: What's this `if` thing? I've not seen that before

nigel-lowry (approved):

app/models/automated_checks/claim_verifiers/census_subjects_taught.rb
vassyz: I have, but was trying out stuff and forgot to change it back.

vassyz: Done in c9a72dd9d6c09b35c54fd0822083b924a7b1d76c

spec/models/automated_checks/claim_verifiers/census_subjects_taught_spec.rb
vassyz: It's a conditional `if`, because these specs run in an `each` iteration. I copied the format from the other policies.

PR #2032, opened 2022-07-12
Prevent the school factory to use existing URN

vassyz:
Fixes `db:reset` `ActiveRecord::RecordNotUnique: PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_schools_on_urn"` error.

**Update:** CI runs `School.pluck(:urn)` before the database gets created. It throws: `PG::ConnectionBad: FATAL:  database "dfe-teachers-payment-service_test" does not exist`. I'll skip this approach for now and update the `urn` value to one with the same amount but closer to the end of `urn_to_award_amount_in_pounds_for_2022_to_2023`. Doing this will prevent `sequence` from getting to that number.

### DIFF
--- a/spec/factories/schools.rb
+++ b/spec/factories/schools.rb
@@ -36,9 +36,11 @@ FactoryBot.define do
 
     trait :levelling_up_premium_payments_eligible do
       # this is a huge array but if it ever cycles, there'll be a message about duplicate URNs
-      sequence :urn, LevellingUpPremiumPayments::Award.urn_to_award_amount_in_pounds(AcademicYear.new(2022))
-                                                      .keys
-                                                      .cycle
+      sequence :urn, LevellingUpPremiumPayments::Award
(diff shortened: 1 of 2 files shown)


spec/factories/schools.rb
```diff
@@ -36,7 +36,11 @@
 
     trait :levelling_up_premium_payments_eligible do
       # this is a huge array but if it ever cycles, there'll be a message about duplicate URNs
-      sequence :urn, LevellingUpPremiumPayments::Award.urn_to_award_amount_in_pounds(AcademicYear.new(2022)).keys.cycle
+      sequence :urn, LevellingUpPremiumPayments::Award
+        .urn_to_award_amount_in_pounds(AcademicYear.new(2022))
+        .keys
+        .excluding(School.pluck(:urn))
```
nigel-lowry, line 42: Brilliant

nigel-lowry (approved):

PR #2056, opened 2022-07-27
[CAPT-518] Add degree question to check your answers page

vassyz:
Screenshot 2022-07-27 at 11 23 28

### DIFF
--- a/app/helpers/claims_helper.rb
+++ b/app/helpers/claims_helper.rb
@@ -1,6 +1,6 @@
 module ClaimsHelper
-  def eligibility_answers(claim)
-    claim.policy::EligibilityAnswersPresenter.new(claim.eligibility).answers
+  def eligibility_answers(current_claim)
+    current_claim.policy::EligibilityAnswersPresenter.new(current_claim.eligible_eligibility).answers
   end
 
   def identity_answers(claim)
--- a/app/models/concerns/early_career_payments_helpable.rb
+++ b/app/models/concerns/early_career_payments_helpable.rb
@@ -36,12 +36,4 @@ module EarlyCareerPaymentsHelpable
   def trainee_teacher?
     nqt_in_academic_year_after_itt == false
   end
-
-  def qualification_name
-    return qualification.gsub("_itt", " initial teaching training") if qualification.split("_").last == "itt"
-
-    qualification_attained = qualification.humanize.downcase
-
-    qualification_attained == "assessment only" ? qualification_attained : qualification_attained + " qualification"
-  end
 end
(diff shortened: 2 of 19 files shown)


app/models/current_claim.rb
```diff
@@ -144,6 +144,12 @@ def policy_year
     end
   end
 
+  def eligible_eligibility
```
kenfodder, line 147: Wondering if this could just be `def eligibility`?

app/helpers/claims_helper.rb
```diff
@@ -1,6 +1,6 @@
 module ClaimsHelper
-  def eligibility_answers(claim)
-    claim.policy::EligibilityAnswersPresenter.new(claim.eligibility).answers
+  def eligibility_answers(current_claim)
+    current_claim.policy::EligibilityAnswersPresenter.new(current_claim.eligible_eligibility).answers
```
kenfodder, line 3: If it's just `def eligibility` then this can just stay the same.

kenfodder (approved):

app/models/concerns/early_career_payments_helpable.rb
```diff
@@ -36,4 +36,12 @@ def no_entire_term_contract?
   def trainee_teacher?
     nqt_in_academic_year_after_itt == false
   end
+
+  def qualification_name
```
nigel-lowry, line 40: We don't need to worry about this now, and my module naming wasn't great here, but I planned to delete this entire module. Currently, it's where the methods are for the implementation of the ineligible screens. That whole lot is going to be replaced when I tackle the updated ineligible screens. We could keep this module around but it'll literally only have your new method in it.

Is there anywhere else your new method could go?

app/models/current_claim.rb
nigel-lowry: @vassyz do you have to maintain an interface or something? Is that why it's got this name?

nigel-lowry (approved):

app/models/current_claim.rb
vassyz: No, it's literally an eligible eligibility (if there is one).

app/models/concerns/early_career_payments_helpable.rb
vassyz: When you guys had a go at this, you coped the method from the ECP to the LUP eligibility model: https://github.com/DFE-Digital/claim-additional-payments-for-teaching/compare/master...capt-518-degree-question-response-in-teacher-journey#diff-d5a4e998dcde35819d9253160606c2aebbedcab062cdf7e8306e51172fdf142fR77. I just looked for a module that both eligibilities have access to so I don't have duplication. Happy to move it somewhere else. Or have the duplication if you think that works better.

nigel-lowry: Would `app/helpers/early_career_payments_helper.rb` be alright?

kenfodder (approved):

PR #2074, opened 2022-08-02
[CAPT-540] DQT API checks LUP and ECP claims for eligibility - Qualifications

vassyz:


nigel-lowry (approved):

app/models/early_career_payments/dqt_record.rb
```diff
@@ -376,5 +51,9 @@ def itt_subject_group
             }&.first
       end
     end
+
+    def eligible_subject?
+      (ELIGIBLE_ITT_SUBJECTS[claim.eligibility.eligible_itt_subject.to_sym] & itt_subjects).present?
```
nigel-lowry, line 56: I'm surprised `present?` works here. I'd have thought it'd need `any?`

vassyz: Both would work. I'm checking if the intersection is not empty. `[].present?` is `false`. I can change it to `.any?` I don't mind.

nigel-lowry: No it's okay, `[].present? == false` in the rails console so that's alright (it's an error in `irb` though which is why I thought it was a problem)

vassyz: I changed it to `.any?` in bb5f7b141fea2623fc15b482c4bd3750b312f9d6
You probably got the error because `.present?` is a Rails method and `.any?` works in plain Ruby.

PR #2095, opened 2022-08-15
[CAPT-604] Update azurerm terraform provider

saliceti:
## What
Update azurerm provider to clear some tech debt and hope it will solve some small ramdom errors we've experienced

## How to review
Successful run: https://dev.azure.com/dfe-ssp/S118-Teacher-Payments-Service/_build/results?buildId=286267&view=results
Review app: https://s118d02-app-pr-2095-as.azurewebsites.net/additional-payments/landing-page

## ⚠️ Important
As the resource is replaced, it will create a small downtime the first time it is applied to an environment

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ set-azure-account:
 
 terraform-init: set-azure-account
 	$(if $(IMAGE_TAG), , $(error Missing environment variable "IMAGE_TAG"))
-	terraform -chdir=azure/terraform init -reconfigure \
+	terraform -chdir=azure/terraform init -reconfigure -upgrade \
 		-backend-config=resource_group_name=${RESOURCE_GROUP_NAME} \
 		-backend-config=storage_account_name=${STORAGE_ACCOUNT_NAME} \
 		-backend-config=container_name=${CONTAINER_NAME} \
--- a/azure/terraform/modules/app_service/app_service.tf
+++ b/azure/terraform/modules/app_service/app_service.tf
@@ -1,79 +1,15 @@
-resource "azurerm_app_service" "app_as" {
+resource "azurerm_linux_web_app" "app_as" {
   name                = local.app_service_name
   resource_group_name = var.app_rg_name
   location            = var.rg_location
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -34,7 +34,7 @@ set-azure-account:
 
 terraform-init: set-azure-account
 	$(if $(IMAGE_TAG), , $(error Missing environment variable "IMAGE_TAG"))
-	terraform -chdir=azure/terraform init -reconfigure \
+	terraform -chdir=azure/terraform init -reconfigure -upgrade \
```
welshman1970, line 37: where does it get the upgrade file?

azure/terraform/modules/app_service/app_service.tf
```diff
@@ -1,78 +1,15 @@
-resource "azurerm_app_service" "app_as" {
+resource "azurerm_linux_web_app" "app_as" {
```
welshman1970, line 1: Why has it changed from azurerm_app_service to azurerm_linux_web_app

azure/terraform/modules/app_service/app_service.tf
```diff
@@ -1,78 +1,15 @@
...
   name                = local.app_service_name
   resource_group_name = var.app_rg_name
   location            = var.rg_location
-  app_service_plan_id = data.azurerm_app_service_plan.app.id
+  service_plan_id     = data.azurerm_service_plan.app.id
```
welshman1970, line 5: why has this changed?

welshman1970 (commented):
just a few comments regarding some changes to some of the resources

azure/terraform/modules/app_service/app_service.tf
saliceti: azurerm_app_service has been deeprecated in the new provider version

azure/terraform/modules/app_service/app_service.tf
saliceti: service_plan_id has been renamed in the new provider version

slorek (approved):

PR #2099, opened 2022-08-15
CAPT-515 content changes to email passcode page

kenfodder:


app/views/claims/email_address.html.erb
```diff
@@ -16,7 +16,7 @@
 
         <%= errors_tag current_claim, :email_address %>
 
-        <%= form.email_field :email_address,
```
kenfodder, line 19: `email_field` is actually better, but browsers like chrome then apply their own inline validation and error messaging, reminders uses `text_field` so this is just to keep it consistent so validation actually hits the model giving our desired error message.

slorek (approved):

PR #2105, opened 2022-08-17
Upgrade to ruby 2.7.5 because dfe-analytics requires it

kenfodder:


kenfodder:
Going to push for 2.7.6

PR #2106, opened 2022-08-17
Capt 615 review app amendments

welshman1970:
test PR

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -104,6 +104,7 @@ stages:
               WebAppName: s118d02-app-$(app_name)-as
               ExtraTerraformVariables: -var="app_name=$(app_name)"
               ServiceConnection: azdo.pipelines.cip.S118D.armfe1ef140-8bef-4043-b5ee-c449e6f951ef
+              pullRequestNumber : $(System.PullRequest.PullRequestNumber)
 
 - stage: ProductionRelease
   condition: and(succeeded('TestRelease'), eq(variables['Build.SourceBranch'], 'refs/heads/master'))
--- a/azure/terraform/modules/app_service/app_service.tf
+++ b/azure/terraform/modules/app_service/app_service.tf
@@ -60,6 +60,7 @@ resource "azurerm_app_service" "app_as" {
     "NOTIFY_API_KEY"                                 = data.azurerm_key_vault_secret.NotifyApiKey.value
     "ORDNANCE_SURVEY_CLIENT_PARAMS"                  = data.azurerm_key_vault_secret.ordnancesurveyclientparms.value
     "ORDNANCE_SURVEY_API_BASE_URL"                   = data.azurerm_key_vault_secret.ordnancesurveyapibaseurl.value
+    "PR_NUMBER"                                      = var.pr_number
     "RAILS_ENV"                                      = "production"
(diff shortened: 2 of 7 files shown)


azure-pipelines.yml
```diff
@@ -15,6 +15,8 @@ variables:
   - group: docker-settings
   - name: app_name
     value: pr-$(System.PullRequest.PullRequestNumber)
+  - name: TF_VAR_PR_NUMBER
+    value : pr-$(System.PullRequest.PullRequestNumber)
```
saliceti, line 19: Shouldn't it be only `$(System.PullRequest.PullRequestNumber)` ?

azure-pipelines.yml
```diff
@@ -15,6 +15,8 @@ variables:
   - group: docker-settings
   - name: app_name
     value: pr-$(System.PullRequest.PullRequestNumber)
+  - name: TF_VAR_PR_NUMBER
```
saliceti, line 18: Terraform variables are usually lowercase, so here we should have `TF_VAR_pr_number`

azure/terraform/modules/app_service/app_service.tf
```diff
@@ -60,6 +60,7 @@ resource "azurerm_app_service" "app_as" {
     "NOTIFY_API_KEY"                                 = data.azurerm_key_vault_secret.NotifyApiKey.value
     "ORDNANCE_SURVEY_CLIENT_PARAMS"                  = data.azurerm_key_vault_secret.ordnancesurveyclientparms.value
     "ORDNANCE_SURVEY_API_BASE_URL"                   = data.azurerm_key_vault_secret.ordnancesurveyapibaseurl.value
+    "PR_NUMBER"                                      = var.PR_NUMBER
```
saliceti, line 63: The app environment variable is uppercase but the terraform variable should be lowercase, as in `var.pr_number`

azure/terraform/modules/app_service/variable.tf
```diff
@@ -49,6 +49,11 @@ variable "bypass_dfe_sign_in" {
   description = "Bypass DFE Sign-In authentication and use a default role"
 }
 
+variable "PR_NUMBER" {
+  type        = string
+  description = "PR number from Azure Devops"
```
saliceti, line 54: "Github pull request number provided by Azure Devops" ?

azure-pipelines.yml
welshman1970: I used capitals according to this doc  : 

"Then, in your Terraform configuration, you have to define the variables with capital letters.

Why? Because Azure DevOps will always transform pipeline variables to uppercase environment variables."

azure/terraform/modules/app_service/app_service.tf
welshman1970: I used capitals according to this doc  : 

"Then, in your Terraform configuration, you have to define the variables with capital letters.

Why? Because Azure DevOps will always transform pipeline variables to uppercase environment variables."

azure/terraform/modules/app_service/variable.tf
welshman1970: its stored in the azure devops variable $(System.PullRequest.PullRequestNumber)
 this is where i am transferring the value to terraform from

saliceti (approved):

PR #2107, opened 2022-08-17
Upgrade to ruby 2.7.6 because dfe-analytics requires it

kenfodder:


Gemfile.lock
```diff
@@ -522,7 +522,7 @@ DEPENDENCIES
   webmock
 
 RUBY VERSION
-   ruby 2.7.3p183
+   ruby 2.7.6p219
 
 BUNDLED WITH
    2.3.7
```
vassyz, line 528: Thoughts on also upgrading bundler to `2.3.20`.?

vassyz (approved):

PR #2113, opened 2022-08-22
[CAPT-646] Fix for reminder service crash

zarembas:


vassyz:
Closed as it's not the solution we're looking for.

PR #2133, opened 2022-08-31
CAPT-690 Add Review App Link to Pull Request

welshman1970:
Add Githun Action to automatically post the link to the review app when a pull request when a PR is opened.

### DIFF
--- a/.github/workflows/ga_publish_ra_link.yml
+++ b/.github/workflows/ga_publish_ra_link.yml
@@ -1,0 +1,12 @@
+on: pull_request
+
+jobs:
+  post_comment_on_pr:
+    runs-on: ubuntu-latest
+    name: Post review app url in comment on PR
+    steps:
+      - name: Post sticky pull request comment
+        uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          message: |
+            Review app deployed to <https://s118d02-app-pr-${{ github.event.pull_request.number }}-as.azurewebsites.net/additional-payments/claim>


.github/workflows/ga_publish_ra_link.yml
```diff
@@ -0,0 +1,17 @@
+on: pull_request
+
+jobs:
+  example_comment_pr:
+    runs-on: ubuntu-latest
+    name: An example job to comment a PR
+    steps:
+      - name: Checkout
```
saliceti, line 8: Is checkout really required?

.github/workflows/ga_publish_ra_link.yml
```diff
@@ -0,0 +1,17 @@
+on: pull_request
+
+jobs:
+  example_comment_pr:
```
saliceti, line 4: This is from an example, can it be more meaningful for us

.github/workflows/ga_publish_ra_link.yml
```diff
@@ -0,0 +1,17 @@
...
+    runs-on: ubuntu-latest
+    name: An example job to comment a PR
```
saliceti, line 6: This is from an example, can it be more meaningful for us

.github/workflows/ga_publish_ra_link.yml
```diff
@@ -0,0 +1,17 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Comment PR
+        uses: thollander/actions-comment-pull-request@v1
```
saliceti, line 12: This adds a comment everytime the job runs. We only really want one. Can we use a sticky comment instead as in https://github.com/DFE-Digital/teaching-vacancies/blob/main/.github/workflows/build_and_deploy.yml#L268

.github/workflows/ga_publish_ra_link.yml
welshman1970: removed

.github/workflows/ga_publish_ra_link.yml
welshman1970: ok i will look into this

saliceti (approved):

PR #2142, opened 2022-09-05
[CAPT-702] Add missing PR number to delete review app workflow

saliceti:
## What
The destroy review app workflow gets cancelled after 60min because terraform is missing an input variable


saliceti:
Closing to test destroy

PR #2170, opened 2022-09-14
Capt 608 delete review app terraform state files

welshman1970:


azure/delete_review_app.yml
```diff
@@ -6,6 +6,7 @@ pool:
   vmImage: ubuntu-latest
 
 variables:
+  - group: azure-review-app
```
saliceti, line 9: Please store the value in keyvault, not in a variable group

PR #2193, opened 2022-09-22
CAPT-727 remove house no field from address search

kenfodder:


spec/features/auto_populating_home_address_from_postcode_search_spec.rb
```diff
@@ -314,6 +314,7 @@
       fill_in :claim_address_line_1, with: "Penthouse Apartment, Millbrook Tower"
       fill_in :claim_address_line_2, with: "Windermere Avenue"
       fill_in "Town or city", with: "Southampton"
+      fill_in "County", with: "Southampton"
```
slorek, line 317:
```suggestion
      fill_in "County", with: "Hampshire"
```

app/controllers/claims_controller.rb
```diff
@@ -23,10 +23,13 @@ def show
     if params[:slug] == "postcode-search" && postcode
       redirect_to claim_path(current_policy_routing_name, "select-home-address", {"claim[postcode]": params[:claim][:postcode], "claim[address_line_1]": params[:claim][:address_line_1]}) and return unless invalid_postcode?
     elsif params[:slug] == "select-home-address" && postcode
+      session[:claim_postcode] = params[:claim][:postcode]
+      session[:claim_address_line_1] = params[:claim][:address_line_1]
```
slorek, line 27: Will this param ever be present? We might be introducing a new bug otherwise (where the user has manually entered their address). Suggest removing this line.

slorek (approved):

app/controllers/claims_controller.rb
kenfodder: Yeah it's used, this fixes a bug where pressing back the form picks a previous stale session value if the user goes forward and back more than once.

spec/features/auto_populating_home_address_from_postcode_search_spec.rb
kenfodder: Ah yeah, I'll tweak that :)

app/controllers/claims_controller.rb
```diff
@@ -23,10 +23,13 @@ def show
...
       if address_data.nil?
```
vassyz, line 28: Shouldn't we clear the postcode and address from the session here?

vassyz (approved):

app/controllers/claims_controller.rb
kenfodder: Ideally yes, but if people press back those pages pick up values from the session, this completely needs rework if we want to ditch the session issues with pressing the back button. This keeps the session in sync and avoids that issue you mentioned about having a stale value when you press back.

PR #2214, opened 2022-10-04
[CAPT-662] Update email validation regex

vassyz:
The previous validation regex `URI::MailTo::EMAIL_REGEXP` was allowing emails similar to `name@example`. This was causing the journey page to crash as Notify can't send emails to an invalid email address.


config/application.rb
```diff
@@ -63,5 +63,7 @@ class Application < Rails::Application
     }
 
     config.active_record.yaml_column_permitted_classes = [BigDecimal, Date, Symbol]
+
+    config.email_regexp = /\A[a-zA-Z0-9.!\#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)+\z/
```
slorek, line 67: Might be worth matching the regex used on the Notifications API: https://github.com/alphagov/notifications-utils/blob/ea93a8989028de2529f651a696bab13f9599a44c/notifications_utils/__init__.py#L14

vassyz: If it were Python but since I'd have to modify it, I'd rather keep this more explicit version.

PR #2219, opened 2022-10-05
Capt 23 26 claim status and filter

kenfodder:


app/controllers/admin/claims_controller.rb
```diff
@@ -4,10 +4,16 @@ class Admin::ClaimsController < Admin::BaseAdminController
   before_action :ensure_service_operator
 
   def index
-    @claims = Claim.includes(:decisions, eligibility: [:claim_school, :current_school]).awaiting_decision.order(:submitted_at)
+    @claims = Claim.approved if params[:status] == 'approved'
+    @claims = Claim.rejected if params[:status] == 'rejected'
```
vassyz, line 8: Feels like this should be an `if/else` or `||=` like below.

app/helpers/admin/claims_helper.rb
```diff
@@ -150,14 +150,25 @@ def claim_summary_heading(claim)
...
+      elsif claim.latest_decision&.rejected?
+        "Rejected"
       else
-        "Awaiting payroll"
+        "Awaiting decision"
       end
     end
 
+    def index_status_filter(status)
+      return "approved" if status == "approved"
```
vassyz, line 166: I'd prefer a `case` here, but doesn't really matter.

vassyz (approved):

app/controllers/admin/claims_controller.rb
```diff
@@ -4,10 +4,16 @@ class Admin::ClaimsController < Admin::BaseAdminController
   before_action :ensure_service_operator
 
   def index
-    @claims = Claim.includes(:decisions, eligibility: [:claim_school, :current_school]).awaiting_decision.order(:submitted_at)
+    @claims = Claim.approved if params[:status] == 'approved'
```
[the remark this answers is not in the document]
kenfodder: 🤦

app/controllers/admin/claims_controller.rb
kenfodder: It feels like it but no, the if/else looks really hard to ready

vassyz: It's just that `params[:status]` can't be both `approved` and `rejected`. If `@claims = Claim.approved` it will never become `Claim.rejected`.

PR #2229, opened 2022-10-11
CAPT-23 - filters claims to current academic year

kenfodder:


app/controllers/admin/claims_controller.rb
```diff
@@ -59,4 +59,8 @@ def filtered_team_member
     name = params[:team_member].split("-")
     DfeSignIn::User.find_by(given_name: name.shift, family_name: name).id
   end
+
+  def current_academic_year_claims
+    Claim.by_academic_year(AcademicYear.for(Date.today))
```
vassyz, line 64: `AcademicYear.for(Date.today)` is being called in a few places. worth having it as `AcademicYear.current`?

vassyz (approved):

app/controllers/admin/claims_controller.rb
kenfodder: Yeah it's used in one other place, don't want to yak shave this further.

PR #2231, opened 2022-10-12
[CAPT-751] Fix remaining hard-coded academic year

vassyz:

### DIFF
--- a/app/models/levelling_up_premium_payments/eligibility.rb
+++ b/app/models/levelling_up_premium_payments/eligibility.rb
@@ -1,13 +1,13 @@
 module LevellingUpPremiumPayments
   class Eligibility < ApplicationRecord
     include EligibilityCheckable
+    include ActiveSupport::NumberHelper
 
     self.table_name = "levelling_up_premium_payments_eligibilities"
     has_one :claim, as: :eligibility, inverse_of: :eligibility
     belongs_to :current_school, optional: true, class_name: "School"
 
-    # TODO: use first year of LUP for now but this must come from a PolicyConfiguration
-    validates :award_amount, on: :amendment, award_range: {max: LevellingUpPremiumPayments::Award.max(AcademicYear.new(2022))}
+    validate :award_amount_must_be_in_range, on: :amendment
     validates :eligible_degree_subject, on: [:"eligible-degree-subject"], inclusion: {in: [true, false], message: "Select yes if you have a degree in an eligible subject"}
 
     delegate :name, to: :current_school, prefix: true, allow_nil: true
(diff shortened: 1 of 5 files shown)


app/models/levelling_up_premium_payments/eligibility.rb
```diff
@@ -36,6 +36,20 @@ class Eligibility < ApplicationRecord
...
+    # {
+    #   <AcademicYear:0x00007f7d874293c8 @start_year=2017, @end_year=2018> => "2017/2018",
+    #   <AcademicYear:0x00007f7d87429288 @start_year=2018, @end_year=2019> => "2018/2019",
+    #   <AcademicYear:0x00007f7d87429260 @start_year=2019, @end_year=2020> => "2019/2020",
+    #   <AcademicYear:0x00007f7d87429238 @start_year=2020, @end_year=2021> => "2020/2021",
+    #   <AcademicYear:0x00007f7d87429210 @start_year=2021, @end_year=2022> => "2021/2022",
+    #   <AcademicYear:0x00007f7d87428c98 @start_year=nil, @end_year=nil> => "None"
+    # }
+    SELECTABLE_ITT_ACADEMIC_YEARS =
+      JourneySubjectEligibilityChecker.selectable_itt_years_for_claim_year(AcademicYear.for(Date.today)).each_with_object({}) do |year, hsh|
```
slorek, line 49: I think we need to get the academic year for the current `PolicyConfiguration`, rather than the current calendar academic year. E.g. `PolicyConfiguration.for(LevellingUpPremiumPayments).current_academic_year`

app/models/levelling_up_premium_payments/eligibility.rb
```diff
@@ -36,6 +36,20 @@ class Eligibility < ApplicationRecord
...
+        hsh[year] = AcademicYear::Type.new.serialize(year)
+      end.merge({AcademicYear.new => AcademicYear::Type.new.serialize(AcademicYear.new)})
```
slorek, line 51: It's not obvious what this blank instance is used for. I would expect it to be the current year but that is returned by `selectable_itt_years_for_claim_year` so may now be redundant, as 2022 wasn't in the original implementation's list.

app/views/early_career_payments/claims/itt_year.html.erb
```diff
@@ -33,8 +33,7 @@
             <% end %>
 
             <div class="govuk-radios">
-<!--              TODO this needs to get the year from the PolicyConfiguration-->
-              <% JourneySubjectEligibilityChecker.selectable_itt_years_for_claim_year(AcademicYear.new(2022)).each do |year| %>
+              <% JourneySubjectEligibilityChecker.selectable_itt_years_for_claim_year(AcademicYear.for(Date.today)).each do |year| %>
```
slorek, line 36: Same comment as above: I think we need to get the academic year for the current `PolicyConfiguration`, rather than the current calendar academic year. E.g. `PolicyConfiguration.for(LevellingUpPremiumPayments).current_academic_year`

slorek (changes requested):

app/models/levelling_up_premium_payments/eligibility.rb
vassyz: Forgot to answer that the blank instance is for the `None of the above` option.

slorek (approved):

PR #2242, opened 2022-10-17
Enable basic auth on review apps

slorek:
https://dfedigital.atlassian.net/browse/CAPT-624


app/controllers/concerns/http_auth_concern.rb
```diff
@@ -0,0 +1,12 @@
+module HttpAuthConcern
+  extend ActiveSupport::Concern
+
+  included do
+    if ENV["BASIC_AUTH_USERNAME"].present? && ENV["BASIC_AUTH_PASSWORD"].present?
```
kenfodder, line 5: @slorek Do we need to add the env vars into Terraform etc...?

slorek: This has already been done by Vito

kenfodder (approved):

PR #2247, opened 2022-10-20
CAPT-763 - catch and handle email address error from Notify

kenfodder:


app/controllers/claims_controller.rb
```diff
@@ -168,8 +168,10 @@ def prepend_view_path_for_policy
   def one_time_password
     case params[:slug]
     when "email-address"
-      ClaimMailer.email_verification(current_claim, otp.code).deliver_now if current_claim.valid?(:"email-address")
-      session[:sent_one_time_password_at] = Time.now
+      if current_claim.valid?(:"email-address")
```
slorek, line 171: Suggestion: we could use `current_claim.validate!(:"email-address")` here instead and rescue the exception. Otherwise it will fail silently if there is a validation error.

app/controllers/reminders_controller.rb
```diff
@@ -122,8 +133,10 @@ def reminder_params
   def one_time_password
     case current_slug
     when "personal-details"
-      ReminderMailer.email_verification(current_reminder, otp.code).deliver_now if current_reminder.valid?(:"email-address")
-      session[:sent_one_time_password_at] = Time.now
+      if current_reminder.valid?(:"personal-details")
```
slorek, line 136: Suggestion: raise exception if invalid and rescue to prevent silent failure.

kenfodder: @slorek Yes I noticed that option, but on L31 it will attempt to save with the same context and that will render the inline form error as it can't save, basically it duplicates running the validation with the way the code is laid out.

app/controllers/claims_controller.rb
kenfodder: @slorek Same as the other comment, I wasn't going to touch this file as it has nothing to do with the reminders, but tidying up, really should be another PR to address the OTP in the same way for this controller to catch the Notify exception.

slorek (approved):

PR #2287, opened 2022-11-24
Allow admins to manage LUPP school award amounts

slorek:
https://dfedigital.atlassian.net/browse/CAPT-754

The last uploaded status for the currently configured academic year is displayed on the relevant manage service page:

Screenshot 2022-11-25 at 18 28 59

If no data is present, there will instead be a warning message:

Screenshot 2022-11-25 at 18 31 15

Admin users can download the current list for any academic year where data has been uploaded in CSV format, and upload award data in the same CSV format:

Screenshot 2022-11-25 at 18 30 26

The CSV file requires two columns: `school_urn` and `award_amount`. A CSV file containing any other data, or with missing headers, will be rejected. It should contain data for schools which are eligible only, but will skip rows where the `award_amount` is `0`.

Error messages show the affected line numbers to aid correction:

Screenshot 2022-11-25 at 18 32 56

When a CSV file is uploaded, it will _replace_ all of the data for that academic year currently in the service. A confirmation message will be displayed:

Screenshot 2022-11-25 at 18 34 16

Note: the list for 2022 which is in the code is needed for the initial data migration and to ensure the change can be deployed with zero downtime. Once deployed it will be safe to remove this code.

### DIFF
--- a/app/models/dfe_sign_in/user_data_importer.rb
+++ b/app/models/dfe_sign_in/user_data_importer.rb
@@ -32,7 +32,7 @@ module DfeSignIn
     end
 
     def users_no_longer_present
-      DfeSignIn::User.all.map(&:dfe_sign_in_id) - dfe_sign_in_users.map(&:user_id)
+      (DfeSignIn::User.all.map(&:dfe_sign_in_id) - dfe_sign_in_users.map(&:user_id)).compact
     end
   end
 end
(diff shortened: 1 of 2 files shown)


app/controllers/admin/levelling_up_premium_payments_awards_controller.rb
```diff
@@ -0,0 +1,37 @@
+module Admin
+  class LevellingUpPremiumPaymentsAwardsController < BaseAdminController
+    helper_method :policy_configuration
+
+    def index
+      response.headers["Content-Type"] = "text/csv"
+      response.headers["Content-Disposition"] = "attachment; filename=awards_#{academic_year}.csv"
+
+      send_data LevellingUpPremiumPayments::Award.csv_for_academic_year(academic_year), filename: "awards_#{academic_year}.csv"
```
kenfodder, line 9: Super minor, wondering if those headers could be passed into send_data as parameters?

kenfodder (approved):

PR #2360, opened 2023-02-17
minimum file changes to add a policy to pay every teacher 10 pounds

mattonfoot:
This is a test branch designed to see how quickly it is to reuse existing components to create a very minimal policy to give every teacher £10.00

I have successfully made a claim checked the answers and approved it on my local machine. Have I missed anything?


kenfodder:
@mattonfoot yep looks like you got the general idea to fit a new policy, let me know if there is anything further you want to discuss, you did mention school selector I think.

PR #2393, opened 2023-05-09
CAPT-948 flash message shows chosen AY not current AY

kenfodder:
In addition:
- For some reason the spec file had `_spec` missing so file renamed.


your (approved):
Needs rebase - codewise looks good to me 👍.

your (approved):

PR #2395, opened 2023-05-12
CAPT-971 - Add a confirmation button when users try to unallocate claims

your:
Refer to the ticket and commit messages


kenfodder (approved):
All looks good, minor point could add a test to check the cancel link goes to the right place.

PR #2401, opened 2023-06-13
CAPT-918 Change slug sequence order for TSLR

your:
https://dfedigital.atlassian.net/browse/CAPT-918


mattonfoot (approved):
LGTM, One step closer

PR #2405, opened 2023-07-12
Add issuer value to DfE Sign-In omniauth config

your:
This is now required due to an update to the DfE Sign-In OIDC dependency.

Fixed by other teams, see: https://github.com/DFE-Digital/access-your-teaching-qualifications/pull/232 and https://github.com/DFE-Digital/apply-for-teacher-training/pull/8385


plisovin (approved):
LGTM

PR #2410, opened 2023-08-02
CAPT-1059 CAPT-1102 Review list of eligible LUP ITT subjects and codes

your:
For simplicity, this PR addresses both https://dfedigital.atlassian.net/browse/CAPT-1059 and https://dfedigital.atlassian.net/browse/CAPT-1102

These are minor changes, and there are no associated tests that need to be amended / are worth adding.


martin-dfe (approved):
It looks good to me 👍

PR #2418, opened 2023-08-12
CAPT-1172 Remove TRN and qualification questions if user is logged in with DfE Identity

martin-dfe:
[CAPT-1172
](https://dfedigital.atlassian.net/browse/CAPT-1172)


your (commented):
A few comments / suggestions

PR #2421, opened 2023-08-16
CAPT-1121 Amend exit screen for users who select 'none of the above' for their ITT year - LUP schools

martin-dfe:
[CAPT-1121](https://dfedigital.atlassian.net.mcas.ms/browse/CAPT-1121)

[CAPT-1121]: https://dfedigital.atlassian.net/browse/CAPT-1121?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/early_career_payments/claims/_ineligibility_generic.html.erb
+++ b/app/views/early_career_payments/claims/_ineligibility_generic.html.erb
@@ -1,5 +1,9 @@
 <p class="govuk-body">
-  You are not eligible for any additional payments.
+  You are not eligible for an
+  <%= link_to("early-career payments", EarlyCareerPayments.eligibility_page_url, class: "govuk-link") %>
+  or a
+  <%= link_to("levelling up premium payments", LevellingUpPremiumPayments.eligibility_page_url, class: "govuk-link") %>.
+  because of the year you studied.  
 </p>
 
 <p class="govuk-body">


app/views/early_career_payments/claims/_ineligibility_generic.html.erb
```diff
@@ -1,5 +1,9 @@
 <p class="govuk-body">
-  You are not eligible for any additional payments.
+  You are not eligible for an
+  <%= link_to("early-career payments", EarlyCareerPayments.eligibility_criteria_url, class: "govuk-link") %>
```
your, line 3: Looking at the ticker more closely, I believe we need to use `eligibility_page_url` rather than `eligibility_criteria_url` here.

app/views/early_career_payments/claims/_ineligibility_generic.html.erb
```diff
@@ -1,5 +1,9 @@
...
+  or a
+  <%= link_to("levelling up premium payments", LevellingUpPremiumPayments.eligibility_criteria_url, class: "govuk-link") %>.
```
your, line 5: Looking at the ticker more closely, I believe we need to use `eligibility_page_url` rather than `eligibility_criteria_url` here.

your (changes requested):
Minor comment on the actual link that needs to be used. Looking at the ticket I think we need to use the main url for each policy in the paragraph that is being changed, and not those with the anchor to the criteria.

On a side note, I believe the `eligibility_criteria_url` for ECP and LUP do not work correctly (the anchor links seem wrong, maybe something to check with Laura)

your (approved):

PR #2423, opened 2023-08-16
CAPT-931 Consume induction data as part of DQT API return for ECP only

your:
Broken down from: https://dfedigital.atlassian.net.mcas.ms/browse/CAPT-931

To be used by: https://dfedigital.atlassian.net.mcas.ms/browse/CAPT-1093 (PR: #2425, not ready for final review)

The wider context is that we're adding a new question to the ECP/LUP journey, which is being worked on as part of https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2420 (Prototype, not ready for review)

### DIFF
--- a/app/models/early_career_payments/dqt_record.rb
+++ b/app/models/early_career_payments/dqt_record.rb
@@ -10,6 +10,9 @@ module EarlyCareerPayments
       :itt_start_date,
       :degree_codes,
       :qualification_name,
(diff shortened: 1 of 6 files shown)


martin-dfe (approved):

martin-dfe (approved):
Looks Good

PR #2427, opened 2023-08-21
CAPT-1073 Upload the 2022 (latest) School Workforce Census onto the Admin site + new columns for automation

martin-dfe:
[CAPT-1073
](https://dfedigital.atlassian.net.mcas.ms/browse/CAPT-1073)


app/models/school_workforce_census.rb
```diff
@@ -2,21 +2,7 @@ class SchoolWorkforceCensus < ApplicationRecord
...
-    :subject_7,
-    :subject_8,
-    :subject_9,
-    :subject_10,
-    :subject_11,
-    :subject_12,
-    :subject_13,
-    :subject_14,
-    :subject_15
+    :subject_description_sfr
```
your, line 5: ❓ Presumably, this `SUBJECT_ATTRIBUTES` is now semantically useless. Also, am I right in saying that the `subjects` method will need to be changed / removed? 

I'd imagine we'd want to retrieve all subjects for the selected teacher (found by URN), so we'd need a relationship or a join with this table by URN and retrieve all subjects from the returned rows?

app/models/school_workforce_census.rb
```diff
@@ -115,7 +97,7 @@ class SchoolWorkforceCensus < ApplicationRecord
     ]
   }.freeze
 
-  def subjects
-    SUBJECT_ATTRIBUTES.map { |attr| send(attr) }.reject(&:blank?)
+  def subjects(trn)
```
your, line 100: This would make sense if it was a class method like `subjects_by_trn`; or simply try the other suggestion above and get rid of this method completely

spec/models/automated_checks/claim_verifiers/census_subjects_taught_spec.rb
```diff
@@ -179,15 +179,15 @@ module ClaimVerifiers
                   subject(:body) { note.body }
 
                   it "returns 'Eligible' with the school workforce census subjects", if: policy == EarlyCareerPayments do
-                    expect(subject).to eq("[School Workforce Census] - Eligible:\n<pre>Subject 1: Mathematics / Mathematical Development (Early Years)\nSubject 2: Statistics\n</pre>\n")
```
your, line 182: Looking at this test and AC n.6, shouldn't we keep the multiple eligible subjects in this and the following expectations? Basically test that we get multiple rows and subjects that are eligible for one TRN

your (commented):
Looking good, a few more tests might need to change though (see comment)

db/migrate/20230829115907_update_school_workforce_censuses.rb
```diff
@@ -0,0 +1,42 @@
...
+      }
+    )
+
+    (1..15).each do |n|
+      add_column_if_not_exist(
+        :school_workforce_censuses,
+        "subject_#{n}".to_sym,
+        :string
+      )
+    end
```
your, line 21: Shouldn't these columns be removed? I can see they are being removed in the schema but added here instead? 🤔

your: When I run `rake db:schema:load` all looks good, but `rake db:migrate` doesn't produce the same effect.

PR #2428, opened 2023-08-24
CAPT-1055 Add test scenarios to personal data scrubber spec

your:
These extra test scenarios are being added as part of the spike at https://dfedigital.atlassian.net/browse/CAPT-1055

At least one of them covers claims that are currently held in prod until the next window. Technically, these are claims without a decision, so the scrubber excludes them, and that test scenario exists already, but thought I'd make it more explicit.


martin-dfe:
Looks good to me:+1:

PR #2436, opened 2023-09-11
[Capt 1200] create sign in or continue page

sam-harris-dev:
[CAPT-1200](https://dfedigital.atlassian.net/browse/CAPT-1200)


martin-dfe (approved):
Looks good to me

PR #2449, opened 2023-09-22
Suggest current school in combined journey when signed in

sam-harris-dev:
CAPT-1178 Which school do you teach at

* correct-school slug if TID logged in and recent school from TPS
* allows ineligibile school page to change school to flow back to current-school page


app/controllers/claims_controller.rb
```diff
@@ -297,4 +304,19 @@ def save_details_check
...
+    if (school = current_claim.recent_tps_school)
+      session[:tps_school_id] = school.id
+      session[:tps_school_name] = school.name
+      session[:tps_school_address] = school.address
+    end
+  end
+
+  def check_correct_school_params
+    updated_claim_params = claim_params
+    CorrectSchoolFormChecker.call(updated_claim_params, change_school: params[:change_school])
```
mark-pkgflow, line 318: Does this mutate the `updated_claim_params`?

spec/features/correct_school_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+    expect(page).to have_text(eligible_school.name)
+    expect(page).not_to have_text("Enter the school name or postcode. Use at least three characters.")
+
+    # - Select the suggested school
+    choose(eligible_school.name)
+    click_on "Continue"
+
+    expect(page).to have_text(I18n.t("early_career_payments.questions.nqt_in_academic_year_after_itt.heading"))
+
+    Claim.order(created_at: :desc).limit(2).each do |c|
```
mark-pkgflow, line 35: Not so much a PR review question, but a gap in my knowledge. Why do we expect 2 claims here?

mark-pkgflow (approved):
LGTM

app/controllers/claims_controller.rb
kenfodder: I was dubious as well...

```
irb(main):005:0> p = {}
=> {}
irb(main):006:0> CorrectSchoolFormChecker.call(p, change_school: "true")
=> {:current_school_id=>nil, :school_somewhere_else=>true}
irb(main):007:0> p
=> {:eligibility_attributes=>{:current_school_id=>nil, :school_somewhere_else=>true}}
```

Also...

`spec/forms/correct_school_form_checker_spec.rb`

Proves it does mutate it.

Thoughts?

spec/features/correct_school_spec.rb
kenfodder: Let's chat about that.

kenfodder: @mark-pkgflow 

https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/0a884dcfaa920e7ff28f669ada0381ba10c71165/app/controllers/concerns/part_of_claim_journey.rb#L57-L68

app/controllers/claims_controller.rb
mark-pkgflow: I saw the test too, but maybe the fact that we both questioned "does that work like that" it could be refactored a little for readability, maybe. Either way, it is not blocking.

app/controllers/claims_controller.rb
```diff
@@ -297,4 +304,18 @@ def save_details_check
...
+  def update_session_with_tps_school
+    if (school = current_claim.recent_tps_school)
+      session[:tps_school_id] = school.id
+      session[:tps_school_name] = school.name
+      session[:tps_school_address] = school.address
+    end
+  end
+
+  def check_correct_school_params
+    updated_claim_params = CorrectSchoolForm.extract_params(claim_params, change_school: params[:change_school])
```
mark-pkgflow, line 317: Maybe the method should be `extract_attributes` instead of `extract_params`.

And we assign to `updated_claim_attributes`, or `updated_attributes`, `claim_attributes`. I think any of these three names would work.

WDYT?

kenfodder: We can always do this in a separate PR.

PR #2451, opened 2023-09-25
Improve performance of school search

slorek:
The school search endpoint is invoked asynchronously every time a user types a character in the school search field. This could result in a lot of requests.

This change improves the performance of the school search by approximately 10x (on my machine):

Before:

```
> Benchmark.measure { 500.times { School.search("Pen").count }}
 => #<Benchmark::Tms:0x000000010c5e2128 @cstime=0.0, @cutime=0.0, @label="", @real=27.749531999696046, @stime=0.13657800000000003, @total=0.723946, @utime=0.587368>
```

After:

```
> Benchmark.measure { 500.times { School.search("Pen").count }}
 => #<Benchmark::Tms:0x000000010aab6d68 @cstime=0.0, @cutime=0.0, @label="", @real=2.881784999743104, @stime=0.128483, @total=0.6241749999999999, @utime=0.49569199999999997>
```

This was due to the lack of any index on the `name` or `postcode` columns and the use of `REPLACE()` which was forcing a full table scan.

### DIFF
--- a/config/analytics_blocklist.yml
+++ b/config/analytics_blocklist.yml
@@ -1,5 +1,5 @@
 ---
-shared:
+:shared:
   :delayed_jobs:
   - id
   - priority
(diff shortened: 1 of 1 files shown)


db/migrate/20230925095916_add_indexes_to_schools.rb
```diff
@@ -0,0 +1,16 @@
+class AddIndexesToSchools < ActiveRecord::Migration[7.0]
+  def up
+    add_column :schools, :postcode_sanitised, :string
+
+    execute("UPDATE schools SET postcode_sanitised=REPLACE(postcode, ' ', '')")
```
your, line 5: Nitpick: Ideally, this should (❔) be a data migration

your (approved):
Awesome 🥇

db/migrate/20230925095916_add_indexes_to_schools.rb
your: 👍

db/migrate/20230925095916_add_indexes_to_schools.rb
```diff
@@ -0,0 +1,16 @@
+class AddIndexesToSchools < ActiveRecord::Migration[7.0]
+  def up
+    add_column :schools, :postcode_sanitised, :string
```
your, line 3: This may need to be added to [the dfe analytics config](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/e6570fb748f6cc5f23a5e647e1a9caaa0b9849a3/config/analytics.yml#L223-L225)

PR #2468, opened 2023-10-04
CAPT-1175 Amend payment confirmation emails (Currently says Friday, days change depending on payment date)

your:
https://dfedigital.atlassian.net/browse/CAPT-1175

This is a consequence of #2407. In short, we would previously always send a payment on the last Friday of the month. With batching, this is not the case anymore, and the scheduled payment date for each claim is saved as part of the confirmation statement upload. Previous calculation: https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2407/files#diff-998b01cd74114e4a520c747ef35924646b7983f441ecd143df06be0cb4551ce3L41-L45


mark-pkgflow (approved):
LGTM

PR #2476, opened 2023-10-10
Sign in with DfE Identity

slorek:
Implements the front end journey changes required for sign in via DfE Identity.


kenfodder (approved):

your (approved):
Impossible to review in detail – skimmed through and couldn't see anything obvious. I'm hoping that the existing journey is not affected in any way i.e. it's been tested manually and rigorously as well, in which case 👍 LGTM 🚀

PR #2482, opened 2023-10-23
Capt-1204 - Create new screen - Playback email address from Teacher ID

mark-pkgflow:
CAPT-1204 - Create new screen - Playback email address from Teacher ID

- Check for signed-in with teacher id and `email` from teacher id 
  - If either or both are false or missing go straight to the old email verification flow 
  - If both are true, show the select email page and pre-populate the `email` from teacher id
    - If email is confirmed set
      - the `email_address` on the claim,
      - `email_verified` to true and
      - `email_address_check` to true.
      - go to the next question
    - If email is not confirmed set
      - the `email_address` on the claim to `nil`,
      - `email_verified` to`nil`,
      - `email_address_check` to `false` and
      - go to the old email verification flow

### DIFF
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -29,6 +29,8 @@ class ClaimsController < BasePublicController
       return redirect_to claim_path(current_policy_routing_name, "eligible-itt-subject")
     elsif params[:slug] == "sign-in-or-continue"
       update_session_with_current_slug
+    elsif params[:slug] == "select-email"
+      session[:email_address] = current_claim.teacher_id_user_info["email"]
     elsif params[:slug] == "correct-school"
       update_session_with_tps_school
     elsif params[:slug] == "nqt-in-academic-year-after-itt" && page_sequence.in_sequence?("correct-school")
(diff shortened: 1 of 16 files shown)


app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -2,7 +2,19 @@ class OmniauthCallbacksController < ApplicationController
   def callback
     auth = request.env["omniauth.auth"]
 
-    session[:user_info] = auth.extra.raw_info
+    session[:user_info] = if Rails.env.development?
```
your, line 5: Maybe one for another day, but is this behaviour documented somewhere e.g. in the README?

your (commented):
Codewise lgtm, but will tests be added at a second stage? Conscious this PR is not testing the transition between new and old screens, or smaller things like the new email validation on the `UserInfo` object.

app/controllers/omniauth_callbacks_controller.rb
mark-pkgflow: Sorry, that was meant to stay local. I will pull it back out

mark-pkgflow: Also, RE, the tests, I was making more changes now. Let me push them up too.

your: 👍

slorek (approved):
Tested on the review app, code looks good.

PR #2487, opened 2023-10-26
CAPT-1205-playback-phone-number-from-tid

mark-pkgflow:
[CAPT-1205](https://dfedigital.atlassian.net/browse/CAPT-1205?atlOrigin=eyJpIjoiMDAyMmVhNDk4Mjc2NDIxN2IyNDYwZmEzYTRjMWM4OTkiLCJwIjoiaiJ9) Create new screen - Playback mobile number from Teacher ID

- Check for signed-in with teacher id and phone number from teacher id
  - If either or both are `false` or missing go straight to the old mobile verification flow
  - If both are `true`, show the select mobile page and pre-populate the `phone_number` from teacher id
  - If mobile is confirmed set
    - the `mobile_number` on the claim,
    - `provide_mobile_number` to `true` and
    - `mobile_check` to "use".
    - go to the next question (bank or building society)
  - If mobile is not confirmed set
    - the `mobile_number` on the claim to nil,
    - `provide_mobile_number` to `nil`,
    - `mobile_check` to "alternative" and
     - go to the old mobile verification flow
 - if declined to provide a mobile
    - the `mobile_number` on the claim to nil,
    - `provide_mobile_number` to `false`,
    - `mobile_check` to "declined" and
    - go to the next question (bank or building society)

[CAPT-1205]: https://dfedigital.atlassian.net/browse/CAPT-1205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/forms/select_mobile_number_form.rb
+++ b/app/forms/select_mobile_number_form.rb
@@ -1,0 +1,33 @@
+class SelectMobileNumberForm
+  def self.extract_attributes(claim, mobile_check:)
+    new(claim, mobile_check).extract_attributes
+  end
+
+  def initialize(claim, mobile_check)
+    @claim = claim
+    @mobile_check = mobile_check
+  end
+
+  def extract_attributes
+    case @mobile_check
+    when "use"
+      {
+        mobile_number: @claim.teacher_id_user_info["phone_number"],
+        provide_mobile_number: true,
+        mobile_check: @mobile_check
+      }
+    when "alternative"
+      {
+        mobile_number: nil,
+        provide_mobile_number: nil,
+        mobile_check: @mobile_check
+      }
+    when "declined"
+      {
+        mobile_number: nil,
+        provide_mobile_number: false,
+        mobile_check: @mobile_check
+      }
+    end
+  end
+end
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -39,7 +39,8 @@ class Claim < ApplicationRecord
     :one_time_password,
     :logged_in_with_tid,
     :details_check,
-    :email_address_check
+    :email_address_check,
+    :mobile_check
   ].freeze
   AMENDABLE_ATTRIBUTES = %i[
     teacher_reference_number
@@ -107,6 +108,7 @@ class Claim < ApplicationRecord
     teacher_id_user_info: false,
     details_check: true,
     email_address_check: true,
+    mobile_check: true,
     qa_required: false,
     qa_completed_at: false
   }.freeze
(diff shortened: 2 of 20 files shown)


spec/forms/select_mobile_number_form_spec.rb
```diff
@@ -0,0 +1,43 @@
+require "rails_helper"
+
+RSpec.describe SelectMobileNumberForm do
+  let(:claim) { double("Claim", teacher_id_user_info: {"phone_number" => "123-456-7890"}) }
```
your, line 4: Could we use `instance_double` to ensure this test breaks if `teacher_id_user_info` is ever removed? I would use pure doubles when mocking external interfaces. As an alternative, you could use FactoryBot's `build_stubbed`

app/models/claim.rb
```diff
@@ -208,6 +210,7 @@ class Claim < ApplicationRecord
 
   validates :details_check, on: [:"teacher-detail"], inclusion: {in: [true, false], message: "Select an option to whether the details are correct or not"}
   validates :email_address_check, on: [:"select-email"], inclusion: {in: [true, false], message: "Select an option to indicate whether the email is correct or not"}
+  validates :mobile_check, on: [:"select-mobile"], presence: {message: "Select an option to indicate whether the mobile number is correct or not"}
```
your, line 213: Is it also worth adding validation for `mobile_check` allowed values (`use`, `alternative`, `declined`) and a test for it

your (commented):
Overall great work, just a couple of questions

app/models/claim.rb
mark-pkgflow: Done, thanks!

db/schema.rb
```diff
@@ -86,12 +86,13 @@
     t.boolean "held", default: false
     t.boolean "hmrc_bank_validation_succeeded", default: false
     t.json "hmrc_bank_validation_responses", default: []
-    t.boolean "logged_in_with_tid", default: false
+    t.boolean "logged_in_with_tid"
```
kenfodder, line 89: Why has this changed?

app/forms/select_mobile_number_form.rb
```diff
@@ -0,0 +1,32 @@
...
+        provide_mobile_number: nil,
+        mobile_check: @mobile_check
+      }
+    elsif @mobile_check == "declined"
+      {
+        mobile_number: nil,
+        provide_mobile_number: false,
+        mobile_check: @mobile_check
+      }
+    end
```
kenfodder, line 30: Since all 3 ifs are for `@mobile_check` case-statement would work well here I think.

db/schema.rb
mark-pkgflow: Noise from hopping between branches. I will tidy.

app/forms/select_mobile_number_form.rb
mark-pkgflow: Nice catch. Done 👍

spec/forms/select_mobile_number_form_spec.rb
mark-pkgflow: Good idea. `instance_double` is used now instead 👍

kenfodder (approved):

your (approved):

PR #2492, opened 2023-11-07
CAPT-1226 New rejection filter for failing induction eligibility (ECP only)

your:
https://dfedigital.atlassian.net/browse/CAPT-1226

This PR adds a new rejected reason that, unlike all the other reasons, can be selected only when rejecting ECP claims.

The related Notify template is being updated when this is deployed; please see the Jira ticket comments section for more info.


app/models/decision.rb
```diff
@@ -33,21 +35,29 @@ class Decision < ApplicationRecord
     rejected: 1
   }
 
+  def self.rejected_reasons_for(policy, prefix: nil)
```
slorek, line 38: `prefix` seems to be presentational so I suggest moving the adding of the prefix back to https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2492/files#diff-97f8a78688ed92b6ad6239a753880d1242ab506e93381578f7342d5c11f39decR92

slorek (approved):

app/models/decision.rb
your: @slorek good point, addressed with [928efc8](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2492/commits/928efc851b86475a126925a70795a1f0720e0783)

PR #2494, opened 2023-11-09
Add specs for removal of TRN page

slorek:
https://dfedigital.atlassian.net/browse/CAPT-1387

While the functionality had been implemented, the existing spec was not enabled and also broken.


mark-pkgflow (approved):
LGTM

PR #2496, opened 2023-11-12
CAPT-1385 - Personal details page with TID

kenfodder:
* Skip page if all details are valid from TID
* Conditionally show personal details page with specific questions if anything is missing from TID
* Change first/surname valid lengh to 1..100 inline with TID validation

### DIFF
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -168,11 +168,11 @@ class Claim < ApplicationRecord
 
   validates :payroll_gender, on: [:gender, :submit], presence: {message: "Select the gender recorded on your school’s payroll system or select whether you do not know"}
 
-  validates :first_name, on: [:"personal-details", :submit], presence: {message: "Enter your first name"}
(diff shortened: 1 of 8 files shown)


app/models/claim.rb
```diff
@@ -554,6 +554,54 @@ def recent_tps_school
...
+    !claim_dup.errors.include?(:date_of_birth)
+  end
+
+  # dup - because we don't want to pollute the claim.errors by calling this method
+  # Check errors hash for key because we don't care about the non-context validation errors
+  def has_valid_nino?
+    claim_dup = dup
+    claim_dup.valid?(:"personal-details-nino")
+    !claim_dup.errors.include?(:national_insurance_number)
+  end
```
mark-pkgflow, line 603: nit: maybe we could extract this out into another model/service

mark-pkgflow (approved):

app/models/claim.rb
kenfodder: Good point, didn't think to split it out as this thing is already massive!

PR #2520, opened 2023-11-29
CAPT-1407 - fix change school link

kenfodder:
If the school playback (from TPS) was suggested and the user clicked on the suggested school the change school link goes back to the playback, all other scenarios it goes to the current-school where the user types in the school name.

### DIFF
--- a/app/models/early_career_payments/eligibility_answers_presenter.rb
+++ b/app/models/early_career_payments/eligibility_answers_presenter.rb
@@ -51,7 +51,7 @@ module EarlyCareerPayments
       [
         translate("early_career_payments.questions.current_school_search"),
         eligibility.current_school_name,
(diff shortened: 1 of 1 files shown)


app/models/early_career_payments/eligibility_answers_presenter.rb
mark-pkgflow, line 1: Nice 👍

your (approved):
Is it worth adding a test to cover this scenario? Although it's a bit of an edge case. LGTM otherwise

kenfodder:
> Is it worth adding a test to cover this scenario? Although it's a bit of an edge case. LGTM otherwise

Nah just wasn't worth it, also just testing the presenter seemed pointless.

PR #2528, opened 2023-12-07
CAPT-1357 TSLR claim school playback

kenfodder:


app/models/teachers_pensions_service.rb
```diff
@@ -30,6 +31,36 @@ def self.recent_tps_school(claim)
...
+  end
+
+  def self.has_tps_school_for_student_loan_in_previous_financial_year?(claim)
+    tps_school_for_student_loan_in_previous_financial_year(claim).present?
+  end
+
+  def self.tps_school_for_student_loan_in_previous_financial_year(claim)
+    previous_academic_year = PolicyConfiguration.for(StudentLoans).current_academic_year - 1
+    start_of_previous_financial_year = Date.new(previous_academic_year.start_year, 4, 6)
+    end_of_previous_financial_year = Date.new(PolicyConfiguration.for(StudentLoans).current_academic_year.start_year, 4, 5)
```
slorek, line 44: Suggestion: it would be useful to have these functions on the `AcademicYear` model.

app/models/teachers_pensions_service.rb
```diff
@@ -30,6 +31,36 @@ def self.recent_tps_school(claim)
...
     return nil unless tps_record&.school_urn
 
+    school_for_tps_record(tps_record)
+  end
+
+  def self.has_tps_school_for_student_loan_in_previous_financial_year?(claim)
+    tps_school_for_student_loan_in_previous_financial_year(claim).present?
+  end
+
+  def self.tps_school_for_student_loan_in_previous_financial_year(claim)
```
slorek, line 41: Suggestion: I think the only property of the `claim` we use in this method is the `teacher_reference_number` so we probably only need to supply this instead of the whole claim.

slorek (approved):

app/models/teachers_pensions_service.rb
kenfodder: Stretch goal for another PR. Pressed for time.

app/models/teachers_pensions_service.rb
kenfodder: I think I need to do this for the most of the functions. Again going to do it in another PR, as I have another PR based off this one in flux. But yeah I had considered as well, it's a good point.

your (approved):

PR #2530, opened 2023-12-07
CAPT-1177: Automate employment check

slorek:
https://dfedigital.atlassian.net/browse/CAPT-1177

This change implements:

* Automatically perform employment check on all claims on submission, using pre-loaded TPS data.
* Re-run failed checks only on claims submitted in the past 3 months when TPS data is uploaded in the admin

When the feature branch is deployed there will be claims awaiting decision for which the employment check was never performed, or was performed manually but not passed. The next time TPS data is uploaded these tasks will be deleted and re-run automatically.


app/controllers/admin/tps_data_uploads_controller.rb
```diff
@@ -15,7 +15,7 @@ def create
         if @tps_data_importer.errors.any?
           render :new and return
         end
-        EmploymentCheckJob.perform_later
+        EmploymentCheckJob.perform_later(AutomatedChecks::ClaimVerifiers::Employment)
```
your, line 18: Can I just check – are we expecting to invoke this job with different verifiers in the future? Asking because the anatomy of `EmploymentCheckJob` still seems very "employment"-specific

Incidentally, I'm introducing another similar job here https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2529/files#diff-9df32c9e4877c1c6582a206eee60cb9b4377bc1fd071f09d334acb8911a6c4f3

slorek: This change was mostly to allow me to introduce a test double (dependency injection). I understand it appears a little strange. What do you think?

your: Hey Steve, I see – I'd probably do it this way instead https://github.com/DFE-Digital/claim-additional-payments-for-teaching/commit/1d24919b343b0064288021d1475c376938a1f979 unless you reckon it's cleaner the dependency injection route

your (approved):

PR #2531, opened 2023-12-07
Fix CSRF error

slorek:
https://dfedigital.atlassian.net/browse/CAPT-1414

On review apps we were seeing some HTTP auth errors on the Omniauth callback controller, so I have removed it from that controller only.

I have also configured the cookie domain to match the callback URL which was suggested by various people when researching the problem. On its own it did not seem to fix the problem, but I think it is better to have it than not.

I've also removed some old/redundant Omniauth configuration which are either deprecated or the same as default..

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,4 @@
 class ApplicationController < ActionController::Base
-  include HttpAuthConcern
-
   TIMEOUT_WARNING_LENGTH_IN_MINUTES = 2
 
   helper_method :timeout_warning_in_minutes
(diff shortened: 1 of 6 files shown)


app/controllers/application_controller.rb
```diff
@@ -1,6 +1,4 @@
 class ApplicationController < ActionController::Base
-  include HttpAuthConcern
```
your, line 2: Am I right in saying that this will expose the admin, where we also use the sign-in bypass functionality? Could we include the concern in the base admin controller? I am not clear how that would affect the TID omniauth callbacks controller

PR #2537, opened 2023-12-13
Capt 1388 tslr remove trn page

kenfodder:


slorek (approved):
I think we would want to combine the feature specs for the TSLR journey in order to reduce duplication of expensive tests. We can do that in a follow-up tech debt exercise as there are a lot of other examples.

kenfodder:
> I think we would want to combine the feature specs for the TSLR journey in order to reduce duplication of expensive tests. We can do that in a follow-up tech debt exercise as there are a lot of other examples.

@slorek I agree, needs thought though because of permutations once the target page is reached. I'm not sure I have a solution at the moment.

PR #2555, opened 2024-01-12
Implement policy end year for Student Loans claims

slorek:
https://dfedigital.atlassian.net/browse/CAPT-1420

### DIFF
--- a/app/models/student_loans.rb
+++ b/app/models/student_loans.rb
@@ -11,6 +11,7 @@ module StudentLoans
   extend self
 
   POLICY_START_YEAR = AcademicYear.new(2013).freeze
+  POLICY_END_YEAR = AcademicYear.new(2020).freeze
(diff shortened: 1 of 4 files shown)


app/models/student_loans.rb
```diff
@@ -81,6 +82,10 @@ def first_eligible_qts_award_year(claim_year = nil)
     ].max
   end
 
+  def last_eligible_qts_award_year
+    POLICY_END_YEAR
+  end
```
your, line 87: Nice, this method could also be used to generate [this](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/71dcdb5cc27de088b77519a20b6090f4a3778def/config/locales/en.yml#L276) text but probably not worth the effort.

your (approved):
🎖️

PR #2559, opened 2024-01-16
Switch on and schedule entity table check job

ericaporter:
**Context**
EntityTableCheck is a job which sends a row_count, a checksum and a timestamp for each entity table sent to Big Query and it’s purpose is to serve as a tool for confirming production database tables match Big Query tables and then help to debug where they don’t match.

**Changes proposed in this pull request**
This PR enables the job and schedules it to run daily at 00:30

**Guidance to review**
As I am not familiar with Claim - please check the scheduling looks correct and advise if there is a better time for the job to run. It might be a good idea to test this on a review app or staging if possible - I can help with this

N.B. 
Wait until Claims is on the latest version of dfe-analytics before merging i.e. 1.11.6 (separate PR for this)


slorek (approved):

spec/jobs/entity_table_check_job_spec.rb
```diff
@@ -0,0 +1,11 @@
+require "rails_helper"
+
+RSpec.describe "EntityTableCheckJob" do
+  describe "#perform" do
+    it "runs the entity table check job" do
+      expect_any_instance_of(DfE::Analytics::EntityTableCheckJob).to receive(:perform)
+
+      DfE::Analytics::EntityTableCheckJob.new.perform
```
slorek, line 8: This is calling the wrong class. It may be useful to make the naming more distinct.

slorek (changes requested):
Spec needs fixing.

app/jobs/dfe_analytics/entity_table_check_job.rb
```diff
@@ -1,4 +1,4 @@
-class DfE::Analytics::EntityTableCheckJob < CronJob
+class DfeAnalytics::EntityTableCheckJob < CronJob
   self.cron_expression = "30 0 * * *" # Every day at 00:30
```
ericaporter, line 3: Is this really correct?

spec/jobs/entity_table_check_job_spec.rb
ericaporter: Now updated 👍

slorek (approved):

PR #2573, opened 2024-02-01
Fix LUPP subject matching against DQT record

slorek:
Fixes a bug in the LUPP DQT record matching where the ITT subject was not matched against the list of all possible strings listed in https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/master/lib/dqt/matchers/levelling_up_premium_payments.rb#L238-L317. It would only match if it was an exact match for the key, e.g. `computing`.

This would incorrectly determine a user as ineligible in the teacher ID journey.


your (approved):
Didn't work on this so I might be missing something from the wider picture, but the explanation and change I see look both legit 👍

PR #2584, opened 2024-02-07
Fix ECP subject matching against DQT record

slorek:
Similar to https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2573

Fixes a bug in the ECP DQT record matching where the ITT subject was not matched against the list of all possible strings listed in https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/master/lib/dqt/matchers/early_career_payments.rb. It would only match if it was an exact match for the key, e.g. `computing`.

This would incorrectly determine a user as ineligible in the teacher ID journey.

### DIFF
--- a/app/models/early_career_payments/dqt_record.rb
+++ b/app/models/early_career_payments/dqt_record.rb
@@ -36,16 +36,16 @@ module EarlyCareerPayments
 
     # TODO: May need to prioritise subject chosen by highest award amount?
     def eligible_itt_subject_for_claim
-      return nil if !itt_subjects || itt_subjects.empty?
+      raise if !itt_subject_groups || itt_subject_groups.empty?
 
       year = itt_year || claim.eligibility.itt_academic_year # The user may have supplied this manually if it was missing from the DQT record
 
-      return nil unless year
+      raise unless year
 
       itt_subject_checker = JourneySubjectEligibilityChecker.new(claim_year: claim.policy.configuration.current_academic_year, itt_year: year)
 
-      itt_subjects.delete_if do |itt_subject|
-        !itt_subject.to_sym.in?(itt_subject_checker.current_and_future_subject_symbols(claim.policy))
+      itt_subject_groups.delete_if do |itt_subject_group|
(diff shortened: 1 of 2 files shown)


app/models/early_career_payments/dqt_record.rb
```diff
@@ -36,16 +36,16 @@ def eligible_induction?
 
     # TODO: May need to prioritise subject chosen by highest award amount?
     def eligible_itt_subject_for_claim
-      return nil if !itt_subjects || itt_subjects.empty?
+      raise if !itt_subject_groups || itt_subject_groups.empty?
```
your, line 39: Now that you've removed the `find` after creating the array in `itt_subject_groups`, I think this can be reduced to

```
raise if itt_subject_groups.empty?
```

app/models/early_career_payments/dqt_record.rb
```diff
@@ -36,16 +36,16 @@ def eligible_induction?
...
 
       year = itt_year || claim.eligibility.itt_academic_year # The user may have supplied this manually if it was missing from the DQT record
 
-      return nil unless year
+      raise unless year
```
your, line 43: It's not great that we are are using an exception to control flow here, but acceptable. As an alternative, we could just use guard clauses with `return :none_of_the_above`.

your (approved):
Some suggestions - looks good anyway

app/models/early_career_payments/dqt_record.rb
slorek: I've combined the guards so there is less duplication when removing the exceptions.

your: I was curious so I had a look at `JourneySubjectEligibilityChecker`, something definitely for another time, but it could/should be refactored to raise custom exceptions that we can then rescue instead of catching all... otherwise that fallback could be a nightmare to debug one day. Looks better now anyway 👍

slorek: Agreed!

PR #2592, opened 2024-02-13
Redact PII in teacher_id_user_info column

slorek:


config/analytics_pii.yml
```diff
@@ -25,6 +25,7 @@ shared:
   - one_time_password
   - hmrc_bank_validation_responses
   - dqt_teacher_status
+  - teacher_id_user_info
```
your, line 28: Does it need to be added to [this](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/8b35ef6b2bf2c51d394f5dc598acb1eb08069f92/app/models/claim/personal_data_scrubber.rb#L9) list as well?

slorek: Yes, thanks!

PR #2596, opened 2024-02-16
Capt 1453 delete claims function

kenfodder:
* An column that should have been deleted when `claim_payments` table was introduced, the lingering foreign key constraint is causing issues for deleting Maths and Physics claims. So cleaning up now.
* Add dependent destroy to claim_payments to sever link to payments if a claim is deleted. The payment is not destroyed via a cascade as a payment could be shared amongst multiple claims.
* Function to delete all claims for a policy type

### DIFF
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -344,11 +344,14 @@ class Claim < ApplicationRecord
     (current_academic_year.approved.qa_required.count.to_f / claims_approved_so_far) * 100 <= MIN_QA_THRESHOLD
   end
 
-  # e.g. Claim.destroy_all_for_policy(MathsAndPhysics)
+  # This is be used EXCLUSIVELY MathsAndPhysics
   # Execute in a rails console
+  # Usage: Claim.destroy_all_for_policy(MathsAndPhysics)
   # Payments are deleted first due to HABTM (claims -< claim_payments >- payments)
-  # Also a payment might encompass 2 claims from multiple policies
+  # Also a payment might encompass 2 claims from multiple policies, those payments are NOT deleted
   def self.destroy_all_for_policy(policy)
+    raise "Claims for policy #{policy} cannot be destroyed" unless policy == MathsAndPhysics
+
     # Load all claims that are to be deleted
     claims = Claim.by_policy(policy)
     claim_ids = claims.map(&:id)
--- a/spec/factories/claims.rb
+++ b/spec/factories/claims.rb
@@ -291,5 +291,31 @@ FactoryBot.define do
     trait :logged_in_with_tid do
       logged_in_with_tid { true }
     end
+
+    trait :has_amendments do
+      after(:create) do |claim, _|
+        create_list(:amendment, 2, claim:)
+      end
+    end
+
+    trait :has_all_passed_tasks do
+      after(:create) do |claim, _|
+        ClaimCheckingTasks.new(claim).applicable_task_names.map do |task|
+          create(:task, :automated, :passed, name: task, claim:)
+        end
+      end
+    end
+
+    trait :has_notes do
+      after(:create) do |claim, _|
+        create_list(:note, 2, claim:)
+      end
+    end
+
+    trait :has_support_ticket do
+      after(:create) do |claim, _|
+        create(:support_ticket, claim:)
+      end
+    end
   end
 end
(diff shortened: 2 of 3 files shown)


app/models/claim.rb
```diff
@@ -345,6 +344,37 @@ def self.below_min_qa_threshold?
...
+  # Payments are deleted first due to HABTM (claims -< claim_payments >- payments)
+  # Also a payment might encompass 2 claims from multiple policies
+  def self.destroy_all_for_policy(policy)
+    # Load all claims that are to be deleted
+    claims = Claim.by_policy(policy)
+    claim_ids = claims.map(&:id)
+
+    # Load all the payments linked to those claims
+    payment_ids = ClaimPayment.where(claim_id: claim_ids).select(:payment_id).map(&:payment_id)
+    payments = Payment.includes(:claim_payments).where(id: payment_ids)
```
Reyko, line 358:
```suggestion
    # Load all the payments linked to those claims
    payments = Payment
      .joins(:claims)
      .merge(Claim.by_policy(policy))
```

app/models/claim.rb
```diff
@@ -345,6 +344,37 @@ def self.below_min_qa_threshold?
...
+    filtered_payments = Payment.where(id: filtered_payment_ids)
+
+    # Delete payments that are just linked with a single claim via claim_payments
+    filtered_payments.destroy_all
+
+    # Delete all the claims and all dependent models
+    claims.destroy_all
+
+    # Delete the PolicyConfiguration
+    policy_configuration = PolicyConfiguration.where(policy_types: [policy.to_s])
```
Reyko, line 372: Will this work if the `PolicyConfiguration` has more than one `policy_types`?

Reyko, line 372: Based on `PolicyConfiguration::for`

```suggestion
    policy_configuration = PolicyConfiguration.where("? = ANY (policy_types)", policy.to_s)
```

Reyko: In case of multiple policies, would require to delete from the `policy_types` Array the `policy` instead of the whole `PolicyConfiguration` record

slorek (commented):
I suspect it should be possible to destroy the records the normal way e.g. `Claim.by_policy(policy).destroy_all` if we created a new ActiveRecord relationship identifying those payments which would otherwise be orphaned. However this is going to be rarely used so I'm happy to leave those refactoring until later. I do think we should have some unit tests covering this as it's a dangerous operation.

app/models/claim.rb
```diff
@@ -345,6 +344,37 @@ def self.below_min_qa_threshold?
...
+    # Exclude payments that linked to another claim
+    payments_with_just_one_policy = payments.reject { |p| p.claim_payments.size > 1 }
+    filtered_payment_ids = payments_with_just_one_policy.map(&:id)
+    filtered_payments = Payment.where(id: filtered_payment_ids)
+
+    # Delete payments that are just linked with a single claim via claim_payments
+    filtered_payments.destroy_all
+
+    # Delete all the claims and all dependent models
+    claims.destroy_all
```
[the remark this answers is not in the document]
Reyko: 1. Sure, I was referring to call `destroy_all` on the `filtered_payments` like you do which would have a single `Claim`
2. Apologies, you're right 👍

app/models/claim.rb
```diff
@@ -345,6 +344,37 @@ def self.below_min_qa_threshold?
     (current_academic_year.approved.qa_required.count.to_f / claims_approved_so_far) * 100 <= MIN_QA_THRESHOLD
   end
 
+  # e.g. Claim.destroy_all_for_policy(MathsAndPhysics)
+  # Execute in a rails console
+  # Payments are deleted first due to HABTM (claims -< claim_payments >- payments)
+  # Also a payment might encompass 2 claims from multiple policies
+  def self.destroy_all_for_policy(policy)
```
your, line 351:
```suggestion
  def self.destroy_all_for_policy(policy)
    raise "claims for policy #{policy} cannot be destroyed" unless policy == MathsAndPhysics
```


From what I understand this is going to be used only once (for the foreseeing future); in that case I think we should prevent accidental usage of this method, otherwise it might be safer to move this method to a data migration to invoke once with `policy = MathsAndPhysics` as argument

Reyko (approved):

your (approved):

PR #2598, opened 2024-02-20
Upgrade rails and dependencies

slorek:
https://dfedigital.atlassian.net/browse/CAPT-1457

Dependabot did not catch this because of an incorrect version constraint on the rails gem.


your (approved):
I tested this a few weeks ago as well and seemed to work 👍

PR #2602, opened 2024-02-22
Capt 1421 remove maths and physics

kenfodder:
This is NOT TO BE merged until the Maths and Physics data has been deleted on PROD (and other environments but not critical)

This has now been rebased of `master`.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -217,65 +217,6 @@ en:
     payroll_gender: "gender"
     address: "address"
     full_name: "Full name"
-  maths_and_physics:
-    policy_name: "Claim a payment for teaching maths or physics"
-    policy_short_name: "Maths and Physics"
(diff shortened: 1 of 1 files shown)


slorek:
This PR removes the `MathsAndPhysics` class, however we must also remove the `PolicyConfiguration` record(s) which reference this in the `policy_types` column, as https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/capt-1421-remove-maths-and-physics/app/models/policy_configuration.rb#L77 attempts to refer to the now deleted class.

slorek:
Remove the `maths_and_physics` locale key.

slorek (approved):

slorek:
There's a few spec factories/helpers still - search for `maths_and_physics` in the codebase, but these shouldn't be blocking this.

PR #2628, opened 2024-03-08
Remove redundant mail templates and policy checks

slorek:
As spotted by @your on https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2625#discussion_r1516477228

These mailer templates are no longer in use as we use GOV.UK Notify templating.

Additionally there were some duplicated policy checks (it was already checking via the `unknown_policy_check` method) which I have removed.


your (approved):
🙇

Reyko (approved):

PR #2634, opened 2024-03-11
CAPT-1484 TID - Allow users to change their mobile number on 'Check your answers' - LUP/ECP/TSLR

your:
https://dfedigital.atlassian.net/browse/CAPT-1484

Refer to the main commit; this PR ensures that the correct answers and change slugs are displayed in "Check your answers" for the email and mobile number-related questions. TID creates, _de facto_, a separate virtual journey for each policy driven primarily by the `claim.logged_in_with_tid?` return value.


app/helpers/claims_helper.rb
```diff
@@ -5,49 +5,55 @@ def eligibility_answers(current_claim)
...
+      a << [t("questions.teacher_reference_number"), claim.teacher_reference_number, "teacher-reference-number"] if show_trn(claim)
+      a << [t("questions.national_insurance_number"), claim.national_insurance_number, "personal-details"] if show_nino(claim)
+
+      a << if claim.logged_in_with_tid? && claim.teacher_id_user_info["email"].present?
+        [t("questions.select_email.heading"), claim.email_address, "select-email"]
+      else
+        [t("questions.email_address"), claim.email_address, "email-address"]
+      end
+
+      if claim.logged_in_with_tid? && claim.teacher_id_user_info["phone_number"].present?
```
slorek, line 21: I think the `email_address_check` and `mobile_check` columns might be useful here to determine if the user provided their own details or the ones from teacher ID.

your: @slorek I think you are right, but when I looked at the slug sequence [here](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/9c65314e7c4ec07f27fe9d32f01849a797916a21/app/models/journeys/additional_payments_for_teaching/slug_sequence.rb#L113-L127), it somehow seems that it's driven by `teacher_id_user_info["phone_number"]` and `teacher_id_user_info["email_address"]` (as well), so it almost seemed useless to check those `_check` flags, which would be set only if the `select-email` and `select-mobile` slugs can be rendered – which is when the user is on TID!

I'll have another look, it generally feels quite messy what's happening there and I found it easier to reason in terms of whether the mobile and email are present or not in TID.

your: @slorek I had another go at this; I spent some time refactoring and adding comments, and I ended up relying on those `_check` flags as you suggested. Hopefully it's easier to understand what is the changes are trying to achieve and the various scenarios: [e7edf90](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2634/commits/e7edf9062edfdcb791da558d1f334a975d87f667)

slorek (approved):

PR #2635, opened 2024-03-11
CAPT-1509 - Form object for the current-school school search

kenfodder:
* ClaimsController now allows forms to be moved over to use form objects individually
* current-school form is using a form object
* claim-school was sharing partials with current-school, that has been copied over so claim-school works as before for now
* change the I18n name spacing so they grouped by form and by the policy

TODO:

- [x] Add tests for form.rb
- [x] JS todo comment, not sure how to solve that one yet
- [x] Needs rebasing
- [x] Remove the validation on current-school on the claim.rb model
- [x] Make claim to use `update!` so any errors on the claim is reported for now

### DIFF
--- a/app/forms/current_school_form.rb
+++ b/app/forms/current_school_form.rb
@@ -21,7 +21,7 @@ class CurrentSchoolForm < Form
   def save
     return false unless valid?
 
-    claim.update!({eligibility_attributes: {current_school_id: current_school_id}})
+    update!({eligibility_attributes: {current_school_id: current_school_id}})
   end
 
   def current_school_name
--- a/app/forms/form.rb
+++ b/app/forms/form.rb
@@ -15,6 +15,10 @@ class Form
     super
   end
 
+  def update!(attrs)
+    claim.update!(attrs)
+  end
+
   def view_path
     journey::VIEW_PATH
   end


app/assets/javascripts/components/school_search.js.erb
```diff
@@ -126,14 +126,8 @@ document.addEventListener("DOMContentLoaded", function () {
...
 
-      if (currentPolicy == "additional-payments") {
-        methodInput.value = "patch";
-        setSubmitValue(defaultFormSubmitValue);
-      } else {
-        form.method = "post";
-        methodInput.value = "patch";
-        setSubmitValue("Continue");
-      }
+      methodInput.value = "patch";
```
slorek, line 129: I think the `currentPolicy` variable is now not used anywhere and can be removed. Thanks for consolidating this.

app/controllers/claims_controller.rb
```diff
@@ -12,6 +12,10 @@ class ClaimsController < BasePublicController
   before_action :clear_claim_session, only: [:new]
   before_action :prepend_view_path_for_journey
 
+  SLUG_TO_FORM = {
+    "current-school" => CurrentSchoolForm
+  }
```
slorek, line 17: I think we might want to define this per journey, so that we can have different behaviours as necessary. I suggest we could move this to the relevant `SlugSequence` class, or the relevant `Journey` module. If we want to have a default behaviour, this could be defined in `Journeys::Base` perhaps.

app/controllers/claims_controller.rb
```diff
@@ -21,7 +25,12 @@ def create
   end
 
   def show
-    search_schools if params[:school_search]
+    if @form ||= form_from_slug(params)
```
slorek, line 28:
```suggestion
    # TODO: Migrate the remaining slugs to form objects.
    if @form ||= form_from_slug(params)
```

app/controllers/claims_controller.rb
```diff
@@ -156,6 +178,10 @@ def set_backlink_path
     @backlink_path = claim_path(current_journey_routing_name, previous_slug) if previous_slug.present?
   end
 
+  def set_any_backlink_override
+    @backlink_path = @form.backlink_path if @form.backlink_path
```
slorek, line 182: The banklink path will likely be different for different journeys, so this highlights the need to be able to override forms on a per-journey basis.

app/forms/current_school_form.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  def initialize(claim:, journey:, params:)
+    super
+
+    load_schools
+    self.current_school_id = permitted_params[:current_school_id]
+  end
+
+  def backlink_path
+    params[:slug] if params[:school_search]
```
slorek, line 18: What happens if there is no `params[:school_search]`?

app/forms/current_school_form.rb
```diff
@@ -0,0 +1,58 @@
...
+    return false unless valid?
+
+    claim.update!({eligibility_attributes: {current_school_id: current_school_id}})
+  end
+
+  def current_school_name
+    claim.eligibility.current_school_name
+  end
+
+  def i18n_errors_path(msg)
```
slorek, line 31: This can probably go in the `Form` base class, and can probably be a private method.

app/forms/current_school_form.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def backlink_path
+    params[:slug] if params[:school_search]
+  end
+
+  def save
+    return false unless valid?
+
+    claim.update!({eligibility_attributes: {current_school_id: current_school_id}})
```
slorek, line 24: It might be worth implementing a common re-usable method in the base `Form` class which updates the claim, as that will make it easier to swap between storing input in a `Claim` record to in a payload in the session when we are able to do that. Otherwise we would need to update every form object.

app/views/claims/correct_school.html.erb
```diff
@@ -7,7 +7,7 @@
 
     <legend class="govuk-fieldset__legend govuk-fieldset__legend--<%= shared_view_css_size %>">
       <h1 class="govuk-fieldset__heading">
-        <%= t("additional_payments.questions.current_school_search") %>
+        <%= t("additional_payments.forms.current_school.questions.current_school_search") %>
```
slorek, line 10: We are using a journey-specific translation in a non-journey-specific generic template. We probably need to have a I18n namespace for non-specific translations, or we might need to duplicate or move this template to the journey folder if it needs to be specific.

app/views/student_loans/claims/_school_search_results.html.erb
```diff
@@ -58,7 +58,4 @@
   <% end %>
 
   <%= form.submit "Continue", class: "govuk-button" %>
-  <% if ! current_claim.has_ecp_or_lupp_policy? %>
-    <%= link_to "Search again", claim_path, class: "govuk-button govuk-button--secondary", role: "button", data: {module: "govuk-button"} %>
-  <% end %>
```
slorek, line 63: 👏

app/controllers/claims_controller.rb
slorek: OK thanks, sorry the method above wasn't visible in the diff.

app/views/claims/correct_school.html.erb
slorek: OK, in that case this template should live in the `views/additional_payments/claims` folder and not in the `views/claims` folder.

your (approved):

slorek (approved):
Looks in a good place to let loose on the other forms. Thanks.

PR #2638, opened 2024-03-13
Refactor journey configuration

slorek:
Continuation of https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2625 and https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2606

This change does the following:

* Create a new module for each journey - refactors the existing `SERVICES` constant and various helper methods that used to live in `JourneyConfiguration`.
* Remove the definition of policies from `JourneyConfiguration` database records
* The primary way to access journey config is via the new module, e.g. `Journeys.for_routing_name("additional-payments")`
* You only need to access the configuration database record when you need to know the current academic year, whether it's open for submissions, etc. This will cut down on excessive database queries. There is a convenience access method for this e.g. `Journeys::AdditionalPaymentsForTeaching.configuration`
* Makes the primary key for the database record the routing name of the journey instead of a random UUID, which doesn't guarantee one record per journey
* Moves the slug sequence classes to the relevant journey namespace instead of `Policies`. Slug sequence should be accessed via the journey module rather than the policy.

I followed the naming convention already used for `Policies` when naming it `Journeys`; I'd probably prefer `Journey` instead but this is open for discussion.


spec/models/journey_configuration_spec.rb
```diff
@@ -2,47 +2,11 @@
...
-    describe "#routing_name" do
-      it "returns routing for JourneyConfiguration" do
-        expect(described_class.for(Policies::StudentLoans).routing_name).to eq "student-loans"
-
-        # Same routing_name for ECP and LUP
-        expect(described_class.for(Policies::EarlyCareerPayments).routing_name).to eq "additional-payments"
-        expect(described_class.for(LevellingUpPremiumPayments).routing_name).to eq "additional-payments"
-      end
-    end
-
```
your, line 45: I believe there's a new instance method `#journey` that has been covered here

app/controllers/claims_controller.rb
```diff
@@ -133,7 +133,10 @@ def start_new
       clear_claim_session
       redirect_to(new_claim_path(current_journey_routing_name))
     else
-      redirect_to(claim_path(current_claim.policy.routing_name, slug: slug_to_redirect_to))
+      new_journey = Journeys.for_policy(current_claim.policy)
+      new_claim_slug_sequence = new_journey.slug_sequence.new(current_claim)
+      new_page_sequence = PageSequence.new(current_claim, new_claim_slug_sequence, session[:slugs], params[:slug])
+      redirect_to(claim_path(new_journey::ROUTING_NAME, slug: new_page_sequence.next_required_slug))
```
your, line 139: A lot is happening here; I wonder if these statements could be extracted to a separate method, e.g. `redirect_to_new_journey_page`. Just an idea

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -8,7 +8,7 @@ def callback
 
     session[:user_info] = auth.extra.raw_info
 
-    redirect_to claim_path(journey: policy.routing_name, slug: "teacher-detail")
+    redirect_to claim_path(journey: Journeys.for_policy(policy)::ROUTING_NAME, slug: "teacher-detail")
```
your, line 11: I think we might have some circular dependency going on here; I spotted a new failure in the pipeline yesterday; just FYI, I'll look into that when I get a chance; unrelated to this PR anyway

your (commented):
Cool! Apart from some questionable suggestions/ideas, there might be just a couple of unit tests missing. I also have been thinking whether we should use `Journey` or `Journeys`, but something tells me the latter is more adequate here

spec/features/sign_in_or_continue_spec.rb
```diff
@@ -19,7 +19,7 @@
   end
 
   scenario "Teacher makes claim for 'Early-Career Payments' claim and select continue without signing in" do
-    visit landing_page_path(Policies::EarlyCareerPayments.routing_name)
+    visit landing_page_path(Journeys::AdditionalPaymentsForTeaching::ROUTING_NAME)
```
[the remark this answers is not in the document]
slorek: Thanks, I do agree and implemented it at one point, but it was literally just:

```ruby
def routing_name
  ROUTING_NAME
end
```

It didn't appear to be adding much value. I'm happy to do it if there is a strong preference.

spec/models/base_policy_spec.rb
```diff
@@ -38,28 +38,4 @@ module TestPolicy
...
-
-    it do
-      Policies::TestPolicy.routing_name
-
-      expect(JourneyConfiguration).to have_received(:routing_name_for_policy)
-        .with(Policies::TestPolicy)
-    end
-  end
-
-  describe "::configuration" do
```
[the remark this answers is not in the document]
slorek: Thanks, removed.

app/controllers/claims_controller.rb
slorek: Thanks, done.

app/models/journeys/configuration.rb
```diff
@@ -0,0 +1,29 @@
...
+    self.table_name = "journey_configurations"
+
+    # Use AcademicYear as custom ActiveRecord attribute type
+    attribute :current_academic_year, AcademicYear::Type.new
+
+    validates :current_academic_year_before_type_cast, format: {with: AcademicYear::ACADEMIC_YEAR_REGEXP}
+
+    def additional_payments?
+      journey == Journeys::AdditionalPaymentsForTeaching
+    end
```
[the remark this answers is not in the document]
slorek: Good idea, though so far only `additional_payments?` is used anywhere (in the admin manbage services section). Perhaps when we add the new policies this will be useful.

app/controllers/omniauth_callbacks_controller.rb
your: Yes, I thought it was gone thanks to your changes, but I did get another related failure yesterday, and after rebasing, see [here](https://dfe-ssp.visualstudio.com/S118-Teacher-Payments-Service/_build/results?buildId=455097&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&s=6884a131-87da-5381-61f3-d7acc3b91d76&t=68dc271a-6b90-554f-dc73-9843e30016f8&l=8892)

spec/features/sign_in_or_continue_spec.rb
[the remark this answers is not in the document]
your: No strong preference 👍 Feel free to resolve the comment

app/models/journeys/configuration.rb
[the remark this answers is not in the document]
your: Definitely, we'll probably end up doing it at some point anyway, I think it's good to keep as much abstraction as possible with this new model

your (approved):
No blockers 👍

app/controllers/omniauth_callbacks_controller.rb
slorek: Thanks, yes I could see how that could happen if `claim_from_session` returned `nil`, which it shouldn't do in practice, but it would be worth fixing this.

PR #2644, opened 2024-03-15
Feature/capt 1518 namespace levelling up premium payments to policies - part 2

Reyko:
Convert `Claim#eligibility_type` from `LevellingUpPremiumPayments::Eligibility` to `Policies::LevellingUpPremiumPayments::Eligibility`

### DIFF
--- a/db/migrate/20240315091658_change_claims_eligibility_type_to_policies_levelling_up_premium_payments.rb
+++ b/db/migrate/20240315091658_change_claims_eligibility_type_to_policies_levelling_up_premium_payments.rb
@@ -1,0 +1,13 @@
+class ChangeClaimsEligibilityTypeToPoliciesLevellingUpPremiumPayments < ActiveRecord::Migration[7.0]
+  def change
+    reversible do |dir|
+      dir.up do
+        Claim.where(eligibility_type: "LevellingUpPremiumPayments::Eligibility").update_all(eligibility_type: "Policies::LevellingUpPremiumPayments::Eligibility")
+      end
+
+      dir.down do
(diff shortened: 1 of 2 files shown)


db/migrate/20240315091658_change_claims_eligibility_type_to_policies_levelling_up_premium_payments.rb
```diff
@@ -0,0 +1,13 @@
+class ChangeClaimsEligibilityTypeToPoliciesLevellingUpPremiumPayments < ActiveRecord::Migration[7.0]
+  def change
+    reversible do |dir|
+      dir.up do
+        Claim.where(eligibility_type: "LevellingUpPremiumPayments::Eligibility").update_all(eligibility_type: "Policies::LevellingUpPremiumPayments::Eligibility")
+      end
```
slorek, line 6: You could add this:

```ruby
entity_tag = Time.now.strftime('%Y%m%d%H%M%S')
DfE::Analytics::LoadEntities.new(entity_name: "claims").run(entity_tag:)
```
So that we don't have to re-import manually (same on the `down` method).

slorek (approved):

db/migrate/20240315091658_change_claims_eligibility_type_to_policies_levelling_up_premium_payments.rb
Reyko: Used `DfE::Analytics::LoadEntities.new(entity_name: "claims").run` instead as the gem's tag that we use does not support arguments [v1.11.6](https://github.com/DFE-Digital/dfe-analytics/blob/v1.11.6/lib/dfe/analytics/load_entities.rb#L13)

PR #2657, opened 2024-03-27
Further separation of policy and journey concerns

slorek:
Continuation of https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2625 and https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2606 and https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2638

This is a further refactoring to separate journey concerns. Key changes:

* Provides a pattern for journey-specific controllers (starting with the reminders controller)
* Separates templates where there is conditional logic depending on the claim policy
* Moves the `PageSequence` model to the `Journeys` namespace and provides convenience methods to access this
* Use the relevant journey answers presenter for all questions

It is best to review the pull request on a per-commit basis as each commit is a discrete change.


spec/models/journeys/teacher_student_loan_reimbursement_spec.rb
```diff
@@ -32,4 +32,25 @@
...
+      expect(page_sequence.claim).to eq(claim)
+      expect(page_sequence.current_slug).to eq(current_slug)
+      expect(page_sequence.completed_slugs).to eq(completed_slugs)
+    end
+
+    it "creates a slug sequence" do
+      expect(Journeys::TeacherStudentLoanReimbursement::SlugSequence).to receive(:new).with(claim)
+      page_sequence
+    end
+  end
```
your, line 55: We could probably test this method in just one place, like in a base spec, unless we think it can easily be overridden and have different behaviour for each journey. Perhaps that's what you are thinking.

app/models/claim.rb
```diff
@@ -531,10 +531,6 @@ def has_ecp_policy?
     policy == Policies::EarlyCareerPayments
   end
 
-  def has_tslr_policy?
-    policy == Policies::StudentLoans
-  end
-
```
your, line 537: mmmh I am using this method here https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/ccab16afbab9c3bebda50ecec77c35f72a141289/app/models/claim_student_loan_details_updater.rb#L12

app/controllers/school_search_controller.rb
```diff
@@ -1,4 +1,7 @@
-class SchoolSearchController < BasePublicController
+class SchoolSearchController < ApplicationController
+  include DfE::Analytics::Requests
+  include HttpAuthConcern
```
your, line 3: Can I just check why we are not using the `BasePublicController` here?

app/models/journeys/additional_payments_for_teaching/slug_sequence.rb
```diff
@@ -64,6 +64,12 @@ class SlugSequence
         "ineligible"
       ].freeze
 
+      REMINDER_SLUGS = %w[
+        personal-details
+        email-verification
+        set
+      ].freeze
```
your, line 71: 👍

your (commented):
Love the direction 👍 First pass, added some questions/comments

spec/models/journeys/teacher_student_loan_reimbursement_spec.rb
slorek: Yes I have been thinking that we can use shared examples for some of these tests, but for now I wanted to keep it simple and readable due to the heavy refactoring.

app/controllers/school_search_controller.rb
slorek: School search sits outside of the journey routes so `journey` will resolve to `nil`. `BasePublicController` assumes it is used in the journey context and does some things like altering the view path which won't work and is not needed here.

app/models/claim.rb
slorek: Thanks, I will remove this commit in that case.

app/helpers/claims/show_helper.rb
```diff
@@ -1,19 +1,7 @@
...
-        content_tag(:h1, text, class: "govuk-panel__title", id: "submitted-title")
-      else
-        content_tag(:h1, "Claim submitted", class: "govuk-panel__title", id: "submitted-title")
-      end
-    end
-
-    def shared_view_css_class_size(claim)
-      claim.has_ecp_or_lupp_policy? ? "l" : "xl"
+    def fieldset_legend_css_class_for_journey(journey)
+      (journey == Journeys::AdditionalPaymentsForTeaching) ? "govuk-fieldset__legend--l" : "govuk-fieldset__legend--xl"
```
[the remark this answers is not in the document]
slorek: I agree. I felt this was an improvement but not a complete solution I'm happy with. Perhaps we can make this consistent if there is consensus and remove this.

app/views/additional_payments/claims/current_school.html.erb
```diff
@@ -0,0 +1,24 @@
...
+    <% else %>
+      <%= render "school_search_results",
+                 question: t("additional_payments.questions.current_school_results"),
+                 school_param: :current_school,
+                 school_id_param: :current_school_id,
+                 exclude_closed: true
+      %>
+    <% end %>
+  </div>
+</div>
```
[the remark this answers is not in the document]
slorek: The old version had two local variables passed to the partial templates; `question` and `ecp_question`. The reason was that of it was an ECP claim the partial should show different text, and there were conditionals in those partials. This change just splits the template so that only `question` is passed to the partial with the correct translation for that journey, so I could remove the conditional logic from the partials.

your (approved):

PR #2664, opened 2024-04-04
CAPT 1567 - Form object for /supply-teacher

AbigailMcP:
## Context

- We're refactoring the handling of individual journey pages into form objects 
- This will allow us to slim down the ClaimsController which is doing To Much Work

First PR to do this: https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2635

## Changes in this PR

- Created new `SupplyTeacherForm` which handles the validations and setting of `employed_as_supply_teacher` boolean attribute on a `Claim`'s `Eligibility`.
  - No custom back routing logic for this page (i don't think!)
- Removed the validation from the `Eligibility` model
- Updated the view and its error rendering to use the new form object
- Listed the form object in a new `def forms` method in the `Journeys::AdditionalPayments` module (rather than `Journeys::Base`), since it's specific to this journey.
  - This need to be compatible with the work in this PR: https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2663

## Guidance for review

https://dfedigital.atlassian.net/jira/software/projects/CAPT/boards/132?selectedIssue=CAPT-1567

- Run through the claim steps until you reach the 'Are you currently employed as a supply teacher?' form
- Check content, page title
- Check answering 'Yes' / 'No' takes you to the correct (different) pages
- Go back to the supply teacher page and check your answer is persisted

Screenshot 2024-04-04 at 11 14 33

Screenshot 2024-04-04 at 11 14 24

PR #2666, opened 2024-04-05
CAPT-1560 Form object for /poor-performance

your:
https://dfedigital.atlassian.net/browse/CAPT-1560

The primary goals of this PR are to reduce the boilerplate when creating a new form and the number of tests (test-what-you-can approach in the base/superclass spec—introduced with this PR). More info on each commit


app/forms/form.rb
```diff
@@ -3,9 +3,11 @@ class Form
...
   include ActiveModel::Serialization
 
-  attribute :claim
-  attribute :journey
-  attribute :params
+  attr_accessor :claim
+  attr_accessor :journey
+  attr_accessor :params
+
+  delegate :persisted?, to: :claim
```
rjlynch, line 10: 👍

app/forms/form.rb
```diff
@@ -35,9 +37,13 @@ def i18n_errors_path(msg)
     I18n.t("#{i18n_namespace}.forms.#{i18n_form_namespace}.errors.#{msg}")
   end
 
+  def permitted_params
+    @permitted_params ||= params.fetch(:claim, {}).permit(*attributes)
```
rjlynch, line 41: thoughts on changing the initialiser to something like
```
  def initialize(claim:, journey:, params:)
    @claim = claim
    @journey = journey
    super(permitted_params)
  end
```
then the subclasses don't have to do the `self.some_attribute = permitted_params[:some_attribute]` ?

rjlynch (approved):
Looks good to me.
Might be best to get someone else with more context of the claim app to take a look

app/forms/form.rb
your: @rjlynch good idea! this seems to do the trick:

```ruby
  def initialize(claim:, journey:, params:)
    super
    assign_attributes(permitted_params)
  end
```

rjlynch (approved):
👍

app/forms/form.rb
```diff
@@ -35,9 +37,13 @@ def i18n_errors_path(msg)
...
+  end
+
   private
 
   def i18n_form_namespace
-    raise "Form#i18n_form_namespace requires all sub-classes to override"
+    self.class.name.demodulize.gsub("Form", "").underscore
```
slorek, line 48: Nice

app/forms/form.rb
slorek, line 42: Nice

app/models/journeys/additional_payments_for_teaching.rb
```diff
@@ -9,5 +9,8 @@ module AdditionalPaymentsForTeaching
     VIEW_PATH = "additional_payments"
     I18N_NAMESPACE = "additional_payments"
     POLICIES = [Policies::EarlyCareerPayments, Policies::LevellingUpPremiumPayments]
+    FORMS = {
+      "poor-performance" => ::PoorPerformanceForm
+    }.freeze
```
slorek, line 14: Suggestion: going with Rails' 'convention over configuration' principle perhaps the form class name could be inferred from the slug name.

app/forms/poor_performance_form.rb
```diff
@@ -0,0 +1,22 @@
+class PoorPerformanceForm < Form
```
slorek, line 1: What do you think about namespacing this form under `Journeys:AdditionalPaymentsForTeaching`?

spec/models/early_career_payments/eligibility_spec.rb
```diff
@@ -330,23 +330,6 @@
       end
     end
 
-    context "when saving in the 'poor-peformance' context" do
```
slorek, line 333: I think we can remove the validations from the class. They are also run on the submit action but I'm not sure this is necessary.

app/forms/form.rb
rjlynch: Actually thinking about it we might not be able to do this.
We probably want to support defaulting to any attributes on the claim / eligibility if they're not set by the params so we can support remembering the user's options when they navigate back. If we were just working with a Claim we could do something like
`assign_attributes(permitted_params.reverse_merge(claim.attributes.slice(attributes))` but as we're potentially dealing with Eligibility defaulting the params might get ugly. Maybe best to leave this for the minute and let subclasses initialise their own attributes 🤔

spec/models/early_career_payments/eligibility_spec.rb
your: @slorek mmh I removed those [here](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2666/files#diff-0ae6631c29911caa072ad79b782f98e861fa65cbbdd0cdf8d49ad606c51b1f76)

Ye the submit context is a little bit of a tricky one but I guess we are saying we can get rid of it for now

app/forms/form.rb
rjlynch: Also in the `PoorPerformanceForm` I think we probably need to set the default values from the eligibility record sort of like

```ruby
def initialize(journey:, claim:, params:)
  self.subject_to_formal_performance_action = permitted_params.fetch(:subject_to_formal_performance_action, claim.eligibility.subject_to_formal_performance_action)
 self.subject_to_disciplinary_action = permitted_params.fetch(:subject_to_disciplinary_action, claim.eligibility.subject_to_disciplinary_action)
end
```
so navigating back doesn't lose the previous option (if we care about that!)

your: I see what you are saying; you are right that it loses the values; they should be pre-filled from the model if they exist. I see that for the first form object (current school), we did a bit of a hack, as there is an instance method on the form object exposing a certain attribute from the claim/eligibility, so I completely missed the defaulting.

your: > as we're potentially dealing with Eligibility defaulting the params might get ugly. Maybe best to leave this for the minute and let subclasses initialise their own attributes until a pattern emerges

definitely my thinking, also I know we actually want to remove that model soon.

rjlynch: Video as I'm not sure if I'm being clear 😂 

Without setting attribute defaults from eligibility record
without-default-attributes



With setting attribute defaults from eligibility record
with-default-attributes

your: Got it, you were clear thanks. I'll fix that

your: @rjlynch I agree with Steve here: https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2664/files#r1553830035

> [...] Perhaps a convenience method in the base class to read previous answers would be helpful and allow us to swap out the underlying data source more easily.

I'll try to look at this first thing next week, unless you come up with something like that before... the eligibility model is a bit of a problem to deal with but 🤷

app/forms/poor_performance_form.rb
slorek: I was thinking maybe of a system similar to the way the views are loaded; first it looks for the view in the relevant journey folder and then it looks in the shared `claims` folder.

app/forms/form.rb
slorek: I think you're right, but bear in mind we want to get rid of the `Claim` record in the next phase so will want to pull in the previously filled answers from the replacement data source (possibly just a Hash in the session), so we could have a convenience method to access the existing answers in the same way we have an `update!` method in the base `Form` class to abstract access to this.

your: @rjlynch I've reverted the changes that I had introduced as part of this conversation so we can tackle it properly separately but made sure we get the default values from the eligibility model where needed

app/models/journeys/additional_payments_for_teaching.rb
your: It's worth postponing this to later so this PR can unblock @rjlynch's work. It's definitely on the radar

rjlynch (approved):
Looks good 👍

PR #2668, opened 2024-04-05
Extract NqtInAcademicYearAfterIttForm

rjlynch:
Please feel free to nit pick as much as possible!

I'm unsure about the backlink path piece, I've kept the logic for rendering that the same as it was in the claim controller. I think it could be simplified and we could remove `return unless page_sequence.in_sequence?("correct-school")` if we can assume that `nqt-in-academic-year-after-itt` always follows the `correct-school` step - I'm not sure on that assumption though.

I hope the specs are ok! For the past year or so I've been writing specs avoiding using `let` and `before` entirely and just repeating the test setup in each `it` block, which I've found works really well, so I'm a bit rusty on `let` best practices.

walk-through

---
Extracts a new form object and moves the logic associated with the
`nqt_in_academic_year_after_itt` step out of claims controller into the
form.
The view no longer needs to worry about the `eligibility` attributes
being nested, so it has been updated to remove the eligibility
namespacing on fields.

I'm not stoked about having to call url helpers in the form model for
the backlink path but was unsure how else to do this. Maybe we want to
consider introducing a new object `BacklinkPath` and pass the form
object to that? Something like
```
BacklinkPath.new(form).to_s #=> "/the-path/we/want"
```

Also I think if we're going to keep the permitting params logic in the
form object, we'll probably want to make it a bit more ergonomic to
define what params are permitted. We could add a class attribute to the
base form and then child forms could do something like
```
class SomeForm < Form
  permitted_params %i[some_param]
```
or we could just read off the attributes defined on the form and assume
any defined attribute is safe to permit as a paramerter.

We could also now remove the validation for
`:nqt_in_academic_year_after_itt` from
`Policies::EarlyCareerPayments::Eligibility` as it's no longer needed,
though I've left this in place for the time being, something for a
future commit.

### DIFF
--- a/app/forms/form.rb
+++ b/app/forms/form.rb
@@ -46,4 +46,13 @@ class Form
   def i18n_form_namespace
     self.class.name.demodulize.gsub("Form", "").underscore
   end
+
+  def page_sequence
+    @page_sequence ||= Journeys::PageSequence.new(
+      claim,
+      journey.slug_sequence.new(claim),
+      nil,
+      params[:slug]
+    )
+  end
 end
--- a/app/forms/journeys/additional_payments_for_teaching/nqt_in_academic_year_after_itt_form.rb
+++ b/app/forms/journeys/additional_payments_for_teaching/nqt_in_academic_year_after_itt_form.rb
@@ -1,0 +1,69 @@
+module Journeys
+  module AdditionalPaymentsForTeaching
+    class NqtInAcademicYearAfterIttForm < Form
+      attribute :nqt_in_academic_year_after_itt, :boolean
+
+      validates :nqt_in_academic_year_after_itt,
+        inclusion: {
+          in: [true, false],
+          message: "Select yes if you are currently teaching as a qualified teacher"
+        }
+
+      def initialize(journey:, claim:, params:)
+        super
+
+        self.nqt_in_academic_year_after_itt = permitted_params.fetch(
+          :nqt_in_academic_year_after_itt,
+          claim.eligibility.nqt_in_academic_year_after_itt
+        )
+      end
+
+      def save
+        return false unless valid?
+
+        update!(
+          {
+            eligibility_attributes: {
+              nqt_in_academic_year_after_itt: nqt_in_academic_year_after_itt,
+              induction_completed: determine_induction_answer_from_dqt_record
+            }
+          }
+        )
+      end
+
+      def backlink_path
+        return unless page_sequence.in_sequence?("correct-school")
+
+        Rails
+          .application
+          .routes
+          .url_helpers
+          .claim_path(params[:journey], "correct-school")
+      end
+
+      private
+
+      def permitted_params
+        @permitted_params ||= params.fetch(:claim, {}).permit(
+          :nqt_in_academic_year_after_itt
+        )
+      end
+
+      def determine_induction_answer_from_dqt_record
+        return unless passed_details_check_with_teacher_id?
+        # We can derive the induction_completed value for current_claim using the
+        # ECP DQT record Remember: even if it's only relevant to ECP, the induction
+        # question is asked at the beginning of the combined journey, and the
+        # applicant may end up applying for ECP or LUPP only at a later stage in
+        # the journey, hence we need to store the answer on both eligibilities.
+        claim_for_policy = claim.for_policy(Policies::EarlyCareerPayments)
+        dqt_teacher_record = claim_for_policy.dqt_teacher_record
+        dqt_teacher_record&.eligible_induction?
+      end
+
+      def passed_details_check_with_teacher_id?
+        claim.logged_in_with_tid? && claim.details_check?
+      end
+    end
+  end
+end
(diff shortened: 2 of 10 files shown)


app/forms/form.rb
```diff
@@ -15,6 +15,10 @@ def initialize(claim:, journey:, params:)
     super
   end
 
+  def persisted?
+    true
+  end
```
your, line 20: This could be delegated to `claim`, an instance of `CurrentClaim` that defines `#persisted?`

app/views/additional_payments/claims/nqt_in_academic_year_after_itt.html.erb
```diff
@@ -1,49 +1,56 @@
-<% content_for(:page_title, page_title(t("additional_payments.questions.nqt_in_academic_year_after_itt.heading"), journey: current_journey_routing_name, show_error: current_claim.errors.any?)) %>
+<% content_for(:page_title, page_title(t("additional_payments.questions.nqt_in_academic_year_after_itt.heading"), journey: current_journey_routing_name, show_error: @form.errors.any?)) %>
 <% path_for_form = current_claim.persisted? ? claim_path(current_journey_routing_name) : claims_path(current_journey_routing_name) %>
```
your, line 2: There might be an obvious reason for `path_for_form` to exist, but I removed it from my form as I couldn't figure it out; if needed, it should be changed to call `#persisted?` on `@form`

spec/models/claim_spec.rb
```diff
@@ -165,8 +165,8 @@
 
     # Tests a single attribute, possibly should test multiple attributes
```
your, line 166: That's an interesting test 🤔 - anyway, something for another day

app/models/journeys/base.rb
```diff
@@ -18,14 +18,18 @@ def slug_sequence
...
+      avaliable_forms = SHARED_FORMS.merge(forms)
+
+      form = avaliable_forms[params[:slug]]
 
       form&.new(journey: self, claim: claim, params: params)
     end
 
+    def forms
+      {}
+    end
```
your, line 31: I did something slightly different in my PR and decided that this method would return the content of `FORMS` from a specific journey model, but overriding the method is also a valid approach

your (commented):
good job, good to see some stuff go from the controller. I am a bit concerned about some of the boilerplate (not specific to this PR) and tried to address it in my PR; in any case it should not be a blocker. The only outstanding question for me is whether we need that `path_for_form` variable in the view

app/models/journeys/base.rb
rjlynch: Yeah I'll rebase this once your pr is merged and pick up the change you made to `persisted?` too

spec/models/claim_spec.rb
rjlynch: Yeah I just updated the spec so this functionality is still tested by something. Once everything has been moved over to form objects we should be able to lose this spec

app/views/additional_payments/claims/nqt_in_academic_year_after_itt.html.erb
rjlynch: Yeah I'm not sure why that's there. With the form checking `persisted?` it should go to the right place 🤔 will remove it

rjlynch:
> @rjlynch Yeah the backlink_path is a bit janky if you look at the `claims_controller` you'll see in the before_actions it calls `set_backlink_path`, my initial attempt was if the form did have custom logic for the back link, it overrode it without interrupting what it was doing before. Since it was doing it in the if statements in the controller.
> 
> I feel this might be easier once all the form objects are done and do this one sweep. Or we could introduce it one of them, make it work in parallel and migrate them as we go along. Thoughts?

I think leaving it on the form for now is best and once all the form objects have been added we can reassess and see if we need to change it. It _feels_ like the backlink is a concern of the page sequence but I'm not sure.

app/forms/nqt_in_academic_year_after_itt_form.rb
```diff
@@ -0,0 +1,65 @@
+class NqtInAcademicYearAfterIttForm < Form
```
AbigailMcP, line 1: I namespaced `SupplyTeacherForm` under `Journeys:AdditionalPaymentsForTeaching` since it's journey specific.

rjlynch: Makes sense. Will reissue this pr with name spacing sorted once I've rebased it on Giuseppe's work

your (approved):
No blockers from me and sorry for the wait 👍

app/forms/journeys/additional_payments_for_teaching/nqt_in_academic_year_after_itt_form.rb
```diff
@@ -0,0 +1,69 @@
...
+          .claim_path(params[:journey], "correct-school")
+      end
+
+      private
+
+      def permitted_params
+        @permitted_params ||= params.fetch(:claim, {}).permit(
+          :nqt_in_academic_year_after_itt
+        )
+      end
```
your, line 50: For another PR: this method can be removed as it's now on the base form and, by default, permits all the attributes defined on the form

PR #2674, opened 2024-04-09
capt 1571/extract form object for teaching subject now

rjlynch:
Extract form for teaching_subject_now

Extracts a form object for the teaching subject now step.

Depending on whether or not there's an eligible subject upon reaching
the `teaching-subject-now` step, we may redirect the subject to the
`eligbible-itt-subject` step rather than render the
`teaching-subject-now` form.
`eligible-itt-subject` can be reset if the `qualification` changes, and
`eligible-itt-subject` resets the `teaching-subject-now` answer.

For now we've preserved the existing behaviour of determining whether to
render the `teaching-subject-now` in the `ClaimsController#show` action.
This feels like the wrong place for this though. It makes more sense to
handle the redirect in the `ClaimsController#update` action or better
yet in the `PageSequence#next_slug` method, something to address in a
future commit.

This line from claims controller:
```ruby
if params[:slug] == "teaching-subject-now" && no_eligible_itt_subject?
  return redirect_to claim_path(current_journey_routing_name, "eligible-itt-subject")
end
```
mixes a couple of concerns.
This line needs to know if a dependent answer has been reset (a concern of
the Claim/Eligibility model) and the order of steps in the journey (a concern
of `PageSequence`).

Ideally we'd keep all this information in one place so the controller could
just do something like `redirect_to next_step_path`.

To make this tricker we need to handle the case where a user navigates to a
step via the browser url (see this test case
`spec/features/combined_teacher_claim_journey_dependent_answers_spec.rb:74`)
so we need to preform some slug checking in the `ClaimsController#show` action.

Options as I see them

1.) Do nothing, leave this line as is and wait for a pattern to emerge -
usually the best option!

2.) Add some logic to remove slugs from completed slugs if
`reset_dependent_answers` has invalidated a previously answered page and then
add a check to the show action that the current slug matches the expected slug
given the state of the claim. This may mean introducing a PageSequence for each
journey.

3.) Something else. I'm not super familar with the code base so there could be
another option I'm over looking.

Something to address in a future ticket

### DIFF
--- a/app/forms/journeys/additional_payments_for_teaching/teaching_subject_now_form.rb
+++ b/app/forms/journeys/additional_payments_for_teaching/teaching_subject_now_form.rb
@@ -1,0 +1,40 @@
+module Journeys
+  module AdditionalPaymentsForTeaching
+    class TeachingSubjectNowForm < Form
+      attribute :teaching_subject_now, :boolean
+
+      validates :teaching_subject_now,
+        inclusion: {
+          in: [true, false],
+          message: "Select yes if you spend at least half of your contracted hours teaching eligible subjects"
+        }
+
+      def initialize(journey:, claim:, params:)
+        super
+
+        self.teaching_subject_now = permitted_params.fetch(
+          :teaching_subject_now,
+          claim.eligibility.teaching_subject_now
+        )
+      end
+
+      def eligible_itt_subject
+        @eligible_itt_subject ||= claim.eligibility.eligible_itt_subject
+      end
+
+      def teaching_physics_or_chemistry?
+        eligible_itt_subject == "physics" || eligible_itt_subject == "chemistry"
+      end
+
+      def save
+        return false unless valid?
+
+        update!(
+          eligibility_attributes: {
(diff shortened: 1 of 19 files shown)


app/forms/journeys/additional_payments_for_teaching/teaching_subject_now_form.rb
```diff
@@ -0,0 +1,48 @@
...
+        update!(
+          eligibility_attributes: {
+            teaching_subject_now: teaching_subject_now
+          }
+        )
+      end
+
+      private
+
+      def permitted_params
```
AbigailMcP, line 41: I think this method can go away now it's in the base `Form` model

AbigailMcP (approved):
This looks good to me! Just one comment in line.

I would be tempted to leave this line:

```
if params[:slug] == "teaching-subject-now" && no_eligible_itt_subject?
  return redirect_to claim_path(current_journey_routing_name, "eligible-itt-subject")
end
```

as it is (as you've done) and wait until similar routing logic bits emerge.

PR #2679, opened 2024-04-11
CAPT-1555 The SLC CSV upload uses British DoB formatting

your:
https://dfedigital.atlassian.net.mcas.ms/browse/CAPT-1555

Context: We use the SLC (Student Loans Company) data to playback the student loan amount to users claiming Student Loan repayments and fetch the correct plan type before sending the payments to payroll. The data is uploaded via the admin portal and can trigger the execution of certain admin checks (each policy has different tasks and rules) for claims currently awaiting a decision. This was one of the latest automations to be introduced.

With this PR, we change the accepted valid DoB date format to be DD/MM/YYYY instead of MM/DD/YYYY (as previously assumed and coded in)

### DIFF
--- a/app/models/student_loans_data_importer.rb
+++ b/app/models/student_loans_data_importer.rb
@@ -18,7 +18,7 @@ class StudentLoansDataImporter < CsvImporter::Base
   private
 
   def skip_row_conditions?(row)
-    row.fetch("NINO").blank?
+    row.fetch("NINO").blank? || cast_as_date(row.fetch("Date of birth")).nil?
   end
 
   def row_to_hash(row)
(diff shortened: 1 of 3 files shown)


rjlynch (approved):
Don't have any context 😂 but these changes look safe 👍 Probably best to get Ken to take a look!

your:
Thank you, @rjlynch; I have added a small description to give more context. We can go through this next week.

app/models/student_loans_data_importer.rb
```diff
@@ -18,27 +18,25 @@ class StudentLoansDataImporter < CsvImporter::Base
...
     }
   end
 
-  def date_of_birth(dob_str)
-    return unless dob_str.present?
-
-    Date.strptime(dob_str, "%m/%d/%Y")
-  rescue Date::Error
+  def cast_as_date(string)
+    Date.strptime(string, I18n.t("date.formats.day_month_year"))
```
slorek, line 38: I think this is OK as it's unlikely we'll change, but strictly speaking if we are dealing with data from an external source we should specify their formatting rather than assuming it matches ours.

slorek (approved):

PR #2680, opened 2024-04-12
CAPT 1573/form object for eligible itt subject

rjlynch:
Extracts a form object for the `eligible_itt_subject` step.
I think the view could be cleaned up a bit more but it requires more 
substantial changes outside the scope of this ticket

There's a couple of existing issues with this page not addressed in this pr.
If there's no `available_subjects` present (or on master if there's no
`subject_symbols`) present, then an empty page can render. One scenario that
can cause this is the user choosing "No" on the "Do you spend at least half of
your contracted hours teaching eligible subjects?", reaches the ineligible page
then navigates back with the browser button. In this scenario the
`JourneySubjectEligibilityChecker#selectable_subject_symbols` method returns an
empty array as there are no eligible policies. Looking in the
`Claims::IttSubjectHelper#subject_symbols` method it also seems possible for it
to return an empty list of options for a trainee teacher, with a policy outside
the supported range, whether or not we can trigger thhis scenario in practice I
don't know!
I also considered moving the logic from
`Claims::IttSubjectHelper#subject_symbols` into the
`JourneySubjectEligibilityChecker#selectable_subject_symbols` method but opted
against doing so in case it is making a bad assumption more embedded in the
code (`JourneySubjectEligibilityChecker` is used in more places that
`subject_symbols`)

### DIFF
--- a/spec/forms/journeys/additional_payments_for_teaching/eligible_itt_subject_form_spec.rb
+++ b/spec/forms/journeys/additional_payments_for_teaching/eligible_itt_subject_form_spec.rb
@@ -65,6 +65,145 @@ RSpec.describe Journeys::AdditionalPaymentsForTeaching::EligibleIttSubjectForm, 
     end
   end
 
+  describe "#available_subjects" do
+    let(:form) do
+      described_class.new(
+        journey: additional_payments_journey,
+        claim: current_claim,
+        params: ActionController::Parameters.new
+      )
+    end
+
+    subject(:available_subjects) { form.available_subjects }
+
+    context "when qualified teacher" do
+      let(:claim) { ecp_qualified_teacher_claim }
+
+      # EarlyCareerPayments policy, claim year 2023, itt_year 2020
+      # see `lib/journey_subject_eligibility_checker.rb`
+      it do
+        is_expected.to contain_exactly(
+          "chemistry",
+          "foreign_languages",
+          "mathematics",
+          "physics"
+        )
+      end
+    end
+
+    context "when trainee teacher" do
+      context "when in ECP and LUP policy year range" do
+        let(:claim) { ecp_trainee_teacher_claim }
+
+        it do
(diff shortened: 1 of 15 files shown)


spec/forms/journeys/additional_payments_for_teaching/eligible_itt_subject_form_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+      let(:claim) { ecp_qualified_teacher_claim }
+
+      # EarlyCareerPayments policy, claim year 2023, itt_year 2020
+      # see `lib/journey_subject_eligibility_checker.rb`
+      it do
+        is_expected.to contain_exactly(
+          "chemistry",
+          "foreign_languages",
+          "mathematics",
+          "physics"
```
your, line 81: This list is likely to change dynamically if the factories are (as I remember) dynamically filling the academic year on the claim and ITT year on the eligibility. So this test might break at some point suddenly (something like this happened last year). Perhaps we can set a fixed `claim.current_year` and `claim.eligibility.itt_academic_year` for this test? Otherwise, another option is to mock and test that `subject_symbols` is being called with `claim`

spec/forms/journeys/additional_payments_for_teaching/eligible_itt_subject_form_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+          allow(current_claim).to(
+            receive(:policy_year).and_return(AcademicYear.new(2000))
+          )
+        end
+
+        it { is_expected.to be_empty }
+      end
+    end
+  end
+
```
your, line 113: I think that there are a few more instance methods on the form that we can probably test here

your (commented):
Much cleaner view; some questions/suggestions

your (approved):
LGTM

PR #2683, opened 2024-04-15
Move ITT Academic Year to a Form object

vacabor:
Transition ITT academic year to a form object

- Moves validation to `IttAcademicYearForm`
- Moves some logic from the view to the Form
- Adds a spec

### DIFF
--- a/app/forms/form.rb
+++ b/app/forms/form.rb
@@ -15,6 +15,8 @@ class Form
 
   def initialize(claim:, journey:, params:)
     super
+
+    assign_attributes(attributes_with_current_value)
   end
 
   def update!(attrs)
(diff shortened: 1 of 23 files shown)


app/forms/journeys/additional_payments_for_teaching/itt_academic_year_form.rb
```diff
@@ -0,0 +1,33 @@
+module Journeys
+  module AdditionalPaymentsForTeaching
+    class IttAcademicYearForm < Form
+      attribute :itt_academic_year
+
+      validates :itt_academic_year, presence: {message: ->(object, _) { I18n.t("activerecord.errors.models.early_career_payments_eligibilities.attributes.itt_academic_year.blank.qualification.#{object.qualification}") }}
```
slorek, line 6: Please change: move errors from the model namespace (which is a legacy from before LUPP was added to the journey) to the `additional_payments.forms` namespace. There is an `i18n_errors_path` helper method for this.

slorek (changes requested):
There are a couple of failing specs. I think at least one of them can be removed.

slorek (approved):

PR #2685, opened 2024-04-15
CAPT 1570/form object for qualification details

rjlynch:
Extracts a form object for qualification details

This PR extracts a form for the qualification details step and moves
the logic related to setting eligibility attributes from dqt records
into the form. We also remove some helper methods and move their logic into the
form object.

## Review notes
The pr is split into two commits. The first commit extracts the form object and moves the logic for setting DQT record attributes on the eligibility into the form object. It's worth paying extra attention to the form object's `save` method and making sure the logic there looks good. The second commit is just refactoring the view a bit. Probably easier to review the two commits separately as the second commit introduces a bit of "noise" into the form object.

## Additional payments journey
additional_payments_details_check

## Student loans journey
student_loans_details_check


app/controllers/claims_controller.rb
```diff
@@ -121,10 +120,6 @@ def update
       check_still_teaching_params
     else
       current_claim.attributes = claim_params
-
-      # If some DQT data was missing and the user fills them manually we need
-      # to re-populate those answers which depend on the manually-entered answer
-      set_dqt_data_as_answers if current_claim.qualifications_details_check
```
slorek, line 127: 🤔 is this not needed when DQT returns incomplete data and the user is directed to enter it manually?

app/controllers/claims_controller.rb
```diff
@@ -27,6 +27,10 @@ def show
       return redirect_to claim_path(current_journey_routing_name, "eligible-itt-subject")
     end
 
+    if params[:slug] == "qualification-details"
+      return redirect_to claim_path(current_journey_routing_name, next_slug) if current_claim.has_no_dqt_data_for_claim?
+    end
```
slorek, line 32: 🤔 we should find a solution to remove this logic from the controller. This is here to catch instances where a user has navigated to the page manually. Perhaps the form object should tell us that it can't be instantiated?

app/controllers/claims_controller.rb
rjlynch: Best I can tell I don't think this actually does anything. Quite hard to follow what with the answer reseting and all but, on master, if DQT returns missing data, eg it returns a payload like
```
      {
        trn: 123456,
        ni_number: "AB123123A",
        name: "Rick Sanchez",
        dob: "66-06-06T00:00:00",
        active_alert: false,
        state: 0,
        state_name: "Active",
}
```
then the journey just kind of breaks. It assumes the `qualification` (ultimately set by `claim.dqt_teacher_record&.route_into_teaching`) is `:postgraduate_itt`, so as `claim.dqt_teacher_record&.route_into_teaching` is returning a value, `qualification` is removed from the slug sequence. Itt year ends up being in the slug sequence as it's not returned from DQT, however as we set `itt_academic_year` on the `itt-year` page,  that resets `eligible_itt_subject`, but the slug sequence is checking for `claim.dqt_teacher_record&.eligible_itt_subject_for_claim` which returns `none_of_the_above` if no value is set ¹, so `eligible-itt-subject` is removed from the slug sequence. (apologies, this was mostly a brain dump for me!)

I'll try and write some feature specs to cover this but I'm not sure if these fixes want breaking out into a separate ticket.

https://github.com/DFE-Digital/claim-additional-payments-for-teaching/assets/9936028/6885f86a-d830-4b21-a6aa-46b6d040a70f

1. For the ECP journey, for LUP it returns `nil`

rjlynch: @slorek No, reset dependent attributes doesn't kick in if there's data from DQT
https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/901e46fa0f77da29da162e65e382dd8ffbc4ec23/app/models/policies/levelling_up_premium_payments/eligibility.rb#L93
https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/901e46fa0f77da29da162e65e382dd8ffbc4ec23/app/models/policies/early_career_payments/eligibility.rb#L124

Though there's quite a few bugs if DQT only returns partial data, eg:

https://github.com/DFE-Digital/claim-additional-payments-for-teaching/assets/9936028/d74da0c4-0824-46be-94f3-8c444239bb08

so we'll probably want to do something like only show `qualification-details` step if DQT returns sufficient data (not sure how much of a problem this is IRL)

rjlynch: There's also some assumptions made about missing DQT data I'm not sure about, eg for an ECP claim, if DQT is missing `initial_teacher_training.qualification` then it assumes the qualification is `:postgraduate_itt`, ultimately due to the `nil` [here](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/901e46fa0f77da29da162e65e382dd8ffbc4ec23/lib/dqt/matchers/general.rb#L6). There's a few other things like that such as a missing `qualified_teacher_status.qts_date` setting `eligible_itt_subject` on the claim's eligibility to `none_of_the_above` and as the value isn't nil the teacher isn't presented with the option to change it. I don't have enough business context to say if these are bugs but they do seem a bit odd!

slorek: Thanks @rjlynch - I think it would be worth creating a ticket for each of these to investigate further.

app/forms/qualification_details_form.rb
```diff
@@ -0,0 +1,159 @@
+class QualificationDetailsForm < Form
+  attribute :qualifications_details_check, :boolean
+
+  validates :qualifications_details_check,
+    inclusion: {
+      in: [true, false],
+      message: "Select yes if your qualification details are correct"
```
slorek, line 7: Please change: put this in the en.yml file in the new forms namespace.

slorek (changes requested):

app/controllers/claims_controller.rb
rjlynch: Done 👍 Only made one ticket as all the issues manifest in a single scenario https://dfedigital.atlassian.net/browse/CAPT-1624

app/forms/qualification_details_form.rb
```diff
@@ -0,0 +1,159 @@
...
+      fail "Unknown eligibility type #{eligibility.class}"
+    end
+  end
+
+  def itt_academic_year(dqt_teacher_record, eligibility)
+    if dqt_teacher_record&.itt_academic_year_for_claim
+      dqt_teacher_record.itt_academic_year_for_claim
+    else
+      eligibility.itt_academic_year
+    end
```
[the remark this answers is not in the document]
rjlynch: Is this a blocker? Just really pains me to go over 80 characters 😂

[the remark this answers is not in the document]
slorek: It's not a blocker, but in general I think we should prefer succinctness over writing extra code to conform to an arbitrary column limit which is not really relevant anymore. This is a lot of repeated redundant code across the various methods. The existing code you replaced used the suggested style. And you can still keep to 80 characters using this way if you really need to. but happy to go with the consensus here.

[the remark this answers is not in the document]
rjlynch: I just like having some constraints like ~80 chars lines as I think it has the net effect of making the code more straight forward, eg it's hard to add a difficult to read compound ternary if you've only got an 80 char line length! Have inlined these methods though 👍

app/forms/qualification_details_form.rb
```diff
@@ -0,0 +1,143 @@
...
+      # Teacher has confirmed the details in the dqt record are correct, update
+      # the eligibility with these details
+      claim.claims.each { |c| set_qualifications_from_dqt_record(c.eligibility) }
+    else
+      # Teacher has said the details don't match what they expected so
+      # nullify them
+      claim.claims.each { |c| set_nil_qualifications(c.eligibility) }
+    end
+
+    claim.save!
```
your, line 36: I think we should try to use the inherited `update!` method as much as possible, as it's going to make it easier to migrate the persistence layer for the claim (`CurrentClaim`) and minimise changes around. It might not be easy or doable right now, so just checking

app/forms/qualification_details_form.rb
```diff
@@ -0,0 +1,143 @@
...
+  def dqt_academic_date
+    AcademicYear.for(claim.dqt_teacher_record.academic_date)
+  end
+
+  # Often the DQT record will represent subject names in all lowercase
+  def dqt_itt_subjects
+    claim.dqt_teacher_record.itt_subjects.map do |subject|
+      (subject.downcase == subject) ? subject.titleize : subject
+    end.join(", ")
+  end
```
your, line 52: I don't understand why we cannot `titlteize` all of them 🤔

app/forms/qualification_details_form.rb
```diff
@@ -0,0 +1,143 @@
...
+    claim.claims.any? do |c|
+      c.dqt_teacher_record.eligible_itt_subject_for_claim == :none_of_the_above
+    end && claim.dqt_teacher_record.degree_names.any?
+  end
+
+  def dqt_degree_subjects
+    claim.dqt_teacher_record.degree_names.map do |subject|
+      (subject.downcase == subject) ? subject.titleize : subject
+    end.join(", ")
+  end
```
your, line 64: It looks like the same implementation as for `dqt_itt_subjects`

app/forms/qualification_details_form.rb
```diff
@@ -0,0 +1,143 @@
...
+        itt_academic_year: itt_academic_year(dqt_record, eligibility),
+        eligible_itt_subject: eligible_itt_subject(dqt_record, eligibility),
+        qualification: qualification(dqt_record, eligibility),
+        eligible_degree_subject: eligible_degree_subject(dqt_record, eligibility)
+      )
+    when Policies::StudentLoans::Eligibility
+      eligibility.qts_award_year = qts_award_year(dqt_record, eligibility)
+    else
+      fail "Unknown eligibility type #{eligibility.class}"
+    end
```
your, line 93: I feel a bit conflicted here; it feels like there should be at least two different forms behaving differently here, one per journey, and they should be namespaced (?) 🤔 🤷 Thoughts?

app/forms/qualification_details_form.rb
rjlynch: I'm not sure of the business rules maybe some are all in CAPS and we want to keep that format?

app/forms/qualification_details_form.rb
rjlynch: Yeah that makes sense, will split this into separate forms

app/forms/qualification_details_form.rb
rjlynch: I'm not sure if we can as on the additional payments journey we set different attributes on the ECP and LUP, with the main issue being the `eligible_degree_subject` attribute. Setting `eligible_degree_subject` requires calling `Policies::LevellingUpPremiumPayments::DqtRecord#eligible_degree_code?` which doesn't have a corresponding method on the ECP DqtRecord. Will see if there's some other way to reuse the update method though

app/forms/qualification_details_form.rb
your: The only other option I can think of is to delegate this pre-filling to something else; I can see pros and cons for both solutions, though... I am also a bit concerned that the pre-filling, in general, is a way of bypassing model validation rules that do exist if we were to input those attributes manually (when/if the slugs -forms- are presented). Raising this for consideration. The older part of me is more concerned about data consistency than coupling, but I don't see this a blocker

app/forms/qualification_details_form.rb
your: Never seen that :D Safe to keep in case, I originally thought it was for some optimisation reason but if anything it probably makes it even worse. Not a blocker

app/forms/qualification_details_form.rb
rjlynch: Yeah that works much better as separate form objects 👍

rjlynch:
@your I've split out the form in to a separate one for each journey, much nicer thanks! Not sure how to avoid looping through the eligibilities in the additional payments journey as the policies have different attributes so we can't just pass `eligibility_attributes` to the base form's update method. Any suggestions?

your:
@rjlynch 
> Not sure how to avoid looping through the eligibilities in the additional payments journey as the policies have different attributes so we can't just pass eligibility_attributes to the base form's update method. Any suggestions?

We could check if the model `has_attribute?` or `respond_to?` when constructing an `eligibility_attributes` hash to pass there... I would have to try it and see how it looks. If you don't like it let me know, I think we can still get this merged otherwise

rjlynch:
> @rjlynch
> 
> > Not sure how to avoid looping through the eligibilities in the additional payments journey as the policies have different attributes so we can't just pass eligibility_attributes to the base form's update method. Any suggestions?
> 
> We could check if the model `has_attribute?` or `respond_to?` when constructing an `eligibility_attributes` hash to pass there... I would have to try it and see how it looks. Maybe it just moves the problem to another place... If you don't like it let me know, I think we can still get this merged otherwise

I'm not sure. I think the base form is the right place for that sort of thing to hide the fact we're working with multiple claims from the child forms, however I think it's going to get a bit weird as we'll have to delete eligibility attributes if one of the claim's eligibilities doesn't have a specific attribute and I'm not sure how we could handle the case where you want to set the same attribute to different values on the eligibility (not sure if that's ever the case but would be hard to handle). I'll take a look at what's involved with doing this, but I think it's probably better as a separate pr so the rest of the team can give feedback on the approach without the noise of the other changes.

your (approved):

rjlynch:
> > @rjlynch
> > > Not sure how to avoid looping through the eligibilities in the additional payments journey as the policies have different attributes so we can't just pass eligibility_attributes to the base form's update method. Any suggestions?
> > 
> > 
> > We could check if the model `has_attribute?` or `respond_to?` when constructing an `eligibility_attributes` hash to pass there... I would have to try it and see how it looks. Maybe it just moves the problem to another place... If you don't like it let me know, I think we can still get this merged otherwise
> 
> I'm not sure. I think the base form is the right place for that sort of thing to hide the fact we're working with multiple claims from the child forms, however I think it's going to get a bit weird as we'll have to delete eligibility attributes if one of the claim's eligibilities doesn't have a specific attribute and I'm not sure how we could handle the case where you want to set the same attribute to different values on the eligibility (not sure if that's ever the case but would be hard to handle). I'll take a look at what's involved with doing this, but I think it's probably better as a separate pr so the rest of the team can give feedback on the approach without the noise of the other changes.

@your I took another look at this. Tried updating current claim `update!` method to hide the fact that this form deals with multiple claims by checking if a given claim's eligibility had an attribute on it before trying to save the eligibility but there's a couple of issues with doing that, at least for this PR.
1.) We call methods in the form that are specific to a given policy, eg `dqt_teacher_record&.eligible_degree_code?`. So even if we move the update logic this form still has to know about there being multiple claims.
2.) I think we may risk subtlety changing the update logic here if we treat both the ECP and LUP claims as the same (though probably not an issue in practice)

Not sure what the right approach is so leaving this as is for the minute because at least it keeps the weirdness in a single file rather than polluting current_claim. Hopefully once the rest of the forms are done and we come back to this we'll have a better idea of what to do

slorek (approved):

PR #2686, opened 2024-04-16
CAPT-1617 Form object for /eligible-degree-subject

your:
https://dfedigital.atlassian.net/browse/CAPT-1617

This slug is presented only to LUPP trainees who might be eligible in the future. To test it, follow:

Select a school with a LUPP award -> select "No, I'm a trainee teacher" -> select "None of the above"

As part of this PR, I'm changing the initialiser in the base form so that it can fetch pre-existing answers when they are not present in the params, instead of leaving that to each form initialiser. Please check the commits for a more detailed breakdown of the changes


spec/support/feature_helpers.rb
```diff
@@ -32,7 +32,7 @@ def choose_school(school)
     click_button "Continue"
 
     # flaky test workaround in case the first click on Continue submitted the form
-    click_button "Continue" unless /(claim|current)-school\?_method=patch/.match?(current_url)
+    click_button "Continue" unless /(claim|current)-school\?_method=get/.match?(current_url)
```
rjlynch, line 35: Is this related to the changes in the pr or is it just a test fix?

rjlynch (commented):
Looks good, really like that we can remove the initialiser from most forms. Let me know when this is ready for review and I'll take another look 👍

spec/support/feature_helpers.rb
your: It shouldn't have been there; I was playing with a test during the rebase. Corrected, thanks

app/forms/form.rb
```diff
@@ -60,4 +62,32 @@ def page_sequence
...
+      attributes[attribute] = load_current_value(attribute)
+    end
+  end
+
+  def load_current_value(attribute)
+    # TODO: re-implement when the underlying claim and eligibility data sources
+    # are moved to an alternative place e.g. a session hash
+
+    # Some, but not all attributes are present directly on the claim record.
+    return claim.public_send(attribute) if claim.has_attribute?(attribute)
```
AbigailMcP, line 84: Currently thinking about this in the `PersonalDetailsForm` - there are attributes on the form (`day`, `month`, `year` which don't map directly to an attribute on the claim (they need to be derived from `date_of_birth`).

As it stands, this method would return something unexpected if no attribute is found on either the `claim` or its `eligibility` - an array of claims!

your: @AbigailMcP just realised the same as I started working on `eligibility-confirmed`, so I moved this back to draft; will think about a way around it

your: Quick thought, there should be a `nil` after the loop, I must have lost it in one of the many iterations

your: Don't make it a blocker for your branch, I can always bring in your changes in and deal with it if this gets stuck for a bit longer – appreciate the heads up, thanks! 🙇

AbigailMcP: no worries! not a blocker for mine, i'm also mulling things over and this approach is really useful to see.

your: I should have addressed it now – I was also considering making the pre-filling "optional", and for non-standard use cases we can always keep overriding `initialize`

PR #2690, opened 2024-04-17
Form object for teacher detail

rjlynch:
Probably best reviewing each commit separately.

Introduces a form object for the teacher details page.
As part of this work introduces a form to move capturing the details from Dfe identity and storing them in the claim's `teacher_id_user_info` column out of the claims controller and into the Dfe identity oauth call back controller. Whether or not we want to store the details on the model before they're confirmed is [being decided](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2690#discussion_r1570647909) this pr preserves the original behaviour, just moves it all into one place.
There's some other questions around what we should do when a teacher says the details don't match what they expect, we should probably address those (in a separate ticket). Currently if a teach says the details from Dfe identity don't match what they expected we keep the details in the `teacher_id_user_info` column but we don't set other attributes on the claim from them. `teacher_id_user_info` is also used in `Claim#has_all_valid_personal_details?` so moving the storing of unapproved Dfe identity attributes off of the model and into the session will effect other form object like personal details.

## Re-sign in with Dfe identify resets details check
https://github.com/DFE-Digital/claim-additional-payments-for-teaching/assets/9936028/4d8b2817-20ad-49ed-a65b-fbe2fe9d4b97

## Navigating back and choosing skip sign in resets details check
https://github.com/DFE-Digital/claim-additional-payments-for-teaching/assets/9936028/b91dc567-9628-49d6-9e26-ecdda330d1aa


app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -4,7 +4,13 @@ class OmniauthCallbacksController < ApplicationController
   def callback
     auth = request.env["omniauth.auth"]
 
-    session[:user_info] = auth.extra.raw_info
+    DfeIdentityCallbackForm.new(
```
slorek, line 7: Suggestion: It's odd to use a form object for callback handling - perhaps we should call this `SignInWithDfeIdentityForm` to reflect that it's part of the handling of the preceding form where the user selects where or not to sign in. That would make a bit more sense to me.

However, I'm not sure there's any need to set `teacher_id_user_info` in the claim unless the user has confirmed their details are correct on the `teacher-detail` page. I think in the case they select 'No' we would not need to retain it. So potentially we could just keep this in the session until we decide to store it on the claim (on the `teacher-detail` form).

rjlynch: `SignInWithDfeIdentityForm` - Yeah I think that's a better name 👍 

> However, I'm not sure there's any need to set teacher_id_user_info in the claim unless the user has confirmed their details are correct on the teacher-detail page

Yeah it is a bit odd storing this before confirming it's correct, though that was the [previous behaviour](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/4a9320e66b55dd03f91766061aadaef5eb031e5f/app/controllers/claims_controller.rb#L39) before this form was introduced and it seems like we don't clear the `teacher_id_user_info` when [details are incorrect](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/4a9320e66b55dd03f91766061aadaef5eb031e5f/app/models/dfe_identity/claim_user_details_reset.rb#L14). It looks like we always stored this info on the claim in a json column but only transferred it over to the claim's attributes proper (eg `first_name`, `surname`) if the teacher confirmed their details. Not sure if that's something we want to keep doing?

rjlynch: 🤔 An option to avoid storing these details in the claim until they're approved* is to redirect to the "teacher-details" page with the values from oauth as parameters

```ruby
    redirect_to claim_path(
      journey: current_journey_routing_name,
      slug: "teacher-detail",
      claim: {
        details_check: nil,
        logged_in_with_tid: true,
        teacher_id_user_info: auth.extra.raw_info.to_h.with_indifferent_access
      }
    )
```
then the `TeacherDetailForm` can read the values from the params, present them to the Teacher for them to confirm and store them on the claim if they do / loose them if they don't. This might be the most "railsy" way to do this as we're not changing any data in a GET request, though I'm not sure about GDPR with putting these details in the url.


*though I think that should be a separate PR as it seems `teacher_id_user_info` is used elsewhere

slorek: Clever idea, but I'm a bit worried about logging as it would be a GET request, and we won't control the list of parameters that could slip through filtering.

slorek: > Yeah it is a bit odd storing this before confirming it's correct, though that was the [previous behaviour](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/4a9320e66b55dd03f91766061aadaef5eb031e5f/app/controllers/claims_controller.rb#L39) before this form was introduced and it seems like we don't clear the `teacher_id_user_info` when [details are incorrect](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/4a9320e66b55dd03f91766061aadaef5eb031e5f/app/models/dfe_identity/claim_user_details_reset.rb#L14). It looks like we always stored the teacher_id_user_info on the claim in a json column but only transferred it over to the claim's attributes proper (eg `first_name`, `surname`) if the teacher confirmed their details. Not sure if that's something we want to keep doing? Didn't want to change any behaviour in this pr!

I don't think there is any need to store it in that scenario, and if it makes it easier and makes more sense then we should change it.

rjlynch: > I don't think there is any need to store it in that scenario, and if it makes it easier and makes more sense then we should change it.

Do you want this doing as part of this PR or do you want me to do it as a separate ticket? My thoughts on doing this as a separate ticket is that we use `current_claim.teacher_id_user_info["email"]` in [a few places](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/cbef1edd799cc32bc79b409e69ef5acaec39c53a/app/controllers/claims_controller.rb#L48) but this doesn't depend on whether or not the details are confirmed as correct or not, as `teacher_id_user_info` [isn't cleared if the teacher says the details don't match](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/cbef1edd799cc32bc79b409e69ef5acaec39c53a/app/models/dfe_identity/claim_user_details_reset.rb#L2) so I'm unsure on the size of the changes required to clear this. Happy to do this either as part of this PR or a separate ticket, which ever you prefer 👍

rjlynch:
@kenfodder Some context for our call
# Context
When a teacher signs in with DfE identity we set the `logged_in_with_tid` flag
to `true`. The `logged_in_with_tid` flag is checked by the [slug sequence](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/cbef1edd799cc32bc79b409e69ef5acaec39c53a/app/models/journeys/additional_payments_for_teaching/slug_sequence.rb#L105) to
determine what slugs are in sequence, so this flag needs to be set _before_ the
slug sequence is created.

## Current approach
The current approach is:
* [Store extra info in the session when returning from DfE identity](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/cbef1edd799cc32bc79b409e69ef5acaec39c53a/app/controllers/omniauth_callbacks_controller.rb#L7)
* [Check the session for presence of extra info in  a callback in the claims controller](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/cbef1edd799cc32bc79b409e69ef5acaec39c53a/app/controllers/claims_controller.rb#L6)
* [If the extra info is present, set the logged_in_with_tid flag on the claim](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/cbef1edd799cc32bc79b409e69ef5acaec39c53a/app/models/dfe_identity/claim_user_details_reset.rb#L34)

### Pros
* Nothing is updated in the omniauth callback controller (GET request)

### Cons
* Callback specific to a single step is fired on every action in the claim's
  controller
* Updating the claim with info from DfE identity happens in two places, the callback and [the teacher-details step](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/cbef1edd799cc32bc79b409e69ef5acaec39c53a/app/controllers/claims_controller.rb#L334)

## Approach in pr #2690
* [Set the logged_in_with_tid flag on the claim and store additional attributes on the claim in the omniauth callback controller](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/9fdbc64600fc6f8cd1cf2f52fa347c34006c99bb/app/forms/sign_in_with_dfe_identity_form.rb#L2)

### Pros
* All updates to the claim related to info from DfE identity happen in a single
  place
* Remove a before action from the claims controller

### Cons
* Claim is updated in a GET request in the oauth callback controller

## Other issues
In both approaches we store the info from DfE identity on the claim record
(`teacher_id_user_info`) before the teacher has confirmed that the details
match what they expect. Ideally we'd only store this info once a match is
confirmed and otherwise throw it away. This might be out of scope for
refactoring the login flow, as `teacher_id_user_info` is used in a few places 
without checking if it's been confirmed.

rjlynch:
Superseded by https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2701

PR #2691, opened 2024-04-17
Move eligible-later view to a Form object

vacabor:
- Remove references to current_claim and exchange for an empty Form object
- Very minimal view change but helps to follow the new Form object pattern

### DIFF
--- a/app/forms/journeys/additional_payments_for_teaching/eligible_later_form.rb
+++ b/app/forms/journeys/additional_payments_for_teaching/eligible_later_form.rb
@@ -1,0 +1,6 @@
+module Journeys
+  module AdditionalPaymentsForTeaching
+    class EligibleLaterForm < Form
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


slorek (commented):
🤔 I'd like to see others' thoughts on whether a form object for a page without a form is a useful change. I'm not so sure it's adding any value.

AbigailMcP:
> 🤔 I'd like to see others' thoughts on whether a form object for a page without a form is a useful change. I'm not so sure it's adding any value.

I went the other way with a previous PR and removed a form that wasn't needed: https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2677

rjlynch:
>  I'd like to see others' thoughts on whether a form object for a page without a form is a useful change. I'm not so sure it's adding any value.

If we're not updating / creating anything I don't think we need to add a form here.

PR #2692, opened 2024-04-18
CAPT-1576 Form object for /eligibility-confirmed

your:
https://dfedigital.atlassian.net/browse/CAPT-1576

As part of this PR, I'm experimenting with injecting the session into the form object in question and creating a simple or dummy form for pages with forms that do not submit anything in particular. Please let me know what you think.

The eligibility-confirmed will (probably) always be a simple page with content to confirm, but for the combined journey (ECP/LUPP), the user has the option to choose the policy to claim, were they found eligible for both, so it's a bit more involved than the average use case for this page

Temporarily branching this off #2686


app/forms/journeys/additional_payments_for_teaching/eligibility_confirmed_form.rb
```diff
@@ -0,0 +1,42 @@
...
+      def initialize(**)
+        super
+
+        self.selected_claim_policy = permitted_params.fetch(:selected_claim_policy, session[:selected_claim_policy])
+      end
+
+      def save
+        return false unless valid?
+
+        session[:selected_claim_policy] = selected_claim_policy
```
rjlynch, line 25: I feel like we're missing a table, possibly a `User/Teacher`, that we would be updating the selected claim on. Obviously that's quite a big change so this is fine!

rjlynch (commented):
I think this is pretty easy to follow so it think it's fine and it's a big improvement on accessing the session in the view! Though passing the session into the form object _feels_ a bit odd, as manipulating the session in the form is tricky and the session is a responsability of the http layer. I can't think of a better solution without introducing a new model that we'd hang the multiple claims off.

rjlynch (approved):
Nice, I think this is the most sensible approach 👍

PR #2693, opened 2024-04-18
CAPT-1561 - Form object for sign-in-or-continue

kenfodder:

### DIFF
--- a/app/forms/sign_in_or_continue_form.rb
+++ b/app/forms/sign_in_or_continue_form.rb
@@ -1,0 +1,43 @@
+class SignInOrContinueForm < Form
+  # - Sign in with Teacher ID does NOT save or interact with this Form and POSTs straight to Teacher ID
+  # - If Teacher ID is disabled for the journey, we still mimick if a user clicked on "Continue without signing in".
+
+  def initialize(claim:, journey:, params:)
+    super
+
+    skip_form_if_teacher_id_not_enabled_for_journey
+  end
+
+  def save
+    continue_without_teacher_id
+
+    true
+  end
+
+  # This is needed because logging in with Teacher ID doesn't submit to this form.
+  # Therefore the slug/page needs to be considered submitted to avoid it being the next slug
+  # after callback redirect back to Claim from Teacher ID.
+  def force_update_session_with_current_slug
+    true
+  end
+
+  private
+
+  def skip_form_if_teacher_id_not_enabled_for_journey
+    if !journey.configuration.teacher_id_enabled?
+      update_claim_with_teacher_id_was_skipped
+      @redirect_to_next_slug = true
+    end
+  end
+
+  def continue_without_teacher_id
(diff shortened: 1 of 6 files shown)


app/forms/sign_in_or_continue_form.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def save
+    continue_without_teacher_id
+
+    true
+  end
+
+  # This is needed because logging in with Teacher ID doesn't submit to this form.
+  # Therefore the slug/page needs to be considered submitted to avoid it being the next slug
+  # after callback redirect back to Claim from Teacher ID.
```
rjlynch, line 19: What happens if we remove the `sign-in-or-continue` slug from the sequence and start the journey at `teacher-details`? We'd need to probably need to point the "sign-in-or-continue" form at it's own controller to handle some stuff around reseting details from DfE identify when skip teacher id is true but it might help us simplify the claims controller. Probably too much effort at the minute just an idea!

rjlynch (commented):
What happens if we remove the sign-in-or-continue slug from the sequence and start the journey at teacher-details? We'd need to probably need to point the "sign-in-or-continue" form at it's own controller to handle some stuff around reseting details from DfE identify when skip teacher id is true but it might help us simplify the claims controller. Probably too much effort at the minute just an idea!
Happy to approve this pr though, I think it's better than having all this in the claim's controller. Just giving others a chance to see what they think

app/forms/sign_in_or_continue_form.rb
slorek, line 19: Untested theory, but I wonder if this form should be combined with @rjlynch 's work - see my comment here https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2690#discussion_r1570647909

I think by incorporating the callback functionality into this form object we might be able to avoid these workarounds.

rjlynch: I'm unsure about combining the forms at the minute but I think we could move the logic for updating the slug to [here](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/9fdbc64600fc6f8cd1cf2f52fa347c34006c99bb/app/controllers/omniauth_callbacks_controller.rb#L14) in the call back controller to avoid this form having to care about it.
That controller already knows about the pages in the journey (it redirects to a specific page) so while not ideal I don't think it does much harm modifying the session slugs there.

rjlynch (approved):
@slorek discussed an approach with Ken. We think it's best to merge this pr as it incrementally improves the claims controller and moves related logic together.
Ultimately we think pulling out the sign in page into its own controller as it doesn't _feel_ related to the slug sequence (it's not really part of a wizard) is the way to go. This will also let us remove the callback controller changes in https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2690 . I'll take a look at what's involved in pulling out the sign in controller.

PR #2695, opened 2024-04-18
Remove requirement to list all form objects and their slug names

slorek:
Currently, every time we add a form object we have to update the `Journey::JourneyName::FORMS` and/or `Form::SHARED_FORMS` constants with the corresponding slug name mapping to the class.

In the spirit of convention over configuration, this change removes the need to do this. Instead, the form object is resolved dynamically based on the slug name.

Slug names resolve directly to class names, e.g. `current-school` slug resolves to `CurrentSchoolForm`.

As currently, a journey can still override a form object class with the same name as a shared class in its namespace, and it will load the class in the journey namespace in preference to the shared one. This allows for shared forms with journey-specific behaviour overrides.

Shared forms still work as currently.

Note: Given the slug name is a user supplied parameter it is unsafe, therefore I have retained the current 'form object registry' pattern rather than dynamically loading a class from the slug name.

### DIFF
--- a/app/forms/form.rb
+++ b/app/forms/form.rb
@@ -9,8 +9,22 @@ class Form
 
   delegate :persisted?, to: :claim
 
-  def self.model_name
-    Claim.model_name
+  class << self
+    def model_name
+      Claim.model_name
+    end
+
+    def slug_name
+      name.demodulize.delete_suffix("Form").underscore.dasherize
+    end
+
+    # Returns form classes outside of any journey-specific namespace
+    def all_shared_forms
+      ObjectSpace
+        .each_object(Class)
+        .select { |klass| klass < self }
+        .reject { |klass| klass.to_s.start_with?("Journeys::") }
+    end
   end
 
   def initialize(claim:, journey:, params:)
--- a/app/forms/form.rb
+++ b/app/forms/form.rb
@@ -20,7 +20,7 @@ class Form
 
     # Returns form classes outside of any journey-specific namespace
     def all_shared_forms
-      ObjectSpace
+      @@all_shared_forms ||= ObjectSpace
         .each_object(Class)
         .select { |klass| klass < self }
         .reject { |klass| klass.to_s.start_with?("Journeys::") }
--- a/app/models/journeys/base.rb
+++ b/app/models/journeys/base.rb
@@ -1,11 +1,5 @@
 module Journeys
   module Base
-    # TODO: move app/forms/*_forms to shared and journey specific folders
-    # but needs load_paths sorting
-    SHARED_FORMS = {
-      "current-school" => CurrentSchoolForm
-    }
-
     def configuration
       Configuration.find(self::ROUTING_NAME)
     end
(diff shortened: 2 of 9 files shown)


app/forms/form.rb
```diff
@@ -9,8 +9,22 @@ class Form
...
+      name.demodulize.delete_suffix("Form").underscore.dasherize
+    end
+
+    # Returns form classes outside of any journey-specific namespace
+    def all_shared_forms
+      ObjectSpace
+        .each_object(Class)
+        .select { |klass| klass < self }
+        .reject { |klass| klass.to_s.start_with?("Journeys::") }
+    end
```
rjlynch, line 27: How does this perform? I think these loops are going to do quite a few iterations. Maybe best to do this sort of thing in an initialiser so it doesn't have to be done on each request?

rjlynch:
> Note: Given the slug name is a user supplied parameter it is unsafe, therefore I have retained the current 'form object registry' pattern rather than dynamically loading a class from the slug name.

Could we look at the constants in the slug sequence to know if the provide slug is a known value?

```ruby
def get_form(slug_from_params)
matching_journeys = [
  Journeys::TeacherStudentLoanReimbursement,
  Journeys::AdditionalPaymentsForTeaching
].select { |journey| journey::SLUGS.include?(slug_from_params) }

case matching_journeys.size
when 0 then raise "no tampering"
when 1 then "#{matching_journey.first}::#{slug_from_param}Form".constantise
when 2 then "#{slug_from_params}Form".constantize
else fail "Unsupported number of journeys
end
rescue NameError
nil
end
```

(though tbh I don't think updating a constant with a slug to form mapping is too bad!)

Just a thought

app/forms/form.rb
slorek: Good shout, it was actually extremely slow! I've fixed it by using class variables.

slorek:
@rjlynch I think this is covered by https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/master/config/routes.rb#L26-L45 but I still feel uneasy about trusting the input to load any available class.

app/models/journeys/base.rb
```diff
@@ -19,13 +13,16 @@ def slug_sequence
     end
 
     def form(claim:, params:)
-      form = SHARED_FORMS.merge(forms)[params[:slug]]
-
+      form = all_forms[params[:slug]]
       form&.new(journey: self, claim: claim, params: params)
     end
```
your, line 18: I was playing with this last week and this is what I was doing:

```ruby
    def form(claim:, params:, session:)
      form_class_for_slug(params[:slug])&.new(journey: self, claim:, params:, session:)
    end

    def form_class_for_slug(slug)
      # Search within the journey namespace, fallback to root otherwise
      ["#{self}::", "::"]
        .zip([slug.titleize.delete(" ").concat("Form")].cycle(2))
        .lazy.map(&:join).map(&:safe_constantize).compact.first
    end
```

I think your approach is also valid and potentially faster now. I definitely didn't test speed, etc. Just dropping it here as what was interesting is that this was the only change to make this work

your (approved):
Nice, dropping an alternative approach in the comments but this is solid as well

app/models/journeys/base.rb
slorek: That's interesting (and less code is good); I took an early decision to not start with the parameter as it was unsafe and could potentially be open for exploitation, but maybe it is safe enough given the filtering in the routes file.

slorek:
Closing as the codebase has evolved significantly since this PR was opened and there were outstadning issues. It doesn't appear to be worthwhile getting it to work at this point in time.

PR #2696, opened 2024-04-19
CAPT-1563 Form object for /personal-details

AbigailMcP:
Extracting the /personal-details into a form object.

### Points to note

**Personal details validations being called from slug sequences** 

Before these changes, there was a method on the `Claim` model `has_all_valid_personal_details?` which checked whether the personal details were valid, and that they matched TID:

https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2696/files#diff-177255fb498b99ec75c3866174729f24e11d030819b91655a5913b28669505c2R505-R507

This was called from the SlugSequences to determine whether to remove `/personal-details` completely.

https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/master/app/models/journeys/additional_payments_for_teaching/slug_sequence.rb#L160

Since this PR removes personal details validations from the model and shifts them into the form, I've had to instantiate a PersonalDetailsForm in both SlugSequences in order to perform the check.

**Before_save normalisations** 
The `before_save`s  like `normalise_first_names` are still in the `Claim` model

**Error translations**
I wasn't sure whether to shift these into locales because they are identical for both journeys so would lead to a lot of repetition in en.yml.

**Date of birth validation**
I've  tidied up the `date_of_birth` validation by introducing an `InvalidDate` struct in the form. This way the attribute `date_of_birth` always exists on the form and you can always call `day`, `month` and `year` on it.

### DIFF
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -345,7 +345,7 @@ class ClaimsController < BasePublicController
     # For claims being submitted using the TID-route and where all personal details came through/are
     # valid, the student loan details must be retrieved after the `information-provided` page instead.
     if params[:slug] == "personal-details" || (params[:slug] == "information-provided" &&
-        current_claim.logged_in_with_tid? && current_claim.has_all_valid_personal_details?)
+        current_claim.logged_in_with_tid? && current_claim.all_personal_details_same_as_tid?)
       ClaimStudentLoanDetailsUpdater.call(current_claim)
     end
   end
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -332,9 +332,6 @@ class ClaimsController < BasePublicController
     current_claim.claims.each { |claim| claim.eligibility.set_qualifications_from_dqt_record }
   end
 
-  # TODO: should calling `ClaimStudentLoanDetailsUpdater.call(current_claim)` be the responsibility
-  # of the PersonalDetailsForm? Then when we have a InformationProvidedForm it could use a
-  # PersonalDetailsForm to check the validity?
   def retrieve_student_loan_details
     # student loan details are currently retrieved for TSLR and ECP/LUPP journeys only
     return unless ["student-loans", "additional-payments"].include?(current_journey_routing_name)
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -75,6 +75,7 @@ class ClaimsController < BasePublicController
     # TODO: Migrate the remaining slugs to form objects.
     if (@form = journey.form(claim: current_claim, params: params))
       if @form.save
+        retrieve_student_loan_details
         redirect_to claim_path(current_journey_routing_name, next_slug)
       else
         set_any_backlink_override
@@ -87,8 +88,6 @@ class ClaimsController < BasePublicController
     case params[:slug]
     when "qualification-details"
       set_dqt_data_as_answers
-    when "personal-details"
-      check_date_params
     when "eligibility-confirmed"
       return select_claim if current_journey_routing_name == "additional-payments"
     when "personal-bank-account", "building-society-account"
@@ -177,17 +176,6 @@ class ClaimsController < BasePublicController
 
   def claim_params
     params.fetch(:claim, {}).permit(Claim::PermittedParameters.new(current_claim).keys)
-  end
-
-  def check_date_params
-    dob_params = {
-      "date_of_birth_day" => claim_params.dig("date_of_birth(3i)").to_s,
-      "date_of_birth_month" => claim_params.dig("date_of_birth(2i)").to_s,
-      "date_of_birth_year" => claim_params.dig("date_of_birth(1i)").to_s
-    }
-    current_claim.attributes = claim_params.merge!(dob_params)
-  rescue ActiveRecord::MultiparameterAssignmentErrors
-    current_claim.attributes = claim_params.except("date_of_birth(3i)", "date_of_birth(2i)", "date_of_birth(1i)") if params[:slug] == "personal-details"
   end
 
   def current_template
@@ -344,6 +332,9 @@ class ClaimsController < BasePublicController
     current_claim.claims.each { |claim| claim.eligibility.set_qualifications_from_dqt_record }
   end
 
+  # TODO: should calling `ClaimStudentLoanDetailsUpdater.call(current_claim)` be the responsibility
+  # of the PersonalDetailsForm? Then when we have a InformationProvidedForm it could use a
+  # PersonalDetailsForm to check the validity?
   def retrieve_student_loan_details
     # student loan details are currently retrieved for TSLR and ECP/LUPP journeys only
     return unless ["student-loans", "additional-payments"].include?(current_journey_routing_name)
--- a/app/forms/personal_details_form.rb
+++ b/app/forms/personal_details_form.rb
@@ -58,19 +58,16 @@ class PersonalDetailsForm < Form
     update!({first_name:, middle_name:, surname:, date_of_birth:, national_insurance_number:})
   end
 
-  def has_valid_name?
-    valid?
-    errors.exclude?(:first_name) && errors.exclude?(:surname)
+  def show_name_section?
+    !(claim.logged_in_with_tid? && claim.name_same_as_tid? && has_valid_name?)
   end
 
-  def has_valid_date_of_birth?
-    valid?
-    errors.exclude?(:date_of_birth)
+  def show_date_of_birth_section?
+    !(claim.logged_in_with_tid? && claim.dob_same_as_tid? && has_valid_date_of_birth?)
   end
 
-  def has_valid_nino?
-    valid?
-    errors.exclude?(:national_insurance_number)
+  def show_nino_section?
+    !(claim.logged_in_with_tid? && claim.nino_same_as_tid? && has_valid_nino?)
   end
 
   private
@@ -115,5 +112,20 @@ class PersonalDetailsForm < Form
 
   def number_of_date_components
     [day, month, year].compact_blank.size
+  end
+
+  def has_valid_name?
+    valid?
+    errors.exclude?(:first_name) && errors.exclude?(:surname)
+  end
+
+  def has_valid_date_of_birth?
+    valid?
+    errors.exclude?(:date_of_birth)
+  end
+
+  def has_valid_nino?
+    valid?
+    errors.exclude?(:national_insurance_number)
   end
 end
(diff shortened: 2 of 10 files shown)


app/views/claims/personal_details.html.erb
```diff
@@ -1,14 +1,14 @@
...
-    <%= render("shared/error_summary", instance: current_claim, errored_field_id_overrides: { date_of_birth: "claim_date_of_birth_3i"}) if current_claim.errors.any? %>
-    <%= form_for current_claim, url: claim_path(current_journey_routing_name) do |form| %>
+    <%= render("shared/error_summary", instance: @form, errored_field_id_overrides: { date_of_birth: "claim_date_of_birth_3i"}) if @form.errors.any? %>
+    <%= form_for @form, url: claim_path(current_journey_routing_name) do |form| %>
       <h1 class="govuk-heading-xl">
         <%= t("questions.personal_details") %>
       </h1>
 
-      <% unless current_claim.logged_in_with_tid? && current_claim.has_valid_name? && current_claim.name_same_as_tid? %>
+      <% unless current_claim.logged_in_with_tid? && @form.has_valid_name? && current_claim.name_same_as_tid? %>
```
AbigailMcP, line 11: I thought about reducing these further into helper methods like `<% if show_name_section? %>` but they're a mix of form validations and TID comparisons (still in the model).

app/controllers/claims_controller.rb
```diff
@@ -84,6 +84,7 @@ def update
     # TODO: Migrate the remaining slugs to form objects.
     if (@form = journey.form(claim: current_claim, params: params))
       if @form.save
+        retrieve_student_loan_details
```
AbigailMcP, line 87: See TODO comment on `def retrieve_student_loan_details` below.

rjlynch:
> Currently there's a method on the Claim model has_all_valid_personal_details? which checks whether the personal details are all valid, and that they all match TID:

I think this line could just be
```ruby
sequence.delete("personal-details") if claim.details_check?
```

If the teacher didn't log in with Dfe identity, then `details_check` is `nil` and we'll want to show the "personal-details" page.
If the teacher did log in with Dfe identity but the details are wrong, then `details_check` is false and we'll want to show the "personal-details" page.
If the teacher did log in with Dfe identity and the details are correct but invalid, then `details_check` is false and we'll want to show the "personal-details" page.
If the teacher did log in with Dfe identity and the details are correct and valid, then `details_check` is true and we don't want to show the "personal-details" page as we already have that info.
[setting details check logic](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/9fdbc64600fc6f8cd1cf2f52fa347c34006c99bb/app/forms/teacher_detail_form.rb#L35) - same logic as on master just a bit easier to show here as it's in one place.

I'd double check my assumptions though!

app/forms/personal_details_form.rb
```diff
@@ -0,0 +1,113 @@
...
+  attribute :day
+  attribute :month
+  attribute :year
+  attribute :date_of_birth
+  attribute :national_insurance_number
+
+  validates :first_name, presence: {message: "Enter your first name"}
+  validates :first_name,
+    length: {maximum: 100, message: "First name must be less than 100 characters"},
+    format: {with: NAME_REGEX_FILTER, message: "First name cannot contain special characters"},
```
rjlynch, line 22: I think this is fine as is but might be tempted to pull out Ana activemodel validator for this REGEX. Not a blocker though!

app/forms/personal_details_form.rb
```diff
@@ -0,0 +1,113 @@
...
+    if: -> { first_name.present? }
```
rjlynch, line 23: Nice 👍

app/forms/personal_details_form.rb
```diff
@@ -0,0 +1,113 @@
...
+    format: {with: NAME_REGEX_FILTER, message: "Middle names cannot contain special characters"},
+    if: -> { middle_name.present? }
+  validates :surname, presence: {message: "Enter your last name"}
+  validates :surname,
+    length: {maximum: 100, message: "Last name must be less than 100 characters"},
+    format: {with: NAME_REGEX_FILTER, message: "Last name cannot contain special characters"},
+    if: -> { surname.present? }
+  validate :date_of_birth_criteria
+  validates :national_insurance_number, presence: {message: "Enter a National Insurance number in the correct format"}
+  validate :ni_number_is_correct_format
```
rjlynch, line 35: Another one that might make a good candidate for an active model validation class - again I think this is fine as is

app/forms/personal_details_form.rb
```diff
@@ -0,0 +1,113 @@
...
+  def normalised_ni_number
+    national_insurance_number.gsub(/\s/, "").upcase
+  end
+
+  def date_of_birth_criteria
+    if date_of_birth.future?
+      errors.add(:date_of_birth, "Date of birth must be in the past")
+    elsif number_of_date_components.between?(1, 2)
+      errors.add(:date_of_birth, "Date of birth must include a day, month and year in the correct format, for example 01 01 1980")
+    elsif !date_of_birth.is_a?(Date) && number_of_date_components == 3
```
rjlynch, line 97: Is there away to avoid this type check? Will an `InvalidDate` have all date_components set?

app/models/invalid_date.rb
```diff
@@ -0,0 +1,9 @@
+class InvalidDate
```
rjlynch, line 1: This might be better inlined in the form as it's an internal detail of how the form works.

```ruby
class PersonalDetailsForm < Form
  class InvalidDate
```
What do you think about using a struct for this? Would it work?
```ruby
InvalidDate = Struct.new(:day, :month, :year, keyword_init: true) do
  def future?
    false
   end
end
```

rjlynch (commented):
Overall looks good to me 👍
Will approve once the question about how best to handle the slug sequence is sorted.

app/models/invalid_date.rb
AbigailMcP: ah this is how i had it an hour ago! agreed 👍

app/forms/personal_details_form.rb
```diff
@@ -0,0 +1,119 @@
...
+    length: {maximum: 61, message: "Middle names must be less than 61 characters"},
+    format: {with: NAME_REGEX_FILTER, message: "Middle names cannot contain special characters"},
+    if: -> { middle_name.present? }
+  validates :surname, presence: {message: "Enter your last name"}
+  validates :surname,
+    length: {maximum: 100, message: "Last name must be less than 100 characters"},
+    format: {with: NAME_REGEX_FILTER, message: "Last name cannot contain special characters"},
+    if: -> { surname.present? }
+  validate :date_of_birth_criteria
+  validates :national_insurance_number, presence: {message: "Enter a National Insurance number in the correct format"}
```
AbigailMcP, line 40: The content doesn't look correct for this error - it's the same as the format validation message. I think the standard GOVUK message would be 'Enter your National Insurance number' but I've left it as it for now.

AbigailMcP:
> > Currently there's a method on the Claim model has_all_valid_personal_details? which checks whether the personal details are all valid, and that they all match TID:
> 
> Assuming we don't show the personal-details page if we have the correct details from DfE identity, I think this line could just be
> 
> ```ruby
> sequence.delete("personal-details") if claim.details_check?
> ```
> 
> * If the teacher didn't log in with Dfe identity, then `details_check` is `nil` and we'll want to show the "personal-details" page.
> * If the teacher did log in with Dfe identity but the details are wrong, then `details_check` is false and we'll want to show the "personal-details" page.
> * If the teacher did log in with Dfe identity and the details are correct but invalid, then `details_check` is false and we'll want to show the "personal-details" page.
> * If the teacher did log in with Dfe identity and the details are correct and valid, then `details_check` is true and we don't want to show the "personal-details" page as we already have that info.
> 
> [setting details check logic](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/9fdbc64600fc6f8cd1cf2f52fa347c34006c99bb/app/forms/teacher_detail_form.rb#L35) - same logic as on master just a bit easier to show here as it's in one place.
> 
> I'd double check my assumptions with one of the other devs though!

We discussed whether it was possible to do this, however realised that it's possible for incomplete data to be returned from DfE Identify:

https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/master/spec/features/combined_teacher_claim_journey_with_teacher_id_spec.rb#L216-L252

In this spec ☝️  if the National Insurance number is missing the user can still answer 'yes' on the teacher-details page.. This would set `details_check` to true. But we still need to show them the personal-details page to fill in their National Insurance number.

Therefore a check `claim.details_check?` is not sufficient and we decided that validating a `PersonalDetailsForm` in the slug sequence would be our best bet for now.

app/forms/personal_details_form.rb
AbigailMcP: Hmm I can't think of a way to completely get rid of it nicely.. an `InvalidDate` doesn't necessarily have all the date_components set.

How do you feel about swapping the type check to be a positive check for `InvalidDate` and removing the `number_of_date_components == 3` as per this commit: https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2696/commits/40d075af59f440665529af4a9bbdf911931cb9a3

app/models/journeys/additional_payments_for_teaching/slug_sequence.rb
```diff
@@ -180,6 +180,14 @@ def self.start_page_url
 
       private
 
+      def personal_details_form
+        PersonalDetailsForm.new(
+          claim:,
+          journey: Journeys::AdditionalPaymentsForTeaching,
+          params: ActionController::Parameters.new
+        )
+      end
```
rjlynch, line 189: Not a blocker, but maybe this should be on `Claim`, replacing the `has_all_valid_personal_details?` method 🤷‍♂️

rjlynch (approved):
Looks good 👍

app/models/journeys/additional_payments_for_teaching/slug_sequence.rb
AbigailMcP: This was my original plan until I realised the `Claim` passed into our forms is actually a combined `CurrentClaim` which made things a bit more tricky in certain places e.g. https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/master/app/forms/form.rb#L85.

This would work in the `CurrentClaim` model:

```ruby
def has_all_valid_personal_details?
  PersonalDetailsForm.new(
    claim: self,
    journey: nil,
    params: ActionController::Parameters.new
  ).valid?
end
```
This also made me realise that running the validations on forms are journey-agnostic. Passing in the specific journey isn't actually required..

rjlynch: Ah yeah, would just leave it called from the slug sequence for now then

PR #2701, opened 2024-04-19
Form object for sign-in-or-continue & teacher-details

rjlynch:
## Review notes
The flow starts in `app/views/claims/sign_in_or_continue.html.erb` so might be best to start the review there.
Claims controller loads `app/forms/sign_in_or_continue_form.rb`
If the user skips DfE identity we `PATCH` back to the `ClaimsController`, and hit the else branch in the `SignInOrContinueForm#save` method. If the user signs in with DfE identity they're eventually redirected to `app/controllers/omniauth_callbacks_controller.rb` which sends them back to the `ClaimsController` and re-renders `app/views/claims/sign_in_or_continue.html.erb` with the attributes from DfE identity for the teacher to confirm.

Some of the existing behaviour seems a little odd like we keep the `teacher_id_user_info` even if the teacher has said the details don't match. I've kept this behaviour, as this is just a refactor, but it's something we should look at (see comments in `spec/forms/sign_in_or_continue_form_spec.rb`.

## Form object for sign-in-or-continue

By combining the sign_in_or_continue_form with the teacher_detail_form
we can avoid a lot of weirdness in the claims controller / auth callback, and
delete the teacher detail slug from the journey.

There's now two signin scenarios
User lands on sign in or continue form
1.) User selects "Skip dfe sign in" -> we PATCH to the claims controller
  update action with `logged_in_with_tid` `false` and they continue on
  their way.

2.) User selects "DfE sign-in" -> Users is redirected to DfE idenitiy ->
  User comes back to the omiauth_callbacks_controller with some
  additional info -> We redirect the user along with this additional
  info and the `logged_in_with_tid` flag set as `true` to the
  `sign-in-or-continue` form. As the sign-in-or-continue form sees the
  `logged_in_with_tid` flag as set it renders the confirm detail ui that
  was previously on the teacher-details page. The form can now PATCH to
  the ClaimsController#update action and the slug sequence will continue
  as normal. No need to remove slugs from the journey depending on
  whether or not the teacher as signed in with DfE identity.

This has the benefit of keeping all signin related logic in a single
form and doesn't require us to use the session to pass data between
controller.

We need to be a bit careful about logging parameters and also ensuring
any params from dfe sign in we redirect with are permitted - we don't
want to trigger an unremitted parameter error if dfe sign-in change the
payload. To ensure we're working with the expected parameters we slice
the ones we know about from the auth payload.

The SignInOrContinueForm uses a nested form TeacherIdUserInfoForm as we
need to write the attributes from DfEIdentity to the view, and using a
nested form lets us mimic the behaviour of
`accepts_nested_attributes_for` rather than having to store the JSON
`teacher_id_user_info` as some unusual structure in the view.

There's a potential refactoring we could do where we remove the
`DfeIdentity::UserInfo` class and place it's responsibilities in the
`TeacherIdUserInfoForm` however I avoided doing this to as it's a bit
out of scope.

## Walk throughs

https://github.com/DFE-Digital/claim-additional-payments-for-teaching/assets/9936028/22ba44fd-c5ad-40d2-8995-00e017837ed2

https://github.com/DFE-Digital/claim-additional-payments-for-teaching/assets/9936028/aa722650-2933-45ba-834d-a00352288ff7


rjlynch:
> teacher_detail

Thanks @kenfodder have renamed those files and renamed `navigate_past_teacher_details_page`

app/forms/form.rb
```diff
@@ -45,11 +45,15 @@ def i18n_errors_path(msg)
...
   def permitted_params
-    @permitted_params ||= params.fetch(:claim, {}).permit(*attribute_names)
+    @permitted_params ||= params.fetch(:claim, {}).permit(*permitted_attributes)
   end
 
   private
 
+  def permitted_attributes
+    attribute_names
+  end
```
your, line 55: 👍

your (approved):
LGTM

PR #2702, opened 2024-04-22
CAPT 1622/github actions for ci

rjlynch:
Use github actions for ci.
Runs the linters and test suite.

We run the test over multiple nodes using github actions matrices. The newly
introduced script `bin/test_splitter` handles breaking up the test suite into
sections that run on each node.
There isn't away (at least not one I know of) to cache the set up steps between
nodes, however as the bulk of CI time is spend on running the test suite this
isn't much of an issue (~1min for set up ~9mins for longest node to run).
Splitting the test suite across nodes means we can improve our test suite time
by breaking up some of the slower running feature specs into separate files so
different nodes pick them up.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,86 @@
+name: CI
+
+on: push
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          cache: yarn
+      - name: Install Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+      - name: Linting - ShellCheck
+        run: bin/rake shellcheck
+      - name: Linting - Brakeman
+        run: bundle exec brakeman -c config/brakeman.ignore
+      - name: Linting - Standardrb
+        run: bin/rails standard
+
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ci_node_total: [8]
+        ci_node_index: [0, 1, 2, 3, 4, 5, 6, 7]
+    services:
+      postgres:
+        image: postgres:11-alpine
+        ports:
+          - '5432:5432'
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: password
+    env:
+      RAILS_ENV: test
+      DFE_TEACHERS_PAYMENT_SERVICE_DATABASE_USERNAME: postgres
+      DFE_TEACHERS_PAYMENT_SERVICE_DATABASE_PASSWORD: password
+      DFE_TEACHERS_PAYMENT_SERVICE_DATABASE_HOST: localhost
+      CI_NODE_TOTAL: ${{ matrix.ci_node_total }}
+      CI_NODE_INDEX: ${{ matrix.ci_node_index }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          cache: yarn
+      - name: Install Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+      - name: Restore assets cache
+        id: assets-cache
+        uses: actions/cache/restore@v4
+        with:
+          path: |
+            public/assets
+          key: assets-${{ hashFiles('**/app/assets/**/*', '**/app/javascript/**/*') }}
+          restore-keys: |
+            ${{ runner.os }}-assets-
+      - name: Prepare assets
+        if: ${{ steps.assets-cache.outputs.cache-hit != 'true' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Setup Node
+        uses: actions/setup-node@v4
+      - name: Setup yarn
+        run: npm install -g yarn
+      - name: Install Node packages
+        run: yarn install
+      - name: Install Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 3.2.0
```
slorek, line 40:
```suggestion
```
If we remove this it will read from the `.ruby-version` file which i think would be preferable.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,59 @@
...
+      CI_NODE_INDEX: ${{ matrix.ci_node_index }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Setup Node
+        uses: actions/setup-node@v4
+      - name: Setup yarn
+        run: npm install -g yarn
+      - name: Install Node packages
+        run: yarn install
```
slorek, line 36:
```suggestion
      - name: Setup Node
        uses: actions/setup-node@v4
        with:
          cache: yarn
```
The `setup-node` action provides built in caching using yarn so this will be faster.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Install Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 3.2.0
+          bundler-cache: true
+      - name: Precompile assets
+        run: bundle exec rails assets:precompile
```
slorek, line 43: We probably want to cache this. Also the `test:prepare` task is probably better to use so we don't have to manage other steps if other dependencies are introduced. Here's a snippet I wrote for another project:

```
    - name: Restore assets cache
      id: assets-cache
      uses: actions/cache/restore@v4
      with:
        path: |
          public/assets
        key: assets-${{ hashFiles('**/app/assets/**/*', '**/app/javascript/**/*') }}
        restore-keys: |
          ${{ runner.os }}-assets-

    - name: Prepare assets
      if: ${{ steps.assets-cache.outputs.cache-hit != 'true' }}
      run: bin/rails test:prepare

    - name: Save assets cache
      if: ${{ steps.assets-cache.outputs.cache-hit != 'true' }}
      uses: actions/cache/save@v4
      with:
        path: |
          public/assets
          public/packs-test
        key: ${{ steps.assets-cache.outputs.cache-primary-key }}
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Linting - ShellCheck
+        run: bin/rake shellcheck
+      - name: Linting - Brakeman
+        run: bundle exec brakeman -c config/brakeman.ignore
+      - name: Linting - Standardrb
+        run: bin/rails standard
```
slorek, line 49: I think these steps are being run for all eight runners, which just introduces extra time. I think we should extract the linting to a separate check which can run either first or concurrently with the tests. This will also mean we get quicker feedback as we won't be blocked running the specs just because the linting failed.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Prepare DB
+        run: bin/rails db:schema:load
```
slorek, line 51:
```suggestion
        run: bin/rails db:prepare
```
This will also load seeds.

slorek (changes requested):

.github/workflows/ci.yaml
rjlynch: Yeah this does run for each runner but it saves us having to re-setup the runners for linting and if linting fails the next step wont run. Though given the overhead of setup is minimal I think moving linting into it's own step makes sense

.github/workflows/ci.yaml
rjlynch: We don't have `test:prepare` task, do you think it's better to add a task that calls `assets:precompile` or just leave the call to `assets:precomplie` in action?

slorek: If it’s not there then no; I can’t recall how this is defined and it’s probably not worth worrying about.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,94 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          cache: yarn
+      - name: Setup yarn
+        run: npm install -g yarn
+      - name: Install Node packages
+        run: yarn install
```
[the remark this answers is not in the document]
rjlynch: Nice, didn't know the cache flag handled that for us 👍

slorek (approved):
Looks to be working great now thanks.

I think we will be able to improve the division of tests so they are spread a bit more evenly; at the moment one takes a few seconds ranging up to nearly 10 minutes for another. But we can look at that in a future iteration.

PR #2704, opened 2024-04-23
Add a form object for choosing the correct school

felixclack:
When using DfE Sign In, we present the school currently associated with
the user to check if it is still the correct school.

Currently, we retrieve the associated TPS school, store it in the
session so that we can display it in the view. Then on confirmation from
the user, this value gets persisted as the current school.

The form object allows us to extract this logic and it can define these
references, ensuring that the controller is simpler.

As far as I can tell, the session itself is unnecessary as the values
stored there are already persisted in the DB.


app/forms/journeys/additional_payments_for_teaching/correct_school_form.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+
+      private
+
+      def current_school
+        @current_school ||= (claim.recent_tps_school || claim.school)
+      end
+
+      def change_school?
+        params[:change_school] == "true" ||
```
rjlynch, line 32: Non blocker but should `change_school` be an attribute on this form? `attribute :change_school, :boolean` - would handle type casting for us and we wouldn't have to access the non permitted params

rjlynch (approved):
Looks good 👍

your (commented):
Build failed with:

```
standard: Use Ruby Standard Style (https://github.com/standardrb/standard)
  app/forms/journeys/additional_payments_for_teaching/correct_school_form.rb:24:29: Style/RedundantParentheses: Don't use parentheses around a logical expression.
  app/forms/journeys/additional_payments_for_teaching/correct_school_form.rb:32:17: Layout/SpaceInsideHashLiteralBraces: Space inside { detected.
  app/forms/journeys/additional_payments_for_teaching/correct_school_form.rb:32:69: Layout/SpaceInsideHashLiteralBraces: Space inside } detected.
  app/forms/journeys/additional_payments_for_teaching/correct_school_form.rb:34:10: Layout/SpaceInsideHashLiteralBraces: Space inside { detected.
  app/forms/journeys/additional_payments_for_teaching/correct_school_form.rb:34:59: Layout/SpaceInsideHashLiteralBraces: Space inside } detected.
  app/models/journeys/additional_payments_for_teaching.rb:25:44: Style/TrailingCommaInHashLiteral: Avoid comma after the last item of a hash.
  spec/forms/journeys/additional_payments_for_teaching/correct_school_form_spec.rb:52:10: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
  spec/forms/journeys/additional_payments_for_teaching/correct_school_form_spec.rb:56:10: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
standard: Run `rake standard:fix` to fix up to 8 problems.
rake aborted!
```

Looks good otherwise!

PR #2706, opened 2024-04-23
CAPT-1588 Form object for /provide-mobile-number

AbigailMcP:
**Changes in this PR**

- Extracted form object for `/provide-mobile-number` page.
  This is a simple for saving the boolean `provide_mobile_number` onto a Claim.
- Removed validation from Claim model

**Points to note**
- I've kept the error messages out of locales for now until we decide on an approach for forms shared across all journeys

### DIFF
--- a/app/forms/provide_mobile_number_form.rb
+++ b/app/forms/provide_mobile_number_form.rb
@@ -1,0 +1,18 @@
+class ProvideMobileNumberForm < Form
+  attribute :provide_mobile_number, :boolean
+
+  validates :provide_mobile_number,
+    inclusion: {
+      in: [true, false],
+      message: "Select yes if you would like to provide your mobile number"
+    },
+    if: -> { claim.has_ecp_or_lupp_policy? }
+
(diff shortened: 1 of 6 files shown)


app/forms/provide_mobile_number_form.rb
```diff
@@ -0,0 +1,18 @@
+class ProvideMobileNumberForm < Form
+  attribute :provide_mobile_number, :boolean
+
+  validates :provide_mobile_number,
+    inclusion: {
+      in: [true, false],
+      message: "Select yes if you would like to provide your mobile number"
+    },
+    if: -> { claim.has_ecp_or_lupp_policy? }
```
rjlynch, line 9:
```suggestion
    }
```
Not a blocker but I think we can lose this as all it does is allow the user on the student loans journey to submit the form without selecting `true` or `false` and it couples knowledge of a specific journey in a shared form.

rjlynch (approved):
Looks good 👍

PR #2708, opened 2024-04-24
information-provided - make template policy agnostic

kenfodder:
This page doesn't actually alter anything on the claim on submit.

However there are conditions in the template which are policy specific. Those have been removed but the behaviour is replicated using `en.yml` and policy namespacing.

Similar to the other PR this doesn't have and "empty" form object, instead we'll handle this when we have more of these pages in a generic way/

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -65,8 +65,8 @@ module ApplicationHelper
     "https://www.gov.uk/done/claim-additional-teaching-payment"
   end
 
-  def information_provided_further_details_with_link(policy_str)
-    policy = policy_str.present? ? Policies.constantize(policy_str) : current_claim.policy
+  def information_provided_further_details_with_link
+    policy = current_claim.selected_policy || current_claim.policy
 
     text = I18n.t("#{policy.locale_key}.information_provided_further_details_link_text")
     link = link_to(text, policy.payment_and_deductions_info_url, class: "govuk-link govuk-link--no-visited-state", target: "_blank")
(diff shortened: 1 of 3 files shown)


rjlynch (approved):
LGTM 👍 Left one minor comment.

I think not introducing a form here is fine, though I think there is an argument for a form as this page hits the controller update action, so it updates the slug, and calls the `ClaimStudentLoanDetailsUpdater`. Probably best to think about these pages that don't update the claim once we've tackled the rest of them though.

app/helpers/application_helper.rb
```diff
@@ -64,4 +64,13 @@ def claim_decision_deadline_in_weeks
...
   end
+
+  def information_provided_further_details_with_link(policy_str)
+    policy = policy_str.present? ? Policies.constantize(policy_str) : current_claim.policy
+
+    text = I18n.t("#{policy.locale_key}.information_provided_further_details_link_text")
+    link = link_to(text, policy.payment_and_deductions_info_url, class: "govuk-link govuk-link--no-visited-state", target: "_blank")
+
+    I18n.t("#{policy.locale_key}.information_provided_further_details", link: link)&.html_safe
+  end
```
your, line 75: A part of me is tempted to add a test for this 😁 but I bet this helper doesn't have unit tests?

slorek (approved):

app/helpers/application_helper.rb
slorek: There are some unit tests for helpers, but that wouldn't be a reason not to add one for a new method :)

your (approved):

PR #2709, opened 2024-04-24
CAPT-1585 Form object for /select-email

your:
https://dfedigital.atlassian.net/browse/CAPT-1585

I made two commits. In the first one, I implemented the form object that can be shared across policies. The good thing is that I could eliminate the `session[:email_address]` as it didn't seem needed in the first place. Also, I had to create shared `en.forms` entries, which is just an idea based on a conversation on Slack. In the second commit, I introduced a class method to retrieve error messages, making the validation statements more concise.

### DIFF
--- a/spec/forms/select_email_form_spec.rb
+++ b/spec/forms/select_email_form_spec.rb
@@ -23,16 +23,16 @@ RSpec.describe SelectEmailForm, type: :model do
       it "cannot be nil" do
         form.email_address_check = nil
 
-        expect(form).to be_invalid
+        is_expected.to be_invalid
         expect(form.errors[:email_address_check]).to eq([form.i18n_errors_path(:select_email)])
       end
 
       it "can be true or false" do
         form.email_address_check = true
-        expect(form).to be_valid
+        is_expected.to be_valid
 
         form.email_address_check = false
-        expect(form).to be_valid
+        is_expected.to be_valid
       end
     end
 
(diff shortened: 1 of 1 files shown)


app/forms/form.rb
```diff
@@ -2,6 +2,7 @@ class Form
   include ActiveModel::Model
   include ActiveModel::Attributes
   include ActiveModel::Serialization
+  include ActiveModel::Validations::Callbacks
```
rjlynch, line 5: Nice, once this is in master I'll rebase my pr to use this rather than using define model call backs
https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/c32857dc81622e2338cbe382a94487c386c78357/app/forms/form.rb#L5

spec/forms/select_email_form_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+          {
+            "email_address_check" => true,
+            "email_address" => email_from_teacher_id,
+            "email_verified" => true
+          }
+        end
+
+        before { form.save }
+
+        it { expect(form).to have_received(:update!).with(expected_saved_attributes) }
```
rjlynch, line 83: Not a blocker but I think we can use `is_expected.to` here as the form is the subject
```ruby
it { is_expected.to have_received(:update!).with(expected_saved_attributes) }
```
reads a bit nicer!

rjlynch (approved):
Looks good 👍

spec/forms/select_email_form_spec.rb
your: Missed that and good point, addressed with [5853320](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2709/commits/585332057d5468290ac9b86bfa1daa949af653d2)

PR #2712, opened 2024-04-24
CAPT 1589,1590,1591 mobile screens

rjlynch:
## Form objects for mobile verification screens

This pr covers 3 tickets as I thought tackling all the mobile tickets together would be the best bet.
For this PR it's likely best to review [each commit separately](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2711/commits) as they build on one another.

A possible controversial choice was to add a column to the claim to store the `sent_one_time_password_at`, which was previously stored in the session and accessed via an `attribute` that was set when doing `claim.update(sent_one_time_password_at: ...)`, took me a while to spot this wasn't being persisted to the claim 😂. 

Storing the `sent_one_time_password_at` column in the db has a couple of advantages
* `claim.update(sent_one_time_password_at: ...)` behaves in the standard way, writing a column to the db
*  We don't need to use the session as temporary storage for passing data between controller actions [¹](#note-1)

I've not added a `sent_one_time_password_at` column to Reminders as that seemed a bit out of scope (though happy to do so).

I think for the email validation work we can easily reuse the `sent_one_time_password_at` column on the claim (as both emails and texts share the same key in the session), though it maybe better to have two columns `sent_email_one_time_password_at` and `sent_mobile_one_time_password_at`
(update [this pr](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2715/commits/52de6ccacbbfcac4c344717db920f443e0028291) for email verification reuses the `sent_one_time_password_at` column)

## Walk through

https://github.com/DFE-Digital/claim-additional-payments-for-teaching/assets/9936028/753fb5ed-4735-4337-b0ec-f5212ddf9a7f

<a name="note-1">1.)<a> I'm not against the plan to move the claim into the session, the current issue with using the session to persist data between controller actions is we're storing data in two places, the session and the database, which is awkward to work with. I think moving as much as possible out of the session and on to the claim model will make switching to persisting the claim in the session easier.


app/forms/mobile_number_form.rb
```diff
@@ -0,0 +1,44 @@
+class MobileNumberForm < Form
+  attribute :mobile_number
+
+  validates :mobile_number,
+    presence: {
+      message: "Enter a mobile number, like 07700 900 982 or +44 7700 900 982"
```
slorek, line 6: Please use the i18n file for error messages for consistency. `i18n_errors_path` should specify where these are.

app/forms/mobile_number_form.rb
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+  validates :mobile_number,
+    format: {
+      with: /\A(\+44\s?)?(?:\d\s?){10,11}\z/,
+      message: "Enter a valid mobile number, like 07700 900 982 or +44 7700 900 982"
```
slorek, line 12: As above

app/forms/mobile_number_form.rb
```diff
@@ -0,0 +1,44 @@
...
+      sent_one_time_password_at: sent_one_time_password_at
+    )
+  end
+
+  private
+
+  def send_sms_message
+    NotifySmsMessage.new(
+      phone_number: mobile_number,
+      template_id: "86ae1fe4-4f98-460b-9d57-181804b4e218",
```
slorek, line 36: Suggestion: I realise this was moved 1:1 from the controller, but perhaps this template ID should be a constant in the `NotifySmsMessage` class (similar to those in `ApplicationMailer`), or the form class.

app/forms/mobile_number_form.rb
```diff
@@ -0,0 +1,44 @@
...
+      personalisation: {otp: OneTimePassword::Generator.new.code}
+    ).deliver!
+  end
+
+  def mobile_number_changed?
```
slorek, line 41: Suggestion: may be overkill for this individual case, but `ActiveModel::Dirty` could be helpful if we find we are doing this often.

slorek (changes requested):
Looks great, thanks! Just a minor change required.

I agree with the approach of moving everything to `CurrentClaim` as an interim measure so we have one interface to change later on. I imagine then we would just extract a subset of attributes when submitting the claim and actually creating the `Claim` record.

app/forms/mobile_number_form.rb
rjlynch: I had a quick go at adding ActiveModel::Dirty but it required some tweaks to how we set up defaulting attributes on the base form so have left it for now

slorek (approved):

PR #2713, opened 2024-04-25
CAPT-1592 Extract form objects for banking details

AbigailMcP:
## Changes in this PR

- Extract form object for /bank-or-building-society
- Re-purpose existing `BankDetailsForm` to serve `/personal-bank-account` and `building-society-account` pages
- Remove page-specific logic from `ClaimsController`
- ~Switch from storing the `hmrc_validation_attempt_count` in the session to a form parameter so all logic can be encapsulated in form object~
- Move error messages to locales under shared `forms:` section

## Things to note

- I've kept some validations in the `Claim` model since looks as though these attributes can be updated via the admin interface and therefore should be validated still
  - We could look at consolidating these into shared validators once we pull out admin form objects?

Screenshot 2024-04-30 at 09 51 43
Screenshot 2024-04-30 at 10 12 58

### DIFF
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -64,6 +64,8 @@ class ClaimsController < BasePublicController
   end
 
   def update
+    params[:claim][:hmrc_validation_attempt_count] = session[:hmrc_validation_attempt_count] || 0 if on_banking_page?
+
     # TODO: Migrate the remaining slugs to form objects.
     if (@form = journey.form(claim: current_claim, params:))
       if @form.save
@@ -71,6 +73,7 @@ class ClaimsController < BasePublicController
         update_session_with_selected_policy
         redirect_to claim_path(current_journey_routing_name, next_slug)
       else
+        session[:hmrc_validation_attempt_count] = (session[:hmrc_validation_attempt_count] || 0) + 1 if on_banking_page? && @form.hmrc_api_validation_attempted?
         set_any_backlink_override
         show
       end
@@ -207,6 +210,10 @@ class ClaimsController < BasePublicController
 
   def prepend_view_path_for_journey
     prepend_view_path("app/views/#{current_journey_routing_name.underscore}")
+  end
+
+  def on_banking_page?
+    %w[personal-bank-account building-society-account].include?(params[:slug])
   end
 
   def one_time_password
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -54,8 +54,6 @@ class ClaimsController < BasePublicController
       session[:claim_postcode] = nil
       session[:claim_address_line_1] = nil
       redirect_to claim_path(current_journey_routing_name, "postcode-search") and return
-    elsif ["personal-bank-account", "building-society-account"].include?(params[:slug])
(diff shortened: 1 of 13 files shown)


app/forms/bank_details_form.rb
```diff
@@ -2,7 +2,7 @@ class BankDetailsForm < Form
   # Only validate against HMRC API if number of attempts is below threshold
   MAX_HMRC_API_VALIDATION_ATTEMPTS = 3
 
-  attribute :hmrc_validation_attempt_count
+  attribute :hmrc_validation_attempt_count, :integer
```
rjlynch, line 5: I _think_ you can do `attribute :hmrc_validation_attempt_count, :integer, default: 0` to avoid having to overwrite the initialiser (not a blocker)

rjlynch (approved):
Nicely done 👍 
Prob worth getting one of the devs who've been on claim longer to see if they spot any issues but all looks good to me

slorek (commented):
Thanks for refactoring this.

🤔 the motivation for adding the attempt count to the session was to prevent tampering. We could potentially keep it in the session but merge it with the params on initialisation of the form so we can still manage it in the form object that way.

What do you think?

AbigailMcP:
> Thanks for refactoring this.
> 
> 🤔 the motivation for adding the attempt count to the session was to prevent tampering. We could potentially keep it in the session but merge it with the params on initialisation of the form so we can still manage it in the form object that way.
> 
> What do you think?

@slorek - made some changes in this commit to keep the count in the session, let me know what you think! https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2713/commits/ea87ca408bb633d9ec3c05ac2e177728b3c9b850

app/controllers/claims_controller.rb
```diff
@@ -68,13 +66,16 @@ def show
...
+    params[:claim][:hmrc_validation_attempt_count] = session[:hmrc_validation_attempt_count] || 0 if on_banking_page?
+
     # TODO: Migrate the remaining slugs to form objects.
     if (@form = journey.form(claim: current_claim, params:))
       if @form.save
         retrieve_student_loan_details
         update_session_with_selected_policy
         redirect_to claim_path(current_journey_routing_name, next_slug)
       else
+        session[:hmrc_validation_attempt_count] = (session[:hmrc_validation_attempt_count] || 0) + 1 if on_banking_page? && @form.hmrc_api_validation_attempted?
```
rjlynch, line 78: Could we store `hmrc_validation_attempt_count` as a column on the `Claim`? That should let us keep the attempt counting logic contained in the form object? Should be safe to just set a column default of `0` for existing claims and as the app isn't live for new claims I don't think we need to worry about handling inflight claims

rjlynch (approved):
Looks good to me 👍 . Left an optional comment.

app/controllers/claims_controller.rb
AbigailMcP: thanks Rich! this is a nice idea and would keep the logic inside the form. i think i'll leave it in the session for now since it's mirroring the current behaviour and isn't not something we need to know about once the user has submitted a claim.

slorek (approved):
Thanks for addressing this. We can look at tidying up the session data management separately along with other examples.

PR #2715, opened 2024-04-25
email verification screens

rjlynch:
Following a similar pattern as in https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2712 this pr extracts form object for mobile verifications screens.
Note that both the mobile verification screens and the email verification screens write to the same field on the claim model `sent_one_time_password_at`. This _shouldn't_ be an issue as both steps were writing to the same `sent_one_time_password_at` key in the session, however it's probably better practice for each step to write to it's own `sent_*_one_time_password_at` column. It should be safe to split the `sent_one_time_password_at` column in a future commit as the service isn't in use at the minute.

## Notes for review
This pr covers two tickets CAPT-1586 and CAPT-1587 and does a bit of clean up work so it might be easier to review each commit separately.

## Walk through
https://github.com/DFE-Digital/claim-additional-payments-for-teaching/assets/9936028/6b14654a-513e-4320-9769-760a56f8333f


app/forms/email_verification_form.rb
```diff
@@ -0,0 +1,29 @@
+class EmailVerificationForm < Form
+  attribute :one_time_password
+
+  # Required for shared partial in the view
+  delegate :email_address, to: :claim
+
+  validate :otp_validate
+
+  before_validation do
+    self.one_time_password = one_time_password.gsub(/\D/, "")
```
slorek, line 10: 🤔 do we really want to remove all non-digits? I can understand whitespace, but perhaps this is too forgiving.

rjlynch: I'm not sure, I just went with what was already in `OneTimePassowrdCeckable#normalise_one_time_password`

slorek (approved):

PR #2720, opened 2024-04-29
Add a form object for reset-claim

felixclack:
When there is a discrepancy with the details returned by DfE Sign In, we
present the option to reset the claim.

Resetting the claim doesn't have any validations but does currently have
a hardcoded next slug. We want to remove this coupling to the next slug
by moving this to a form object.

It is a very simple form object, with its only purpose to allow us to
dynamically define the next slug.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -454,7 +454,7 @@ GEM
       simplecov_json_formatter (~> 0.1)
     simplecov-html (0.12.3)
     simplecov_json_formatter (0.1.4)
-    spring (4.1.3)
+    spring (4.2.1)
     spring-watcher-listen (2.1.0)
       listen (>= 2.7, < 4.0)
       spring (>= 4)
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -54,8 +54,6 @@ class ClaimsController < BasePublicController
       session[:claim_postcode] = nil
       session[:claim_address_line_1] = nil
       redirect_to claim_path(current_journey_routing_name, "postcode-search") and return
-    elsif ["personal-bank-account", "building-society-account"].include?(params[:slug])
-      @bank_details_form ||= BankDetailsForm.new(claim: current_claim)
     end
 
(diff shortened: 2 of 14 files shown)


app/controllers/concerns/part_of_claim_journey.rb
```diff
@@ -22,7 +22,7 @@ def send_unstarted_claimants_to_the_start
   end
 
   def skip_landing_page?
-    params[:skip_landing_page] == "true"
+    params.dig(:claim, :skip_landing_page) == "true"
```
rjlynch, line 25: Given this module is included in the `SubmissionsController` is this safe to change here?
Do we need to copy the original implementation of the method into the `SubmissionsController`?

app/views/claims/reset_claim.erb
```diff
@@ -16,7 +16,7 @@
         You can continue to complete an application to check your eligibility and apply for a payment.
       </p>
 
-      <%= form_with(url: claim_path(current_journey_routing_name, "current-school"), method: :get) do %>
+      <%= form_with model: @form, url: reset_claim_path do %>
         <div class="govuk-form-group">
           <%= hidden_field_tag :skip_landing_page, 'true' %>
```
rjlynch, line 21: Does this need updating to be `f.hidden_field` otherwise it will come through in the params as `skip_landing_page` rather than `claims[:skip_landing_page]` and we've changed `def skip_landing_page?` to look for this nested under `claim`

rjlynch (commented):
Will let one of the devs with a bit more context on that app review this before approving, though I think the overall idea looks sensible.
I wonder if the form is the right place for the back_link method? Maybe there should be a helper method that takes the current page sequence (something for a separate ticket)

slorek (approved):
It feels a bit unnecessary to do this, but this is a limitation of the slug sequence implementation. I think https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/master/config/routes.rb#L54 is a route to nowhere and should be removed.

PR #2725, opened 2024-04-30
CAPT-1600 Form object for /subjects-taught (TSLR)

your:
https://dfedigital.atlassian.net/browse/CAPT-1600

As part of this ticket, I'm removing a custom JS component to reset multiple-choice checkboxes when selecting the "None taught" option. This is not needed as the GOV.UK Frontend [includes](https://design-system.service.gov.uk/components/checkboxes/) that functionality (see "Add an option for ‘none’"). Furthermore, in the existing view, there is a mix of checkboxes and radio buttons, which is not ideal.


rjlynch (approved):
Looks good to me
Nice job removing the custom js and getting rid of the radio button.

I'm not sure why we need to introduce the `subject_taught` accessor. If we loop through each of the questions in the view and send each checkbox through as a separate param that might simplify setting the attributes in the form object and we could loose the before_validation callback and the `taught_eligible_subjects` attribute

```
def check_box_options
  subject_attributes + [:none_taught]
end

<% f.object.subject_attributes.each do |subject| %>
  <%= f.check_box subject %>
<% end %>
```
Probably missing something though!

your:
@rjlynch thanks for the review. The checkboxes need to have the same input name (which results in an array of values), because it's a "limitation" of using the GOV.UK frontend "exclusive" checkboxes feature. I was not too happy as it meant more work in the form, but I'd rather do that than maintain a custom JavaScript component.

See description and implementation here: https://github.com/alphagov/govuk-frontend/blob/8729af48207d7513218840669aefcc0271db56c2/packages/govuk-frontend/src/govuk/components/checkboxes/checkboxes.mjs#L149-L158

PR #2726, opened 2024-04-30
WIP - Capt 1628/move resetting answers into the forms

rjlynch:
WIP remove dependent attributes.
There's some forms that need extracting in the student loans journey before we can fully remove the dependent attributes

Very much just a WIP PR to see what everyone thinks about whether or not we should do this

NOTE we need to decide on how to reset `eligible_degree_subject` as only LUP eligibility has that attribute


app/forms/journeys/additional_payments_for_teaching/eligible_itt_subject_form.rb
```diff
@@ -27,15 +27,33 @@ def chemistry_or_physics_available?
...
+
+        if claim.qualifications_details_check?
+          update!(
+            eligibility_attributes: {
+              eligible_itt_subject: eligible_itt_subject
+            }
+          )
+        else
+          # FIXME RL: decide how to handle resetting `eligible_degree_subject`
+          # and add test for it - only LUP eligibility has this field
```
rjlynch, line 40: Not sure how best to tackle this.
Options are
1.) Update current_claim to filter attributes from eligibility_attributes if the eligibility it's setting attributes for doesn't respond to the method
2.) Add a non persisted attribute to `EarlyCareerPayments::Eligibility` for compatibility
3.) Something else

Maybe `eligible_degree_subject` is a sign that the eligibilities can't be treated the same and we shouldn't try and move reseting dependent answers into the form object?

Any suggestions on this?

slorek (commented):
I think refactoring the resetting of dependant answers is much needed.

Just to remind ourselves of some principles around forms:

* Forms should be re-usable on different journeys
* There's no guarantee the dependant answer will exist on another journey
* The ordering of pages may be different on another journey

It seems the changes made so far only apply to the forms in the additional payments journey, which in the short term seems fine, but if we want to re-use those forms in other journeys it is slightly more difficult to do because rather than moving the form class to the `app/forms` top level we would likely have to also override the `save` method on journey-specific child classes.

It could get quite messy in the event that the forms are presented in a different order, as the dependent answers changes across the whole journey.

I think this is fine as an interim solution but that we are missing a bit of glue between the forms and the controller which handles this stuff, among other things.

rjlynch:
> I think refactoring the resetting of dependant answers is much needed.
> 
> Just to remind ourselves of some principles around forms:
> 
> * Forms should be re-usable on different journeys
> * There's no guarantee the dependant answer will exist on another journey
> * The ordering of pages may be different on another journey
> 
> It seems the changes made so far only apply to the forms in the additional payments journey, which in the short term seems fine, but if we want to re-use those forms in other journeys it is slightly more difficult to do because rather than moving the form class to the `app/forms` top level we would likely have to also override the `save` method on journey-specific child classes.
> 
> It could get quite messy in the event that the forms are presented in a different order, as the dependent answers changes across the whole journey.
> 
> I think this is fine as an interim solution but that we are missing a bit of glue between the forms and the controller which handles this stuff, among other things.

OK I think given
> Forms should be re-usable on different journeys
> There's no guarantee the dependant answer will exist on another journey

Then this is probably the wrong approach as it couples knowledge of other steps in the form. Likely best to leave the current reset dependent answers implementation as is for the time being.

PR #2727, opened 2024-05-01
Fixes flakey omniauth spec

rjlynch:
Depending on the order specs ran
`spec/requests/omniauth_callbacks_controller_spec` would fail, the
minimal reproduction case for this failure was

```
bundle exec rspec --order defined \
  spec/features/teacher_identity_sign_in_spec.rb \
  spec/requests/omniauth_callbacks_controller_spec.rb
````

This was due to omniauth mocks not being reset between test runs.
Updating the suite's before hook to nil out the mock_auths fixes the
issue.


AbigailMcP (approved):

slorek (approved):
Good catch, thanks.

PR #2728, opened 2024-05-01
CAPT-1601 Form object for /still-teaching

AbigailMcP:
## Repurpose form for /still-teaching

- Repurpose StillTeachingForm
- Remove logic from claims controller
- KEEP validations from eligibility

## Why is the validation in the eligibility model still?

- Removing the eligibility causes the [submittable?](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/master/app/models/claim.rb#L305-L307) method to return `true` in cases where it should be `false`.
  - This is because removing the validation causes `claim.valid?(:submit)` to be `true`
- This means that in some situations, the [next_slug](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/efeca0423795517f0491467c06e3017de90b52c8/app/models/journeys/page_sequence.rb#L29) method redirects the user to the wrong step (the check answers page when it should be an eligibility step)


app/models/policies/student_loans/eligibility.rb
```diff
@@ -49,7 +49,7 @@ class Eligibility < ApplicationRecord
       belongs_to :current_school, optional: true, class_name: "School"
 
       validates :claim_school, on: [:"select-claim-school"], presence: {message: ->(object, _data) { object.select_claim_school_presence_error_message }}, unless: :claim_school_somewhere_else?
-      validates :employment_status, on: [:"still-teaching", :submit], presence: {message: ->(object, _data) { "Select if you still work at #{object.claim_school_name}, another school or no longer teach in England" }}
+      validates :employment_status, on: [:submit], presence: {message: ->(object, _data) { "Select if you still work at #{object.claim_school_name}, another school or no longer teach in England" }}
```
AbigailMcP, line 52: Question about the existing code - should we have a separate error message for the `logged_in_with_tid_and_has_recent_tps_school?` flow? This error message refers directly to `claim_school_name` rather than tps school...

slorek: It seems like we should :) I would guess this would be blank in that scenario which we can test.

app/forms/journeys/teacher_student_loan_reimbursement/still_teaching_form.rb
```diff
@@ -0,0 +1,40 @@
+module Journeys
+  module TeacherStudentLoanReimbursement
+    class StillTeachingForm < Form
+      attribute :employment_status
+      attribute :current_school_id
+
+      validates :employment_status,
+        presence: {
+          message: ->(form, _) { form.i18n_errors_path("select_which_school_currently", school_name: form.claim_school_name) }
```
slorek, line 9: As per your existing comment, this error message probably should also revert to the TPS school in that scenario.

slorek (approved):

spec/features/tslr_claim_journey_with_teacher_id_still_teaching_spec.rb
```diff
@@ -36,9 +36,13 @@
   end
 
   scenario "Selects school" do
-    # - "Selects suggested school retrieved from TPS" do
     navigate_to_still_teaching_page
 
+    # - Tries to continue without selecting option
+    click_on "Continue"
+    expect(page).to have_text("Select if you still work at #{eligible_school.name}, another school or no longer teach in England")
```
AbigailMcP, line 43: This new test fails on master. The bug is that the current code looks up the school name for the error message from `eligibility.claim_school_name`. But for TID journey users it should be from `claim.recent_tps_school.name`.

spec/forms/journeys/teacher_student_loan_reimbursement/still_teaching_form_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+      )
+    end
+  end
+
+  describe "#error_message" do
+    context "when the school is closed" do
+      let(:claim_school) { build(:school, :student_loans_eligible, close_date: Date.yesterday) }
+
+      it "does not contain the school name" do
+        expect(form.error_message).to eq("Select yes if you are still employed to teach at a school in England")
```
AbigailMcP, line 33: This is also a bug on master. If the school isn't open, we only show two options to the question "Are you still employed to teach at a school in England?", yes or no. We shouldn't be showing the closed school name in the error message if it's not an option to pick.

On master:
Screenshot 2024-05-08 at 10 47 40

AbigailMcP: Now says "Select if you are still employed to teach at a school in England"

spec/features/tslr_claim_journey_with_teacher_id_still_teaching_spec.rb
AbigailMcP: Now includes school name.

PR #2733, opened 2024-05-07
Capt 1604/check answers form

rjlynch:
Adds a form object for claim submission.
This pr introduces a `ClaimSubmissionForm`, which validates a `Claim` is submittable and a `ClaimSubmissionJob` which handles the steps involved in submitting the `Claim`.

The form on the final check your answers page posts to the `SubmissionsController` to create the claim. I've gone for the name `ClaimSubmissionForm` rather than `CheckYourAnswersForm` as this form is responsible for validating a claim is submittable, though it does break our page -> form naming convention, happy to change either the form or the view name if we need to keep this convention.

More of the validations on a `Claim` can be moved over to the `ClaimSubmissionForm` once we no longer use `Claim#submittable?` in the `Journeys::PageSequence`. Having the claim submission logic in one place _should_ make it easier to incrementally switch over to the new approach of not building a claim until the end of the journey.

## Successful claim submission
https://github.com/DFE-Digital/claim-additional-payments-for-teaching/assets/9936028/fe5af2fa-b820-470f-b66e-7c38d26adfc8

## Claim becomes ineligible at the last minute
https://github.com/DFE-Digital/claim-additional-payments-for-teaching/assets/9936028/5ff40af3-bb71-4f4d-aabd-c0e6b208c419


slorek (commented):
Thanks for tackling this 👏 

Do you think we still need the `SubmissionsController` rather than just using the `ClaimsController` with this form object? I found this extra controller did introduce complications for me a couple of times and it doesn't seem to be doing anything special now.

My preference would be for a service object to encapsulate the submission logic rather than a background job which we call synchronously. There's a few other bits of functionality which would benefit from this approach in the admin area.

Going forward we would probably want to change the interface to this as part of the 'not creating a claim until submission' work as this interface is tied pretty strongly to the `CurrentClaim`. It would be nice to have a service object which just accepted a list of answers and created a claim on validation. That opens up a number of possibilities with seeds and factories etc. This is fine for now but something to bear in mind.

rjlynch:
> Thanks for tackling this 👏
> 
> Do you think we still need the `SubmissionsController` rather than just using the `ClaimsController` with this form object? I found this extra controller did introduce complications for me a couple of times and it doesn't seem to be doing anything special now.
> 
> My preference would be for a service object to encapsulate the submission logic rather than a background job which we call synchronously. There's a few other bits of functionality which would benefit from this approach in the admin area.
> 
> Going forward we would probably want to change the interface to this as part of the 'not creating a claim until submission' work as this interface is tied pretty strongly to the `CurrentClaim`. It would be nice to have a service object which just accepted a list of answers and created a claim on validation. That opens up a number of possibilities with seeds and factories etc. This is fine for now but something to bear in mind.

I thought I'd leave the submission controller as is for now as I didn't want to change too much, I think this can be revisited as part of refactoring how we create claims.

> It would be nice to have a service object which just accepted a list of answers and created a claim on validation.
That's the plan! I think ideally we'll want to do something like

```ruby
def create
  form = ClaimSubmissionForm.new(params: journey_session.answers)
  if form.save
   redirect_to ...
```

>  My preference would be for a service object to encapsulate the submission logic rather than a background job which we call synchronously

Happy to switch to a service object as they both do the same thing, though I think jobs have advantages over service objects, especially WRT long term maintenance / new devs joining the team, as jobs have conventions around how they're used where as service objects don't have a general conventions (eg can they return results objects? do they preform validations?) and they start to take on extra concerns rather than just executing a series of steps.

rjlynch:
> > Thanks for tackling this 👏
> > Do you think we still need the `SubmissionsController` rather than just using the `ClaimsController` with this form object? I found this extra controller did introduce complications for me a couple of times and it doesn't seem to be doing anything special now.
> > My preference would be for a service object to encapsulate the submission logic rather than a background job which we call synchronously. There's a few other bits of functionality which would benefit from this approach in the admin area.
> > Going forward we would probably want to change the interface to this as part of the 'not creating a claim until submission' work as this interface is tied pretty strongly to the `CurrentClaim`. It would be nice to have a service object which just accepted a list of answers and created a claim on validation. That opens up a number of possibilities with seeds and factories etc. This is fine for now but something to bear in mind.
> 
> I thought I'd leave the submission controller as is for now as I didn't want to change too much, I think this can be revisited as part of refactoring how we create claims.
> 
> > It would be nice to have a service object which just accepted a list of answers and created a claim on validation.
> 
> That's the plan! I think ideally we'll want to do something like
> 
> ```ruby
> def create
>   form = ClaimSubmissionForm.new(params: journey_session.answers)
>   if form.save
>    redirect_to ...
> ```
> 
>  > My preference would be for a service object to encapsulate the submission logic rather than a background job which we call synchronously
> 
> Happy to switch to a service object as they both do the same thing, though I think jobs have advantages over service objects, especially WRT long term maintenance / new devs joining the team. As jobs are a built in layer of Rails they have conventions around how they're used whereas service objects, being a layer we introduce, don't have general conventions (eg in this app can they return results objects? do they perform validations?) and they have a tendency to start to take on extra concerns rather than just executing a series of steps.
> 
> Will send up a commit to switch this to using a service and see what you think, can always drop the commit before merging if we want to stick with jobs 👍

@slorek I've moved it over to a service object in this commit https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2733/commits/5fd67a2e6ecf2690c52712f94f19281d8f2fff67 let me know which approach you want to go with

rjlynch:
@slorek closing this pr, most of this needs to change to create the claim from the journey session so not much use introducing this pr

PR #2734, opened 2024-05-08
Introduces journey session model

rjlynch:
As part of the efforts to simplify multi claim handling we want to move
away from creating a claim at the start of the journey and instead only
create the claim at the final step where the teacher submits their
answers.

This commit introduces a new model `Journeys::Session` which will be
used to store the answers entered by the teacher. The answers will be
stored in a json column on the new model, once validated by the form
objects.

`Journeys::Session` is currently the best name we could come up with but
we're all ears for any better suggestions.

### DIFF
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -159,6 +159,7 @@ class ClaimsController < BasePublicController
 
     current_claim.save!
     session[:claim_id] = current_claim.claim_ids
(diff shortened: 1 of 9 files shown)


slorek (approved):
Don't forget we'll need to expunge old data periodically :)

rjlynch:
> Don't forget we'll need to expunge old data periodically :)

Yeah I've got a ticket for that 👍

your (approved):

PR #2737, opened 2024-05-09
CAPT-1632 Implement review apps

AbigailMcP:
## Changes made in this PR

-  Ran the new_service command in the https://github.com/DFE-Digital/teacher-services-cloud repo and copied config files into this repo, updating where necessary
- Updated Makefile so that we have existing commands still working plus new commands with `-aks` suffixes
- Removed autogenerated `development` config and replaced with `test` env
- Created new `build_and_deploy` workflow to deploy review app when 'deploy' label added
- Created delete review app workflow on PR close

## Guidance to review

- Check that the review app is built correctly
- Check that the review app and all related resources are destroyed when the PR is closed 

## Not included

- We've left the old review app creation/deletion in the codebase to make the review simpler, will remove in follow up PR


terraform/application/config/test.tfvars.json
```diff
@@ -0,0 +1,7 @@
+{
+  "cluster": "test",
+  "namespace": "srtl-test",
+  "environment": "test",
+  "deploy_azure_backing_services": false,
```
RMcVelia, line 5: remove deploy_azure_backing_services for test as we will use azure backing service

terraform/application/config/test.tfvars.json
```diff
@@ -0,0 +1,7 @@
...
+  "enable_postgres_ssl": false
```
RMcVelia, line 6: remove enable_postgres_ssl for test as we want the default (true)

global_config/test.sh
```diff
@@ -0,0 +1,6 @@
+CONFIG=test
+ENVIRONMENT=test
+CONFIG_SHORT=test
```
RMcVelia, line 3: CONFIG_SHORT should be ts

slorek (changes requested):
Good work and nice to see how quickly this works compared to the current review apps!

There were a few issues I have noticed or unable to test:

- Basic auth should be implemented as with the existing review apps. This is primarily to prevent Google indexing the review app.
- If I remove the `deploy` label should the review app be deleted? This is not a blocker but something to consider.
- We need to update the DfE Identity whitelist to include the new domains, as currently it is not possible to sign in
- Ordnance survey API address lookup is not working - possibly missing credentials?
- OTP emails are not being sent to the GOV.UK Notify API - no exception is raised even if I input a non-Notify-team email address
- I can't test the HMRC API integration as I can't get past the email confirmation page currently
- `/admin` gives a `Forbidden` message.
- We need to check if the `BYPASS_DFE_SIGN_IN` environment variable is set and working (may be related to above)
- When I closed my test pull request (https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2744) the 'delete review app' workflow didn't run (possibly due to the branch condition) - see https://github.com/DFE-Digital/claim-additional-payments-for-teaching/actions/runs/9093876927/job/24993757797

Please check all the credentials in the development Azure key vault are being bound to the app.

.github/workflows/build_and_deploy.yml
```diff
@@ -0,0 +1,77 @@
...
+        uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          header: aks
+          message: |
+            ### Deployments
+
+            | Journey             | URL                                  |
+            | ------------------- | ------------------------------------ |
+            | Additional Payments | <${{ env.APP_URL }}/additional-payments/claim> |
+            | Student Loans       | <${{ env.APP_URL }}/student-loans/claim>       |
```
slorek, line 77:
```suggestion
            | Student Loans       | <${{ env.APP_URL }}/student-loans/claim>       |
            | Admin               | <${{ env.APP_URL }}/admin>       |
```

terraform/application/config/review_app_env.yml
```diff
@@ -0,0 +1,37 @@
+---
```
slorek, line 1: Do we need to include dummy secrets in this file? It may be confusing.

slorek (commented):
We should also update the documentation in `/docs` - we can point to https://github.com/DFE-Digital/teacher-services-cloud/blob/main/documentation/developer-onboarding.md but worth identifying the namespaces and cluster names in our documentation. The`first-line-support-developer-runbook.md` would also be useful to provide some quick examples.

.github/workflows/build_and_deploy.yml
```diff
@@ -0,0 +1,77 @@
...
+        uses: DFE-Digital/github-actions/build-docker-image@master
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          context: .
+
+  deploy_review:
+    name: Deploy to review environment
+    concurrency: deploy_review_${{ github.event.pull_request.number }}
+    runs-on: ubuntu-latest
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'deploy') }}
```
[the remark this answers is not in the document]
AbigailMcP: good point! we'll make it so that they only build when the label is added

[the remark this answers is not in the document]
RMcVelia: Our preferred standard is to run the build on push (even if 'deploy' isn't added). As it's a good test for the build to run without error. We don't mandate this though, just a general preference.

AbigailMcP:
Things to do:

* Basic auth should be implemented as with the existing review apps. This is primarily to prevent Google indexing the review app.
* ~If I remove the `deploy` label should the review app be deleted? This is not a blocker but something to consider.~
* ~We need to update the DfE Identity whitelist to include the new domains, as currently it is not possible to sign in. I will request this now.~ Edit: now done
* DfE Identity sign in isn't working: `error_description:The mandatory 'redirect_uri' parameter is missing`. I think this may have to do with the `CANONICAL_HOSTNAME` environment variable (and possily `TID_BASE_URL` as well) not being present on the new platform.
* Ordnance survey API address lookup is not working - possibly missing credentials?
* OTP emails are not being sent to the GOV.UK Notify API - no exception is raised even if I input a non-Notify-team email address
* I can't test the HMRC API integration as I can't get past the email confirmation page currently
* `/admin` gives a `Forbidden` message.
* We need to check if the `BYPASS_DFE_SIGN_IN` environment variable is set and working (may be related to above)
* When I closed my test pull request ([Test AKS review apps #2744](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2744)) the 'delete review app' workflow didn't run (possibly due to the branch condition) - see https://github.com/DFE-Digital/claim-additional-payments-for-teaching/actions/runs/9093876927/job/24993757797
* I then opened a new PR ([Test aks review apps 2 #2745](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2745)) from a new branch (also based off this one) and it then deleted the review app from the previous PR (https://github.com/DFE-Digital/claim-additional-payments-for-teaching/actions/runs/9093876916/job/24993757761?pr=2745), which was on a different branch. I think we need to fix this as it will probablty happen whenever we use a feature branch. It might be related to the branch conditions of the workflow again.
* My deployment activity on the test pull requests (based off this branch) are appearing in the timeline for this PR.
* Please check all the credentials in the development Azure key vault are being bound to the app. It's not clear to me where (or if) this is happenning

terraform/domains/environment_domains/config/production.tfvars.json
```diff
@@ -0,0 +1,12 @@
+{
+  "hosted_zone": {
+    "claim-additional-teaching-payment.service.gov.uk": {
+      "front_door_name": "#FRONT_DOOR_NAME#",
```
saliceti, line 4: The template was not rendered here. Should we remove all the non review enviroment files?

AbigailMcP: good idea, will do!

.github/actions/deploy-environment/action.yml
```diff
@@ -0,0 +1,44 @@
...
+    description: The pull request number which triggered this deploy.
+    required: false
+
+runs:
+  using: composite
+
+  steps:
+    - uses: hashicorp/setup-terraform@v3
+      with:
+        terraform_version: 1.6.4
```
saliceti, line 24: Since we're starting fresh, can we use a more recent version?

.github/workflows/build_and_deploy.yml
```diff
@@ -0,0 +1,77 @@
...
+      - uses: azure/login@v2
+        with:
+          creds: ${{ secrets.AZURE_CREDENTIALS }}
+
+      - name: Seed review app
+        shell: bash
+        if: github.event.number != ''
+        run: |
+          make ci review-aks get-cluster-credentials
+          kubectl exec -n srtl-development deployment/claim-additional-payments-for-teaching-review-${{ github.event.pull_request.number }} -- sh -c "DISABLE_DATABASE_ENVIRONMENT_CHECK=1 bin/rails db:schema:load db:seed"
```
saliceti, line 61: Add this action above to guarantee the kubectl version: https://github.com/DFE-Digital/github-actions/tree/master/set-kubectl

.github/workflows/ci.yaml
```diff
@@ -24,6 +24,33 @@ jobs:
       - name: Linting - Standardrb
         run: bin/rails standard
 
+  lint-dfe-analytics:
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:11-alpine
```
saliceti, line 31: Can we take the opportunity to migrate to a more recent version? 16 is available

Makefile
```diff
@@ -96,6 +133,46 @@ deploy-arm-resources: arm-deployment ## Validate ARM resource deployment. Usage:
 
 validate-arm-resources: set-what-if arm-deployment ## Validate ARM resource deployment. Usage: make domains validate-arm-resources
 
+domains-infra-init: bin/terrafile domains composed-variables set-azure-account-aks
```
saliceti, line 136: The domains-* targets are not required in this PR

terraform/application/application.tf
```diff
@@ -0,0 +1,40 @@
...
+  service_short          = var.service_short
+  config_short           = var.config_short
+  secret_key_vault_short = "app"
+
+  is_rails_application = true
+
+  config_variables = merge(
+    local.app_env_values,
+    {
+      ENVIRONMENT_NAME = var.environment
```
saliceti, line 16: Formatting. I suggest [adding fmt and tflint](https://github.com/DFE-Digital/teacher-services-cloud/blob/main/.github/actions/validate-terraform/action.yml#L35) (but it can be in a different PR)

slorek (changes requested):
👏 Looking very good, only a handful of issues remaining.

These are the things I've tested:

✅ Basic auth
✅ DfE Identity sign in
✅ Claim submission
✅ Exception reporting to Rollbar
✅ Ordnance survey API address lookup
✅ HMRC API integration
✅ e-mails sent
✅ Admin login with DfE Sign In
✅ Levelling Up Premium Payments awards upload
✅ TPS data upload
✅ School data import
✅ Checked application and deployment logs for errors
✅ Opening a PR against another PR already deployed keeps the deployment status and PR comments separate from the base branch PR
✅ Close PR deletes the review app
✅ Reopening a close PR re-deploys the review app

Remaining Issues:

🛑 We should upgrade from Postgres 11 to Postgres 16 which we will be doing for the production environment at the same time as well
⚠️ dfe analytics lint check should be a required check in order to merge a PR. Currently the PR can be merged when the linter fails, which will result in a broken production deployment. We can set this as a required check once the PR is merged but we must remember to do so.
⚠️ If a migration is subsequently added to the branch after the initial deployment, the migration doesn't run. This is not a blocker as the migration can be run manually using kubectl but we would also have to restart the application pod as well. This can be addressed in a follow up ticket as the existing review apps implementation doesn't work either
⚠️ I'm unable to check the log output for migrations in GitHub as the migrations don't run and there's no output (despite it saying running migrations)
🛑 DQT client is not working (I've checked the credentials are present and correct; seems to be a networking issue but works locally):
```
irb(main):002:0> c = Dqt::Client.new
irb(main):005:0> c.teacher.find("1886092")
/usr/local/lib/ruby/3.2.0/net/http.rb:1603:in `initialize': Failed to open TCP connection to preprod.teacher-qualifications-api.education.gov.uk:443 (getaddrinfo: Try again) (Faraday::ConnectionFailed)
```
⚠️ Minor - deleting of a review app shows as a new deployment in GitHub rather than a modification of the existing deployment. Once deleted, it continues to show the deployment as active and 'successfully deployed'. We can fix this in a follow up ticket.
⚠️ We need to update the docs, removing out of date Heroku/Azure/Google account instructions and add some examples and document the namespace name etc.

slorek (approved):

RMcVelia (approved):

PR #2739, opened 2024-05-10
Add a form object for selecting an address

felixclack:
When the user is adding their home address, we have a hard-coded address
selection form that can be migrated to a form object.

This change allows us to remove the AddressDetails concern from the
controller as the functionality it adds is moved into the form objects.


app/models/journeys/additional_payments_for_teaching.rb
```diff
@@ -25,7 +25,8 @@ module AdditionalPaymentsForTeaching
       "eligibility-confirmed" => EligibilityConfirmedForm,
       "correct-school" => CorrectSchoolForm,
       "reset-claim" => ResetClaimForm,
-      "postcode-search" => PostcodeSearchForm
+      "postcode-search" => PostcodeSearchForm,
+      "select-home-address" => SelectHomeAddressForm
```
AbigailMcP, line 29: We also have `SHARED_FORMS` in the `Journeys::Base` model for forms applicable to both journeys

app/controllers/claims_controller.rb
```diff
@@ -37,26 +36,7 @@ def show
...
     end
 
-    if params[:slug] == "select-home-address" && postcode
-      session[:claim_postcode] = postcode
-      session[:claim_address_line_1] = params.dig(:claim, :address_line_1)
-      if address_data.nil?
-        redirect_to claim_path(current_journey_routing_name, "no-address-found") and return
-      else
-        # otherwise it takes you to "no-address-found" on the backlink from the slug sequence
-        @backlink_path = claim_path(current_journey_routing_name, "postcode-search")
```
AbigailMcP, line 47: I _think_ this backlink path logic might be missing now.

AbigailMcP: yep! if it's not just a backlink to the previous slug i.e. there's some logic behind it, we override the `backlink_path` in the base form 👍

AbigailMcP (approved):

PR #2740, opened 2024-05-13
Store poor performance answers on the journey session

felixclack:
As part of the move to using the Journeys::Session model for persisting
answers, we want to store the poor performance answers.

The form objects accept a journey_session object now. This is the target
that we need to store the answers on.


app/views/additional_payments/claims/poor_performance.html.erb
```diff
@@ -7,6 +7,7 @@
       "subject_to_disciplinary_action": "claim_subject_to_disciplinary_action_true"
     }) if @form.errors.any? %>
     <%= form_for @form, url: claim_path(current_journey_routing_name) do |form| %>
+      <%= form.hidden_field :journey_session_id, value: journey_session.id %>
```
rjlynch, line 10: I think we probably want to pass this into the form object's constructor rather than pass it as a hidden field, which will be much easier to do once this pr https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2736 is merged

app/forms/journeys/additional_payments_for_teaching/poor_performance_form.rb
```diff
@@ -10,7 +11,14 @@ class PoorPerformanceForm < Form
       def save
         return false unless valid?
 
-        update!(eligibility_attributes: attributes)
+        update!(eligibility_attributes: attributes.except("journey_session_id"))
+        journey_session.update(answers: {eligibility_attributes: attributes.except("journey_session_id")})
```
rjlynch, line 15: I don't think we want to nest these answers under eligibility_attributes as the journey session is just a bag of answers and doesn't need to know about eligibility.

```suggestion
        journey_session.update(answers: attributes.except("journey_session_id"))
```

rjlynch (dismissed):
I think we'll want to change this to pass the journey session into the forms. I'll merge this https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2736 into main as soon as ci passes which will make it much easier to access the journey session

app/forms/journeys/additional_payments_for_teaching/poor_performance_form.rb
rjlynch: I think we'll want a method like this
```
    def update_answers!(args)
      answers.merge!(args)
      save!
    end
```
on `Journeys::Session`

app/forms/journeys/additional_payments_for_teaching/poor_performance_form.rb
```diff
@@ -10,7 +10,8 @@ class PoorPerformanceForm < Form
       def save
         return false unless valid?
 
-        update!(eligibility_attributes: attributes)
+        update!(eligibility_attributes: attributes.except("journey_session_id"))
+        journey_session.update(answers: attributes.except("journey_session_id"))
```
slorek, line 14: Another option would be something like:
```
journey_session.update(attributes.except("journey_session_id"))
```

So we don't need to worry about merging over overwriting in the client. We just need a generic setter in `Journeys::Session` which handles it.

slorek (approved):

PR #2742, opened 2024-05-13
CAPT-1633 Form objects for reminders | CAPT-1634 Encapsulate pre/post form rendering/submission logic per-journey (Part 1)

your:
This PR is the first iteration of CAPT-1634 applied to an isolated controller that handles a mini-journey for reminders (ECP/LUPP) and converts the reminders slugs to form objects as per CAPT-1633. In a follow-up PR, I will be adding the new concern to the claims controller and iterate with any improvements or changes required.

### How to review / my notes

1. Review the `FormSubmittable` concern and associated request spec; I won't repeat the rationale here as it's explained extensively in various comments inside the concern; worth noting that:
  - It's a first iteration, designed with speed in mind; each GET/POST/PATCH request on a `<path>/:slug` will dynamically define all the pre-post rendering/submission callback methods for that slug only. I have considered defining them as singleton methods so we could tap into them in a more declarative way (i.e. passing a block or a method name and options), but I didn't feel like pursuing it yet (it would probably mean storing all the callbacks and options in a variable for *all* slugs in a single req-res cycle):
  ```
     # potential future iteration...

     after_form_save_success :retrieve_student_loan_details, on: :"personal-details"

     # ...vs current iteration

     def personal_details_after_form_save_success
       retrieve_student_loan_details
     end
  ```
  - I have changed the concern name so many times, but I really like `FormSubmittable` now as it helps define a specific trait of controllers that handle form-based page sequences
  - This approach can allow for cleaner code as long as developers keep the callback implementations readable and slim
  - Callbacks can be difficult to debug; so are the rails controller filters (`*_action`), and are necessary for complex request cycles. I have added logging when filters are triggered; see an example here:
  ```
tail -f log/development.log| grep callback
2024-05-13 20:34:15.434222 I [79219:puma srv tp 005] {request_id: d0d27584-edb4-428d-92fb-cf40c0472e2f} Journeys::AdditionalPaymentsForTeaching::RemindersController -- Executing callback #handle_form_submission
2024-05-13 20:34:15.434457 I [79219:puma srv tp 005] {request_id: d0d27584-edb4-428d-92fb-cf40c0472e2f} Journeys::AdditionalPaymentsForTeaching::RemindersController -- Executing callback #email_verification_after_form_save_failure
2024-05-13 20:34:42.736920 I [79219:puma srv tp 001] {request_id: eb06c4f0-40b3-44f4-90a1-d5ee838d25d8} Journeys::AdditionalPaymentsForTeaching::RemindersController -- Executing callback #email_verification_before_update
2024-05-13 20:34:42.739362 I [79219:puma srv tp 001] {request_id: eb06c4f0-40b3-44f4-90a1-d5ee838d25d8} Journeys::AdditionalPaymentsForTeaching::RemindersController -- Executing callback #handle_form_submission
2024-05-13 20:34:42.752033 I [79219:puma srv tp 001] {request_id: eb06c4f0-40b3-44f4-90a1-d5ee838d25d8} Journeys::AdditionalPaymentsForTeaching::RemindersController -- Executing callback #email_verification_after_form_save_success
  ```
  - The concern is not tight to the `PageSequence` logic (which should probably be revised anyway), and only expects `slugs` to be defined along with `current_slug` and `next_slug`, so it can be used with any sequence
2. Look at the new form objects for the reminders
  - I have realised that our `Form#model_name` should probably be generic and not linked to any particular AR model. I'm using a generic "form" name for the newly converted reminders slugs, but it could be extended to all the other forms
  - I have tried to re-use existing form objects for the reminders slugs, but:
    - `personal-details` is quite different from our "shared" one, as in this case it only focuses on `full_name` and `email_address`. I don't think it's worth extracting validations and sharing, not in this PR
    - `email-confirmation` is similar to the "shared" one, but in the shared one the OTP generation timestamp is an attribute of the `current_claim`... I didn't want that to be an attribute of `current_reminder`, but rather a transient attribute for the form object, hence I have injected it into the session and proxied to the form object. A similar approach could be used for the shared one
  - I'm using a `current_data_object` wrapper as the `claim:` argument to initialising the form object; I think it's a good abstraction
  - Forms needed to be re-organised per controller and not just journey
  - I have realised that our reminders journey is always "open", even if a window is closed... addressed that as well.

https://dfedigital.atlassian.net/browse/CAPT-1633
https://dfedigital.atlassian.net/browse/CAPT-1634

### DIFF
--- a/app/controllers/concerns/form_submittable.rb
+++ b/app/controllers/concerns/form_submittable.rb
@@ -1,0 +1,167 @@
+module FormSubmittable
+  extend ActiveSupport::Concern
+
+  #
+  # This concern provides a way of handling form submissions for a generic slug sequence.
+  #
+  # The `new`, `show`, `create`, and `update` actions can be overridden only if necessary,
+  # but it's discouraged as you could easily break the rendering cycle and callback chain.
+  # If you need to override actions, you are probably not dealing with a form-based page sequence.
+  #
+  # The average use case will most likely only require to override slug-specific callbacks
+  # that are used to execute custom logic before, after, or around certain actions.
+  # In some cases, you may not need to define any callbacks at all.
+  #
+  # Default behaviour summary for each action:
+  #
+  #   controller#new: `redirect_to_first_slug`
+  #   controller#show: `before_show` -> `render_template_for_current_slug`
+  #   controller#update: `before_update` -> `handle_form_submission` ->
+  #     -> `form#save` succeded? ->
+  #       -> `execute_callback_if_exists(:after_form_save_success)` OR `redirect_to_next_slug`
+  #     -> `form#save failed?` ->
+  #       -> `execute_callback_if_exists(:after_form_save_failure)` OR `render_template_for_current_slug
+  #   controller#create: same as controller#update
+  #
+  # When including this concern make sure that these methods are accessible in the controller:
+  #
+  # - `slugs`, `current_slug`, `next_slug` (normally from `PageSequence`, or overridden)
+  # - `journey` (from `PartOfJourneyConcern`)
+  # - `current_data_object`, required to load the form object
+  # - any slug-specific callbacks (via a separate mixin)
+  #
+  # Important: `current_slug` is used to generate callbacks and assumed safe to use, i.e. derived
+  # from constrained, validated or sanitised user input.
+  #
+  # Important: If there are other callbacks in the controller, you should be including this concern
+  # **after** all the callbacks are defined. Placing this at the top is most likely a bad idea.
+  # In most cases, form submission callbacks are meant to kick in last.
+  #
+  # See the implementation of `_set_slug_specific_callbacks` below for more details.
+
+  included do
+    before_action :_set_slug_specific_callbacks, only: [:show, :update, :create]
+    before_action :before_show, only: :show
+    before_action :before_update, only: [:update, :create]
+    before_action :load_form_if_exists, only: [:show, :update, :create]
+    around_action :handle_form_submission, only: [:update, :create]
+
+    def current_data_object
+      # This is the instance of the main resource handled by the form object,
+      # and should always be defined in the controller.
+      nil
+    end
+
+    def new
+      redirect_to_first_slug
+    end
+
+    def show
+      render_template_for_current_slug
+    end
+
+    def create
+      # Note: if implemented, this action will be yielded at the end of `handle_form_submission`
+    end
+
+    def update
+      # Note: if implemented, this action will be yielded at the end of `handle_form_submission`
+    end
+
+    private
+
+    #
+    # Slug-specific callbacks are generated and executed around `show`, `update`, `create` actions,
+    # For example, for the "personal-details" slug, the following callbacks are available:
+    #
+    # `personal_details_before_show`
+    # `personal_details_before_update`
+    # `personal_details_after_form_save_success` (*)
+    # `personal_details_after_form_save_failure` (*)
+    #
+    # Ensure that the callbacks are implemented only where really needed.
+    # Consider organizing the callback methods in one mixin per journey and controller.
+    #
+    # (*) If you need to define these callbacks, the default rendering behaviour will not be
+    # followed, so you'll have to explicitly define what to do next (render/redirect_to).
+
+    def _set_slug_specific_callbacks
+      %i[before_show before_update after_form_save_success after_form_save_failure].each do |callback_name|
+        self.class.send(:define_method, callback_name) do
+          execute_callback_if_exists(callback_name)
+        end
+      end
+    end
+
+    def redirect_to_slug(slug)
+      raise NoMethodError, "End of sequence: you must define #{current_slug.underscore}_after_form_save_success" unless next_slug
+      raise NoMethodError, "Missing path helper for resource: \"#{path_helper_resource}\"; try overriding it with #path_helper_resource" unless respond_to?(:"#{path_helper_resource}_path")
+
+      redirect_to send(:"#{path_helper_resource}_path", current_journey_routing_name, slug)
+    end
+
+    def redirect_to_next_slug
+      redirect_to_slug(next_slug)
+    end
+
+    def redirect_to_first_slug
+      redirect_to_slug(first_slug)
+    end
+
+    def path_helper_resource
+      controller_name.singularize
+    end
+
+    def render_template_for_current_slug
+      render current_template
+    end
+
+    def current_template
+      current_slug.underscore
+    end
+
+    def slugs
+      journey.slug_sequence::SLUGS
+    end
+
+    def first_slug
+      slugs.first.to_sym
+    end
+
+    def execute_callback_if_exists(callback_name)
+      callback_name = :"#{current_slug.underscore}_#{callback_name}"
+      if respond_to?(callback_name)
+        log_event(callback_name) { send(callback_name) }
+        return true
+      end
+      false
+    end
+
+    def handle_form_submission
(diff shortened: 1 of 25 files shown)


app/controllers/journeys/additional_payments_for_teaching/reminders_form_callbacks.rb
```diff
@@ -0,0 +1,95 @@
...
+
+        if current_reminder.email_verified? && current_reminder.save
+          ReminderMailer.reminder_set(current_reminder).deliver_now
+
+          redirect_to reminder_path(current_journey_routing_name, "set")
+        end
+      end
+
+      def inject_sent_one_time_password_at_into_the_form
+        params[:form]&.[]=(:sent_one_time_password_at, session[:sent_one_time_password_at])
```
rjlynch, line 48: Could we store the `sent_one_time_password_at` on the `Reminder`? Then we can access it in the form object which might make things a bit simpler

app/controllers/journeys/additional_payments_for_teaching/reminders_form_callbacks.rb
```diff
@@ -0,0 +1,95 @@
...
+        try_mailer { set_a_reminder_immediately_if_possible }
+      end
+
+      def email_verification_before_update
+        inject_sent_one_time_password_at_into_the_form
+      end
+
+      def personal_details_after_form_save_success
+        update_reminder_id
+        try_mailer { send_verification_email } || return
```
rjlynch, line 18: What are your thoughts on sending the verification email from inside the personal details form, or the other forms? Seems like changing email address and sending a verification email are pretty closely related concepts additionally it would let us move some of the form error setting responsibilities out of this module

app/controllers/journeys/additional_payments_for_teaching/reminders_form_callbacks.rb
your: I understand your question, and I remember seeing you explain why that was a good idea for the `Claim` model.

My thoughts are that storing `sent_one_time_password_at`,  a transient attribute, directly on the `Reminder` model might introduce unnecessary coupling between the model and the form object. While with the existing approach there is coupling between the controller and the form object, this coupling is more acceptable and expected within the context of MVC architecture. Controllers (and the "state", e.g. in the session) are meant to interact with form objects to handle form submissions and orchestrate the flow of data within the application. Either way I don't feel too strong, and it might make sense to have a consistent approach, but I wouldn't consider it a blocker. Will have a think

app/controllers/journeys/additional_payments_for_teaching/reminders_form_callbacks.rb
your: > What are your thoughts on sending the verification email from inside the personal details form, or the other forms? Seems like changing email address and sending a verification email are pretty closely related concepts additionally it would let us move some of the form error setting responsibilities out of this module

@rjlynch, it all boils down to where we decide to put the boundaries of what a form object can and cannot do, which is what prompted the creation of this ticket. I had originally figured that our forms were, in practice, becoming service objects, and I felt ok with that until we decided that it was not ideal and something else had to be done – but as a team, we accepted the fact we were going to complete the form objects epic without a better separation of concerns.

With that context in mind, I think it's best to handle the mailer firing task outside the form object, possibly in controller callbacks or service objects. By separating concerns in this way, we allow the form object to focus on its primary responsibilities (validation, data manipulation and communicating response with the controller), and we avoid violating the single responsibility principle. I understand it's likely we'll always need to verify an email address after changing it, but what should keep those two concepts together is a service object, and in the absence of that, a controller callback – exactly in the same way you wouldn't want to keep those two things together in a model.

rjlynch: > I think it's best to handle the mailer firing task outside the form object,

Agreed, the usual approach would be using a job for things like sending emails etc something like

```
def update
  if @some_form.save
   SomeJob.perform_later(@some_form.record)
   redirect_to @some_form.record
  else
    render "edit"
  end
end
```

A bit tricker in our set up but we could infer the job to call to handle post form submission tasks from the name of the slug. Also would probably want to use `perform_now`. Might have the advantage of being easier to test compared to a controller callback?

app/controllers/journeys/additional_payments_for_teaching/reminders_form_callbacks.rb
rjlynch: I'm not sure `sent_one_time_password_at` is transient 🤔 but I don't think it's too big a deal to keep it in the session

app/controllers/journeys/additional_payments_for_teaching/reminders_form_callbacks.rb
your: > we could infer the job to call to handle post form submission tasks from the name of the slug

My concern with that is that it's a solution for one specific context – sending emails – while we have seen that there are other generic tasks that need to be performed after a form is saved. So the idea of having a callback is that you can put anything you need in it, and hopefully survives future changes of requirements. That said, I'm open to exploring other solutions if they can propose a re-usable and slug-agnostic pattern.

> Might have the advantage of being easier to test compared to a controller callback?

I take the point about testing, and I didn't even make an attempt at unit testing as those behaviours should be captured in request specs (which we should make more use of, in my opinion) or features specs.

> Probably unrelated to the changes in this pr though, maybe one to discuss with the team

This is a good discussion, and happy to extend it to the rest of the team. I'll try to draft how this applies to the claims controller so we can see what it means at scale.

slorek: My view on this is that the form object's responsibility should end at validating user input; the persistence of data and other things such as sending e-mails or scheduling jobs should be handled within a service object. This allows us to execute these processes outside of the context of a web form. I think this is the direction we should iterate towards.

app/controllers/concerns/form_submittable.rb
```diff
@@ -0,0 +1,167 @@
...
+
+    private
+
+    #
+    # Slug-specific callbacks are generated and executed around `show`, `update`, `create` actions,
+    # For example, for the "personal-details" slug, the following callbacks are available:
+    #
+    # `personal_details_before_show`
+    # `personal_details_before_update`
+    # `personal_details_after_form_save_success` (*)
```
slorek, line 79: I think we will want to iterate towards the forms _not_ being responsible for persistence, therefore moving those responsibilities to one of these callbacks might make sense. Given that this is likely to be repeated on everyt form I don't mind if we decide to keep as-is for simplicity, but I suspect we may find a way to do this in a re-usable way when we don't have the unnecessary complexity of multiple policy `Claim` records and `::Eligibility` records to work with. I would imagine we would only need a very simple interface to save the form's attributes to the `Journey:Session` record in this new `handle_form_submission` method. 

The naming of `after_form_save_success` might be inappropriate in that case. Whether we decide to change it now or later I don't mind, but I anticipate we would change it fairly soon.

slorek (approved):
👏 I think this is awesome work, thank you.

I do understand there will initially be a bit of extra cognitive load during debugging, but once you understand the callback structure (which you have documented well) it might actually make it a lot easier since the concerns are more clearly separated.

This is not the same problem as callbacks which are inserted in non-obvious ways since there is a fixed structure which will be commonly used as it unlocks the ability to support multiple journeys.

app/controllers/concerns/form_submittable.rb
your: I see what you are saying; we can try to do that in the next iteration

PR #2751, opened 2024-05-17
Split journey session per journey

rjlynch:
**There's quite a lot of files changed but the bulk of them are just updating the specs to use the journey session factory for the appropriate journey session. So don't worry about having to do a massive review** 😂 Probably easier to review each commit separately.

This is a follow up to https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2738 I've put these changes in a separate PR so we can discuss them more easily.

There's three changes in this pr.

The first is adding a `Journeys::Session` for each journey, as both journeys have different eligibility attributes and will likely have different behaviour hanging off the journey session it seems sensible to have separate objects. The shared base class contains the claim attributes that are common between the two journeys.

The next change is adding attributes and type casting to the answers hash. The type casting makes it easier to work with Academic years. Having methods rather than accessing values with `[ ]` makes it easier to mimic the `Claim`'s api in places where we need to access the same data pre and post claim submission (eg working with DqtRecords).

The third change is cleaning up some of the noise in the claim submission form, we don't really need the list of attributes in each of the claim subclasses as it's now just duplicating info stored on the journey sessions.

Let me know if we want to go with this approach or if this a bit ott (happy to jump on call to discuss anything)
There's still the outstanding question from PR #2738 as to whether we want to store a reference to the journey session on the claim or whether we want to add a submitted_at time stamp to the journey session to detect if a session has already been submitted.

If this PR looks good I'll merge it into #2738 and then ship #2738


app/forms/journeys/additional_payments_for_teaching/claim_submission_form.rb
```diff
@@ -1,36 +1,15 @@
...
+      def journey
+        Journeys::AdditionalPaymentsForTeaching
       end
 
       def selected_claim_policy
-        if selected_policy.present?
-          "Policies::#{selected_policy}".constantize
+        if answers.selected_policy.present?
+          "Policies::#{answers.selected_policy}".constantize
         end
```
slorek, line 13: Please change: Since this method can only ever safely use two policies' classes, we should be able to avoid using `constantize` from user input here.

```suggestion
        answers.selected_policy == "EarlyCareerPayments" ? Policies::EarlyCareerPayments : Policies::LevellingUpPremiumPayments
        end
```

We could even consider raising an exception if it's not one of the two expected values.

app/forms/journeys/additional_payments_for_teaching/claim_submission_form.rb
```diff
@@ -1,36 +1,15 @@
...
-        attribute name, type
-      end
-
       private
 
-      def eligibility_attribute_names
-        ELIGIBILITY_ATTRIBUTES.map(&:first)
+      def journey
+        Journeys::AdditionalPaymentsForTeaching
       end
```
slorek, line 8: I'm not sure if this is necessary as this class lives in the relevant journey namespace. If we still need the class to tell us this (I'm not sure where it's used) we could potentially implement it in `ClaimSubmissionBaseForm` using `Module.nesting.first`.

app/models/journeys/additional_payments_for_teaching/session.rb
```diff
@@ -0,0 +1,11 @@
+module Journeys
+  module AdditionalPaymentsForTeaching
+    class Session < Journeys::Session
+      attribute :answers, SessionAnswersType.new
+
+      def journey_module
+        Journeys::AdditionalPaymentsForTeaching
+      end
```
slorek, line 8: I'm not sure if this is necessary as this class lives in the relevant journey namespace. If we still need the class to tell us this (I'm not sure where it's used) we could potentially implement it in `ClaimSubmissionBaseForm` using `Module.nesting.first`.

slorek (changes requested):
👏 This looks really good.

I do think it adds a little complexity but it does solve a number of issues we identified. Overall I'm comfortable with it.

As you mentioned I think there's an open question about whether data in the session table is ephemeral or retained for future reference. I see advantages and disadvantages.

With these changes we now have a model class around the answer set data, so we could potentially re-use it regardless of where the data is stored, e.g. if it's copied to the claim on submission.

Just a couple of minor points added to inline comments.

app/forms/journeys/additional_payments_for_teaching/claim_submission_form.rb
rjlynch: It's used to find the policies we need to build eligibilities. I've now moved it into the base form as we can work it out from there 👍

app/forms/journeys/additional_payments_for_teaching/claim_submission_form.rb
rjlynch: Good point. I think raising if an unexpected policy makes it here is best. Have gone for a case statement as we need to support the `selected_policy` being `nil`

app/models/journeys/teacher_student_loan_reimbursement/session_answers_type.rb
```diff
@@ -0,0 +1,38 @@
+module Journeys
+  module TeacherStudentLoanReimbursement
+    class SessionAnswersType < ActiveModel::Type::Value
```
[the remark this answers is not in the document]
rjlynch: I left them separate to avoid any meta programmy type stuff as they each need to access the `SessionAnswers` class in their respective modules. I'm not opposed to moving to a single SessionAnswersType and doing something with `module_parent` to get at the correct `SessionAnswers` class if you think that's better. Thought I think these SessionAnswersType classes are unlikely to be changed. What do you think?

rjlynch:
> 👏 This looks really good.
> 
> I do think it adds a little complexity but it does solve a number of issues we identified. Overall I'm comfortable with it.
> 
> As you mentioned I think there's an open question about whether data in the session table is ephemeral or retained for future reference. I see advantages and disadvantages.
> 
> With these changes we now have a model class around the answer set data, so we could potentially re-use it regardless of where the data is stored, e.g. if it's copied to the claim on submission.
> 
> Just a couple of minor points added to inline comments.

@slorek Thanks for the review, have addressed your comments. I'm going to merge this PR into the base branch and we can pick up any issues from there. @felixclack probably set to base any work off of `submit-claim-from-journey-session` and then rebase onto master once that PR has the green light

PR #2756, opened 2024-05-20
CAPT-1625 Move teacher_reference_number from claim to eligibility

kenfodder:

### DIFF
--- a/config/analytics.yml
+++ b/config/analytics.yml
@@ -151,7 +151,6 @@ shared:
     - had_leadership_position
     - mostly_performed_leadership_duties
     - claim_school_somewhere_else
-    - teacher_reference_number
   :early_career_payments_eligibilities:
     - id
     - nqt_in_academic_year_after_itt
@@ -170,7 +169,6 @@ shared:
     - award_amount
     - induction_completed
     - school_somewhere_else
-    - teacher_reference_number
   :levelling_up_premium_payments_eligibilities:
     - id
     - created_at
@@ -190,7 +188,6 @@ shared:
     - eligible_degree_subject
     - induction_completed
     - school_somewhere_else
-    - teacher_reference_number
   :international_relocation_payments_eligibilities:
     - id
     - created_at
--- a/config/analytics_blocklist.yml
+++ b/config/analytics_blocklist.yml
@@ -19,7 +19,7 @@
   - building_society_roll_number
   - banking_name
   - mobile_number
-  - teacher_reference_number
+  - column_to_remove_teacher_reference_number
   - email_address
   - payroll_gender
   - one_time_password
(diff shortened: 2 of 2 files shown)


slorek (approved):

app/models/journeys/page_sequence.rb
```diff
@@ -119,5 +119,13 @@ def claim_submittable?
     def journey_ineligible?
       @journey_ineligible ||= journey::EligibilityChecker.new(journey_session: @journey_session).ineligible?
     end
+
+    def completed?
+      if journey == Journeys::AdditionalPaymentsForTeaching
```
rjlynch, line 124: Is TRN not required for student loans now?

Is there a reason this can't just be `claim_submittable?`. The claim submission form (which `claim_submittable?` checks) constructs an eligibility* and checks if it's valid so if the eligibility requires a TRN then `claim_submittable?` will return false if the TRN isn't set, so we don't need to check the answers for the presence of a TRN. If the eligibility doesn't require the TRN to be set then `claim_submittable` will handle that too. I might be missing something, in which case this is fine.


*(I actually want to change this so it just uses the policy checkers and only creates an eligibility on save but that's a separate matter!)

rjlynch (commented):
Looks good to me but will hold off approving until Steve's had a look as I'm not super familiar with the admin side of things!

app/models/claim/claims_preventing_payment_finder.rb
```diff
@@ -22,14 +22,20 @@ def claims_preventing_payment
     private
 
     def find_claims_preventing_payment
-      payrollable_claims_from_same_claimant = Claim.payrollable.where(teacher_reference_number: claim.teacher_reference_number)
+      eligibility_ids = [
```
slorek, line 25: Suggestion: I think this is likely to be overlooked and may be best configured in the policies. Also all of the new policies being added won’t have a TRN so we should create a new ticket to decide on a solution for matching payees.

app/models/claim/matching_attribute_finder.rb
```diff
@@ -1,29 +1,52 @@
 class Claim
   class MatchingAttributeFinder
-    ATTRIBUTE_GROUPS_TO_MATCH = [
-      ["teacher_reference_number"],
+    STUDENT_LOANS_AND_ADDITIONAL_PAYMENTS_GROUP = [
```
slorek, line 3: I think this configuration location could be missed and may be best configured in the Policies namespaces.

app/models/claim/search.rb
```diff
@@ -1,14 +1,23 @@
 class Claim
   # Accepts a search term and returns all Claims that match against any of the
-  # attributes defined in the `SEARCHABLE_ATTRIBUTES` constant. Both subject
+  # attributes defined in the `SEARCHABLE_ATTRIBUTES` or `SEARCHABLE_ELIGIBILITY_ATTRIBUTES` constant. Both subject
   # and attribute are downcased, so the search is case-insensitive.
   class Search
     attr_accessor :search_term
 
-    SEARCHABLE_ATTRIBUTES = %w[
+    POLICIES = [
```
slorek, line 8: It feels like this could be better configured in the relevant Policy classes.

slorek (approved):

PR #2758, opened 2024-05-21
CAPT 1638/migrate sign in or continue to journey session

rjlynch:
wip


rjlynch:
A note on the implementation of `JourneySession` and `SessionAnswers`: There's quite a few places where you just want to work with the answers only but have to go through the journey session in order to be able to save changes, additionally not having a `created_at` timestamp on the `SessionAnswers` means some methods that naturally belong on there need to be on the `JourneySession` instead. Once we've migrated all the forms over to write answers to the `JourneySession` I think we should take a look at storing the answers in a table rather than a json column and removing the `JourneySession` instead having something like `Journeys::JourneyName::AnswerSet` which has a column for each answer, and something like a `delegated_type` if we want to share the claim fields between the journey models. Hopefully the active model attributes should make it relatively straightforward to switch out how we're persisting the data. Not one for this PR just something to consider in the future!

rjlynch:
I appreciate there's a lot of changes here so I you want to walk though any of this on a call just ping me on slack 👍

rjlynch:
@felixclack @kenfodder couple of things from this PR might be relevant for you guy's work

* You can specify whether or not a question is considered as answered by the `SessionAnswers`. This is useful for controlling whether to get attributes from `SessionAnswers` or from `CurrentClaim` when setting attributes on a form. By default any attributes change on `SessionAnswers` will be passed into the answers has, but now you can explicitly pass an array to `answered` when updating attributes to only count the attributes you want to as having been answered by the `SessionAnswers`. See [`Form#load_current_value`](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/4a831bb97444b93b0640227fd108525ef0bde132/app/forms/form.rb#L80) and the changes to the [`Journeys::Session#before_save` callback](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/428d74b0e0279096f5e58ce3dadf9c668bb3a270/app/models/journeys/session.rb#L35)

* [We now use the `SessionAnswers` to check if personal details have been changed from those pulled from teacher id](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/428d74b0e0279096f5e58ce3dadf9c668bb3a270/app/models/journeys/session_answers.rb#L68). In CAPT-1659 we'll be removing the `claim` parameter to this method

* [Some factories have been set up for populating `JourneySession`](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/428d74b0e0279096f5e58ce3dadf9c668bb3a270/spec/factories/journeys/additional_payments_for_teaching/session_answers.rb). Can use them like
```ruby
create(
  :additional_payments_session,
  answers: attributes_for(
    :additional_payments_answers,
    :with_personal_details,
    email: "overwritten@email.com"
  )
)
```
Not too bad to use though if we eventually switch to backing the answers with it's own table we can clean this up a bit more

* Changed some bits of the slug sequences to refer to the session answers

* The dqt record and tps changes I don't think should be relevant to your work

slorek (approved):
Thank you. Just some minor suggested tweaks.

PR #2759, opened 2024-05-21
CAPT-1634 Encapsulate pre/post form rendering/submission logic per-journey (Part 2)

your:
This PR completes the work started with #2742 by making the `ClaimsController`... _FormSubmittable_.

The main changes are:

1. Slug-specific actions are moved to a dedicated mixin with callbacks; as a future improvement, they could be organised per journey (better after splitting the controller) but also per action
2. The backlink path override is not expressed by form objects anymore
3. Minor tweaks and claims controller cleanup 

https://dfedigital.atlassian.net/browse/CAPT-1634


slorek (approved):
So much better, thank you for your hard work on this.

PR #2760, opened 2024-05-22
Store the correct school on the session

felixclack:
As part of the migration to storing all answers on the
`Journeys::Session` object, we want to store the correct school
attributes there as part of the current journey.

In a future change we will remove the claim.update call and rely solely
on the journey session.


rjlynch (approved):
This change looks fine 👍 
Is updating this form to not write the current school to the eligibility going to be a separate pr or tacked on to this?

slorek (approved):

PR #2766, opened 2024-05-24
CAPT-1419 ECP exit screen showing incorrect text for AY24-25 (when induction not completed)

your:
The rationale for the changes is explained in the commits; in short, we do have the concept of start/end date for ECP (and LUP), but it's used as a "combined" value in the `EligibilityCheckable` concern... this is a problem when the end year of a particular policy affects the eligibility criteria for that policy only (like with the induction question).

https://dfedigital.atlassian.net/browse/CAPT-1419

Needs manual testing

### DIFF
--- a/app/models/journeys/additional_payments_for_teaching/slug_sequence.rb
+++ b/app/models/journeys/additional_payments_for_teaching/slug_sequence.rb
@@ -183,13 +183,14 @@ module Journeys
         )
       end
 
-      def replace_ecp_only_induction_not_completed_slugs(sequence)
+      def replace_ecp_only_induction_not_completed_slugs(sequence, ecp_claim)
+        dead_end_slug = (ecp_claim.eligibility.status == :eligible_later) ? "eligible-later" : "ineligible"
+
         slugs = %w[
           current-school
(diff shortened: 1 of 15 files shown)


app/models/journeys/additional_payments_for_teaching/slug_sequence.rb
```diff
@@ -184,12 +184,16 @@ def personal_details_form
       end
 
       def replace_ecp_only_induction_not_completed_slugs(sequence)
+        # TODO: Remove this workaround to the shim class not reloading the answers after they're updated
+        eligibility = Policies::EarlyCareerPayments::PolicyEligibilityChecker.new(answers:)
+
+        dead_end_slug = (eligibility.status == :eligible_later) ? "eligible-later" : "ineligible"
+
```
rjlynch, line 191: Not a blocker but now the shim has been removed can we just do?
```suggestion
        dead_end_slug = (ecp_eligibility_checker.status == :eligible_later) ? "eligible-later" : "ineligible"
```

rjlynch (approved):
LGTM, left one comment where we can remove a work around for the shim as we no longer use the shim

PR #2772, opened 2024-05-30
[CAPT-529] style font for result number in search results

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-529
- Issue with font styling on school search results on iOS

# Changes

- Target css to use `govuk-body` instead of fallback font which looks like times new roman
- Should probably be fixed upstream but is quickly fixed here for now

# Screenshots

## before
before

## after
after


rjlynch (approved):
👍

slorek (approved):

PR #2773, opened 2024-05-30
CAPT 1671/bank or building society

rjlynch:
Another relatively straight forward form conversion pr.
This pr is split into two commits, though it's probably fine to review both
commits together.
Bulk of the changes are just spec updates.


app/forms/bank_details_form.rb
```diff
@@ -26,13 +26,15 @@ class BankDetailsForm < Form
...
     return false unless valid?
 
-    update!(
-      banking_name:,
-      bank_sort_code:,
-      bank_account_number:,
-      building_society_roll_number:,
-      hmrc_bank_validation_succeeded:
+    journey_session.answers.assign_attributes(
+      banking_name: banking_name,
```
asmega, line 30: i'm guessing you're not a fan of the implicit syntax 😂

rjlynch: No! Every time I look at I think it's a syntax error!

PR #2774, opened 2024-05-30
Fix flaky test

asmega:
# Context

- School name generated by factory which is backed by faker sometimes generates a name which includes an apostrophe which should be handled
- The output converts this to an HTML entity

# Changes

- Update spec to consider this and now escapes the string so it can find it on the page


rjlynch (approved):
Nice 👍

PR #2778, opened 2024-05-31
add rspec-retry

asmega:
# Context

- This codebase has flaky tests
- This adds a workaround whilst engineering time can be spent more productively

# Changes

- Add `rspec-retry` gem
- Use by adding labels to specs eg `js: true, retry: 3`


slorek (approved):
Bonus points if this could be applied automatically to the `flaky: true` label.

asmega:
> Bonus points if this could be applied automatically to the `flaky: true` label.

Done

PR #2781, opened 2024-05-31
CAPT 1675/trn form

rjlynch:
Updates trn form to write to session

Another commit in the migrating the forms to write to the session saga.


slorek (approved):
Looks fine, but I think modifying the `update!` action might be a faster way to deal with this repeated work.

PR #2783, opened 2024-06-03
Eligibility checker

kenfodder:
The goal of this PR is to introduce a Eligibility checker that doesn't rely on using the Eligibility class to obtain the `status` of a claim.

The checker:
* app/models/journeys/additional_payments_for_teaching/eligibility_checker.rb

Concerns extracted from the eligibilities, `include Eligible`
* app/models/concerns/policies/early_career_payments/eligible.rb
* app/models/concerns/policies/levelling_up_premium_payments/eligible.rb

Individual policy eligibility checkers, which uses the concern extracted from the eligibilities:
* app/models/policies/early_career_payments/policy_eligibility_checker.rb
* app/models/policies/levelling_up_premium_payments/policy_eligibility_checker.rb

This is doesn't cover everything as we have to yet to rely on using the just the `answers` from the journey session and so far everything is still saved to the Eligibility model until we attack the first one. As soon as one of the form objects uses this it will reveal further work.

Some key areas which I will see and issue:
* EligibilityConfirmedForm - `claims_eligible_now` and `claims_eligible_now_and_sorted` will need to use the Checker in some way, by this relies on `Claims` being returned, but really I think only the policy type is needed here not claim models.
* `return "ineligible" if claim.ineligible?` in `PageSequence` will need to use the Checker. 
* Possibly other things but the tests will test us this.

Bonus - the tests are running quicker now!

### DIFF
--- a/app/models/concerns/policies/early_career_payments/eligible.rb
+++ b/app/models/concerns/policies/early_career_payments/eligible.rb
@@ -5,16 +5,6 @@ module Policies
         Policies::EarlyCareerPayments
       end
 
-      def eligible_later_year
-        # This covers the case for *the other policy* (LUP) which has *no* ITT academic year set for a trainee teacher
-        # but this method will still be called and needs to run without error
-        if itt_academic_year.blank?
-          nil
-        else
-          JourneySubjectEligibilityChecker.new(claim_year: claim_year, itt_year: itt_academic_year).next_eligible_claim_year_after_current_claim_year(CurrentClaim.new(claims: [claim]))
-        end
-      end
-
       def award_amount
         super || BigDecimal(calculate_award_amount || 0)
       end
--- a/spec/models/early_career_payments/eligibility_spec.rb
+++ b/spec/models/early_career_payments/eligibility_spec.rb
@@ -78,55 +78,6 @@ RSpec.describe Policies::EarlyCareerPayments::Eligibility, type: :model do
     end
   end
 
-  describe "#eligible_later_year" do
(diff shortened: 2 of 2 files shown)


app/models/concerns/policies/early_career_payments/eligible.rb
```diff
@@ -0,0 +1,112 @@
...
+        Policies::EarlyCareerPayments
+      end
+
+      def eligible_later_year
+        # This covers the case for *the other policy* (LUP) which has *no* ITT academic year set for a trainee teacher
+        # but this method will still be called and needs to run without error
+        if itt_academic_year.blank?
+          nil
+        else
+          JourneySubjectEligibilityChecker.new(claim_year: claim_year, itt_year: itt_academic_year).next_eligible_claim_year_after_current_claim_year(CurrentClaim.new(claims: [claim]))
```
rjlynch, line 14: I'm working on a pr, based off this branch, to make this not depend on current claim 👍

rjlynch (approved):
Very nice work, this should unblock the rest of the tickets, thanks!

app/models/concerns/policies/early_career_payments/eligible.rb
```diff
@@ -0,0 +1,112 @@
...
+        end
+      end
```
rjlynch, line 16:
```suggestion
```
This method isn't used any where so we can drop it and lose the dependency on current claim. Test suite passes with out it

rjlynch (approved):
👍

PR #2787, opened 2024-06-04
[CAPT-1647] Port poor performance to journey session

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-1647

# Changes

- The `poor_performance` form now persists answers to journey session rather than eligibility
- Update references to these fields on answers rather than on eligbility
- Also added `.rspec_parallel` with configuration that should increase local test run times for those that run the test suite in parallel

### DIFF
--- a/app/models/journeys/additional_payments_for_teaching/answers_presenter.rb
+++ b/app/models/journeys/additional_payments_for_teaching/answers_presenter.rb
@@ -51,7 +51,7 @@ module Journeys
       def current_school
         [
           t("additional_payments.forms.current_school.questions.current_school_search"),
-          eligibility.current_school_name,
+          journey_session.answers.current_school.name,
           (journey_session.answers.school_somewhere_else == false) ? "correct-school" : "current-school"
         ]
       end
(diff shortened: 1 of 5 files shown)


app/models/journeys/additional_payments_for_teaching/answers_presenter.rb
```diff
@@ -156,12 +156,11 @@ def itt_academic_year
...
       def text_for_subject_answer
-        policy = eligibility.policy
-        subjects = JourneySubjectEligibilityChecker.new(claim_year: Journeys.for_policy(policy).configuration.current_academic_year,
-          itt_year: journey_session.answers.itt_academic_year).current_and_future_subject_symbols(policy)
+        subjects = JourneySubjectEligibilityChecker
+          .current_and_future_subject_symbols(journey_session.answers)
 
         if subjects.many?
-          t("additional_payments.forms.eligible_itt_subject.answers.#{eligibility.eligible_itt_subject}")
+          t("additional_payments.forms.eligible_itt_subject.answers.#{journey_session.answers.eligible_itt_subject}")
```
rjlynch, line 163: I'm not 100% sure we can change this. This eligibility is from the claim submission form (see the method at the top of this file), so it's created using the information from the answers. However it's a _single_ eligibility where as `current_and_future_subject_symbols` can return the subjects for multiple eligibilities. It's probably fine but might be one to check with @kenfodder to see if he can spot an issue with this change

rjlynch:
> with configuration that should increase local test run times
👀 😁

rjlynch (approved):
Looks good but I'd check with Ken or Steve WRT my comment before merging

app/models/journeys/additional_payments_for_teaching/claim_journey_session_shim.rb
```diff
@@ -46,11 +46,11 @@ def has_entire_term_contract
       end
 
       def subject_to_disciplinary_action
-        journey_session.answers.subject_to_disciplinary_action || try_eligibility(:subject_to_disciplinary_action)
+        journey_session.answers.subject_to_disciplinary_action
```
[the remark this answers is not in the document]
asmega: i've inlined everything

app/models/journeys/additional_payments_for_teaching/answers_presenter.rb
asmega: ok i've now reverted this back to the original implementation which now seems to work after rebasing

PR #2788, opened 2024-06-04
CAPT 1680/claim school form

rjlynch:
Migrates claim school to session answers.

Couple of changes to note
1.) We've moved claim_school_form.rb into the TSLR namespace as it's only used on student loans journey
2.) We've introduced an eligibility checker class.

Claim school is used in a bunch of eligibility checks, so we now need to
be able to check the eligibility of the answers _and_ of the eligibility, and
as we're part way through the migration away from current claim we need
to be able to check the eligibility of the combination of both! To do so
we've introduced a new class
`Policies::StudentLoans::EligibilityChecker` which accepts either a
session answers or a claim and can be queried for eligibility status. We
can also pass the "shim" to this class for when we need to check the
eligibility across a combination of answers and current claim.
Eventually this class should probably subsume
`lib/ineligibility_reason_checker` too but that's something for a future
commit.

### DIFF
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -43,6 +43,11 @@ class ClaimsController < BasePublicController
 
   def redirect_to_existing_claim_journey
     new_journey = Journeys.for_policy(current_claim.policy)
+
+    # Set the params[:journey] to the new journey routing name so things like
+    # journey_session that rely on the journey param find the correct journey.
+    params[:journey] = new_journey::ROUTING_NAME
+
     new_page_sequence = new_journey.page_sequence_for_claim(
       current_claim,
       journey_session,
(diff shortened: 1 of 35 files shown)


app/models/journeys/page_sequence.rb
```diff
@@ -116,12 +116,26 @@ def journey
...
+        journey_session: @journey_session
+      )
+    end
+
+    def eligibility_checkers
+      if journey == Journeys::TeacherStudentLoanReimbursement
+        [student_loans_eligibility_checker]
+      else
+        [claim]
+      end
```
rjlynch, line 138: another one where we'll eventually want to do something like
```ruby
journey::POLICIES.map { |p| p::EligibilityChecker.new(answers) }
```

once we've finished the migration and don't require the shim

app/views/claims/ineligible.html.erb
```diff
@@ -2,7 +2,25 @@
...
+      # we require this
+    if current_journey_routing_name == "student-loans"
+      shim = Journeys::TeacherStudentLoanReimbursement::ClaimJourneySessionShim.new(
+        current_claim: current_claim,
+        journey_session: journey_session
+      )
+      ineligibility_reason = Policies::StudentLoans::EligibilityChecker.new(shim.answers).ineligibility_reason
+    else
+      ineligibility_reason = current_claim.eligibility.ineligibility_reason
+    end %>
```
rjlynch, line 17: Again one that can be cleaned up once we have finished the migration.

app/controllers/claims_controller.rb
```diff
@@ -43,6 +43,11 @@ def start_new
 
   def redirect_to_existing_claim_journey
     new_journey = Journeys.for_policy(current_claim.policy)
+
+    # Set the params[:journey] to the new journey routing name so things like
+    # journey_session that rely on the journey param find the correct journey.
+    params[:journey] = new_journey::ROUTING_NAME
```
rjlynch, line 49: Annoying but we build a slug sequence for a _different_ journey to the journey specified in the params which was breaking as we were passing the "wrong" answers to the slug sequence. This is ugly but I think there's a deeper problem with how this flow works.

PR #2804, opened 2024-06-06
CAPT-1655 Update eligibility-confirmed to use the EligibilityChecker

kenfodder:
Not done on this PR is the `selected_claim` still is managed by the CurrentClaim - leaving that for a separate PR.


rjlynch (approved):
Nice 👍 we landed on pretty much the same approach to this!

PR #2806, opened 2024-06-06
Store the nqt_in_academic_year_after_itt on the journey session

felixclack:
We want to store the value of the nqt_in_academic_year_after_itt on the
journey session.

There are places in the eligibility check that rely on this value being
stored on the eligibility record. Removing this expectation requires
changing how the eligibility is calculated.

That seems to be beyond the scope of this particular change and will
best be addressed directly rather than as a side-effect of this PR.

### DIFF
--- a/app/models/payroll/payment_csv_row.rb
+++ b/app/models/payroll/payment_csv_row.rb
@@ -41,17 +41,11 @@ module Payroll
     end
 
     def start_date
-      second_monday_of_month.strftime(DATE_FORMAT)
+      Date.today.at_beginning_of_month.strftime(DATE_FORMAT)
     end
 
     def end_date
-      second_monday_of_month.end_of_week.strftime(DATE_FORMAT)
-    end
-
-    def second_monday_of_month
-      day = Date.today.at_beginning_of_month
-      day += 1.days until day.monday?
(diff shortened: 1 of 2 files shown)


app/forms/journeys/additional_payments_for_teaching/eligible_itt_subject_form.rb
```diff
@@ -8,15 +8,15 @@ class EligibleIttSubjectForm < Form
       validates :eligible_itt_subject, inclusion: {in: :available_options, message: i18n_error_message(:inclusion)}
 
       def available_subjects
-        @available_subjects ||= subject_symbols(claim).map(&:to_s)
+        @available_subjects ||= subject_symbols(current_claim: claim, answers: journey_session.answers).map(&:to_s)
```
slorek, line 11: I think we need to refactor `subject_symbols` so we aren't passing a claim object to it. We should only need to pass in the specific answers needed to calculate the available subjects.

slorek (dismissed):

rjlynch (approved):
Looks good 👍 
Seems like there's one failing test. I'd probably delete that test as it's kind of pointless to check that the error is propagated

PR #2807, opened 2024-06-06
port eligible_itt_subject wip

asmega:

### DIFF
--- a/app/forms/journeys/additional_payments_for_teaching/eligible_itt_subject_form.rb
+++ b/app/forms/journeys/additional_payments_for_teaching/eligible_itt_subject_form.rb
@@ -28,16 +28,11 @@ module Journeys
       def save
         return false unless valid?
 
-        # FIXME RL: Once this method writes to the journey session answers we
-        # update the initializer in
-        # AdditionalPaymentsForTeaching::QualificationDetailsForm
-        claim.assign_attributes(
-          eligibility_attributes: {eligible_itt_subject: eligible_itt_subject}
+        journey_session.answers.assign_attributes(
+          eligible_itt_subject:,
         )
-        claim.reset_eligibility_dependent_answers(["eligible_itt_subject"])
-        claim.save!
 
-        true
+        journey_session.save!
       end
     end
   end
--- a/app/forms/journeys/additional_payments_for_teaching/qualification_details_form.rb
+++ b/app/forms/journeys/additional_payments_for_teaching/qualification_details_form.rb
@@ -19,7 +19,6 @@ module Journeys
         # session ansswers, so we need to set these values from the elgiibility
         # here.
         journey_session.answers.itt_academic_year = claim.eligibility.itt_academic_year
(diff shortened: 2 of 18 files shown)


app/forms/journeys/additional_payments_for_teaching/eligible_itt_subject_form.rb
```diff
@@ -28,16 +28,11 @@ def chemistry_or_physics_available?
       def save
         return false unless valid?
 
-        # FIXME RL: Once this method writes to the journey session answers we
-        # update the initializer in
-        # AdditionalPaymentsForTeaching::QualificationDetailsForm
-        claim.assign_attributes(
-          eligibility_attributes: {eligible_itt_subject: eligible_itt_subject}
+        journey_session.answers.assign_attributes(
```
rjlynch, line 31: This will need to set `teaching_subject_now` to `nil` if `eligible_itt_subject` has changed but not if `teaching_subject_now` answer came from `dqt` 🙃. See `Policies::EarlyCareerPayments::Eligibility#reset_dependent_answers` for how this reseting stuff currently works.
To make it more fun as we're part way through the migration we'll want to reset `teaching_subject_now` on both the claim and the journey_session. An example of all this is `app/forms/journeys/additional_payments_for_teaching/qualification_form.rb` in this pr https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2803/files

Once we've finished the migration it's probably best to delete all the reset dependent answers callbacks and just have the appropriate forms handle `nil`ing out dependent answers

app/forms/journeys/additional_payments_for_teaching/qualification_details_form.rb
```diff
@@ -19,7 +19,6 @@ def initialize(...)
         # session ansswers, so we need to set these values from the elgiibility
         # here.
         journey_session.answers.itt_academic_year = claim.eligibility.itt_academic_year
-        journey_session.answers.eligible_itt_subject = claim.eligibility.eligible_itt_subject
```
rjlynch, line 22: 👍

app/models/concerns/policies/early_career_payments/eligible.rb
```diff
@@ -9,8 +9,27 @@ def award_amount
         super || BigDecimal(calculate_award_amount || 0)
       end
 
+      def journey
+        @journey ||= Journeys.for_policy(policy)
+      end
+
+      def journey_session
+        @journey_session ||= journey::Session.find(claim.journeys_session_id)
```
rjlynch, line 17: I don't think we can use this association here, as the claim isn't associated with the journey session until the claim is submitted and once this migration has finished we wont create a claim until the end of the journey when the claim submission form is submitted.
Also it's still up for debate if we even want to associate the claim and the journey session, or if we just throw the journey session away once the claim has been created

asmega:
closing as superseded by https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2826

PR #2808, opened 2024-06-06
Bugfix on everything_ineligible? and add ineligible?

kenfodder:
This should deal with the page sequence ineligibility dropout properly so it it works with the shim.


asmega (approved):

rjlynch (approved):
👍

slorek (approved):

PR #2811, opened 2024-06-07
CAPT-1709: Output shortened policy names in the payroll file

slorek:
https://dfedigital.atlassian.net/browse/CAPT-170

Shortens the policy names output in the payroll CSV file to accommodate a new 30-character limit in the payroll system which ingests them.

The 30 characters needs to include multiple policies where a single payment covers claims from more than one policy.


rjlynch (approved):
Looks good 👍

PR #2816, opened 2024-06-07
Attempt to fix flaky school search specs

slorek:
This is an attempt to fix specs which are regularly failing whenever they use the `choose_school` helper.

This type of failure can happen when the page has not yet finished loading before Capybara tries to run the next matcher. Since this page is relatively large and has Javascript instantiation on the school search field this could be the reason for the intermittent failures. The `have_text` matcher has a built in wait mechanism which usually resolves this.


AbigailMcP (approved):
🙏

PR #2817, opened 2024-06-07
[CAPT-1708] Refactor ineligible reason checker

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-1708
- Refactor `IneligibilityReasonChecker` so it is able to accept either an existing claim or use answers on a `JourneySession`

# Changes

- At first I attempted to use `SessionAnswers` instead of the now `JourneySession`, however unfortunately found that data on the journey is required and simply using `SessionAnswers` was not sufficient
- I had to add a flag `use_claim` which is defaulted to true for one area in order to toggle between using either the claim or answers 
- I have moved the class from `/lib` to `app/models` so zeitwerk loads automagically
- Uncommented a recommended setting in `spec_helper` to allow to focus down on specific tests


app/models/journeys/session.rb
```diff
@@ -65,5 +65,11 @@ def tps_school_for_student_loan_in_previous_financial_year
           teacher_reference_number: answers.teacher_id_user_info["trn"]
         )
     end
+
+    def load_answers(other_answers)
+      other_answers.attributes.each do |field,value|
+        answers.public_send("#{field}=", value)
+      end
+    end
```
asmega, line 73: i added this so i can instantiate the class and load in some custom values as i need to be able to pass certain values to `EligibilityChecker` to determine the outcome. Not sure if there is a better way to do this?

app/views/additional_payments/claims/ineligible.html.erb
```diff
@@ -1,5 +1,4 @@
-<% require "./lib/ineligibility_reason_checker.rb" %>
-<% reason = IneligibilityReasonChecker.new(current_claim).reason %>
+<% reason = IneligibilityReasonChecker.new(current_claim:, journey_session:).reason %>
```
rjlynch, line 1: One thing that might we worth considering is passing the `ClaimSessionShim` into this class, and replacing all the calls to current claim with that which will work with both the session and the claim. That's what I went for https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/CAPT-1650/itt-year-form/lib/ineligibility_reason_checker.rb

app/models/journeys/session_answers.rb
```diff
@@ -44,6 +44,18 @@ class SessionAnswers
...
+    attribute :has_entire_term_contract, :boolean
+    attribute :employed_directly, :boolean
+    attribute :subject_to_formal_performance_action, :boolean
+    attribute :subject_to_disciplinary_action, :boolean
+    attribute :itt_academic_year
+    attribute :teaching_subject_now, :boolean
+    attribute :eligible_itt_subject, :string
+    attribute :eligible_degree_subject, :boolean
+    attribute :nqt_in_academic_year_after_itt, :boolean
+    attribute :induction_completed, :boolean
```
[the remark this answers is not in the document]
rjlynch: Are these specific to the AddtionalPaymentsJourney? If so we probably want to keep them on the `AdditionalPayments::SessionAnswers` as other journeys wont care about these attributes

spec/models/ineligibility_reason_checker_spec.rb
```diff
@@ -0,0 +1,557 @@
...
+
+  let(:current_claim) { CurrentClaim.new(claims: [ecp_claim, lup_claim]) }
+
+  let(:journey) { "additional-payments" }
+  let(:answers) { Journeys::SessionAnswers.new }
+  let(:journey_session) do
+    object = Journeys::AdditionalPaymentsForTeaching::Session.new(journey:)
+    object.load_answers(answers)
+    object
+  end
```
rjlynch, line 33: I think the way to go with this is setting up some factories for the answers for each scenario
https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/CAPT-1650/itt-year-form/spec/lib/ineligibility_reason_checker_spec.rb

app/models/journeys/session.rb
asmega: this has now been removed in favour of using factories in specs

app/models/journeys/session.rb
```diff
@@ -65,5 +65,11 @@ def tps_school_for_student_loan_in_previous_financial_year
           teacher_reference_number: answers.teacher_id_user_info["trn"]
         )
     end
+
+    def load_answers(other_answers)
```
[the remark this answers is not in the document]
asmega: i've removed this now and using factories instead which is probably a better idea

spec/models/ineligibility_reason_checker_spec.rb
asmega: I originally tried but ran into some issues (don't recall the failures), but i can have another play

app/models/journeys/session_answers.rb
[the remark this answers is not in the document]
asmega: I see, I've now removed this as not needed because as you've pointed out this is only for the additional payments journey

app/views/additional_payments/claims/ineligible.html.erb
asmega: I originally tried but ran into some issues (don't recall the failures), but i can have another play

spec/models/ineligibility_reason_checker_spec.rb
asmega: I'm now using the factories as you've suggested and `#load_answers` is no longer needed and has been removed

asmega:
I don't think this PR is needed anymore and should be covered by https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2818

PR #2818, opened 2024-06-08
itt-year-form / change eligibility to work with answers

rjlynch:
# Updates the itt-year-form to write it's answers to the session.

There's a couple of commits here so might be better to review them separately but the first commit is pretty large. Over view of changes in the first commit:

`itt_academic_year` is used through out the code base to calculate eligibility, to facilitate this migration we've had to make some changes.

* Expose `eligible_now_and_sorted`, `eligible_now`, `eligible_now_or_later` on the `ClaimSubmissionForm`
  These methods were used on the `EligibilityConfirmedForm` and as we now can't call these methods on the current claim as it doesn't have `itt_academic_year` set we call them on the `ClaimSubmissionForm` which already handles checking the answers and claim for eligibility

* Update the `EarlyCareerPayments` and `LevelingUpPremiumPayments` `PolicyEligibilityChecker` class to take `answers` rather than the `shim` in their initialiser and update the TSLR journey eligibility checker and policy checker to match additional payments naming conventions. These classes only call method on the `answers` rather than the shim and are easier to call indecently of the journey eligibility checkers if they work with just answers. (Also the `JourneySession` is just an artefact of how we're persisting the data, all we really care about is the answers, so we want to work directly with them as much as possible in case we decide to change to a table backed answers model).

* Update `lib/ineligibility_reason_checker` to work with answers rather than a current claim. As the details required to calculate eligibility will be found on the answers model, this class needs to work with answers (while we're migrating we actually pass this class the shim).

* Update `lib/journey_subject_eligibility_checker.rb` to work with answers. We've also moved the subject symbols method out out the app helpers and into a class method on `journey_subject_eligibility_checker.rb`. The main change on this class is how we find the potentially still eligible policies. Previously we did this be querying the various claims on the current claim, however as we can no longer do this we loop through the policies in the journey. As this class is only used for the additional payments journey and is closely tied to the domain model we should move it out of the lib directory.

* Update the slug sequence to check various eligible conditions against the `PolicyEligibilityChecker` classes. There's still some checks agains the eligibility but these can now be switched over to check the appropriate `PolicyEligibilityChecker` as we progress with the rest of the migration

Apologies for the large first commit, changing the eligibility was very **_very_** involved and I didn't have a clear plan for how to break this up into separate commits. Ping me on slack if you want to jump on a call to discuss this.

Additionally this pr contains some commits that remove unused methods and bring the TSLR journey eligibility checker and student loans checker inline with the naming conventions in AdditionalPayments


app/forms/journeys/additional_payments_for_teaching/claim_submission_form.rb
```diff
@@ -1,6 +1,20 @@
...
+      end
+
+      def eligible_now
+        eligibilities.select { |e| e.status == :eligible_now }
+      end
+
+      def eligible_now_or_later
+        eligibilities.select do |e|
+          e.status == :eligible_now || e.status == :eligible_later
+        end
```
rjlynch, line 15: I think eventually these methods should be on the journey eligibility checker but moving these is probably too involved at the minute

app/forms/journeys/additional_payments_for_teaching/itt_academic_year_form.rb
```diff
@@ -8,12 +8,18 @@ class IttAcademicYearForm < Form
...
-        claim.save!
+        if reset_dependent_answers?
+          journey_session.answers.assign_attributes(eligible_itt_subject: nil)
+          claim.update!(
+            eligibility_attributes: {eligible_itt_subject: nil}
+          )
+        end
+
+        journey_session.answers.assign_attributes(
+          itt_academic_year: itt_academic_year
```
rjlynch, line 19: This was the change to the form that caused all the other changes in this pr 😂

spec/features/reminders_spec.rb
```diff
@@ -47,21 +48,21 @@
           click_on "Continue"
 
           # - Which subject did you do your postgraduate initial teacher training (ITT) in?
-          choose I18n.t("additional_payments.forms.eligible_itt_subject.answers.#{args[:subject]}")
+          choose "Yes"
```
rjlynch, line 51: On master we're eligible for both policies, but the test set up indicates we only care about ECP.
This test loops twice, once with itt_academic_year of "2018 to 2019" and claim year of 2022, and once with itt_academic_year of "2019 to 2020" and claim year of 2023.  Given a claim year of 2022 and itt year of 2018 the displayed subjects should only be `mathematics` for ECP policies, same for 2023 claim year and 2019 itt year. What's happening on master is the subjects for the LUP policy are also being returned so there are multiple options available. I'm not sure if this was intended as we don't set the LUP policy up to be eligible. Choosing yes doesn't change what we care about in this test (that a reminder is created) so wen't with that option rather than setting up the answers to be LUP eligible too.

PR #2826, opened 2024-06-11
[CAPT-1651] Port eligible ITT subject form

asmega:


app/controllers/claims_controller.rb
```diff
@@ -74,7 +74,7 @@ def check_page_is_in_sequence
       return redirect_to new_claim_path
     end
 
-    raise ActionController::RoutingError.new("Not Found") unless page_sequence.in_sequence?(params[:slug])
+    raise ActionController::RoutingError.new("Not Found for #{params[:slug]}") unless page_sequence.in_sequence?(params[:slug])
```
rjlynch, line 77: 👍

app/controllers/claims_form_callbacks.rb
```diff
@@ -84,7 +84,7 @@ def on_school_search_results?
   end
 
   def no_eligible_itt_subject?
-    !current_claim.eligible_itt_subject
+    !journey_session.answers.eligible_itt_subject && !journey_session.answers.early_career_payments_dqt_teacher_record&.eligible_itt_subject_for_claim
```
rjlynch, line 87: How come we need to check the dqt_teacher_record here?
I thought we'd only need to access the dqt teacher record when setting the attributes on the answers and otherwise we'd just check the answers for the eligible_itt_subject.
Does the test suite pass if we remove the right hand side of the `&&`?

app/forms/journeys/additional_payments_for_teaching/eligible_itt_subject_form.rb
```diff
@@ -33,23 +33,23 @@ def subject_symbols
...
-        # answers) (and remove this comment!)
-        # Also upudate the IttAcademicYearForm to no longer reset the
-        # eligible_itt_subject on the claim as it's not needed once this form
-        # writes to the answers
-        claim.assign_attributes(
-          eligibility_attributes: {eligible_itt_subject: eligible_itt_subject}
+        if eligible_itt_subject_changed? && !journey_session.answers.qualifications_details_check
+          journey_session.answers.assign_attributes(
+            teaching_subject_now: nil
+          )
```
rjlynch, line 39: I think we also need to reset the `eligible_degree_subject` here (see polices/levelinguppremiumpayments/eligibility)
Also probably should update the tests to check this gets reset

app/forms/journeys/additional_payments_for_teaching/eligible_itt_subject_form.rb
```diff
@@ -33,23 +33,23 @@ def subject_symbols
...
-        # and update
-        # QualificationForm#save to not reset eligible_itt_subject subject on
-        # the claim, as it's no longer needed (still keep resetting it on the
-        # answers) (and remove this comment!)
-        # Also upudate the IttAcademicYearForm to no longer reset the
-        # eligible_itt_subject on the claim as it's not needed once this form
-        # writes to the answers
-        claim.assign_attributes(
-          eligibility_attributes: {eligible_itt_subject: eligible_itt_subject}
+        if eligible_itt_subject_changed? && !journey_session.answers.qualifications_details_check
```
rjlynch, line 36: 👍

spec/forms/journeys/additional_payments_for_teaching/eligible_itt_subject_form_spec.rb
```diff
@@ -218,21 +218,87 @@
...
+          )
+        )
+      end
+
+      it "does not reset dependent attributes" do
+        expect { form.save }.not_to change { journey_session.reload.answers.teaching_subject_now }
+      end
+    end
+
+    context "when a change and teaching_subject_now from DQT" do
```
rjlynch, line 277: 👍

rjlynch (changes requested):
Nicely done.
Have left a couple of comments, main one is we want to reenable that commented out test.

app/forms/journeys/additional_payments_for_teaching/qualification_details_form.rb
```diff
@@ -93,6 +93,24 @@ def dqt_degree_subjects
...
+        if (dqt_subjects || []).any?
+          overlap = (dqt_subjects & possible_eligible_itt_subjects)
+
+          if overlap.any?
+            overlap.first
+          else
+            :none_of_the_above
+          end
+        end
+      end
```
rjlynch, line 108: I think this will return subjects that are selectable for the policy but not necessarily ones returned from DQT.

I think what we want to do is see if either policy's dqt record returns a subject that is not `none_of_the_above`

```ruby
dqt_subjects = [
  answers.early_career_payments_dqt_teacher_record.ieligible_itt_subject_for_claim,
  answers.levelling_up_premium_payments_dqt_reacher_record.eligible_itt_subject_for_claim
]

return nil if dqt_subjects.empty?

not_none_of_the_above = dqt_subjects.reject { |subject| subject == :none_of_the_above }

if not_none_of_the_above.any?
  not_none_of_the_above.first
else
  :none_of_the_above
 end
```

I _think_ (caveat I don't understand the policy side of this) that both dqt records should return the same value, however for historical reasons they don't.

app/controllers/claims_form_callbacks.rb
asmega: it does seem to pass so totally removing it. it'll be from my many attempts at getting things to pass

app/models/journeys/additional_payments_for_teaching/answers_presenter.rb
```diff
@@ -51,31 +51,31 @@ def has_entire_term_contract
       def current_school
         [
           t("additional_payments.forms.current_school.questions.current_school_search"),
-          eligibility.current_school_name,
-          (eligibility.school_somewhere_else == false) ? "correct-school" : "current-school"
+          journey_session.answers.current_school.name,
```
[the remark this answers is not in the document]
asmega: gotcha, i've revert this bit back

app/forms/journeys/additional_payments_for_teaching/qualification_details_form.rb
asmega: i've updated to a similar implementation you suggested

app/forms/journeys/additional_payments_for_teaching/eligible_itt_subject_form.rb
asmega: good spot, done

spec/features/admin_claim_tasks_update_with_dqt_api_spec.rb
```diff
@@ -162,7 +162,8 @@ def task_outcome
             first_name: "Fred",
             national_insurance_number: "QQ100000C",
             surname: "ELIGIBLE",
-            teacher_reference_number: "1234567"
+            teacher_reference_number: "1234567",
+            eligible_itt_subject: "mathematics"
```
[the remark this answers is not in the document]
asmega: `eligible_itt_subject` must have been set some other way previously and now for test setup is needed here for it to be present. i've tried your suggestion but seems to open a can of worms. If you remove it, it expects it to be there with a sea of red.

```
  23       claim.update!(
  24         attributes_for(
  25           :claim,
  26           :submittable,
  27           **claim_attributes
  28         )
```
Tests say claim should submittable but is it not submitted at this point? I tried that avenue with no luck.

I feel like I've missed something that should have automatically setup whatever was needed therefore added this line as a workaround. I currently have no idea what did/does that and therefore what I should change.

spec/features/ineligible_levelling_up_premium_payments_spec.rb
```diff
@@ -41,7 +41,7 @@
     expect(page).to have_text("The school you have selected is not eligible")
   end
 
-  scenario "When subject 'none of the above' and user does not have an eligible degree" do
+  xscenario "When subject 'none of the above' and user does not have an eligible degree" do
```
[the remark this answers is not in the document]
asmega: now that I've rebased and `eligible_degree_subject` was ported over, I've uncommented this and works as expected

spec/features/admin_claim_tasks_update_with_dqt_api_spec.rb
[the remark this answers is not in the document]
rjlynch: I think, thought these tests are really hard to follow, the eligible itt subject should be being set from the hash of DQT data (line 84). I think that blob of data ends up being used in the qualification details form when we set the eligible itt subject from the dqt record. Probably fine to just update these tests to set the subject explicitly, though it does remove a layer of indirect testing

spec/factories/journeys/additional_payments_for_teaching/session_answers.rb
```diff
@@ -67,6 +71,7 @@
       with_qualification
       with_itt_academic_year
       with_teaching_subject_now
+      with_eligible_itt_subject
```
rjlynch, line 74: 👍

rjlynch (approved):
Nice work 👍 
Left a comment about that test but I don't think it's a blocker

PR #2828, opened 2024-06-11
CAPT 1697/set student loans data on answers

rjlynch:
Probably slightly easier to review both commits separately

Set student loans on session

There's a call back that sets student loans data on the claim, we need
to change this to set the student loans data on the session. The
ClaimStudentLoanDetailsUpdater is also used for setting data on
submitted claims so we've introduced an answers specific version and a
presenter class to share the data between the claim and answers
versions.

### DIFF
--- a/app/models/claim_student_loan_details_updater.rb
+++ b/app/models/claim_student_loan_details_updater.rb
@@ -11,11 +11,6 @@ class ClaimStudentLoanDetailsUpdater
     claim.transaction do
       eligibility.update!(eligibility_student_loan_attributes) if claim.has_tslr_policy?
 
-      # When the claim hasn't been submitted yet, we need a way of knowing if the student loan
-      # details on the claim were found using the SLC data we held before submission;
-      # after submission, the `submitted_using_slc_data` value must not change
-      claim.assign_attributes(submitted_using_slc_data: student_loans_data.found_data?) unless claim.submitted?
-
       claim.assign_attributes(claim_student_loan_attributes)
       claim.save!(context: :"student-loan")
     end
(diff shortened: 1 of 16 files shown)


app/models/claim_student_loan_details_updater.rb
```diff
@@ -14,7 +14,7 @@ def update_claim_with_latest_data
       # When the claim hasn't been submitted yet, we need a way of knowing if the student loan
       # details on the claim were found using the SLC data we held before submission;
       # after submission, the `submitted_using_slc_data` value must not change
-      claim.assign_attributes(submitted_using_slc_data: found_data?) unless claim.submitted?
+      claim.assign_attributes(submitted_using_slc_data: student_loans_data.found_data?) unless claim.submitted?
```
rjlynch, line 17: Question: Do we need to set `submitted_using_slc_data` on the journey session and write that to the claim in the claim submission form? If so probably shouldn't merge this pr yet!

rjlynch:
Think I need to change something to write submitted using slc data to the answers- so holding off on merging this until that's done

app/models/journeys/answers_student_loans_details_updater.rb
```diff
@@ -0,0 +1,41 @@
...
+      # held before submission; after submission, the
+      # `submitted_using_slc_data` value must not change
+      journey_session.answers.assign_attributes(
+        has_student_loan: student_loans_data.found_data?,
+        student_loan_plan: student_loans_data.student_loan_plan,
+        submitted_using_slc_data: student_loans_data.found_data?
+      )
+
+      journey_session.save!
+    rescue => e
```
rjlynch, line 23: Not stoked on a bare rescue but that's how it's done in the claim student loans details updater and I'm unsure on what errors can be raised in production.

PR #2830, opened 2024-06-11
Migrate employed directly to the journey session

felixclack:
We want to store the employed_directly answer on the journey session and
move it away from the eligibility object.

This does impact how the ineligibility reason is calculated. For now I
opted to include the answers to the ineligibility checker class. This
seemed like the simplest change for the short-term.

Other work is progressing to move the logic from
`IneligibilityReasonChecker` to the journey session.


rjlynch (commented):
Once this is rebased on master should be good to go. If there's failing tests just update the session answers factories to set employed directly in the `submittable` trait and that should handle most failures

rjlynch (approved):
LGTM 👍

PR #2836, opened 2024-06-13
CAPT-1655 - selected policy using session answers only, remove select…

kenfodder:
…ed policy cookie


rjlynch (approved):
Really nice 👍 this is so much cleaner than storing the policy in the session!

PR #2845, opened 2024-06-13
CAPT-1720/Remove shim

rjlynch:
Remove shim

Now now forms are writing to the journey session we can remove the shim
classes


app/forms/journeys/additional_payments_for_teaching/reminders/email_verification_form.rb
```diff
@@ -47,6 +47,22 @@ def sent_one_time_password_at?
         rescue Date::Error
           false
         end
+
+        def load_current_value(attribute)
```
rjlynch, line 51: Have a ticket to sort this https://dfedigital.atlassian.net.mcas.ms/browse/CAPT-1719

PR #2848, opened 2024-06-14
Refactor: remove email helper by inlining

asmega:
# Context

- Refactor by removing excessive helper and inlining
- helper was only used in a single location


rjlynch (approved):
👍 Nice

PR #2849, opened 2024-06-14
Fix deprecation warning

asmega:
# Context

- Fix deprecation warning in tests
- Also removed someones personal note that don't belong


rjlynch (approved):
👍

PR #2864, opened 2024-06-17
CAPT-1702 Add slack notification on failure

AbigailMcP:
**Context**

For permantent environments we want to notify on Slack when a deploy fails.

**Changes in this PR**

Add an extra step to the `deploy_test` job to notify on slack if is fails.


slorek (approved):

AbigailMcP:
Tested with this commit :https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2864/commits/cc4279f6f0dbd4250847880fc254419f368d2166

Posted to Slack 👍

PR #2865, opened 2024-06-17
[CAPT-1729] Use GOVUK formbuilder for email address form

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-1729
- Ultimate goal is to remove `module GovukFormHelper` which houses custom code for forms

# Changes

- Use `GOVUKDesignSystemFormBuilder::FormBuilder` for email address form
- Updated tests to target fields via labels rather than their `id`


app/views/claims/email_address.html.erb
```diff
@@ -9,35 +9,29 @@
...
-        <div id="email-address-hint" class="govuk-hint">
+      <%= f.govuk_text_field :email_address,
+        autocomplete: "email",
+        spellcheck: "false",
+        caption: {
+          text: I18n.t("questions.personal_details"),
+          size: "xl"
+        },
+        label: {
+          text: "Email address",
```
rjlynch, line 23: Should this be
```suggestion
          text: t("questions.email_address")
```

(though I prefer plain text, i18n is more trouble than it's worth!)

rjlynch (approved):
Much nicer. Left one comment about i18n

app/views/claims/email_address.html.erb
asmega: I've just left these as is from before, but yes we can probably just start hard coding these

asmega: Wait, I misunderstood, yeah i can change this. I'm not sure if form builder uses I18n out the box it would be could to not have to set these if formbuilder reads the correct location in the i18n values

PR #2869, opened 2024-06-17
[CAPT-1731] Fix missing back link

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-1731
- When hitting the `ClaimsController` with a form submission, if there is an error the back link is missing

# Changes

- back link is generated on form `show` and `update`


rjlynch (approved):
👍

PR #2871, opened 2024-06-17
POC - integrate IRP into claim

rjlynch:
Need to sort out the slug sequence, check some personal details differences between IRP and claim with Laura, and fix up the other feature tests, but the initial question page renders. Not sure what we want to see on the admin side when we successfully submit a claim

**Note** will abandon this pr and issue a separate pr to set up the initial scaffolding. This was just a proof of concept to see what adding a new journey would involve. Currently this just works up to rendering the start page and the initial form

### DIFF
--- a/app/views/get_a_teacher_relocation_payment/claims/application_route.html.erb
+++ b/app/views/get_a_teacher_relocation_payment/claims/application_route.html.erb
@@ -1,0 +1,68 @@
+<% content_for(
+  :page_title,
+  page_title(
+      t(
+        "get_a_teacher_relocation_payment.forms.application_route.questions.which_route"
+      ),
+      journey: current_journey_routing_name,
+      show_error: @form.errors.any?
+  )
+) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render("shared/error_summary", instance: @form) if @form.errors.any? %>
+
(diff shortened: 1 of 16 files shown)


app/views/get_a_teacher_relocation_payment/claims/application_route.html.erb
```diff
@@ -0,0 +1,68 @@
...
+      journey: current_journey_routing_name,
+      show_error: @form.errors.any?
+  )
+) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= render("shared/error_summary", instance: @form) if @form.errors.any? %>
+
+    <%= form_for @form, url: claim_path(current_journey_routing_name) do |f| %>
```
asmega, line 16:
```suggestion
    <%= form_for @form, url: claim_path(current_journey_routing_name), builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
```
The formbuilder has been installed in this app for a long time but never used. Would be good to actually start using it and therefore it should simplify a lot of the view code. I'm working on removing all the custom code we have built up by not using GOVUK formbuilder at the moment

rjlynch: 👍 yeah makes sense, I'm not planning on merging this pr!

PR #2873, opened 2024-06-18
Remove unused method reset_dependent_answers

felixclack:
Now that we're storing answers on an object separate from a claim, we no
longer need to reset attributes on the claim itself.

The method that handles this is redundant and can be safely removed.


slorek (approved):

rjlynch (approved):
Looks good 👍 current claim has been deleted on master so you can just remove that file and spec

rjlynch (commented):
Can we also remove the ATTRIBUTE_DEPENDENCIES constants as I don't think they're referenced anymore

PR #2880, opened 2024-06-19
Initial IRP journey wire up

rjlynch:
This commit adds the initial set up for IRP along with a some feature
tests.
So far the IRP journey lets a user click "Start now" and then
shows the check your answers page where they can submit a claim.
The feature test has some temporary hacks, where we update the database
model, this is just required to get the answers in a state that passes
the validations on the claim. We'll be removing the
`teacher_reference_number` question and (probably) the `payroll_gender`
question from the IRP journey but they're currently required to submit a claim.

[follow up PR that adds the first form to the journey](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2886)


slorek (approved):
Looks great. Perhaps we need a migration or rake task to create the journey configuration record as well.

rjlynch:
> Looks great. Perhaps we need a migration or rake task to create the journey configuration record as well.

👍 Have added a ticket to get that done

PR #2885, opened 2024-06-19
POC form level callbacks

asmega:
WIP


asmega:
after some discussion the root cause of needing arbitrary redirects in the controller seems to be due to the sequencing classes not able to reliably determine the next and previous pages. fixing that should mean we can remove such controller level redirects altogether and gain reliable sequencing without the need for random controller redirects therefore removing the need for this feature

PR #2876, opened 2024-06-18
Configure the Semantic Logger to output JSON

vacabor:
As per https://technical-guidance.education.gov.uk/infrastructure/monitoring/logit/#ruby-on-rails.

Extracted from #2815.

PR #2815, opened 2024-06-07
CAPT-1702 Migrate test environment to AKS

AbigailMcP:
## Changes in this PR

Create test environment, so we can have proper testing before deploying to production

## Guidance to review

Run `make test-aks terraform-plan-aks` and check output is correct

Testing URLS (created by running make commands locally):

https://claim-additional-payments-for-teaching-test-web.test.teacherservices.cloud/additional-payments/claim
https://claim-additional-payments-for-teaching-test-web.test.teacherservices.cloud/student-loads/claim
https://claim-additional-payments-for-teaching-test-web.test.teacherservices.cloud/admin

PR #2887, opened 2024-06-20
CAPT-1703 Migrate production environment to AKS

vacabor:
Similar to #2815 but targetting the production environment.

Look to merge #2876 after merging this.


terraform/application/variables.tf
```diff
@@ -61,6 +61,18 @@ variable "canonical_hostname" {
   default     = null
 }
 
+variable "web_replicas" {
+  type        = integer
```
AbigailMcP, line 65: this produces an error when i run the terraform plan:

```
│ Error: Invalid type specification
│
│   on variables.tf line 65, in variable "web_replicas":
│   65:   type        = integer
│
│ The keyword "integer" is not a valid type specification.
```

is it `number`?

terraform/application/config/production_app_env.yml
```diff
@@ -0,0 +1 @@
+---
```
AbigailMcP, line 1: we don't really seem to be using these files - i wonder whether we pull the 'non-secret' secrets out of the keyvaults and put them here. Things like `HMRC_API_BANK_VALIDATION_ENABLED`

or just move the couple that we do have in test/review e.g. `BYPASS_DFE_SIGN_IN` into the keyvault? cc @slorek

RMcVelia:
The prod db may need to be enabled for HA, and non-burstable sku (although that depends on requirements)

.github/workflows/build_and_deploy.yml
```diff
@@ -136,3 +136,59 @@ jobs:
...
+        with:
+          environment: production-aks
+          docker-image: ${{ needs.build.outputs.docker-image-tag }}
+          azure-credentials: ${{ secrets.AZURE_CREDENTIALS }}
+
+      - name: Run migrations
+        shell: bash
+        run: |
+          make ci production-aks get-cluster-credentials
+          kubectl exec -n srtl-production deployment/claim-additional-payments-for-teaching-production-worker -- sh -c "DISABLE_DATABASE_ENVIRONMENT_CHECK=1 bin/prepare-database"
```
AbigailMcP, line 171: I wonder whether we should move running migrations into our `deploy-environment` action since we wouldn't consider a deploy complete without them.

We could then decide whether to move the smoke tests into the `deploy-environment` action too (as per a previous suggestion from Rick)

.github/workflows/build_and_deploy.yml
```diff
@@ -137,3 +120,55 @@ jobs:
           SLACK_TITLE: Failure deploying release to test
           SLACK_MESSAGE: Failure deploying release to test - Docker tag ${{ needs.build.outputs.docker-image-tag }}
           SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_URL }}
+
+  deploy_production:
+    name: Deploy to production environment
+    concurrency: deploy_production
+    runs-on: ubuntu-latest
+    # if: github.ref == 'refs/heads/master' && github.event_name == 'push'
+    needs: [deploy_test]
```
RMcVelia, line 129: I think this needs to be needs: [build,deploy_test] otherwise it can't get the docker_image i.e. needs.build.outputs.docker-image-tag, and will keep installing master.
I think I was the one who said to replace build with deploy_test before! It was almost right...

RMcVelia (approved):

terraform/application/config/production_app_env.yml
saliceti: Yes please move the non secret values to yml. It makes the deployment faster, you can track the values more easily, and change them via PR in a controlled way.
But you may do that in another PR.

.github/workflows/build_and_deploy.yml
```diff
@@ -137,3 +120,55 @@ jobs:
           SLACK_TITLE: Failure deploying release to test
           SLACK_MESSAGE: Failure deploying release to test - Docker tag ${{ needs.build.outputs.docker-image-tag }}
           SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_URL }}
+
+  deploy_production:
```
saliceti, line 124: This is the same code as above. Can we use a [matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) instead? With `max-parallel: 1` to make it sequential.

AbigailMcP:
cc @vacabor 

**Tested and working:**

✅ Basic auth
✅ Claim submission
✅ Ordnance survey API address lookup
✅ Notify emails sent
✅ HMRC API integration

**Issues:**

🔴 Exception reporting to Rollbar
Entered a non-team email address and expected the following error to be reported:
'BadRequestError: Can’t send to this recipient using a team-only API key'
But nothing appeared in Rollbar.

🔴 DfE Identity sign in
Error: 'the specified 'redirect_uri' is not valid for this client application'

🔴  Admin login with DfE Sign In
Error: 'invalid_redirect_uri, redirect_uri did not match any of the client's registered redirect_uris'

**Untested (need admin login which isn't working currently):**
🟠 Levelling Up Premium Payments awards upload
🟠 TPS data upload
🟠 School data import

**Untested (waiting for PIM):**
🟠 Checked application and deployment logs for errors
🟠 DQT integration

terraform/application/config/production_app_env.yml
AbigailMcP: addressing in new PR 👍

AbigailMcP (approved):

saliceti (approved):

slorek (approved):

PR #2893, opened 2024-06-20
[LUPEYALPHA-559] Introduce further education payments

asmega:
# Context

- https://dfedigital.atlassian.net.mcas.ms/browse/LUPEYALPHA-559
- Introduce further education payments journey into the application
- We want to add scaffolding for this journey to unblock developers so they can start working in parallel  

# Changes

- Added lots of boilerplate to enable new FE payments journey
- Simple test to go through happy path journey pages
- Added lots of place holder pages with blank content to be added later

# Known issues

- There are multiple issues with this PR but have left out of scope of this PR
- Only added single happy path, many branching pages will be missing
- There is no form submission at the end and will still need implementing
- Not 100% on slugs these can be corrected at a later stage
- Missing lots of translation strings

### DIFF
--- a/app/models/policies/further_education_payments.rb
+++ b/app/models/policies/further_education_payments.rb
@@ -1,0 +1,4 @@
+module Policies
+  module FurtherEducationPayments
+  end
+end
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -75,11 +75,12 @@ jobs:
           message: |
             ### Deployments
 
-            | Journey             | URL                                  |
-            | ------------------- | ------------------------------------ |
-            | Additional Payments | <${{ env.APP_URL }}/additional-payments/claim> |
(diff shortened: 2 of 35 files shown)


app/models/policies/further_education_payments.rb
```diff
@@ -0,0 +1,13 @@
...
+  module FurtherEducationPayments
+    class Eligibility < ApplicationRecord
+      self.table_name = "further_education_payments_eligibilities"
+
+      has_one :claim, as: :eligibility, inverse_of: :eligibility
+
+      def ineligible?
+        false
+      end
+    end
```
rjlynch, line 11:
```suggestion
```
Should this be here?

rjlynch (changes requested):
Overall LGTM but looks like the eligibility model has snuck into the module definition

app/models/policies/further_education_payments.rb
asmega: oh yeah good spot

rjlynch (approved):
Looks good 👍

PR #2905, opened 2024-06-21
CAPT 1689/irp/remove trainee

rjlynch:
Remove trainee pages

IRP is no longer going to be eligible for trainee teachers. This commit
removes the trainee teacher screens we'd migrated so far and updates the
eligibility to make trainee teachers ineligible.
We've also removed the resetting of depenent answers from the
application for as now if the user changes their answer to any of the
options they'll be unable to proceed with the reset of the journey.


app/models/journeys/get_a_teacher_relocation_payment/slug_sequence.rb
```diff
@@ -36,14 +35,7 @@ def initialize(journey_session)
...
       def slugs
-        SLUGS.dup.tap do |sequence|
-          if answers.trainee?
-            sequence.delete("state-funded-secondary-school")
-            sequence.delete("contract-details")
-          else
-            sequence.delete("trainee-details")
-          end
-        end
+        SLUGS.dup
```
asmega, line 38: do you need the `dup` here? It made sense before as it got mutated

asmega (approved):
Looks good, just one minor non blocking thing

app/models/journeys/get_a_teacher_relocation_payment/slug_sequence.rb
rjlynch: Not strictly here, but in future commits we do end up mutating this

app/models/policies/international_relocation_payments/policy_eligibility_checker.rb
```diff
@@ -23,6 +23,8 @@ def ineligible?
 
       def ineligible_reason
         case answers.attributes.symbolize_keys
+        in application_route: "salaried_trainee"
+          "application route salaried trainee not accecpted"
```
slorek, line 27:
```suggestion
          "application route salaried trainee not accepted"
```

slorek (approved):

PR #2909, opened 2024-06-24
remove lefthook gem and config

asmega:
# Context

- No one appears to be using lefthook so removing the dependency so we spend less time maintaining it

# Changes

- Remove lefthook gem and associated configs


rjlynch (approved):
👍

PR #2911, opened 2024-06-24
Setup the admin to display IRP claims

felixclack:
In order to be able to display the IRP claims in the admin, there are
some basic methods that need to be available.

This change ensures the minimum implementation for displaying claims
without being concerned about the correctness of the required methods.

| Name | Screenshot |
|--------|--------|
| Claims Index | Screenshot 2024-06-27 at 10 40 01 am |
| Claim Show |  Screenshot 2024-06-27 at 10 55 52 am |

### DIFF
--- a/app/models/policies/international_relocation_payments/eligibility.rb
+++ b/app/models/policies/international_relocation_payments/eligibility.rb
@@ -3,7 +3,20 @@ module Policies
     class Eligibility < ApplicationRecord
       self.table_name = "international_relocation_payments_eligibilities"
 
+      AMENDABLE_ATTRIBUTES = %i[].freeze
+
       has_one :claim, as: :eligibility, inverse_of: :eligibility
+
+      attr_accessor :teacher_reference_number
+
+      def award_amount
+        0
+      end
+
+      # No current_school attribute on the model. This method is for compatibility with the admin UI.
+      def current_school
+        nil
+      end
 
       def ineligible?
         false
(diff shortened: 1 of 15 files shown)


app/models/policies/international_relocation_payments/eligibility.rb
```diff
@@ -3,7 +3,14 @@ module InternationalRelocationPayments
     class Eligibility < ApplicationRecord
       self.table_name = "international_relocation_payments_eligibilities"
 
+      AMENDABLE_ATTRIBUTES = %i[].freeze
+
       has_one :claim, as: :eligibility, inverse_of: :eligibility
+      belongs_to :current_school, class_name: "School", optional: true
```
rjlynch, line 9: Does the current IRP app set a school? If not, could this just be a method like

```
# required for compatibility with ...
def current_school
  nil
end
```

rjlynch: Actually I'm not sure what we want to do here!
There is a school address page `step/employment-details` 🤔

rjlynch: I think for the minute we can just have `current_school` as a method that returns nil and we'll just capture the school name / address as is currently done in IRP. Easy enough to change if we do need an association

rjlynch:
The claimant journey for IRP is now basically wired up https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2920 - so if you want to test how claims are coming through to the admin area you can either create a branch off that and merge this PR's changes in or rebase this branch off of `CAPT-1736/form-fixes` (might not want to rebase this, I don't think there would be many conflicts but this PR would get blocked waiting on my chain of PRs!)

PR #2913, opened 2024-06-25
[LUPEYALPHA-545] FE provision search

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-545
- Add provision (college) search to FE journey

# Changes

- College search for non-JS version has been written from the ground up
- JS version is mostly a slimmed down version of what we used in other journey
- As part of forms refactor I am slowly chipping away at, I will revisit this JS to make it more reusable but as it stands it works for now
- We simply store the selected institution, no work has been done to determine the eligibility of the selected institution

# Notes

- There is a spreadsheet with eligible FE bodies. I did a spot check and these appear in our usual GIAS import
- However not sure how we determine which bodies will actually be eligible. Are we able to infer these from the GIAS data? or will we have to handle a csv upload to maintain which bodies are eligible?
- We don't currently store the `UKPRN` when we do the GIAS import but we may have to add this though has not been included in this PR


app/forms/journeys/further_education_payments/select_provision_form.rb
```diff
@@ -0,0 +1,42 @@
+module Journeys
+  module FurtherEducationPayments
+    class SelectProvisionForm < Form
+      attribute :school_id, :string # school GUID
+
+      validates :school_id, presence: {message: i18n_error_message(:blank)}
+
+      def radio_options
+        results.map do |school|
```
rjlynch, line 9: How come we need to map these to an open struct? Can we not just return the schools given they have the same api?

rjlynch (approved):
Looks good, left one question but it's not a blocker

app/forms/journeys/further_education_payments/select_provision_form.rb
asmega: ah yes I see your point. let me remove the `map` and `openstruct` as they're not necessary

PR #2914, opened 2024-06-25
CAPT 1689/irp/personal details pages

rjlynch:
Adds personal details screens from claim to the IRP journey.
This PR is split into separate commits which you may want to review separately.
We currently capture the TRN, to allow us to complete the journey via the
browser, but will be removing that step once TRN is no longer a required field
on the claim.


app/models/policies/international_relocation_payments.rb
```diff
@@ -1,4 +1,12 @@
 module Policies
   module InternationalRelocationPayments
+    include BasePolicy
+    extend self
+
+    # NOTE RL: currently IRP only has a single reply to address, so notify
+    # doesn't show the address id
+    def notify_reply_to_id
+      nil
+    end
```
asmega, line 10: if we're migrating to use to the claim notify account i would have expected this to be populated?

slorek: We can migrate the templates separately.

slorek (approved):

PR #2918, opened 2024-06-25
CAPT 1689/irp/employment information

rjlynch:
Add employment details form

Adds the employment details page as per IRP. We've named the fields the
same as in IRP. We could probably provide a school search box for this
but that's something for a future iteration.
Have also added the school details to the PII section to be on the safe
side as they might be enough to identify an claimant in combination with
some of the other non pii answers.

### DIFF
--- a/app/forms/journeys/get_a_teacher_relocation_payment/employment_details_form.rb
+++ b/app/forms/journeys/get_a_teacher_relocation_payment/employment_details_form.rb
@@ -1,0 +1,62 @@
+module Journeys
+  module GetATeacherRelocationPayment
+    class EmploymentDetailsForm < Form
+      attribute :school_headteacher_name, :string
+      attribute :school_name, :string
+      attribute :school_address_line_1, :string
+      attribute :school_address_line_2, :string
+      attribute :school_city, :string
+      attribute :school_postcode, :string
+
+      validates :school_headteacher_name,
+        presence: {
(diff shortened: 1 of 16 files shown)


config/analytics_pii.yml
```diff
@@ -45,3 +45,9 @@ shared:
   - email
   :international_relocation_payments_eligibilities:
   - passport_number
+  - school_headteacher_name
+  - school_name
+  - school_address_line_1
+  - school_address_line_2
+  - school_city
+  - school_postcode
```
slorek, line 53: Please move these to `analytics_blocklist.yml` - see https://dfedigital.atlassian.net/browse/CAPT-1741

slorek (changes requested):
We probably do want to migrate this to the existing school search forms prior to the next claim window to avoid future data schema difference issues.

config/analytics_pii.yml
rjlynch: 👍

rjlynch:
> We probably do want to migrate this to the existing school search forms prior to the next claim window to avoid future data schema difference issues.

That should be covered by this [ticket](https://dfedigital.atlassian.net.mcas.ms/jira/software/projects/CAPT/boards/132?assignee=712020%3A565ea239-10c3-40ce-a2e0-3a46fcb91702&selectedIssue=CAPT-1533). I'll make the schema changes in that so we don't block testing on the current journey.

PR #2919, opened 2024-06-25
CAPT 1689/irp/claim submission form

rjlynch:
Add smoke test

Adds a smoke test to ensure the expect attributes for the journey are
written to the claim and eligibility.


spec/forms/journeys/get_a_teacher_relocation_payment/claim_submission_form_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+      build(
+        :get_a_teacher_relocation_payment_answers,
+        **claim_answers.merge(eligibility_answers)
+      )
+    end
+
+    before { form.save }
+
+    it "sets the expect attributes on the claim" do
+      form.save
```
asmega, line 81: i don't think you need this line if you already have the `before`

slorek (approved):

PR #2927, opened 2024-06-26
[LUPEYALPHA-549] Add teaching hours per week to FE journey

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-549
- Add teaching hours per week to FE journey


rjlynch (approved):
Looks good 👍

app/forms/journeys/further_education_payments/teaching_hours_per_week_form.rb
```diff
@@ -0,0 +1,44 @@
+module Journeys
+  module FurtherEducationPayments
+    class TeachingHoursPerWeekForm < Form
+      attribute :teaching_hours_per_week, :string
+
+      validates :teaching_hours_per_week,
+        inclusion: {in: ->(form) { form.radio_options.map(&:id) }, message: i18n_error_message(:inclusion)}
+
+      def question
+        "On average, how many hours per week are you timetabled to teach at #{college.name} during the current term?"
```
slorek, line 10: Not a blocker, but to note, we should be using I18n for this type of content strings, as we do with error messages.

slorek (approved):

app/forms/journeys/further_education_payments/teaching_hours_per_week_form.rb
asmega: yep I will go back and change these

PR #2928, opened 2024-06-26
CAPT 1742/dont allow future start dates

rjlynch:
Don't allow future dates for start date

IRP feedback from Richard Forrest

>  I am able to enter a contract start date in the future although I can't
>  enter a travel date in the future. Can both these fields only accept
>  dates in the past?

Updates the start date form to not permit dates in the future.


slorek (approved):
Let's double check this is correct with the poroduct/policy team before merging.

rjlynch:
> Let's double check this is correct with the product/policy team before merging.

I spoke to Richard Forest about this - The reasoning this check was introduced was to stop people who hadn't yet moved to England, and hadn't yet started their contract, from opening claims for their upcoming contract. Apparently there were quite a few claims like this and the policy intent is to only give payments for those who have already started teaching.

PR #2930, opened 2024-06-27
[LUPEYALPHA-551] Add subject areas to FE journey

asmega:
# Context

- Add subject areas form to FE journey

# Changes

- Added subject areas form to FE journey
- Added `form#t` to aid in translations on forms
- From what I can tell this is actually the first use of checkboxes in any of the existing journeys and was not supported out of the box
- The issue was around strong parameters and how permitted attributes were whitelisted. In order to `permit` a collection it must be passed `attribute_name: []` and not just `attribute_name` which was the prior implementation
- A collection is discerned with the default value of `[]` which is probably sufficient for our current needs. If this needs to be more robust we'll need to register a custom type eg:
```
ActiveModel::Type::Collection # does not exist and we'd have to implement

class MyForm
  attribute :my_field, :collection
end
```

### DIFF
--- a/app/forms/form.rb
+++ b/app/forms/form.rb
@@ -38,6 +38,10 @@ class Form
     I18n.t("#{i18n_namespace}.#{base_key}", default: base_key, **args)
   end
 
+  def t(key, args = {})
+    I18n.t(key, scope: "#{i18n_namespace}.forms.#{i18n_form_namespace}", **args)
+  end
+
   def permitted_params
     @permitted_params ||= params.fetch(model_name.param_key, {}).permit(*permitted_attributes)
   end
@@ -49,7 +53,16 @@ class Form
   private
 
   def permitted_attributes
-    attribute_names
+    attributes.keys.map do |key|
+      field = @attributes[key]
+
+      case field.value_before_type_cast
+      when []
+        {key => []}
+      else
+        key
+      end
+    end
   end
 
   def i18n_form_namespace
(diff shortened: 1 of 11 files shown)


app/forms/form.rb
```diff
@@ -49,19 +53,28 @@ def persisted?
...
   def attributes_with_current_value
-    attributes.each_with_object({}) do |(attribute, _), attributes|
-      attributes[attribute] = permitted_params[attribute]
-      next unless attributes[attribute].nil?
+    attributes.each_with_object({}) do |(attribute, _), hash|
+      hash[attribute] = permitted_params[attribute]
+      next unless hash[attribute].nil?
 
-      attributes[attribute] = load_current_value(attribute)
+      hash[attribute] = load_current_value(attribute)
```
asmega, line 77: no change. change local variable name to `hash` to disambiguate from `attributes` which is a method on the object

app/forms/journeys/further_education_payments/subjects_taught_form.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      attribute :subjects_taught, default: []
+
+      before_validation :clean_subjects_taught
+
+      validates :subjects_taught,
+        presence: {message: i18n_error_message(:inclusion)},
+        inclusion: {in: ->(form) { form.radio_options.map(&:id) }, message: i18n_error_message(:inclusion)}
+
+      def radio_options
```
rjlynch, line 14:
```suggestion
      def checkbox_options
```

app/views/further_education_payments/claims/subjects_taught.html.erb
```diff
@@ -0,0 +1,28 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: @form, url: claim_path(current_journey_routing_name), method: :patch, builder: GOVUKDesignSystemFormBuilder::FormBuilder, html: { novalidate: false } do |f| %>
+      <%= f.govuk_error_summary %>
+
+      <%= f.govuk_check_boxes_fieldset :subjects_taught,
+        legend: {
+          text: @form.t(:question),
```
rjlynch, line 8: 👍 much nicer

rjlynch (approved):
Looks good, one minor naming comment but it's not a blocker

PR #2932, opened 2024-06-27
Setup the admin to display IRP claims

felixclack:
In order to be able to display the IRP claims in the admin, there are some basic methods that need to be available.

This change ensures the minimum implementation for displaying claims without being concerned about the correctness of the required methods. Specifically the `award_amount` is a placeholder.

In order to ensure we can successfully check for duplicate claims, I
made an assumption that we would use passport number rather than teacher
reference number as a unique identifier.

I don't think claimants for IRP will have TRNs, therefore we need some
other way of matching.

We can change this assumption easily in the future but for now this
allows us to ship the admin changes and keep moving forward.

### DIFF
--- a/app/models/policies/international_relocation_payments.rb
+++ b/app/models/policies/international_relocation_payments.rb
@@ -3,6 +3,9 @@ module Policies
     include BasePolicy
     extend self
 
+    ELIGIBILITY_MATCHING_ATTRIBUTES = [["passport_number"]].freeze
+    OTHER_CLAIMABLE_POLICIES = []
+
     # NOTE RL: currently IRP only has a single reply to address, so notify
     # doesn't show the address id
     def notify_reply_to_id
--- a/app/models/policies/international_relocation_payments/eligibility.rb
+++ b/app/models/policies/international_relocation_payments/eligibility.rb
@@ -3,7 +3,16 @@ module Policies
     class Eligibility < ApplicationRecord
       self.table_name = "international_relocation_payments_eligibilities"
 
+      AMENDABLE_ATTRIBUTES = %i[].freeze
+
       has_one :claim, as: :eligibility, inverse_of: :eligibility
+      belongs_to :current_school, class_name: "School"
+
+      attr_accessor :teacher_reference_number
+
+      def award_amount
+        0
+      end
 
       def ineligible?
         false
--- a/app/models/policies/international_relocation_payments/eligibility_admin_answers_presenter.rb
+++ b/app/models/policies/international_relocation_payments/eligibility_admin_answers_presenter.rb
@@ -1,0 +1,28 @@
+module Policies
+  module InternationalRelocationPayments
+    class EligibilityAdminAnswersPresenter
+      include Admin::PresenterMethods
+
+      attr_reader :eligibility
+
+      def initialize(eligibility)
+        @eligibility = eligibility
+      end
+
+      def answers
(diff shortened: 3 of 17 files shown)


app/models/policies/international_relocation_payments/eligibility_admin_answers_presenter.rb
```diff
@@ -0,0 +1,28 @@
...
+          a << current_school
+        end
+      end
+
+      private
+
+      def current_school
+        [
+          translate("admin.current_school"),
+          eligibility.current_school.present? ? display_school(eligibility.current_school) : "No"
```
rjlynch, line 23: 👍

rjlynch (approved):
Looks good to me 👍
We'll be able to add in the current school soon too, which should be pretty straight forward as the admin eligibility presenter is already set up for it 👍

rjlynch:
FYI - Have manually tested this with a claim on my local machine, looks like it works well and the claims show up in the payroll run too 👍 
On `master` the frontend journey is now setting the `current_school_id` on the IRP eligibility, so if you want to show the school in the admin area it should be good to go (just add the `belongs_to :current_school` to the eligibility and it should all work) - but I don't think it's a blocker to shipping this

app/models/policies/international_relocation_payments/eligibility.rb
```diff
@@ -3,8 +3,21 @@ module InternationalRelocationPayments
...
+      attr_accessor :teacher_reference_number
+
+      def award_amount
+        0
+      end
+
+      # No current_school attribute on the model. This method is for compatibility with the admin UI.
+      def current_school
+        nil
+      end
```
rjlynch, line 19: Not a blocker but we can now change this to

```ruby
belong_to :current_school, class_name: "School"
```

as the current_school_id is being set on master

app/models/policies/international_relocation_payments/eligibility_admin_answers_presenter.rb
```diff
@@ -0,0 +1,28 @@
...
+        ]
```
rjlynch, line 24: Again not a blocker but we can now assume a current_school is always set on the eligiblity

spec/support/admin_view_claim_feature_shared_examples.rb
```diff
@@ -34,7 +34,12 @@
   }
 
   let!(:similar_claim) {
-    eligibility = create(:"#{policy.to_s.underscore}_eligibility", :eligible, teacher_reference_number: multiple_claim.eligibility.teacher_reference_number)
+    duplicate_attribute = if policy == Policies::InternationalRelocationPayments
+      {passport_number: multiple_claim.eligibility.passport_number}
```
rjlynch, line 38: 👍 
I was wondering if we could use the national insurance number as it's required for all journeys and it should be unique. Obv passport number is fine too

app/models/policies/international_relocation_payments.rb
```diff
@@ -3,6 +3,9 @@ module InternationalRelocationPayments
     include BasePolicy
     extend self
 
+    ELIGIBILITY_MATCHING_ATTRIBUTES = [["passport_number"]].freeze
```
rjlynch, line 6: 👍

rjlynch (approved):
Looks good, left a couple of comments but none of them are blockers 👍

PR #2935, opened 2024-06-27
Fix visiting landing page for same journey

rjlynch:
Previously if a claimant started a journey then visited the landing page
for that journey and choose to continue with their existing session we'd
throw an error. The `other_journey_sessions` array contains the journey
sessions for the other journeys, which if you're still on the same
journey would be empty, so we need to check the current journey session
(the session for the journey we're currently on).

How we handle managing which journey you're on is subtle and spread
across many places in the controllers / concerns, the overall approach
likely needs revisiting.

## Before

https://github.com/DFE-Digital/claim-additional-payments-for-teaching/assets/9936028/2d71a455-5f6f-4f1b-9aca-6ec26d834659

## After

https://github.com/DFE-Digital/claim-additional-payments-for-teaching/assets/9936028/2a4fbf42-ae5d-43a4-b461-2c09c6862010

### DIFF
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -39,7 +39,11 @@ class ClaimsController < BasePublicController
   delegate :slugs, :current_slug, :previous_slug, :next_slug, :next_required_slug, to: :page_sequence
 
   def redirect_to_existing_claim_journey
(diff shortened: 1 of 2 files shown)


7AXEL:
Check out this project: https://github.com/7AXEL/Web2APK

slorek (approved):
Good catch

PR #2936, opened 2024-06-27
CAPT 1533/school select

rjlynch:
Replace employment details with select school

In the original IRP journey the claimant had to manually enter their
school's address. We've replaced manual address entering with a school
select form. We still ask for the name of the head teacher as the GIAS
database doesn't contain head teacher information for all schools.

## Changes
* Removes the employment details form
* Adds a head teacher name form
* Wires up current school form into IRP journey

Note that we can get the head teacher name, for most schools, from GIAS however there's a business decision to ask the claimant for this information.

Note that we don't reset the head teacher name if the school changes - I'll check with the business team whether that's something we need to do or not. If we do want to do so I'll address this as a separate PR.

## Walk through

https://github.com/DFE-Digital/claim-additional-payments-for-teaching/assets/9936028/be61bb7c-d6f1-4df3-981d-4118cb89482e


app/views/get_a_teacher_relocation_payment/claims/_current_school_details.html.erb
```diff
@@ -0,0 +1 @@
+<%# NOOP %>
```
rjlynch, line 1: This is required by the current school form

spec/features/get_a_teacher_relocation_payment/teacher_route_completing_the_form_spec.rb
```diff
@@ -41,31 +45,33 @@
...
-        then_the_check_your_answers_part_page_shows_my_answers
+        then_the_check_your_answers_part_page_shows_my_answers(school)
         and_i_submit_the_application
         then_the_application_is_submitted_successfully
       end
     end
 
     context "without postcode search" do
-      it "submits an application" do
+      it "submits an application", js: true do
```
rjlynch, line 63: Figured we want to hit the js version of the current school form in one of the IRP tests to make sure that's covered by a spec

PR #2939, opened 2024-06-27
[LUPEYALPHA-555] Add poor performance form to FE journey

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-555
- Add poor performance form to FE journey

# Changes

- there is now a single `poor_performance` form that is shared between
- `poor_performance` form now uses GOVUK formbuilder


slorek (commented):
Can we re-use the `Journeys::AdditionalPaymentsForTeaching::PoorPerformanceForm` form object? We can move that out of the journey specific namespace so it can be re-used.

Generally, we would want to do this and then override where necessary. This includes view templates and copy strings as well.

slorek (approved):

PR #2942, opened 2024-06-28
[LUPEYALPHA-545] FE journey school search only returns FE bodies

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-545
- Missing part of FE journey where school search should only return FE bodies

# Changes

- School search backend now capable of returning FE bodies only
- This is based of `PhaseOfEducation` is `16 plus` this is encoded as `6` 
- Wired up FE journey to this so only return FE bodies


rjlynch (approved):
Nice approach 👍

PR #2943, opened 2024-06-28
Add eligibility check for contract start date

rjlynch:
We want to reject claimants if their contract start date is earlier than
6 months before the autumn term. This commit updates the eligibility
checker to do so and also adds some missing tests.

## Walk through

https://github.com/DFE-Digital/claim-additional-payments-for-teaching/assets/9936028/2ac61505-a9a7-4d84-9990-403650dc26ec

### DIFF
--- a/app/models/policies/international_relocation_payments/eligibility.rb
+++ b/app/models/policies/international_relocation_payments/eligibility.rb
@@ -3,6 +3,15 @@ module Policies
     class Eligibility < ApplicationRecord
       self.table_name = "international_relocation_payments_eligibilities"
 
+      PRE_ACADEMIC_YEAR_WINDOW_LIMIT = 6.months
+
+      def self.earliest_eligible_contract_start_date
+        Journeys::GetATeacherRelocationPayment
+          .configuration
+          .current_academic_year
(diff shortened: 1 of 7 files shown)


app/models/policies/international_relocation_payments/eligibility.rb
```diff
@@ -3,6 +3,15 @@ module InternationalRelocationPayments
...
       self.table_name = "international_relocation_payments_eligibilities"
 
+      PRE_ACADEMIC_YEAR_WINDOW_LIMIT = 6.months
+
+      def self.earliest_eligible_contract_start_date
+        Journeys::GetATeacherRelocationPayment
+          .configuration
+          .current_academic_year
+          .start_of_autumn_term - PRE_ACADEMIC_YEAR_WINDOW_LIMIT
+      end
```
slorek, line 13: I think this method could probably live in the eligibility checker class more comfortably. For context, I think we need to change the name of these `Eligibility` classes to something like `SupportingInformation` - or move their data into the `Claim` itself.

slorek (approved):

app/models/policies/international_relocation_payments/eligibility.rb
rjlynch: Makes sense, have moved it to the checker class 👍

PR #2947, opened 2024-06-28
[LUPEYALPHA-615] FE journey handle ineligible as lack teaching

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-615
- For FE journey add ineligible path when user lacks teaching responsibility

# Changes

- Added new ineligible page
- This is the first ineligible page for this journey so some plumbing was required to set everything up

### DIFF
--- a/app/forms/journeys/further_education_payments/ineligible_form.rb
+++ b/app/forms/journeys/further_education_payments/ineligible_form.rb
@@ -1,0 +1,9 @@
+module Journeys
+  module FurtherEducationPayments
+    class IneligibleForm < Form
+      def journey_eligibility_checker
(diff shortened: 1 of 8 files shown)


slorek (commented):
I'm not sure we need a form object for the ineligible page.

asmega:
> I'm not sure we need a form object for the ineligible page.

The benefit of using a form object is it acts a point to hang stuff for things that happen in the view. At the moment theres not much there and I could delete the form object and just inline everything into the view. The downside though is that the next person that comes along is likely to stuff more things into the view than desired if more business logic comes along.

I think its a very low price to pay for the potential benefits down the line. Also it just makes the pattern of the other pages in the journey much the same.

slorek (approved):

PR #2949, opened 2024-07-01
Add the Home Office identity confirmation task

felixclack:
The first admin task for the IRP journey is verifying the Home Office
identity check.

We request the nationality and passport number for the individual
submitting the claim and present that as the check for the admin task.

| Header | Header |
|--------|--------|
| Claim view | Screenshot 2024-07-01 at 7 41 36 am |
| Task view | Screenshot 2024-07-01 at 7 41 48 am |
| Confirmed | Screenshot 2024-07-01 at 7 42 12 am |


rjlynch (approved):
Looks good 👍

PR #2951, opened 2024-07-01
Purge unsubmitted journey sessions

rjlynch:
Now we only create a claim once the claimant has completed the journey
we need to update the purge job to work on journey sessions rather than
claims as all claims are submitted claims.

### DIFF
--- a/spec/jobs/purge_unsubmitted_claims_job_spec.rb
+++ b/spec/jobs/purge_unsubmitted_claims_job_spec.rb
@@ -6,17 +6,37 @@ RSpec.describe PurgeUnsubmittedClaimsJob do
     let(:four_hours_ago) { 4.hours.ago }
 
     it "destroys any unsubmitted claims that have not been updated in the last 24 hours" do
-      expired_unsubmitted_claim = create(:claim, updated_at: over_24_hours_ago)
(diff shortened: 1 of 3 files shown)


spec/jobs/purge_unsubmitted_claims_job_spec.rb
```diff
@@ -6,17 +6,37 @@
...
-      expired_unsubmitted_claim = create(:claim, updated_at: over_24_hours_ago)
-      active_unsubmitted_claim = create(:claim, updated_at: four_hours_ago)
+      submitted_journeys = []
+      unsubmitted_fresh_journeys = []
+      unsubmitted_expired_journeys = []
 
-      old_submitted_claim = create(:claim, :submitted, updated_at: over_24_hours_ago)
+      Journeys::JOURNEYS.each do |journey|
+        submitted_journeys << journey::Session.create!(
+          journey: journey::ROUTING_NAME,
```
rjlynch, line 15: I think we might want to consider switching from storing the journey on the journey session like this to proper STI. (Not sure if it's worth the effort as I don't expect us to do much querying on the journey session table)

slorek (approved):

PR #2954, opened 2024-07-02
Add a visa check task for IRP

felixclack:
For the international relocation payments journey we want to add a new
admin task for checking the visa type provided.

We collect the visa type as part of the journey, so we want an admin
task to confirm this.

Screenshot 2024-07-02 at 10 02 53 am


app/views/admin/tasks/visa.html.erb
```diff
@@ -0,0 +1,27 @@
...
+  </div>
+
+  <div class="govuk-grid-column-two-thirds">
+    <%= render "admin/claims/answers", answers: @tasks_presenter.visa %>
+  </div>
+
+  <div class="govuk-grid-column-two-thirds">
+    <% if !@task.passed.nil? %>
+      <%= render "task_outcome", task: @task do %>
+      <% end %>
```
rjlynch, line 19: Do we need to pass a block here, doesn't seem like it's used
```suggestion
      <%= render "task_outcome", task: @task %>
```

rjlynch (approved):
Looks good

PR #2959, opened 2024-07-02
Remove hostname redirect check

slorek:
This code redirects requests to the 'canonical hostname'. It was [implemented in 2019](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/400) when it appears there were some publicly accessible Azure URLs  causing problems. In production, it also redirected all requests to `www.` if this was not prefixed on the request URL.

With the new DNS changes it has now been possible to implement this at the infrastructure level and so this code is now redundant.


rjlynch (approved):
👍

slorek:
> Please also remove the associated terraform to set this environment variable if it's no longer used, e.g. https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/master/terraform/application/config/production.tfvars.json#L6

Checked with @saliceti and confirmed this can't be removed.

PR #2961, opened 2024-07-02
Capt 1527/keep irp national insurance and name for 2 years

rjlynch:
## Background
International relocation payments have different rules around scrubbing PII from submitted claims
> We need to amend the purge rules for IRP only so that applicant name and NiNo are retained for 2 years. 

## Changes
There's a few commits in this pr, here's an over view of the main changes:

### Changes how we identify claims from the same claimant.
Previously we used TRN to identify claims from the same claimant, however new policies, such as IRP, don't capture a TRN. We've now switched to using national insurance number to identify multiple claims belonging to the same claimant. This is a bit unrelated to the main work in this pr but was required to fix a test and we'll run into the issue eventually. See [this commit](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2961/commits/08741e81f998f1184acf73f1f8ccbc27b7cd6018) for where these changes are implemented. **I think this commit deserves the most scrutiny as I'm not 100% sure of the knock on effects of using NINO to identify claimants**

### Adds policy specific data scrubbers
IRP has a different retention policy for some PII fields. We've replaced the `Claim::PersonalDataScrubber` with policy specific `Policies::<PolicyName>::ClaimPersonalDataScrubber`s to allow overwriting how PII is scrubbed on a per policy level. See [this commit](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2961/commits/091f4a27540992059a1d915a93efb8f4cde7bba3)

### Refactor `ClaimPersonalDataScrubber`
We've moved some methods around and separated out finding claims to be scrubbed from scrubbing the claim, extracting a `Claim::Scrubber` class. This makes it easier for policy specific claim scrubbers to override the behaviours in the base class. See these commits [1](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2961/commits/1d717b0f5e4607445b4096894bd761357a88b184), [2](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2961/commits/8444f18170fd9c9a4f3cd1cab89b034e56078d37), [3](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2961/commits/8e81a0008a3617ba8c9cb83c35282da8d5dd9d3e)

### Retain NINO and claimant name for IRP claims
Scrub national insurance number and claimant name after 2 years. See [this commit](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2961/commits/9d67cfa554595330f4a7692e657c952f2cab2879)

### Scrub PII from journey session
We may want to consider doing this when we submit the claim but [this commit](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2961/commits/95544ccdc40265376e7056da994190474a4bcbf2) updates the claim scrubber to also remove all answers from the journey session


slorek (commented):
I think the approach of defining the behaviours in policy specific classes is the right one. I would lean to moving these to the `Policies` namespace as well so that all of the policy specific code is housed together rather than dotted around the codebase and having to know where to look. Also agree with implementing the NOOP classes before merging. Thanks.

app/models/claim/claims_preventing_payment_finder.rb
```diff
@@ -15,28 +15,21 @@ def initialize(claim)
...
     def find_claims_preventing_payment
-      return [] if claim.policy == Policies::InternationalRelocationPayments
-
-      eligibility_ids = claim.policy.policies_claimable.map { |policy|
-        policy::Eligibility.where(teacher_reference_number: claim.eligibility.teacher_reference_number)
-      }.flatten.map(&:id)
+      return [] if claim.personal_data_removed?
 
-      payrollable_claims_from_same_claimant = Claim.payrollable.where(eligibility_id: eligibility_ids)
+      payrollable_claims_from_same_claimant = Claim.payrollable.with_same_claimant(claim)
```
rjlynch, line 27: I'm not sure if we need to preserve filtering the claims here by matching `policies_claimable` or not?

app/models/claim/scrubber.rb
```diff
@@ -0,0 +1,47 @@
...
+      amendment.save!
+    end
+
+    def scrub_claim!
+      personal_data_mask = attributes_to_delete.to_h { |attr| [attr, nil] }
+      attributes_to_set = personal_data_mask.merge(
+        personal_data_removed_at: Time.zone.now
+      )
+      claim.update_columns(attributes_to_set)
+    end
```
asmega, line 39: why do we call `save!` on the amendment but `update_columns` on the claim.
so `save!` will fire validation and callbacks but `update_columns` will not
is there a reason for the difference? or should they both be calling the same thing?

app/models/claim/scrubber.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def scrub_session!
+      return unless claim.journey_session
+
+      claim.journey_session.update!(answers: {})
```
asmega, line 44: are we sure we want to remove all answers and not just the answers that need scrubbing?

one example i can think of for FE is that claimants are supposed to have a teaching qualification or on course to attain a teaching qualification. this answer may want to be kept around so this can be verified at a future date

asmega: should be fine as we use a different implementation there i guess, since for this change technically only applies to IRP

asmega (approved):

app/models/claim/scrubber.rb
rjlynch: We use `update_columns` to skip any callbacks and validations. The previous version of this used `update_all` for persisting the changes to the claim and `save!` for persisting the changes to the amendment so I wanted to preserve that behaviour.

app/models/claim/scrubber.rb
rjlynch: > are we sure we want to remove all answers and not just the answers that need scrubbing?
I think it's safe as the answers should be copied to the claim upon submission. We probably will want to move scrubbing the session answers at the point of submission/claim creation instead, I've added scrubbing the session here in case we decide not to do that

app/models/claim/scrubber.rb
alkesh: Using `save!` here will mean the callbacks on the claim fail, as there are a bunch of `normalise_*` callbacks that assume those fields are populated.

slorek: It would be preferable to use `save` because the [dfe_analytics](https://github.com/DFE-Digital/dfe-analytics) gem  does not hook into `update_all` or `update_columns`. This causes data inconsistencies in the BigQuery database (although PII is not sent there in the first place).

app/models/policies/claim_personal_data_scrubber.rb
```diff
@@ -29,46 +32,65 @@ class PersonalDataScrubber
     ]
 
     def scrub_completed_claims
-      Claim.transaction do
-        scrub_claims(old_rejected_claims)
-        scrub_claims(old_paid_claims)
+      old_rejected_claims
+        .where(personal_data_removed_at: nil)
```
slorek, line 36: Suggestion: a named scope for this condition would be a small improvement.

slorek (approved):

app/models/claim/scrubber.rb
rjlynch: I don't think we can use `save` due to the callbacks Alkesh highlighted. Does the existing data scrubbing implementation create data inconsistency issues in big query? We could try using `save` in a suppress callbacks block.

spec/factories/journeys/get_a_teacher_relocation_payment/get_a_teacher_relocation_payment_answers.rb
```diff
@@ -80,9 +80,21 @@
...
+
+    trait :submittable do
+      eligible_teacher
+      with_personal_details
+      with_nationality
+      with_passport_number
+      with_email_details
+      with_mobile_details
+      with_bank_details
+      with_employment_details
```
alkesh, line 97: this trait `with_employment_details` seems to be missing

app/models/claim/scrubber.rb
alkesh: could we update the 3 affected callbacks (for `ni_number`, `first_name` and `surname` so they don't get called if that field is nil?
e.g.
`before_save :normalise_ni_number, if: [:national_insurance_number, :national_insurance_number_changed?]`

db/migrate/20240701143154_add_award_amount_to_international_relocation_payments_eligibilities.rb
alkesh, line 1: this is a duplicate - it already exists in `db/migrate/20240709110147_add_award_amount_to_international_relocation_payments_eligibilities.rb`

app/models/claim/scrubber.rb
rjlynch: > could we update the 3 affected callbacks (for `ni_number`, `first_name` and `surname` so they don't get called if that field is nil? e.g. `before_save :normalise_ni_number, if: [:national_insurance_number, :national_insurance_number_changed?]`

That seems to do the trick 👍

PR #2965, opened 2024-07-03
[LUPEYALPHA-547] FE contract type and branching

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-547
- Add contract type forms and associated forms with branching

# Changes

- Added all pages related to contract types for FE journey
- Also includes an ineligible possibility as an outcome 
- Resetting of dependent answers as its possible to go forwards then backtrack, then forwards again and get to page that has been answered which is handle by resetting the answers on the change

### DIFF
--- a/app/forms/journeys/further_education_payments/taught_at_least_one_term_form.rb
+++ b/app/forms/journeys/further_education_payments/taught_at_least_one_term_form.rb
@@ -13,11 +13,11 @@ module Journeys
         [
           OpenStruct.new(
             id: true,
-            name: "Yes, I have taught at #{school.name} for at least one academic term"
+            name: t("options.true", school_name: school.name)
           ),
           OpenStruct.new(
             id: false,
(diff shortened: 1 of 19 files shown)


app/forms/journeys/further_education_payments/taught_at_least_one_term_form.rb
```diff
@@ -0,0 +1,37 @@
+module Journeys
+  module FurtherEducationPayments
+    class TaughtAtLeastOneTermForm < Form
+      attribute :taught_at_least_one_term, :boolean
+
+      validates :taught_at_least_one_term,
+        inclusion: {
+          in: [true, false],
+          message: ->(object, data) { i18n_error_message(:inclusion, school_name: object.school.name).call(object, data) }
```
asmega, line 9: this was a bit quirky not 100% what the api should be but this seems to work and my tests pass

slorek (approved):

PR #2969, opened 2024-07-03
Add the payroll admin tasks for IRP

felixclack:
We want to allow the existing payroll tasks to the IRP journeys in the
admin.

I opted to remove the early return for the IRP task list in order to
avoid duplicating logic around displaying the payroll tasks.


rjlynch (approved):
👍

PR #2971, opened 2024-07-04
CAPT-1751 Add review and test app ENV variables

AbigailMcP:
## Changes in this PR

As per this comment: https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2887#discussion_r1660801819

Move non-secret variables into code so the deployment is faster and they can be tracked / updated more easily.

Guided by what was already disclosed here: https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/master/.env.production

### DIFF
--- a/terraform/application/config/review_app_env.yml
+++ b/terraform/application/config/review_app_env.yml
@@ -1,3 +1,21 @@
 ---
 BYPASS_DFE_SIGN_IN: true
+
+DFE_SIGN_IN_API_CLIENT_ID: teacherpayments
+DFE_SIGN_IN_API_ENDPOINT: https://pp-api.signin.education.gov.uk
+DFE_SIGN_IN_IDENTIFIER: teacherpayments
+DFE_SIGN_IN_ISSUER: https://pp-oidc.signin.education.gov.uk:443
+
+DQT_API_URL: https://preprod.teacher-qualifications-api.education.gov.uk/v1
+DQT_BASE_URL: https://test-api-customerengagement.platform.education.gov.uk/dqt-crm/v1/
+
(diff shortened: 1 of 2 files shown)


AbigailMcP:
Tested by deleting ORDNANCE_SURVEY_API_BASE_URL from azure key vault and re-deploying test with new in-code variables.

Checked ENV in rails console and ORDNANCE_SURVEY_API_BASE_URL was correctly populated.

Also checked manually that the postcode search still worked.

So as not to break pipeline for others I've temporarily replaced ORDNANCE_SURVEY_API_BASE_URL in key vault whilst this PR is reviewed.

.github/workflows/build_and_deploy.yml
```diff
@@ -2,7 +2,7 @@ name: Build and deploy to AKS cluster
 
 on:
   push:
-    branches: [master]
+    branches: [CAPT-1751-move-non-secret-secrets-out-of-key-vault]
```
slorek, line 5: Please update before merging.

.github/workflows/build_and_deploy.yml
```diff
@@ -86,7 +86,7 @@ jobs:
     name: Deploy to test environment
     concurrency: deploy_test
     runs-on: ubuntu-latest
-    if: github.ref == 'refs/heads/master' && github.event_name == 'push'
+    if: github.ref == 'refs/heads/CAPT-1751-move-non-secret-secrets-out-of-key-vault' && github.event_name == 'push'
```
slorek, line 89: Please update before merging.

slorek (approved):

PR #2977, opened 2024-07-05
CAPT-1751 Move production env variables into code

AbigailMcP:
## Changes in this PR

Following on from: https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/2971

Move non-secret variables into code.


AbigailMcP:
> @AbigailMcP I presume it's mostly copy and paste, just my OCD, some have trailing slashes and some don't wondering if they need them, I guess if they use plain old string concatenation it could cause issues if we remove/add them for consistency.

I'll test out removing the trailing slashes today on the test env.

AbigailMcP:
> @AbigailMcP I presume it's mostly copy and paste, just my OCD, some have trailing slashes and some don't wondering if they need them, I guess if they use plain old string concatenation it could cause issues if we remove/add them for consistency.

@kenfodder I tested removing the trailing slash from TID_BASE_URL and it broke the DfE Identity Sign In... so I'm leaving them copy-and-pasted as is!

PR #2978, opened 2024-07-05
[LUPEYALPHA-551] FE subjects and courses

asmega:
# Context

- Add subject areas and courses to FE journey
- When a user selects subject areas they are then shown course pages related to their selected subject areas

# Changs

- Sorry for the biggish change, most of the work is fairly repetitive which should make reviewing easier
- I've camel cased the FE email, seems like this is how we want to display it
- If a user selects `none` to subject area question they are ineligible
- If a user does not select at least one eligible course they are ineligible
- If a user unchecks any subject areas we reset dependent answers therefore any courses they selected are blanked
- Minor tweak where on the ineligible page the `start again` link goes to the landing page rather than `/claim`
- `Mathematics` is now `Maths` or `maths` despite being different in other journeys this is the requirement for this journey


app/forms/journeys/further_education_payments/building_construction_courses_form.rb
```diff
@@ -0,0 +1,65 @@
...
+
+      def course_field
+        :building_construction_courses
+      end
+
+      def checkbox_options
+        [
+          OpenStruct.new(
+            id: "esfa_buildingconstruction",
+            name: "ESFA-funded qualifications at level 3 and below in the #{govuk_link_to "building and construction", "https://www.qualifications.education.gov.uk/Search?Status=Approved&Level=0,1,2,3,4&Sub=7"} sector subject area"
```
asmega, line 26: i tried a few different approaches to get link into label and personally found this one to be the best. if you don't fancy this approach please suggest a better approach

asmega: ok, i've now moved all of these in to i18n

app/views/further_education_payments/claims/_courses.html.erb
```diff
@@ -0,0 +1,36 @@
...
+      <%= f.govuk_error_summary %>
+
+      <%= f.govuk_check_boxes_fieldset @form.course_field,
+        legend: {
+          text: @form.t(:question),
+          tag: "h1",
+          size: "l"
+        },
+        hint: {
+          text: @form.t(:hint, email: govuk_link_to(t("support_email_address", scope: @form.i18n_namespace))).html_safe
```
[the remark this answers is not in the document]
asmega: i've added the line break

PR #2979, opened 2024-07-08
Add StatusCake for the test environment

vacabor:
Enables the StatusCake module and uses the existing contact groups for alerting.


AbigailMcP (commented):
Do we need to set the `external_url` variable in `test.tfvars.json` to be our healthcheck url?

AbigailMcP:
> > > Do we need to set the `external_url` variable in `test.tfvars.json` to be our healthcheck url?
> > 
> > 
> > `statuscake.tf` takes care of setting a check on the healthcheck URL by getting it from the application module, but I've also set `external_url` so we can have an extra SSL check on `/additional-payments/claim`
> 
> I've reverted this - it's something we can set on production, but for test the endpoint is protected by HTTP basic (so the StatusCake check always fails).

makes sense!

AbigailMcP (approved):

PR #2980, opened 2024-07-08
Fix missing translation

rjlynch:
We were missing a translation for the matching details task

before

after


asmega (approved):

asmega:
should probably throw exceptions in test to catch these?

PR #2981, opened 2024-07-08
[LUPEYALPHA-639] Add unhappy teaching hours paths to FE journey

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-639
- Handle unhappy paths for teaching hour pages for FE journey

# Changes

- Added new page to capture teaching hours for next term if user is not on a permanent contract
- Although technically you could store as a boolean due the way the question is structured i think storing as a string is far more flexible should they decide to change the question from a simple yes/no answer


rjlynch (approved):
Looks good 👍

PR #2984, opened 2024-07-08
[LUPEYALPHA-640] Add ineligible state for non recent FE teacher

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-640
- Add ineligible state for non-recent FE teachers for FE journey

### DIFF
--- a/app/models/policies/further_education_payments/policy_eligibility_checker.rb
+++ b/app/models/policies/further_education_payments/policy_eligibility_checker.rb
@@ -24,6 +24,8 @@ module Policies
           :lack_teaching_responsibilities
         elsif answers.taught_at_least_one_term == false
           :must_teach_at_least_one_term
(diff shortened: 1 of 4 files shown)


app/models/policies/further_education_payments/policy_eligibility_checker.rb
```diff
@@ -24,8 +24,16 @@ def ineligibility_reason
...
           :must_teach_at_least_one_term
+        elsif !recent_further_education_teacher?
+          :must_be_recent_further_education_teacher
         end
       end
+
+      private
+
+      def recent_further_education_teacher?
+        !answers.further_education_teaching_start_year&.start_with?("pre-")
```
rjlynch, line 35: Minor, but may want to move this method on to the answers - obv not a blocker

rjlynch (approved):
Looks good 👍

PR #2987, opened 2024-07-08
[LUPEYALPHA-645] Add FE ineligible state due to problematic actions

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-645
- Add ineligible state due to problematic actions for FE journey


rjlynch (approved):
👍

PR #2988, opened 2024-07-09
import schools on seeding review apps

asmega:
# Context

- When review apps are created we currently do not import schools from GIAS
- Importing schools from GIAS aids in the review process with less managing of custom seeds

# Changes

- When a review app spins up as part of the seeding process also enqueue `SchoolDataImporterJob` which the worker will pick up to import schools from GIAS

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -15,6 +15,8 @@ if Rails.env.development? || ENV["ENVIRONMENT_NAME"].start_with?("review")
   ENV["FIXTURES_PATH"] = "spec/fixtures"
   ENV["FIXTURES"] = "local_authorities,local_authority_districts,schools"
   Rake::Task["db:fixtures:load"].invoke
+
+  SchoolDataImporterJob.perform_later if School.count < 100
 end
 
 if Rails.env.development?


rjlynch (approved):
Nice 👍

.github/workflows/build_and_deploy.yml
```diff
@@ -64,6 +64,7 @@ jobs:
         run: |
           make ci review-aks get-cluster-credentials
           kubectl exec -n srtl-development deployment/claim-additional-payments-for-teaching-review-${{ github.event.pull_request.number }}-worker -- sh -c "DISABLE_DATABASE_ENVIRONMENT_CHECK=1 bin/prepare-database"
+          kubectl exec -n srtl-development deployment/claim-additional-payments-for-teaching-review-${{ github.event.pull_request.number }}-worker -- sh -c "bin/rails runner 'SchoolDataImporterJob.perform_later'"
```
slorek, line 67: We have a rake task: `schools_data:import`. It would probably be good to create a separate bin script which can be used to perform seeding on review apps specifically, rather than specifying it in the workflow file.

slorek (changes requested):

PR #2992, opened 2024-07-09
[LUPEYALPHA-650] Add ineligible state for FE journey

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-650
- For FE journey add ineligible state when teacher does not spend at least half their hours teaching FE


rjlynch (approved):
Looks good 👍

PR #3001, opened 2024-07-11
CAPT-1613 One Login proof of concept

alkesh:
done:
✅ call to /authorize endpoint works - returns an authorisation token
✅ call to /token endpoint
✅ call to identity verification - returns JWT with first-name and last-name
✅ decode and verify returned JWT

to do (not really part of proof of concept - will be done later):
* continue button should continue with journey
* store returned user info in journey answers
* failure cases

⚠️ the build has one failure - the happy path feature spec

useful info:
https://docs.sign-in.service.gov.uk/test-your-integration/using-integration-for-testing/

things to note when running locally:
* you can create your own One Login user accounts for initial log in, and then for the identity verification details, use one of the test users
* the app needs to be running on http://localhost:3000 (not https) as that is what the configured redirect URL is
* the one login integration environment is protected with basic auth (ask me for the credentials)
* you will need ONELOGIN_SIGN_IN_CLIENT_ID, ONELOGIN_SIGN_IN_SECRET_BASE64, and ONELOGIN_IDENTITY_JWT_PUBLIC_KEY_BASE64 environment variables


config/initializers/omniauth.rb
```diff
@@ -73,4 +87,45 @@ def self.bypass?
...
+        port: onelogin_sign_in_issuer_uri&.port,
+        redirect_uri: onelogin_sign_in_redirect_uri&.to_s,
+        scheme: onelogin_sign_in_issuer_uri&.scheme,
+        secret: onelogin_sign_in_secret_key
+      },
+      discovery: true,
+      issuer: ENV["ONELOGIN_SIGN_IN_ISSUER"],
+      response_type: :code,
+      scope: %i[openid email phone],
+      send_scope_to_token_endpoint: false
```
slorek, line 108: I think we need to send the `vtr` parameter as `["Cl.Cm"]` according to https://docs.sign-in.service.gov.uk/integrate-with-integration-environment/authenticate-your-user/#make-a-request-for-authentication-and-identity

alkesh: that parameter is optional.
"If you do not specify the vtr parameter, your service will automatically log your users in at the medium level of authentication (Cl.Cm)."
https://docs.sign-in.service.gov.uk/integrate-with-integration-environment/authenticate-your-user/#replace-the-placeholder-values-in-your-example

I'm happy to put that parameter in, to make it explicit.

slorek: great, thanks for clarifying.

PR #3002, opened 2024-07-12
CAPT-1762 Add database backup workflow

AbigailMcP:
## Changes in this PR

- Add backup database workflow via GitHub Actions
- Uses this DfE GitHub Action: https://github.com/DFE-Digital/github-actions/blob/master/backup-postgres/README.md
- Schedule to run at 3am nightly (disabled until after migration), or via manual workflow dispatch

### DIFF
--- a/.github/workflows/database_backup.yml
+++ b/.github/workflows/database_backup.yml
@@ -3,8 +3,9 @@ concurrency: build_and_deploy_main
 
 on:
   workflow_dispatch:
-  schedule: # 03:00 UTC
-    - cron: "0 3 * * *"
+  # TODO: Uncomment after migration
+  # schedule: # 03:00 UTC
+  #   - cron: "0 3 * * *"
 
 jobs:
   backup:
@@ -23,5 +24,5 @@ jobs:
           resource-group: s189p01-capt-pd-rg
           app-name: claim-additional-payments-for-teaching-production
           cluster: production
-          azure-credentials: ${{ secrets.AZURE_CREDENTIALS}}
+          azure-credentials: ${{ secrets.AZURE_CREDENTIALS }}
           backup-file: capt_prod_$(date +"%F").sql
--- a/.github/workflows/database_backup.yml
+++ b/.github/workflows/database_backup.yml
@@ -13,9 +13,6 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-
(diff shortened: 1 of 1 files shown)


.github/workflows/database_backup.yml
```diff
@@ -0,0 +1,28 @@
...
+on:
+  workflow_dispatch:
+  # TODO: Uncomment after migration
+  # schedule: # 03:00 UTC
+  #   - cron: "0 3 * * *"
+
+jobs:
+  backup:
+    name: Backup AKS Database (production)
+    if: ${{ github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' }}
```
saliceti, line 13: Is that necessary? The only triggers are schedule and workflow_dispatch

.github/workflows/database_backup.yml
```diff
@@ -0,0 +1,28 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Backup postgres
+        uses: DFE-Digital/github-actions/backup-postgres@master
+        with:
+          storage-account: s189p01captdbbkppdsa
+          resource-group: s189p01-capt-pd-rg
+          app-name: claim-additional-payments-for-teaching-production
```
saliceti, line 25: Can you double check the exat app name? I think it includes "web" at least

.github/workflows/database_backup.yml
```diff
@@ -0,0 +1,28 @@
...
+  #   - cron: "0 3 * * *"
+
+jobs:
+  backup:
+    name: Backup AKS Database (production)
+    if: ${{ github.event_name == 'schedule' || (github.event_name == 'workflow_dispatch' }}
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
```
saliceti, line 17: I'm not sure this is necessary. Can you try without?

AbigailMcP: will give it a go 👍 
this was as per the README: https://github.com/DFE-Digital/github-actions/blob/master/backup-postgres/README.md

.github/workflows/database_backup.yml
AbigailMcP: Good spot 👀

.github/workflows/database_backup.yml
AbigailMcP: Removed!

saliceti (approved):

PR #3006, opened 2024-07-12
[LUPEYALPHA-551] Admins can upload + download eligible FE providers

asmega:
# Context

- Admins can upload a CSV to update eligible FE providers
- Admins can also download a CSV to view the current eligible FE providers

# Changes

- Added to admin layout, `content_for` for back link. ALL the current back links in the admin area are not in the correct position. Once this is merged we can start slowly moving these to the correct position
- Added back link to edit journey configuration screen
- Added class `EligibleFeProvidersImporter` to handle upload of CSV of eligible FE providers. This does all the heavy lifting, we could call this in a job if we need it to be async
- The uploading of eligible FE providers is done synchronously. This feature is behind authentication and authorisation so don't expect any abuse of the feature. The number of records to update at the moment is `288` so not expecting long running SQL queries. And also don't expect the dataset to change very often

# Screenshots

feature in situ
Screenshot 2024-07-16 at 14 11 27

upload success screen
Screenshot 2024-07-16 at 14 13 39

upload error screen
Screenshot 2024-07-16 at 14 12 02

upload error screen 2
Screenshot 2024-07-22 at 13 30 52


spec/features/further_education_payments/happy_js_path_spec.rb
```diff
@@ -18,6 +18,7 @@
     expect(page).to have_content("Which FE provider are you employed by?")
     fill_in "Which FE provider are you employed by?", with: college.name
     within("#claim-provision-search-field__listbox") do
+      sleep(1)
```
asmega, line 21: seems to help with the success rate of this test

app/models/importers/eligible_fe_providers.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  attr_reader :file, :academic_year
+
+  validate :validate_first_row_headers
+
+  def self.headers
+    %w[
+      ukprn
+      max_award_amount
+      lower_award_amount
```
slorek, line 14: I think we only need the first figure. As I understand, the lower amount is always 50% of the higher amount.

We will also need to include additional fields such as contact email in this CSV to facilitate provider verification.

app/models/importers/eligible_fe_providers.rb
```diff
@@ -0,0 +1,55 @@
+require "csv"
+
+class Importers::EligibleFeProviders
```
slorek, line 3: Take a look at the `CsvImporter::Base` class which the other importers inherit from. This was introduced to consolidate several differing implementations so it would be preferable to use this.

db/schema.rb
```diff
@@ -175,6 +175,16 @@
     t.index ["teacher_reference_number"], name: "index_ecp_eligibility_trn"
   end
 
+  create_table "eligible_fe_providers", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.integer "ukprn", null: false
+    t.text "academic_year", null: false
+    t.decimal "max_award_amount", precision: 7, scale: 2
+    t.decimal "lower_award_amount", precision: 7, scale: 2
```
slorek, line 182: I'm not sure we'll need a column for the lower amount as it's always 50% of the higher amount.

spec/features/further_education_payments/happy_js_path_spec.rb
slorek: Try using `expect(page).to have_content("some content")` because Capybara has a built-in wait mechanism which eliminates the need for sleep calls.

slorek (changes requested):
Looks good. The main thing I would check with the design team is the design of the upload/download pages as these are inconsistent with those of the LUP policy, which might cause confusion for users.

spec/features/further_education_payments/happy_js_path_spec.rb
asmega: i'm aware of that but don't think it's the issue. as the line after this
```
find("li", text: college.name).click
```
should wait for the selector then click on it. this suggests that element appears first but the js event has yet to be bound. I've not investigated upstream yet to check if they add the element prior to the event being bound.

if this is the case it would make total sense as the element gets clicked but no js event handler is present so nothing happens instead of all the js that needs to happen

app/models/importers/eligible_fe_providers.rb
asmega: > I think we only need the first figure. As I understand, the lower amount is always 50% of the higher amount.

I have removed `lower_award_amount` and can calculate it when we need it later

> We will also need to include additional fields such as contact email in this CSV to facilitate provider verification.

I don't have any requirements for that at this stage. are we able to kick out of scope for this PR? should be fairly trivial to add after though

app/models/importers/eligible_fe_providers.rb
asmega: The new class now inherits from `CsvImporter::Base`. I've had to make some minor changes. Out of scope of this PR but we should probably namespace this grouping of work. Another nice to have would be to make it `Actvemodel::Errors` compliant which current it is not

db/schema.rb
asmega: I've now removed this column

spec/features/further_education_payments/happy_js_path_spec.rb
asmega: I've had a look upstream but can't determine if there is a delay between when the element is added versus when the event handler is added. However adding the `sleep` clause on this specific line and not other locations seems to drastically improve success rate. which still suggests to me that the click always happens but its whether or not the js is fired

asmega:
> Looks good. The main thing I would check with the design team is the design of the upload/download pages as these are inconsistent with those of the LUP policy, which might cause confusion for users.

I don't have any new designs to refer to for this page, I'm happy to put this ticket on hold if we want to wait for designs.

I'm confident to say the existing design and existing implementation is kludged together with no design thought applied as it already has multiple responsibilities on the page. I've added this new functionality on the new page rather than append it to an already busy page. I'm hoping the other responsibilities could be extracted out into different pages further down the line.

app/models/importers/eligible_fe_providers.rb
slorek: technically that's no problem. Let's check with the stakeholders as changing the CSV specification may cause confusion in testing for example.

slorek:
> I don't have any new designs to refer to for this page, I'm happy to put this ticket on hold if we want to wait for designs.
> 
> I'm confident to say the existing design and existing implementation is kludged together with no design thought applied as it already has multiple responsibilities on the page. I've added this new functionality on the new page rather than append it to an already busy page. I'm hoping the other responsibilities could be extracted out into different pages further down the line.

Absolutely it was because I did it ;) But I think we would want a consistent design, regardless. Let's ask them what they think.

asmega:
> Absolutely it was because I did it ;) But I think we would want a consistent design, regardless. Let's ask them what they think.

I've now updated so the upload and download sections are inline at the bottom of the journey configuration screen

app/controllers/admin/eligible_fe_providers_controller.rb
```diff
@@ -0,0 +1,45 @@
+module Admin
+  class EligibleFeProvidersController < BaseAdminController
+    before_action :ensure_service_operator
+
+    helper_method :journey_configuration
+
+    def create
+      @download_form = EligibleFeProvidersForm.new
```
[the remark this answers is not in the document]
asmega: i'll leave this for now, reads a bit more easily despite it not being needed

slorek (approved):

PR #3010, opened 2024-07-15
FE personal details

kenfodder:
Add all the Personal Details pages to the FE journey.

### DIFF
--- a/app/models/journeys/base_answers_presenter.rb
+++ b/app/models/journeys/base_answers_presenter.rb
@@ -13,7 +13,7 @@ module Journeys
         a << [t("questions.name"), answers.full_name, "personal-details"] if show_name?
         a << [t("forms.address.questions.your_address"), answers.address, "address"]
         a << [t("questions.date_of_birth"), date_of_birth_string, "personal-details"] if show_dob?
-        a << [t("forms.gender.questions.payroll_gender"), t("answers.payroll_gender.#{answers.payroll_gender}"), "gender"]
+        a << payroll_gender
         a << [t("questions.teacher_reference_number"), answers.teacher_reference_number, "teacher-reference-number"] if show_trn?
         a << [t("questions.national_insurance_number"), answers.national_insurance_number, "personal-details"] if show_nino?
         a << [t("questions.email_address"), answers.email_address, "email-address"] unless show_email_select?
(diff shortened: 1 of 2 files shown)


app/views/further_education_payments/claims/check_your_answers.html.erb
```diff
@@ -1,7 +1,34 @@
...
+) %>
 
-<%= form_with model: @form, url: claim_path(current_journey_routing_name), method: :patch, builder: GOVUKDesignSystemFormBuilder::FormBuilder, html: { novalidate: false } do |f| %>
-  <%= f.govuk_submit %>
-<% end %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <% if @form.errors.any? %>
+      <%= render("shared/error_summary", instance: @form) %>
+    <% end %>
```
asmega, line 14: can probably delete this? or are we expecting errors here?

app/views/further_education_payments/claims/information_provided.html.erb
```diff
@@ -0,0 +1,78 @@
...
+      Your financial incentive payment is taxable income, but DfE will pay Income Tax and National Insurance contributions due to HMRC for you.
+      This means the payment amount you receive may be lower following tax deductions.
+    </p>
+
+    <p class="govuk-body">
+      For more details, you can read about payments and deductions for the
+      <%= govuk_link_to "levelling up premium payment", "https://www.gov.uk/guidance/levelling-up-premium-payments-for-fe-teachers", new_tab: true %>.
+    </p>
+
+    <%= button_to "Continue", claim_path(current_journey_routing_name), method: :patch, class: "govuk-button", role: :button, data: { module: "govuk-button" } %>
```
asmega, line 76: not a blocker but you can use the govuk components helper https://govuk-components.netlify.app/helpers/button/#regular-button to remove the extra arguments

asmega (approved):

app/forms/mobile_number_form.rb
```diff
@@ -33,6 +33,11 @@ def save
   private
 
   def send_sms_message
+    if Rails.env.development?
+      Rails.logger.info("SMS CODE: #{OneTimePassword::Generator.new.code}")
```
slorek, line 37: I think this probably won't yield the same value as that sent in the Notify SMS message below. I'm also not sure how we're storing the OTP code to check it's correct on the next 2FA page?

PR #3021, opened 2024-07-19
[LUPEYALPHA-653] FE award amount

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-653
- Populate FE award amount placeholder with actual award amount value

# Changes

- Also fixed a slug sequence issue from a previous rebasing mistake
- Also moved superfluous placeholder page to the end of journey for now

### DIFF
--- a/app/models/journeys/further_education_payments/session_answers.rb
+++ b/app/models/journeys/further_education_payments/session_answers.rb
@@ -119,6 +119,17 @@ module Journeys
       def less_than_half_hours_teaching_eligible_courses?
         hours_teaching_eligible_subjects == false
       end
+
+      def award_amount
+        case teaching_hours_per_week
+        when "more_than_12"
+          school.eligible_fe_provider.max_award_amount
+        when "between_2_5_and_12"
+          school.eligible_fe_provider.lower_award_amount
+        else
+          0
+        end
+      end
     end
   end
 end
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -143,6 +143,10 @@ class School < ApplicationRecord
     sql
   end
 
+  def eligible_fe_provider(academic_year: AcademicYear.current)
+    EligibleFeProvider.find_by(ukprn:, academic_year:)
+  end
+
   def address
     [street, locality, town, county, postcode].reject(&:blank?).join(", ")
   end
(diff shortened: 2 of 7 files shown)


app/models/journeys/further_education_payments/session_answers.rb
```diff
@@ -96,6 +96,17 @@ def fe_provider_closed?
 
         school.closed?
       end
+
+      def award_amount
+        case teaching_hours_per_week
+        when "more_than_12"
+          school.eligible_fe_provider.max_award_amount
+        when "between_2_5_and_12"
+          school.eligible_fe_provider.max_award_amount / 2
```
slorek, line 105: Do we need to round this to either a whole number or 2 decimals? I realise the view helper will do this for display purposes, but we might want to implement it here instead for safety.

app/models/school.rb
```diff
@@ -116,6 +116,8 @@ class School < ApplicationRecord
   enum school_type_group: SCHOOL_TYPE_GROUPS
   enum school_type: SCHOOL_TYPES
 
+  has_one :eligible_fe_provider, foreign_key: :ukprn, primary_key: :ukprn
```
slorek, line 119: We'll need a database index on the `eligible_fe_providers` table for `ukprn` column only. At the moment it's only a secondary column in another index.

asmega: it's actually only used in the test for convenience. although now that you've mentioned it and i've given it some thought i think i'll remove this association and code by hand in the tests. i don't think we need the index as we should never look up just by `ukprn` and should always in tandem with `academic_year`

asmega: no, actually there is bug as it is actually used in the code and will fix

asmega: this is now an instance method and defaults to use `AcademicYear.current`, which should be fine for now. If there needs to be any processing in another academic year we may want to store the academic year as part of their answers when they start their journey

app/models/journeys/further_education_payments/session_answers.rb
asmega: i think if we just took the CSV uploaded values instead of dividing by 2 we could have avoided this scenario in the first place. I feel now we've simply created more work and complexity for ourselves. my recommendation would be to take the static values that are fed to us rather than having to write any code. otherwise i now have heaps of questions:
- what are the rounding rules?
- do we always display pence even when zero pence?
- are the rounding rules merely for display purposes? or also used for award amount?

slorek: As discussed, we will supply the lower amount in the CSV file.

slorek (approved):

PR #3023, opened 2024-07-22
CAPT-1766 one login bypass

alkesh:
The env var `BYPASS_ONELOGIN_SIGN_IN` is set to`true`,
which means the user will be logged in with email: `user@example.com` and name: `TEST USER`.
(note: the name from One Login currently comes back all capitalised).

This branch is built on top of the changes from https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/3001


slorek (approved):

slorek (commented):
If this is to be merged we will first need to move the sign in slug to the correct position in the sequence.

asmega (approved):

PR #3026, opened 2024-07-24
[LUPEYALPHA-760] EY scaffolding + landing page

AbigailMcP:
## Context

- We are adding a new journey and policy to the service: 'Claim an early years financial incentive payment'.
- This PR contains the bare-bones models for the provider journey only.
- In the future, there will also be a practitioner journey which will support the same claim submitted via this provider journey.

## Changes in this PR

- Adds models (journey, eligibility, session, session_answers, slug_sequence, policy) for
 the new `EarlyYearsPayment::Provider` journey.
- Adds basic happy path spec that visits the landing page.
- Adds basic view and content for landing page as per the prototype
- Includes the new journey config in the seeds
- Adds migration to create early_years_payment_eligibilities
- Hides the start button on the landing page when the journey is closed via the admin.

## Out of scope
- Whole documents section
- Links to consent form and privacy notice

## Guidance to review

How do we feel about the journey name `Journeys::EarlyYearsPayment::Provider`?
What about the journey name as it appears on the admin 'Manage service' page?

Prototype: https://claim-additi-main-9yrds6vyxcsd.herokuapp.com/landing-page

- Head to `/early-years-payment-provider/landing-page` and compare against the prototype
- Head to `/admin/journey_configurations`, turn the journey off
- Head back to the landing page and check the 'Start now' button has disappeared

Screenshot 2024-07-26 at 11 19 37

Screenshot 2024-07-26 at 11 20 00

Screenshot 2024-07-26 at 10 52 50


slorek (commented):
Note there will be two journeys supporting this policy, so we may consider naming this journey clearly as the Provider journey.

app/views/early_years_payment/provider/landing_page.html.erb
```diff
@@ -0,0 +1,139 @@
...
+      "zero-hour contracts",
+      "guaranteed minimum hour contracts",
+      "self-employed or freelance contracts",
+      "fixed term contracts (FTCs)",
+      "agency work and apprenticeship jobs"
+    ], type: :bullet %>
+
+    <p class="govuk-body">
+      You can find more information about employment contracts on the
+      <%= govuk_link_to("ACAS website", "#") %>.
```
AbigailMcP, line 70: @cwadsworthdfe - could you let me know where this should link to?

cwadsworthdfe: Its this one: 
https://www.acas.org.uk/employment-contracts

cwadsworthdfe:
I've reviewed this and happy to approve. 

- Landing page replicates the prototype
- Service can be turned on and off and start now button behaviour is correct
- Journey name split is clear too in manage services

Thanks! @AbigailMcP

rjlynch (approved):
This all looks good to me 👍

app/models/journeys/early_years_payment/provider/slug_sequence.rb
```diff
@@ -0,0 +1,29 @@
+module Journeys
+  module EarlyYearsPayment
+    module Provider
+      class SlugSequence
+        SLUGS = %w[].freeze
+
+        def self.start_page_url
+          if Rails.env.production?
+            "https://www.example.com" # TODO: update to correct guidance
```
asmega, line 9: For FE i ended up removing this. from what I could tell, stakeholders wanted the landing page to be the start page

app/views/early_years_payment/provider/landing_page.html.erb
AbigailMcP: thanks! updated

app/models/journeys/early_years_payment/provider/slug_sequence.rb
AbigailMcP: ahh thank you! i think that's the same for EY - i'll remove too 👍

PR #3027, opened 2024-07-24
[LUPEYALPHA-778] FE journey: add ineligible state when under hours teaching courses

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-778
- Add ineligible state for when teaching under half hours on eligible courses

# Changes

- Tweaked `Form#t` to now accept an array as a key, this just get converted internally to `whatever.the.key.is` which makes the api a bit nicer
- Moved some code from eligibility checker to session answers, which is where i think it belongs better
- Use i18n for the ineligible screens. Could probably extract to reusable wrapper partial but not really much gain for the effort

### DIFF
--- a/app/forms/form_helpers.rb
+++ b/app/forms/form_helpers.rb
@@ -9,9 +9,16 @@ module FormHelpers
   end
 
   def t(key, args = {})
+    key_string = case key
+    when Array
+      key.join(".")
+    else
+      key
+    end
+
     i18n_form_namespace_dup = args.delete(:i18n_form_namespace) || i18n_form_namespace
 
-    base_key = :"forms.#{i18n_form_namespace_dup}.#{key}"
+    base_key = :"forms.#{i18n_form_namespace_dup}.#{key_string}"
     I18n.t("#{i18n_namespace}.#{base_key}", default: base_key, **args)
   end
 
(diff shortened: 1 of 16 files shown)


app/models/journeys/further_education_payments/session_answers.rb
```diff
@@ -105,6 +105,20 @@ def fe_provider_closed?
...
+        groups = subjects_taught.reject { |e| e == "none" }
+
+        return if groups.empty?
+
+        groups.all? do |subject|
+          public_send(:"#{subject}_courses").include?("none")
+        end
+      end
+
+      def less_than_half_hours_teaching_eligible_courses?
```
rjlynch, line 119: 👍

app/forms/form_helpers.rb
```diff
@@ -9,9 +9,16 @@ def i18n_errors_path(key, args = {})
...
+    when Array
+      key.join(".")
+    else
+      key
+    end
+
     i18n_form_namespace_dup = args.delete(:i18n_form_namespace) || i18n_form_namespace
 
-    base_key = :"forms.#{i18n_form_namespace_dup}.#{key}"
+    base_key = :"forms.#{i18n_form_namespace_dup}.#{key_string}"
```
rjlynch, line 21: Nice 👍

rjlynch (approved):
Looks good 👍

PR #3031, opened 2024-07-24
CAPT-1765 - Allow admins to update the address on claims

rjlynch:
It is expected that many IRP claims will need the address changing so we
want to update the admin interface to allow admins to do this themselves.

Screenshot 2024-07-24 at 16 42 44


asmega (approved):
Not sure why this form is so differently styled and away from govuk style guide. it would be great to get it back into the correct shape

PR #3035, opened 2024-07-26
CAPT-1773 remove 30min timeout

alkesh:
https://dfedigital.atlassian.net.mcas.ms/browse/CAPT-1773
`ClaimSessionTimeout` concern is no more.
Here is a breakdown of the `session` entries it used to delete:
* `claim_postcode` and `claim_address_line_1` were used in the view `app/views/claims/no_address_found.html.erb` which was never actually used
* `no_address_selected` - not used
* `reminder_id` - has been moved into journey session answers
* `slugs` - remains in the session, cleared by `ClaimsController` when the user changes to a different journey, or for new claims. Also cleared by `SubmissionsController` when the claim is submitted.
* `bank_validation_attempt_count` - not used
* `user_info` - not used
* `tps_school_id`, `tps_school_name` and `tps_school_address` - not used

Also `session[:hmrc_validation_attempt_count]` has been moved into journey session answers.


app/views/claims/no_address_found.html.erb
alkesh, line 1: this view was never used anywhere

app/controllers/concerns/claim_session_timeout.rb
```diff
@@ -1,25 +1,6 @@
-module ClaimSessionTimeout
-  CLAIM_TIMEOUT_LENGTH_IN_MINUTES = 30
-
+module ClaimSessionTimeout # TODO: rename this class - SessionManager?
   def clear_claim_session
-    session.delete(:claim_postcode)
-    session.delete(:claim_address_line_1)
```
alkesh, line 6: these were only ever used by the `app/views/claims/no_address_found.html.erb` template, which was never called

app/views/static_pages/cookies.html.erb
```diff
@@ -130,7 +130,7 @@
         <tr class="govuk-table__row">
           <td class="govuk-table__cell">_dfe_teachers_payment_service_session</td>
```
alkesh, line 131: hmm - this seems to be the wrong cookie name - should be `_claim_session` - is there another ticket to address this?

spec/features/changing_answers_spec.rb
```diff
@@ -64,7 +64,6 @@
     expect(find("#claim_qts_award_year_on_or_after_cut_off_date").checked?).to eq(true)
 
     choose_qts_year :before_cut_off_date
-    click_on "Continue"
```
alkesh, line 67: This was actually clicking on the "Continue session" button in the timeout banner at the top of the page.
This `click_on` is not needed, because the `choose_qts_year` helper does it's own `click_on "Continue"`

rjlynch (approved):
Looks good. Left one question about whether or not we can remove the sessions controller

app/controllers/sessions_controller.rb
```diff
@@ -1,9 +1,5 @@
 class SessionsController < BasePublicController
-  skip_before_action :end_expired_claim_sessions
-
   def refresh
-    clear_claim_session if claim_session_timed_out?
```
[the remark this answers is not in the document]
alkesh: This controller is still used by the admin site: 
https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/master/config/routes.rb#L116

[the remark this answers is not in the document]
rjlynch: Is that not hitting the admin sessions controller?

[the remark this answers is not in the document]
alkesh: ooh yeah, you're right.
I'll remove this controller.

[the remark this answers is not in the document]
asmega: theres a javascript thing that happens?
```
app/assets/javascripts/components/timeout_dialog.js
```
is that calling this?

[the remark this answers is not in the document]
alkesh: yes - I've removed the line in the main application layout that rendered that:
https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/3035/files#diff-f43fe075643e681b2c01c2f853bb0c4299d135b47fcbd4da96890d521c49e3ebL45

That dialog is still used and rendered in the admin layout.

app/views/static_pages/cookies.html.erb
asmega: if not, defo raise one

app/controllers/concerns/session_management.rb
```diff
@@ -0,0 +1,6 @@
+module SessionManagement
+  def clear_claim_session
+    session.delete(:slugs)
```
asmega, line 3: something i noticed too, i think we want to get rid of this in future when we get a moment as we currently track all previous slugs and instead should be able to work out where we came from

app/views/static_pages/cookies.html.erb
alkesh: raised: https://dfedigital.atlassian.net/browse/CAPT-1774

asmega (approved):

PR #3038, opened 2024-07-29
ignore session answer attributes in the stored JSON that have not been defined on the SessionAnswers class

alkesh:
Before this change, an `UnknownAttributeError` is raised on any page in the journey, hence breaking the app totally, in the following scenarios:
* we've added an attribute to `Journeys::SessionAnswers`, deployed it, but then want to roll back that deployment
* we want to remove an attribute from `Journeys::SessionAnswers` and deploy that
* for a developer running locally, and having completed journeys locally, switching between branches that have added new attributes

There is a workaround that doesn't involve this change - and that is to purge all journey sessions.
This change would mean that is no longer necessary.

### DIFF
--- a/app/models/journeys/session_answers_type.rb
+++ b/app/models/journeys/session_answers_type.rb
@@ -12,7 +12,10 @@ module Journeys
         rescue
           nil
         end
-        answers_class.new(decoded) unless decoded.nil?
+        unless decoded.nil?
+          decoded_with_allowed_attributes = decoded.slice(*answers_class.attribute_names)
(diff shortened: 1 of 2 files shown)


app/models/journeys/session_answers_type.rb
```diff
@@ -12,7 +12,10 @@ def cast_value(value)
         rescue
           nil
         end
-        answers_class.new(decoded) unless decoded.nil?
+        unless decoded.nil?
```
rjlynch, line 15: 👍 nice

rjlynch (approved):
Nice 👍

PR #3044, opened 2024-07-30
[LUPEYALPHA-798] Update copy for FE select provision

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-798
- Update copy for FE select provision page

### DIFF
--- a/app/views/further_education_payments/claims/select_provision.html.erb
+++ b/app/views/further_education_payments/claims/select_provision.html.erb
@@ -11,7 +11,7 @@
           tag: "h1",
           size: "l"
         },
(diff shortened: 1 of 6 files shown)


app/views/further_education_payments/claims/select_provision.html.erb
```diff
@@ -11,7 +11,7 @@
           tag: "h1",
           size: "l"
         },
-        hint: { text: "Select your college from the search results." } %>
+        hint: { text: @form.t("hint") } %>
```
rjlynch, line 14: 👍

rjlynch (approved):
👍

PR #3049, opened 2024-07-31
LUPEYALPHA-786 - add courses to check your answers part one

kenfodder:
* Displays all the courses selected
* Hyperlinks in the courses descriptions are not shown like they are in the initial radio buttons or course list playback 
* If "I do not teach any of these courses" was selected for a course field, that will show the question, to give opportunity to go back and change
* If a subject/course field was NOT selected meaning the question was NOT asked the question is NOT shown as there is no page to go back to


rjlynch (approved):
LGTM 👍 

Probably not worth the effort to implement but it seems like a "course" could be it's own object with an id, description, and a url attribute 🤔

PR #3054, opened 2024-08-02
LUPEYALPHA-762 enter email address

alkesh:
The first 2 pages, where you enter your email address and an email is sent, are now part of a separate journey (`Journeys::EarlyYearsPayment::Start`).

Then, a magic link is created (e.g. `/early-years-payment-provider/claim?code=123456&email=test@example.com`)
Clicking on that link will start the actual provider journey (`Journeys::EarlyYearsPayment::Provider`)

Things not implemented in this PR:
* the 'send another link' link will be done in a future ticket (https://dfedigital.atlassian.net/browse/LUPEYALPHA-842)
* only allowing access to provider journey if the user has come via the magic link, this will be done in https://dfedigital.atlassian.net/jira/software/projects/LUPEYALPHA/issues/LUPEYALPHA-790


terraform/application/application.tf
```diff
@@ -16,6 +16,7 @@ module "application_configuration" {
       ENVIRONMENT_NAME   = var.environment
       PGSSLMODE          = local.postgres_ssl_mode
       CANONICAL_HOSTNAME = local.canonical_hostname
+      APP_BASE_URL       = local.app_base_url
```
alkesh, line 19: Is this a good way to set the URL base to use for the magic link?
If it is, then a future optimisation would be to replace `DFE_SIGN_IN_REDIRECT_BASE_URL`, `TID_BASE_URL` and `ONELOGIN_REDIRECT_BASE_URL` with this `APP_BASE_URL`.

spec/factories/claims.rb
```diff
@@ -17,7 +17,7 @@
       journey = Journeys.for_policy(evaluator.policy)
 
       begin
-        Journeys.for_policy(evaluator.policy)&.configuration.present?
+        raise ActiveRecord::RecordNotFound unless Journeys.for_policy(evaluator.policy)&.configuration.present?
```
alkesh, line 20: bug with the existing factory - it is possible to have `.present` return false - previously no exception was raised, so no journey_configuration gets created

app/controllers/claims_controller.rb
```diff
@@ -79,9 +79,19 @@ def check_page_is_in_sequence
 
     raise ActionController::RoutingError.new("Not Found for #{params[:slug]}") unless page_sequence.in_sequence?(params[:slug])
 
+    handle_magic_link if page_sequence.magic_link?
     redirect_to claim_path(current_journey_routing_name, next_required_slug) unless page_sequence.has_completed_journey_until?(params[:slug])
   end
 
+  def handle_magic_link
+    otp = OneTimePassword::Validator.new(params[:code], answers.sent_one_time_password_at)
+    if otp.valid? # && EligibleEyProvider.eligible_email?(journey_session.answers.email_address) # TODO eligible check once model merged: https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/3050
```
alkesh, line 88: is it ok to handle the eligibility check like this? 
we only have magic links for this scenario currently - can be moved to something more generic if we have a magic link for another scenario.

.env.development
```diff
@@ -24,3 +26,5 @@ TID_SIGN_IN_CLIENT_ID=claim
 
 ONELOGIN_SIGN_IN_ISSUER=https://oidc.integration.account.gov.uk/
 ONELOGIN_REDIRECT_BASE_URL=http://localhost:3000
+
+EY_MAGIC_LINK_SECRET=some-secret-for-ey-magic-link
```
asmega, line 30: given our current `SECRET` is a constant that is random does this mean every time the app reboots OTPs used elsewhere are no longer valid?

db/seeds.rb
```diff
@@ -12,6 +12,7 @@
   Journeys::Configuration.create!(routing_name: Journeys::GetATeacherRelocationPayment::ROUTING_NAME, current_academic_year: AcademicYear.current)
   Journeys::Configuration.create!(routing_name: Journeys::FurtherEducationPayments::ROUTING_NAME, current_academic_year: AcademicYear.current)
   Journeys::Configuration.create!(routing_name: Journeys::EarlyYearsPayment::Provider::ROUTING_NAME, current_academic_year: AcademicYear.current)
+  Journeys::Configuration.create!(routing_name: Journeys::EarlyYearsPayment::Start::ROUTING_NAME, current_academic_year: AcademicYear.current)
```
asmega, line 15: can we rename `Start` to `Claimant`? I don't think people will refer to it as the "start" journey and more likely call it the "claimant" journey

app/views/early_years_payment/start/claims/email_address.html.erb
```diff
@@ -0,0 +1,27 @@
...
+
+      <h1 class="govuk-heading-l">
+        Enter your email address
+      </h1>
+
+      <p class="govuk-body">
+        You must use the email address previously provided to your local authority to access this service.
+      </p>
+
+      <%= f.govuk_text_field :email_address,
```
asmega, line 18: we're not using it elsewhere but since we are now using govuk form builder maybe we can start using `govuk_email_field`

app/views/early_years_payment/start/claims/check_your_email.html.erb
```diff
@@ -0,0 +1,35 @@
...
+      <h2 class="govuk-notification-banner__title" id="govuk-notification-banner-title">
+        Success
+      </h2>
+    </div>
+    <div class="govuk-notification-banner__content">
+      <h3 class="govuk-notification-banner__heading">
+        Email sent to <%= journey_session.answers.email_address %>
+      </h3>
+    </div>
+  </div>
```
asmega, line 15: theres a component which should reduce the amount of markup you need see https://govuk-components.netlify.app/components/notification-banner/#notification-banner-with-success

asmega: i'd probably also move this one line down so its nested in `govuk-grid-column-two-thirds` as typically all the content is two thirds, unless its been explicitly designed for full width?

asmega:
there seems to be an issue with the review app? https://claim-additional-payments-for-teaching-review-3054-web.test.teacherservices.cloud/early-years-payment-provider/landing-page it seems to be throwing an error.

app/views/early_years_payment/start/claims/check_your_email.html.erb
alkesh: I'm basing the banner on the prototype (https://dfedigital.atlassian.net.mcas.ms/browse/LUPEYALPHA-762) which has it full width.

app/models/policies/early_years_payments.rb
```diff
@@ -1,5 +1,5 @@
 module Policies
-  module EarlyYearsPayment
+  module EarlyYearsPayments
```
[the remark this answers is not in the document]
alkesh: all policies are plural, so I've made it plural for consistency.
Maybe the journey should be plural, similar to FE.

db/seeds.rb
alkesh: even so, I don't think `Start` is the best name - it's all I could come up with - there must be something better

slorek (approved):

PR #3059, opened 2024-08-06
support validating OTP without a generated_at time and with a specified secret

alkesh:
We need 2 changes to the OTP code for the EY magic link:
* allow OTP code to be validated without having a `generated_at` time.
* allow OTP codes to be generated and verified using a specified secret (because we now are verifying codes generated in a separate session)

### DIFF
--- a/lib/one_time_password/base.rb
+++ b/lib/one_time_password/base.rb
@@ -12,6 +12,10 @@ module OneTimePassword
 
     private
 
-    attr_reader :issuer
+    attr_reader :issuer, :secret
+
+    def encode_secret(secret)
+      ROTP::Base32.encode(secret) if secret
+    end
   end
 end
--- a/lib/one_time_password/validator.rb
+++ b/lib/one_time_password/validator.rb
@@ -1,19 +1,20 @@
 module OneTimePassword
   class Validator < Base
-    def initialize(code, generated_at = nil)
+    def initialize(code, generated_at = nil, secret: nil)
       @code = code
       @generated_at = generated_at
+      @secret = encode_secret(secret) || SECRET
     end
 
     def valid?
-      code.present? && !wrong_length? && (generated_at && !expired?) && !incorrect?
+      code.present? && !wrong_length? && !expired? && !incorrect?
     end
 
     def warning
       return "Enter a passcode" if code.blank?
       return "Enter a valid passcode containing #{LENGTH} digits" if wrong_length?
-      return "Your passcode has expired, request a new one" if generated_at && expired?
-      return "Your passcode is not valid or has expired" if !generated_at
+      return "Your passcode has expired, request a new one" if expired?
+      return "Your passcode is not valid or has expired" if !generated_at && incorrect?
       "Enter a valid passcode" if incorrect?
     end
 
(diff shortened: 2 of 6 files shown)


lib/one_time_password/validator.rb
```diff
@@ -1,41 +1,36 @@
...
     end
 
     private
 
     attr_reader :code, :generated_at
 
     def wrong_length?
-      return @wrong_length if defined?(@wrong_length)
-
-      @wrong_length = code.gsub(/\D/, "").length != LENGTH
```
alkesh, line 26: I'm not sure what this was ever doing - ignoring non-digits when checking for length doesn't seem useful.
e.g. code "1234A6B" would have the correct length in this case, but would still not be a valid code (it still fails the `ROTP:TOTP#verify`)

lib/one_time_password/validator.rb
```diff
@@ -21,21 +22,15 @@ def warning
...
+      @wrong_length ||= code.length != LENGTH
     end
 
     def expired?
-      return @expired if defined?(@expired)
-
-      @expired = generated_at < DRIFT.seconds.ago
+      @expired ||= generated_at < DRIFT.seconds.ago
```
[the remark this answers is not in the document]
alkesh: I think using memoization results in the same behaviour in these scenarios.
The only time the `defined?` approach results in different behaviour, is when the value could be `nil`, and you want to cache that `nil` value.
In this scenario, `<` never returns `nil`. Same for #incorrect? - `nil?` never returns nil.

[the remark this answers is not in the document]
alkesh: hmm - you're right - I thought a value of `false` would be cached, but it's not.

lib/one_time_password/validator.rb
```diff
@@ -1,18 +1,19 @@
...
   class Validator < Base
-    def initialize(code, generated_at)
+    def initialize(code, generated_at = nil)
       @code = code
       @generated_at = generated_at
     end
 
     def valid?
-      code.present? && !wrong_length? && !expired? && !incorrect?
+      code.present? && !wrong_length? && (generated_at && !expired?) && !incorrect?
```
asmega, line 9: - if `generated_at` is `nil` won't `valid?` always return `false`?
- in which case why is `generated_at` now an optional parameter?
- also i don't see how brackets makes any difference in a chain of `&&` statements?

alkesh: yeah, you're right. I've moved the check for a nil `generated_at` into the `expired?` method - so this works properly now.

spec/lib/one_time_password/validator_spec.rb
```diff
@@ -0,0 +1,117 @@
+require "rails_helper"
+
+RSpec.describe OneTimePassword::Validator do
+  subject { described_class.new(one_time_passcode, generated_at) }
+
+  let!(:one_time_passcode) { OneTimePassword::Generator.new.code }
+  let!(:generated_at) { Time.now }
```
asmega, line 7: not blocking but don't see why these can't be lazy ie `let` instead of `let!`

alkesh: because I have a spec that uses `travel` to move 20 minutes into the future, to test for passcode expiry.

lib/one_time_password/base.rb
```diff
@@ -12,6 +12,11 @@ def rotp
 
     private
 
-    attr_reader :issuer
+    attr_reader :issuer, :secret
+
+    def encode_secret(secret)
+      return ROTP::Base32.encode(secret) if secret
+      SECRET
+    end
```
asmega, line 20: not blocking and probably a stylistic thing. i think i would of made this return nil if `secret` is nil. then the caller can do the following
```ruby
@secret = encode_secret(secret) || SECRET
```

asmega (approved):

lib/one_time_password/base.rb
alkesh: 👍🏽 done

PR #3062, opened 2024-08-06
Add teaching hours task

rjlynch:
IRP requires we check with the school how much time the claimant spends
teaching specific subjects. We don't ask this question in the
eligibility, so in the admin presenter we reshow the teaching subject
for this task.

### DIFF
--- a/app/models/claim_checking_tasks.rb
+++ b/app/models/claim_checking_tasks.rb
@@ -29,6 +29,7 @@ class ClaimCheckingTasks
         task_names.delete("employment_contract")
         task_names.delete("employment_start")
         task_names.delete("subject")
+        task_names.delete("teaching_hours")
       end
     end
   end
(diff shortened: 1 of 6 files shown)


asmega (approved):

app/models/claim_checking_tasks.rb
```diff
@@ -29,6 +29,7 @@ def applicable_task_names
         task_names.delete("employment_contract")
         task_names.delete("employment_start")
         task_names.delete("subject")
+        task_names.delete("teaching_hours")
```
asmega, line 32: not related to your change but think this class could do with some changes as a whole. be great if we could delegate these to some other classes and switchover from a blacklisting to a whitelist approach, i think would make this so much more readable

rjlynch: Yeah I agree, I think these should be moved into the policy as they're policy specific, and would match how we handle admin claim rejection reasons

PR #3072, opened 2024-08-07
Add maintenance window for DfE Analytics

slorek:
Per @ericaporter:

> Hi Steve, would you be able to help me set a maintenance window on Claim so that we can do some CMEK encrypting of the Big Query production events table? This is part of the ongoing BQ assurance work, we were able to encrypt all other Claims tables week before last, but we want to stop events streaming to the production tables before we do those.


rjlynch (approved):
LGTM left an option to make sure this line gets removed but I think a FIXME + ticket would work the same!

slorek:
> LGTM left an option to make sure this line gets removed but I think a FIXME + ticket would work the same!

Good thinking.

PR #3073, opened 2024-08-07
LUPEYALPHA-818 - Update computing course options

kenfodder:
Screenshot 2024-08-07 at 16 56 10

Screenshot 2024-08-07 at 16 56 29

### DIFF
--- a/app/forms/journeys/further_education_payments/computing_courses_form.rb
+++ b/app/forms/journeys/further_education_payments/computing_courses_form.rb
@@ -42,8 +42,8 @@ module Journeys
             name: course_option_description("tlevel_digitalproduction")
           ),
           OpenStruct.new(
-            id: "ibo_level3",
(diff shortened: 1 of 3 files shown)


app/forms/journeys/further_education_payments/computing_courses_form.rb
```diff
@@ -42,8 +42,8 @@ def checkbox_options
             name: course_option_description("tlevel_digitalproduction")
           ),
           OpenStruct.new(
-            id: "ib_certificate_compsci",
-            name: course_option_description("ib_certificate_compsci")
+            id: "ibo_level3",
```
asmega, line 45: i think its worth suffixing `_compsci` or something similar to that respect, otherwise downstream for reporting and data consumers may find it harder to figure out what was selected

asmega (approved):

PR #3074, opened 2024-08-07
Authorisation spike 2

rjlynch:
# An alternative approach from [this pr](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/3070) for third parties to verify claims

(this is just a spike, I think some of the name spacing could be improved, and I'm not sold on "third parties" as the name for external approvers, and there's a bunch of shared code we could clean up.

## Reasoning
Third parties verifying claims feels quite a lot like the admin approving claims feature. We present a page of questions about an existing claim to a user, who we need to authenticate.
This pr proposes an new concept of a "third party verifier" to approve claims for journeys that require it.

We introduce three new controllers
* third_parties/claims/verifications - for handling the actual approval of a claim
* third_parties/sessions - for authenticating the third party
* third_parties/authorisation_failures - for displaying the reason authentication failed

Each journey will define it's own logic for
* verification views
* authentication views
* authorisation failure views
* serialising / deserialising the third party "user" from the session
* verification form
* authorisation logic

## Notes for review
Best to start reviewing in the third_parties/claims/verifications_controller#new action as that's where the journey kicks off. The two tests in this pr `spec/features/further_education_payments/provider/provider_verifying_claims_spec.rb` pass so you can pull this code down and drop some `binding.pry`s in to get a feel for how this works.

## Happy path
* user visits the link to verify a form
* they land on the Claims::VerificationsController
* they're not signed in so they're kicked to the session controller's new action
* for FE SessionController#new renders a form to post to DfE Sign in
* user is redirected to SessionController#callback
* we capture details from oauth into the session
* we redirect the user back to the Claims::VerificationsController
* we check the user's authorisation
* we render the form for them to approve the claim

## Notes
We've namespaced verifications under claims as there needs to be a claim for them to verify, sessions aren't namespaces under claims as they're not specific to verifying a claim, and we may in future want to support other third party use cases, eg a controller that shows claims awaiting verification. A bunch of shared logic around what journey we're on can be pulled out into a shared controller. We should name space the models better eg `Journeys::FurtherEducationPayments::ThirdParties::Claims::VerificationForm` rather than `Journeys::FurtherEducationPayments:: ThirdPartyVerificationForm` (I was doing this quickly!)

*Note* we're relying on the session being tamper proof!

### DIFF
--- a/app/models/journeys/further_education_payments/third_party_authorisation.rb
+++ b/app/models/journeys/further_education_payments/third_party_authorisation.rb
@@ -1,0 +1,43 @@
+module Journeys
+  module FurtherEducationPayments
+    class ThirdPartyAuthorisation
+      def initialize(user:, record:)
+        @user = user
+        @claim = record
+      end
+
+      def authorised?
+        failure_reason.nil?
+      end
+
+      def failure_reason
+        return :no_service_access unless service_access?
+        return :organisation_mismatch unless organisation_matches?
+        nil
+      end
+
+      private
+
+      attr_reader :user, :claim
+
+      def service_access?
+        dfe_sign_in_user.has_role?("claim-verifier")
+      end
+
+      def organisation_matches?
+        # Temp early return until fe claims have a school
+        return true
+
+        claim.school.ukprn == dfe_sign_in_user.organisation_ukprn
+      end
+
+      def dfe_sign_in_user
+        @dfe_sign_in_user ||= DfeSignIn::Api::User.new(
+          organisation_id: user.organisation_id,
+          organisation_ukprn: user.organisation_ukprn,
+          user_id: user.uid
+        )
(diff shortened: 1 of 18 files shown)


app/models/journeys/further_education_payments/third_party_authorisation.rb
```diff
@@ -0,0 +1,43 @@
...
+        return :organisation_mismatch unless organisation_matches?
+        nil
+      end
+
+      private
+
+      attr_reader :user, :claim
+
+      def service_access?
+        dfe_sign_in_user.has_role?("claim-verifier")
```
rjlynch, line 24: prob should do this once in the callback and cache it in the session, we don't want to make an api to DSI to get this info on every request.

PR #3076, opened 2024-08-08
[LUPEYALPHA-597] One login header + sign out

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-597
- Add one login header and implement sign out functionality within banner 

# Changes

- Added one login header as documented with their css and js
- js is used by mobile for the "burger menu" expansion and collapsing
- i've swapped the default sign out link for a button backed by a form `DELETE` which should help prevent xss or a simple redirect from causing the user to deauth
- i had to add some custom css to in order to keep this form button styled as it was the link
- when handing over to one login to sign out we send over a `state` param which is sent back to us. this is currently set to the current journey name so we can bounce them back to relevant start page. this means that for other journeys we can send to relevant start pages after signing out. if we need even more customisation i recommend delegating to relevant journey class to handle specific behaviour
- in order to be able to sign out of one login as part of the redirect we must also send `id_token_hint` param which is essentially the `id_token` given to us in the auth callback. therefore when we do the initial auth, the auth credentials are now stored in the journey session
- i have refactored the original one login omniauth callback code. we no longer surface any params in the redirects which expose the auth state. instead these are now handled directly in the callback and nothing is sent back to the client (browser)

# Review notes

- one login is not enabled for review apps, therefore after the one login flow in review apps you're not actually logged into one login and this is faked by us
- as a result clicking on `Sign out` will cause an error as you attempt to sign out of one login which you were never signed into in the first place

# Screenshots

desktop
Screenshot 2024-08-08 at 12 51 41

mobile
Screenshot 2024-08-08 at 12 52 00

mobile - expanded via js
Screenshot 2024-08-08 at 12 52 15


app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -54,35 +54,38 @@ def current_journey_routing_name
...
 
   def process_one_login_identity_verification_callback(core_identity_jwt)
     first_name, surname = extract_name_from_jwt(core_identity_jwt)
+
+    journey_session.answers.assign_attributes(
+      identity_confirmed_with_onelogin: true
+    )
+    journey_session.answers.first_name ||= first_name
+    journey_session.answers.surname ||= surname
+    journey_session.save!
```
rjlynch, line 83: Not a blocker but one thing we might want to consider is a form object for handling mapping the auth hash to the journey session

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -54,35 +54,38 @@ def current_journey_routing_name
...
 
   def process_one_login_authentication_callback(auth)
-    onelogin_user_info_attributes = auth.info.to_h.slice(
-      *SignInForm::OneloginUserInfoForm::ONELOGIN_USER_INFO_ATTRIBUTES.map(&:to_s)
-    )
+    onelogin_user_info = auth.info.to_h.slice("email", "phone")
+    onelogin_credentials = auth.credentials
 
-    journey_session.answers.assign_attributes(onelogin_user_info: onelogin_user_info_attributes)
+    journey_session.answers.assign_attributes(
```
rjlynch, line 60: We really need to implement a `journey_session.answers.update!` method that does this two step process for us 😂

app/assets/stylesheets/components/button_to.scss
```diff
@@ -0,0 +1,20 @@
...
+  line-height: inherit;
+}
+
+.button-to-as-link:hover {
+  cursor: pointer;
+
+  text-decoration: underline;
+  text-decoration-thickness: 3px;
+  text-underline-offset: 0.1578em;
+}
```
rjlynch, line 20: 👍

config/routes.rb
```diff
@@ -10,6 +10,9 @@
     post "/auth/onelogin_identity", to: "omniauth_callbacks#onelogin"
   end
 
+  delete "deauth/onelogin", to: "deauth#onelogin"
+  get "deauth/onelogin/callback", to: "deauth#onelogin_callback"
```
rjlynch, line 14: Not a blocker / out of scope but if these are nested under the `journey` we get access to the `journey` param in the callback so we know where they came from without relying on the session, which will let us clean up some stuff in the claims controller. Again not a blocker!

rjlynch (approved):
LGTM 👍 - Didn't leave any blocking comments just some thoughts on our omni auth stuff, I think these changes look good.

Out of scope but in future we might want to pull all the one login stuff into an `OneLogin::SessionsController` so we can have the login / logout behaviour in one place (`OneLogin::SessionsController#create` and `OneLogin::SessionsController#destroy` for auth / deauth respectively)

config/routes.rb
asmega: for one login deauth we don't rely on the session. the session is immediately deleted after we redirect them to one login logout flow. we send one login a `state` param which we receive back which we can therefore use to send them to where they should be.

i'm not sure we really want to nest the auth or deauth endpoints? omniauth typically follows the pattern of
- `/auth/:provider`
- `/auth/:provider/callback`
which i've tried to mimic for deauth.

we will have quite a few of these now and we may want to convert home of these different handlers to independent omniauth strategies we can register and each one knows when to handle the request/callback

i can't really comment on the DSI flows as I've not worked with them close enough to know if they'll work with this pattern?

app/controllers/omniauth_callbacks_controller.rb
asmega: I removed that part of the code as it doesn't seem to be form related? my we can substitute for some other object which has the mapping

PR #3078, opened 2024-08-09
FE provider journey

rjlynch:
This PR adds the FE provider journey.
The journey starts from the provider following a link in an email sent at the end of the claimant journey. The sending of this link is covered in https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/3115 This PR deals with what happens after the link is clicked.

## Notes for review
Highly recommend reviewing each commit separately as there's more context in each of the commit messages, especially the commit that introduces the "Authorised Slugs" concept to the app, and the "Store verification details" commit.

## Walk through
(Note the error copy is still TBD)

https://github.com/user-attachments/assets/4ccaa113-cbbe-43e7-bee5-502810ed86cb

Ping me if you want to jump on a call to discuss any design decisions


app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -108,4 +101,51 @@ def test_user_auth_hash
       OmniAuth::AuthHash.new(info: {email: ""}, extra: {raw_info: {ONELOGIN_JWT_CORE_IDENTITY_HASH_KEY => "test"}})
     end
   end
+
+  def further_education_payments_provider_callback(auth)
+    if DfESignIn.bypass?
+      auth = OmniAuth::AuthHash.new({
```
rjlynch, line 107: The built in [developer strategy](https://github.com/omniauth/omniauth/blob/master/lib/omniauth/strategies/developer.rb) doesn't allow for setting `extra` fields. I'm considering sending up a PR to fix that but for the time being we just build the specific hash we want here.

app/models/policies/further_education_payments/eligibility.rb
```diff
@@ -17,6 +17,13 @@ def policy
       def ineligible?
         false
       end
+
+      # FIXME RL add this as a column in a separate pr, just stubbing
+      # for now while I work on the auth stuff. Put the commit adding the TRN
+      # first.
+      def teacher_reference_number
+        nil
+      end
```
asmega, line 26: fyi should be fixed in master now

app/views/further_education_payments/provider/claims/unauthorised.html.erb
```diff
@@ -0,0 +1,33 @@
...
+      the organisation in the claim.
+    </p>
+
+    <p class="govuk-body govuk-!-margin-top-6">
+      If your DfE Sign-in account is linked to multiple organisations, check
+      that you have logged in using the correct one.
+    </p>
+
+    <p class="govuk-body govuk-!-margin-top-6">
+      Email <%= govuk_mail_to("FE-Levellingup.PremiumPayments@education.gov.uk") %>
```
asmega, line 28: i think this is some where in the i18n file

app/views/further_education_payments/provider/claims/sign_in.html.erb
```diff
@@ -0,0 +1,37 @@
...
+      class: "govuk-button govuk-button--start",
+      data: {
+        module: "govuk-button"
+      }
+    ) do %>
+      Start now
+      <svg class="govuk-button__start-icon" xmlns="http://www.w3.org/2000/svg" width="17.5" height="19" viewBox="0 0 33 40" aria-hidden="true" focusable="false">
+        <path fill="currentColor" d="M0 0h13l20 20-20 20H0l20-20z" />
+      </svg>
+    <% end %>
```
asmega, line 35: don't bother changing but fyi for next time https://govuk-components.netlify.app/components/start-button/

lib/dfe_sign_in/api/user.rb
```diff
@@ -58,20 +58,32 @@ def has_role?(role_code)
...
         body["roles"].map { |r| r["code"] }
       end
 
+      def service_access?
+        response.code == "200"
+      end
+
+      def service_error?
+        response.code == "500"
+      end
```
asmega, line 74: don't think you're actually using this any where?

lib/dfe_sign_in/api/user.rb
```diff
@@ -58,20 +58,32 @@ def has_role?(role_code)
       end
 
       def role_codes
+        unless service_access?
+          raise ExternalServerError, "#{response.code}: #{response.body}" unless response.code.eql?("200")
```
asmega, line 62: is the `unless` needed given the line above (61) is guarding already?

asmega (approved):

lib/dfe_sign_in/api/user.rb
rjlynch: No, I'll put a fix for that in a future PR

lib/dfe_sign_in/api/user.rb
rjlynch: No, I guess YAGNI applies here

PR #3087, opened 2024-08-13
[LUPEYALPHA-764] Current Nursery page

vacabor:
https://dfedigital.atlassian.net/browse/LUPEYALPHA-764

Adds the ability for an EY provider to confirm which nursery their employee works at based on the list of providers the logged-in user has access to.

### DIFF
--- a/app/forms/journeys/early_years_payment/provider/authenticated/current_nursery_form.rb
+++ b/app/forms/journeys/early_years_payment/provider/authenticated/current_nursery_form.rb
@@ -28,7 +28,7 @@ module Journeys
           def nursery_is_valid_for_user
             return if nursery_urn.blank? || nursery_urn == "none_of_the_above"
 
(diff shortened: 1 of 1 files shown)


app/models/eligible_ey_provider.rb
```diff
@@ -25,10 +25,14 @@ def self.csv
   end
 
   def self.eligible_email?(email_address)
-    if email_address.present?
-      where(primary_key_contact_email_address: email_address).or(
-        where(secondary_contact_email_address: email_address)
-      ).exists?
-    end
+    email_address.present? && for_email(email_address).exists?
```
alkesh, line 28:
```suggestion
    for_email(email_address).exists?
```
does the `email_address.present?` check in `for_email` mean that we don't need to the same check here?

alkesh (approved):

AbigailMcP (approved):

PR #3088, opened 2024-08-13
Remove debug rescue

asmega:
# Context

- We are silently catching an exception and displaying it back to the user
- This was likely debug code and don't want this to happen under normal circumstances
- We want this exception to bubble up and be caught thru usual means


rjlynch (approved):
👍

PR #3090, opened 2024-08-13
Pass parameters to sesson creation

rjlynch:
On some journeys we'll want to be able to set answers based on url
parameters when starting a journey. This commit adds support for setting
answers when creating a session introducing a journey session form. The
session form is passed the parameters and subclasses of the form permit
various parameters based on their attributes.

---

In the specific case of the further education payments provider journey, the provider starts the journey by following a link with the claim id in it, we need to capture this claim id in the journey session when we create it.
For reference here is the session form for the FE provider journey:

```ruby
module Journeys
  module FurtherEducationPayments
    module Provider
      class SessionForm < Journeys::SessionForm
        attribute :claim_id, :string
      end
    end
  end
end
```

---

Currently `create_journey_session!` is called by a call back before the `ClaimsController#new` action runs. Ideally we'd update the landing pages for each journey to `POST` to the `ClaimsController#create` action and create the journey session there or, better still, have a separate `JourneysController` which handles all the setup actions for a journey and the `ClaimsController` can always assume one exists. Something for the backlog.

### DIFF
--- a/app/forms/journeys/session_form.rb
+++ b/app/forms/journeys/session_form.rb
@@ -1,0 +1,43 @@
+module Journeys
+  class SessionForm
+    include ActiveModel::Model
+    include ActiveModel::Attributes
+
+    def self.create!(params)
+      new(params).tap(&:save!).journey_session
+    end
+
+    attr_reader :journey_session
+
+    def initialize(params)
+      super(permitted_params(params))
+    end
+
+    def save!
+      @journey_session = journey::Session.create!(
+        journey: journey::ROUTING_NAME,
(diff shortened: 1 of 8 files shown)


app/forms/journeys/session_form.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+
+    private
+
+    def journey
+      self.class.module_parent
+    end
+
+    def answers
+      attributes.to_h
```
asmega, line 34: is `attributes` not already a hash?

asmega (approved):

PR #3092, opened 2024-08-14
[LUPEYALPHA-765] EY - Who are you making this claim for?

AbigailMcP:
## Changes in this PR

- New form on `/claimant-name` which captures `first_name` and `surname` of the claimant
- Newly extracted `NameFormatValidator` for regex shared between `PersonalDetailsForm` and new `ClaimantNameForm`

## Guidance to review

Prototype: https://claim-additi-main-9yrds6vyxcsd.herokuapp.com/claimant-name

Using the review app linked in comments below, navigate to url: `/early-years-payment-provider/claimant-name`

Check error states:
- No names entered
- More than 100 characters
- Special characters

Screenshot 2024-08-13 at 15 59 37


app/models/name_format_validator.rb
```diff
@@ -0,0 +1,11 @@
+class NameFormatValidator < ActiveModel::EachValidator
+  NAME_REGEX_FILTER = /\A[^"=$%#&*+\/\\()@?!<>0-9]*\z/
```
alkesh, line 2: it's probably not in the scope of this PR, but this regex seems too lenient.
It's allowing the characters: []^_`|{}~

AbigailMcP: ah good spot! this is taken from the current personal details logic. i'll raise a new PR so it can be product reviewed separately 👍

alkesh (approved):

PR #3095, opened 2024-08-14
Bug - Improve name format regex

AbigailMcP:
## Context

As per this comment: https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/3092#discussion_r1716900855

We realised that the following special characters were still being allowed in name fields:

```
[]^_`|{}~
```

## Changes in this PR

Update the regex to exclude these characters too.

## Guidance to review

Business review - are any of these characters supposed to be allowed for any reason?


rjlynch (approved):
Looks good 👍

alkesh (approved):

PR #3097, opened 2024-08-15
[LUPEYALPHA-851] One login DID

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-851
- One Login uses a public keys which we use to verify certain JWTs
- These public keys are published in a Decentralized Identifier (DID) document
- The keys in this document are prone to rotation therefore must be handled by us to ensure we have the correct keys in place
- See One Login docs for more info https://docs.sign-in.service.gov.uk/integrate-with-integration-environment/prove-users-identity/#validate-the-core-identity-claim-jwt-using-a-public-key

# Changes

- Remove the hardcoded public key which was held in an environment variable `ONELOGIN_IDENTITY_JWT_PUBLIC_KEY_BASE64`
- Implement a wrapper to hold a DID document
- Implement fetching and caching of DID document, this is cached in memory
- We attempt to respect the caching time provided by the DID doc with its response headers
- If we cannot fetch a new DID doc we will keep the existing cached doc
- If the DID doc has expired we will also attempt to fetch a new doc
- If we cannot fetch a doc due to expiry, we extend the expiry we do not hammer the DID doc store and continue to use the cached doc

# Post deploy tasks

- After deployment the env var `ONELOGIN_IDENTITY_JWT_PUBLIC_KEY_BASE64` will no longer be used and there should be cleaned up from all environments

### DIFF
--- a/app/models/one_login/did_cache.rb
+++ b/app/models/one_login/did_cache.rb
@@ -1,0 +1,102 @@
+require "json"
+require "net/http"
+
+class OneLogin::DidCache
+  # public api to use document
+  # if document not present it will fetch
+  # if it determines document has expired it will first refresh the document
+  def self.document
+    @cache ||= new
+
+    @cache.document
+  end
+
+  def self.clear_cache!
+    @cache ||= new
+    @cache.clear_cache!
+  end
+
+  def self.cache
+    @cache
+  end
+
+  attr_reader :document_object, :expires_at
+
+  def initialize(document_object: nil, expires_at: nil)
+    @document_object = document_object
+    @expires_at = expires_at
+  end
+
+  def expired?
+    return true if expires_at.nil?
+
+    Time.now > expires_at
+  end
+
+  def clear_cache!
+    @document_object = nil
+  end
+
+  def document
+    if document_object.nil?
+      refresh_document
+    end
+
+    if expired?
+      refresh_document
+    end
+
+    document_object
+  end
+
+  private
+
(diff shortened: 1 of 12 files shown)


app/models/one_login/did_cache.rb
```diff
@@ -0,0 +1,102 @@
...
+require "net/http"
+
+class OneLogin::DidCache
+  # public api to use document
+  # if document not present it will fetch
+  # if it determines document has expired it will first refresh the document
+  def self.document
+    @cache ||= new
+
+    @cache.document
```
rjlynch, line 11: Should this throw an error if the `@cache.document` is `nil`?

rjlynch (approved):
Really good 👍 
My only comment is maybe we want this code in `lib` as it's independent of our business domain and other projects might find this useful.

asmega:
> Really good 👍 My only comment is maybe we want this code in `lib` as it's independent of our business domain and other projects might find this useful.

not really a fan of `/lib` after zeitwerk was introduced. although happy to make an `/app/lib` folder and stick it in there which is my preferred pattern

app/models/one_login/did_cache.rb
asmega: it'll probably surface as some other error at the moment as it errors elsewhere. there is a user journey where if theres an error we render a certain page. so can probably do that then

alkesh (approved):

PR #3102, opened 2024-08-16
[LUPEYALPHA-770] EY PAYE reference

alkesh:

### DIFF
--- a/app/forms/journeys/early_years_payment/provider/authenticated/paye_reference_form.rb
+++ b/app/forms/journeys/early_years_payment/provider/authenticated/paye_reference_form.rb
@@ -10,12 +10,8 @@ module Journeys
           validates :paye_reference, presence: {message: i18n_error_message(:valid)},
             format: {with: PAYE_REFERENCE_REGEX_FILTER, message: i18n_error_message(:valid)}
 
-          attr_reader :nursery_name
-
-          def initialize(journey_session:, journey:, params:)
-            super
-
-            @nursery_name = EligibleEyProvider.find_by(urn: answers.nursery_urn)&.nursery_name
+          def nursery_name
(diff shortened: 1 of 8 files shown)


app/forms/journeys/early_years_payment/provider/authenticated/paye_reference_form.rb
```diff
@@ -0,0 +1,31 @@
...
+          validates :paye_reference, presence: {message: i18n_error_message(:valid)},
+            format: {with: PAYE_REFERENCE_REGEX_FILTER, message: i18n_error_message(:valid)}
+
+          attr_reader :nursery_name
+
+          def initialize(journey_session:, journey:, params:)
+            super
+
+            @nursery_name = EligibleEyProvider.find_by(urn: answers.nursery_urn)&.nursery_name
+          end
```
asmega, line 19: not easier with just a getter method?
```suggestion
          def nursery_name
            @nursery_name ||= EligibleEyProvider.find_by(urn: answers.nursery_urn)&.nursery_name
          end
```

config/locales/en.yml
```diff
@@ -1117,6 +1117,11 @@ en:
         question: Confirm that your employee works at the location below
         errors:
           presence: Select the nursery where your employee works
+      paye_reference:
+        question: "What is %{nursery_name}'s employer PAYE reference?"
```
asmega, line 1121: from my experience there is typically a preference to use smart quotes where possible including apostrophes https://designnotes.blog.gov.uk/2015/09/16/tips-for-creating-good-typography/

asmega (approved):

config/locales/en.yml
alkesh: ah, interesting - we have both in the codebase, but mostly the smart quote.

PR #3103, opened 2024-08-19
Depending on environment link to correct OL home

asmega:
# Context

- For one login integration environment there is a separate `home` or account area
- The One Login home link now sends to integration environment if indeed using integration environment


rjlynch (approved):
👍

PR #3106, opened 2024-08-19
[LUPEYALPHA-869] [LUPEYALPHA-870] One login FE failure

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-869
- https://dfedigital.atlassian.net/browse/LUPEYALPHA-870
- Handle failure from One Login auth

# Changes

- When user is not logged in and there is an auth error from one login show `Sorry, there is a problem with the service` page
- When user is logged in and there is an auth error from one login show `We cannot progress your application` page
- Should be relatively trivial to adapt this to handle EY journey too
- Not all parameters passed in are used, but are available should they be needed by the next person

### DIFF
--- a/app/controllers/omniauth_callbacks_controller.rb
+++ b/app/controllers/omniauth_callbacks_controller.rb
@@ -24,7 +24,16 @@ class OmniauthCallbacksController < ApplicationController
   end
 
   def failure
-    render layout: false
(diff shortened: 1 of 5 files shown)


rjlynch (approved):
Looks good 👍

PR #3109, opened 2024-08-20
One Login save UID + check it

asmega:
# Context

- At the moment the One Login UID is not saved in against the record which makes it harder later to marry up accounts
- We also do not check if the user who auth-ed is the same user that passes IDV. This means that potentially one person could auth then use another account to pass IDV.

# Changes

- After the OL auth callback also save their OL UID
- After their IDV callback ensure UID in this callback matches the UID we saved in the auth step


rjlynch (approved):
Looks good 👍

PR #3111, opened 2024-08-21
[LUPEYALPHA-771] check your answers

alkesh:
https://dfedigital.atlassian.net.mcas.ms/browse/LUPEYALPHA-771

The check-your-answers page has a text field where the provider has to enter their name.

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -1,18 +1,11 @@
 class SubmissionsController < BasePublicController
   include PartOfClaimJourney
+  include ClaimSubmission
 
   skip_before_action :send_unstarted_claimants_to_the_start, only: [:show]
 
   def create
-    @form = journey::ClaimSubmissionForm.new(journey_session: journey_session)
-
-    if @form.save
-      session[:submitted_claim_id] = @form.claim.id
-      clear_claim_session
-      redirect_to claim_confirmation_path
-    else
-      render "claims/check_your_answers"
-    end
+    create_and_save_claim_form
   end
 
   def show
--- a/app/models/journeys/early_years_payment/provider/authenticated/answers_presenter.rb
+++ b/app/models/journeys/early_years_payment/provider/authenticated/answers_presenter.rb
@@ -1,0 +1,81 @@
+module Journeys
+  module EarlyYearsPayment
+    module Provider
+      module Authenticated
+        class AnswersPresenter < BaseAnswersPresenter
+          include ActionView::Helpers::TranslationHelper
+
+          def claim_answers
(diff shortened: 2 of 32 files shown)


app/controllers/submissions_controller.rb
```diff
@@ -4,7 +4,7 @@ class SubmissionsController < BasePublicController
   skip_before_action :send_unstarted_claimants_to_the_start, only: [:show]
 
   def create
-    @form = journey::ClaimSubmissionForm.new(journey_session: journey_session)
+    @form = journey::ClaimSubmissionForm.new(journey_session: journey_session, params: params)
```
asmega, line 7: can the check your answers page not be a `Form` like most of the other pages in the journey? that way you won't have to do this params business. I think if you think of the check your answers page as a form for `submitter_name` but with a whole load of read-only content ie the check answers bit

alkesh: 👍🏽 yes - that does work - using a callback `check_your_answers_after_form_save_success` to create and save the `ClaimSubmissionForm`

app/models/journeys/early_years_payment/provider/authenticated/answers_presenter.rb
```diff
@@ -0,0 +1,81 @@
...
+            [
+              "Employer’s PAYE reference number",
+              answers.paye_reference,
+              "paye-reference"
+            ]
+          end
+
+          def employee_name
+            [
+              "Employee's name",
```
AbigailMcP, line 40: Small one, but there's a a few straight apostrophes in this file that need to be curly ’

AbigailMcP, line 40: Small one, but there's a a few straight apostrophes in this file that need to be curly ’

AbigailMcP (commented):
Question for design really - if i try to submit the form without adding my name, it jumps to the top of the page and the error text is off screen. I think we need an error summary at the top of the page to provide a link to the form field in error.

alkesh:
> Question for design really - if i try to submit the form without adding my name, it jumps to the top of the page and the error text is off screen. I think we need an error summary at the top of the page to provide a link to the form field in error.

Checked with Xanthe - yes, we should have an error banner on the check-your-answers page - and also, for consistency, the other pages should also show an error banner.

app/models/journeys/early_years_payment/provider/authenticated/answers_presenter.rb
alkesh: 👍🏽 fixed

AbigailMcP (approved):

PR #3113, opened 2024-08-22
[LUPEYALPHA-890] Admin download report supports FE claims

asmega:


spec/models/claim/data_report_request_spec.rb
```diff
@@ -26,7 +28,7 @@
         expect(report_request_csv[index].fields("Full name")).to include(claim.full_name)
         expect(report_request_csv[index].fields("Email")).to include(claim.email_address)
         expect(report_request_csv[index].fields("Date of birth")).to include(claim.date_of_birth.to_s)
-        expect(report_request_csv[index].fields("ITT subject")).to include(claim.eligibility.eligible_itt_subject)
+        expect(report_request_csv[index].fields("ITT subject")).to include(claim.eligibility.try(:eligible_itt_subject))
```
alkesh, line 31: this `try` seems slightly obscure.
Would it be better to split out the claims that have and ITT subject, and those that don't, and then check the field is explicitly populated or nil ?

asmega: - i've split out the FE test and removed the `try`
- i've also tidied up most of the file so it's more readable

PR #3116, opened 2024-08-22
fe provider dsi bypass

rjlynch:
Add DfE sign in by pass

In review apps we can't use DfE sign in as the urls are dynamic and so
are not registered with DfE sign in. This commit provides a form to
allow testers to set the DfE sign in payload to check various sign in
scenarios.

Screenshot 2024-08-29 at 10 37 57

### DIFF
--- a/app/controllers/omniauth_callbacks_controller.rb
+++ b/app/controllers/omniauth_callbacks_controller.rb
@@ -123,6 +123,8 @@ class OmniauthCallbacksController < ApplicationController
   end
 
   def further_education_payments_provider_callback(auth)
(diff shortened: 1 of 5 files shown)


rjlynch:
I'd investigated using the [developer oauth strategy](https://github.com/omniauth/omniauth/blob/master/lib/omniauth/strategies/developer.rb) but it only lets us set top level fields, can't set `extra` attributes, and it was too much of a faff to add support for more fields so decided to just use a form and add a branch on whether DSI is enabled. Patching the dev strategy is probably better long term, may be worth sending up a PR to them

asmega (approved):

PR #3120, opened 2024-08-23
[LUPEYALPHA-848] Add FE claimant email templates

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-848
- Add missing FE claimant email templates
- Copy within these templates are not final and will still need updating
- This just hooks up the mechanism to makes copy changes much easier


rjlynch (approved):
LGTM 👍

PR #3124, opened 2024-08-27
[LUPEYALPHA-950] teaching-hours-per-week page - content update

alkesh:
change hint on teaching-hours-per-week page from:
> ‘Teaching’ refers to the direct contact time you spend in lessons with students of all ages.

to

> ‘Timetabled teaching hours’ refers to the time you spend teaching lessons to students of all ages.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -894,7 +894,7 @@ en:
       teaching_hours_per_week:
         question:
           On average, how many hours per week are you timetabled to teach at %{school_name} during the current term?
-        hint: ‘Teaching’ refers to the direct contact time you spend in lessons with students of all ages.
+        hint: ‘Timetabled teaching hours’ refer to the time you spend teaching lessons to students of all ages.
         options:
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
```diff
@@ -894,7 +894,7 @@ en:
       teaching_hours_per_week:
         question:
           On average, how many hours per week are you timetabled to teach at %{school_name} during the current term?
-        hint: ‘Teaching’ refers to the direct contact time you spend in lessons with students of all ages.
+        hint: ‘Timetabled teaching hours’ refer to the time you spend teaching lessons to students of all ages.
```
asmega, line 897: think wireframe is wrong and should be `refers` not `refer`. wdyt?

alkesh: yeah, I agree. 
I think the 'refer' is referring to the singular phrase "Timetabled teaching hours", and not the plural hours, so refers is more grammatically correct.

asmega (approved):

PR #3128, opened 2024-08-28
Add heartbeat monitor for statuscake

johnake:
https://trello.com/c/PQgfB2yQ/1961-claims-worker-heartbeat-monitor

### DIFF
--- a/app/jobs/heartbeat_job.rb
+++ b/app/jobs/heartbeat_job.rb
@@ -1,12 +1,18 @@
-# Writes "Heartbeat job performed" to the logs every minute. This message is
-# evidence that the worker is working properly. We can use a log alerting tool
-# to alert us if too much time passes without seeing one of these messages.
+require "net/http"
+
 class HeartbeatJob < CronJob
   self.cron_expression = "* * * * *"
 
   queue_as :heartbeat
 
   def perform
-    logger.info "Heartbeat job performed"
+    if ENV.key?("HEARTBEAT_CHECK_URL") # Not available in dev and review environments
+      uri = URI(ENV["HEARTBEAT_CHECK_URL"])
+      res = Net::HTTP.get_response(uri)
+
+      unless res.is_a?(Net::HTTPSuccess)
+        Rails.logger.error "Error connecting to StatusCake: #{res.code} #{res.msg}"
+      end
+    end
   end
 end
--- a/terraform/application/application.tf
+++ b/terraform/application/application.tf
@@ -17,9 +17,14 @@ module "application_configuration" {
       PGSSLMODE          = local.postgres_ssl_mode
       CANONICAL_HOSTNAME = local.canonical_hostname
   })
-  secret_variables = {
-    DATABASE_URL = module.postgres.url
-  }
+  secret_variables = merge(
+    {
+      DATABASE_URL        = module.postgres.url
+    },
+    var.enable_monitoring ? {
+      HEARTBEAT_CHECK_URL = module.statuscake[0].heartbeat_check_urls[local.heartbeat_check_name]
+    } : {}
+  )
 }
 
 module "web_application" {
--- a/terraform/application/statuscake.tf
+++ b/terraform/application/statuscake.tf
@@ -7,4 +7,6 @@ module "statuscake" {
   ssl_urls    = compact([var.external_url])
 
   contact_groups = var.statuscake_contact_groups
+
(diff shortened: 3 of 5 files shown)


johnake:
[Claims: worker heartbeat monitor](https://trello.com/c/PQgfB2yQ/1961-claims-worker-heartbeat-monitor)

terraform/application/statuscake.tf
```diff
@@ -7,4 +7,7 @@ module "statuscake" {
   ssl_urls    = compact([var.external_url])
 
   contact_groups = var.statuscake_contact_groups
+
+  heartbeat_names  = ["heartbeat-check"]
+  heartbeat_period = 600  # The period in seconds within which a heartbeat must be received
```
saliceti, line 12: The default is 600 so there is no point setting it here too

terraform/application/statuscake.tf
```diff
@@ -7,4 +7,7 @@ module "statuscake" {
...
 }
```
saliceti, line 13: The devs will need an environment variable in the worker with the check URL. Please populate it in https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/master/terraform/application/application.tf#L20 using the module output

saliceti:
Please point at the _testing_ terraform modules branch so devs can test

terraform/application/application.tf
```diff
@@ -17,9 +17,12 @@ module "application_configuration" {
       PGSSLMODE          = local.postgres_ssl_mode
       CANONICAL_HOSTNAME = local.canonical_hostname
   })
-  secret_variables = {
-    DATABASE_URL = module.postgres.url
-  }
+  secret_variables = merge(
+    {
+      DATABASE_URL         = module.postgres.url
+      HEARTBEAT_CHECK_URLS = join(",", values(module.statuscake[0].heartbeat_check_urls))
```
saliceti, line 23: We'll most probably need only 1 check, so something like this would be enough:
```
HEARTBEAT_CHECK_URL = module.statuscake[0].heartbeat_check_urls[0]
```

terraform/application/application.tf
```diff
@@ -17,9 +17,12 @@ module "application_configuration" {
       PGSSLMODE          = local.postgres_ssl_mode
       CANONICAL_HOSTNAME = local.canonical_hostname
   })
-  secret_variables = {
-    DATABASE_URL = module.postgres.url
-  }
+  secret_variables = merge(
+    {
+      DATABASE_URL        = module.postgres.url
+      HEARTBEAT_CHECK_URL = module.statuscake[0].heartbeat_check_urls[local.heartbeat_check_name]
```
saliceti, line 23: This may break if var.enable_monitoring is false. We should only add `{HEARTBEAT_CHECK_URL = module.statuscake[0].heartbeat_check_urls[local.heartbeat_check_name]}` if it's true

app/jobs/heartbeat_job.rb
```diff
@@ -1,12 +1,13 @@
...
+require "net/http"
+
 class HeartbeatJob < CronJob
   self.cron_expression = "* * * * *"
 
   queue_as :heartbeat
 
   def perform
-    logger.info "Heartbeat job performed"
+    url = URI.parse(ENV["HEARTBEAT_CHECK_URL"])
```
saliceti, line 9: This may break in review as there won’t be monitoring and the env var will be absent. Can you plan for that? like if `ENV.key? "HEARTBEAT_CHECK_URL" {…}`

app/jobs/heartbeat_job.rb
```diff
@@ -1,12 +1,15 @@
...
+
 class HeartbeatJob < CronJob
   self.cron_expression = "* * * * *"
 
   queue_as :heartbeat
 
   def perform
-    logger.info "Heartbeat job performed"
+    if ENV.key?("HEARTBEAT_CHECK_URL") # Not available in dev and review environments
+      url = URI.parse(ENV["HEARTBEAT_CHECK_URL"])
```
saliceti, line 10: This code returns a 400 HTTP error
I got it working with:
```ruby
uri = URI(ENV["HEARTBEAT_CHECK_URL"])
res = Net::HTTP.get_response(uri)
```

saliceti: We could also exit with an error if `res.is_a?(Net::HTTPSuccess)`

saliceti (approved):

PR #3129, opened 2024-08-28
[LUPEYALPHA-775] EY previous job in early years setting - unhappy path

AbigailMcP:
## Changes in this PR

- Branching for when the user answers 'yes' to having worked in an early years setting in the past 6 months
- Two new pages and forms
- Ineligible page for when the returner does not qualify for a payment

Also:
- Rename of an attribute `first_job_within_6_months` to `returning_within_6_months` which was named opposite to the question being asked and boolean being stored.

### DIFF
--- a/app/views/early_years_payment/provider/authenticated/claims/_ineligibility_returner.html.erb
+++ b/app/views/early_years_payment/provider/authenticated/claims/_ineligibility_returner.html.erb
@@ -1,0 +1,32 @@
+<% first_name = journey_session.answers.first_name %>
+
+<% content_for(:page_title, page_title("#{first_name} is not eligible for this payment", journey: current_journey_routing_name)) %>
+<% @backlink_path = claim_path("early-years-payment", "landing-page") %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l">
+      <%= first_name %> is not eligible for this payment
+    </h1>
+    <p class="govuk-body">
+      Based on the information you have provided, <%= first_name %> is not eligible to claim an early years incentive payment.
+    </p>
+
+    <p class="govuk-body">
+      For further guidance, you should refer to the
+      <%= govuk_link_to "full eligibility criteria", claim_path("early-years-payment", "landing-page") %>
+      at the start of this service.
+    </p>
+
+    <h2 class="govuk-heading-m">
+      Get help with your claim
+    </h2>
+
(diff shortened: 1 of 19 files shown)


spec/models/policies/early_years_payments/policy_eligibility_checker_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+          returner_worked_with_children: true,
+          returner_contract_type: "permanent"
+        )
+      end
+
+      it "is ineligble as :returner" do
+        expect(subject).to be_ineligible
+        expect(subject.status).to eql(:ineligible)
+        expect(subject.ineligibility_reason).to eql(:returner)
+      end
```
alkesh, line 21: I think we're missing scenarios here for when `returning_within_6_months` and `returner_worked_with_children` are false.
Currently, if the PolicyEligibilityChecker is refactored to remove any 2 out of 3 checks, the spec still passes.

alkesh:
Is there an issue with changing your answers via the check-your-answers page?
If I answer yes on the returner page, 
then, when I get to the check-your-answers, it's not possible to change returner-worked-with-children or returner-contract-type.

AbigailMcP:
> Is there an issue with changing your answers via the check-your-answers page? If I answer yes on the returner page, then, when I get to the check-your-answers, it's not possible to change returner-worked-with-children or returner-contract-type.

Had a chat with Xanthe about this:

> If it's not too much effort to dynamically add additional rows if the user got to those questions, which I think is what you've proposed, I think that would be the best design. If not, we can have a think about another way to design this page. (also I'm sure Chris has mentioned but the copy for those two chunky questions are subject to change when we hand over the content docs)

Now there's a row for each of the three questions, depending on whether the user answered 'yes' to the one before it. This is different to the prototype.

alkesh (approved):

app/views/early_years_payment/provider/authenticated/claims/_ineligibility_returner.html.erb
```diff
@@ -0,0 +1,32 @@
+<% first_name = journey_session.answers.first_name %>
+
+<% content_for(:page_title, page_title("#{first_name} is not eligible for this payment", journey: current_journey_routing_name)) %>
+<% @backlink_path = claim_path("early-years-payment", "landing-page") %>
```
alkesh, line 4: Is this backlink correct?
On the FE journey, the backlink on ineligibility screens actually goes back to the question just asked (same as hitting the back button).

PR #3133, opened 2024-08-29
[LUPEYALPHA-704] FE reminders

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-704
- Able to set reminder as part of FE journey

# Changes

- I don't think the current reminder code is the best and have written a new version of reminder journey with its own controller
- Things are a little rough round the edges but works as needed
- I've tried to keep things simple and also considered if we need to add this to the end of other journeys
- I hope later we can remove the old reminder controller and use this one instead
- I also hope we can later we can model the `ClaimsController` to be more like the controller created here

### DIFF
--- a/app/controllers/reminders_controller.rb
+++ b/app/controllers/reminders_controller.rb
@@ -1,4 +1,6 @@
 class RemindersController < BasePublicController
+  before_action :create_journey_session_if_no_session
+
   def show
     @form = form_from_slug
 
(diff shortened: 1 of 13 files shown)


app/controllers/reminders_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+    journey_session
+  end
+
+  def journey_session_key
+    :"#{journey::ROUTING_NAME}_journeys_session_id"
+  end
+
+  def view_file
+    params[:slug].underscore
+  end
```
rjlynch, line 47: I think this is fine as we throw "form not found" if the slug isn't one we expect

rjlynch (approved):
Looks good 👍 

I wonder if it's possible to lose the journey session from the reminders journey and store everything we need on the reminder model (would need to move otp to a db attribute rather than virtual attribute it currently is 🙃). If we need to pre-fill the personal details we could just pull them off the journey session / claim and pass them via url params when the user clicks the "Set reminder" button.

PR #3139, opened 2024-08-30
Provider confirmation email

rjlynch:
When the provider has verified the claim we want to email their
organisation with a confirmation email, letting them know a claim has
been verified and who did it.

Also fixes the `have_received_email_matcher` to show the template id if
a match cant be found.


spec/forms/journeys/further_education_payments/provider/verify_claim_form_spec.rb
```diff
@@ -164,13 +191,26 @@
           ],
           "verifier" => {
             "dfe_sign_in_uid" => "123",
-            "first_name" => "Seymoure",
+            "first_name" => "Seymour",
             "last_name" => "Skinner",
             "email" => "seymore.skinner@springfield-elementary.edu"
```
asmega, line 196: 😅

asmega (approved):

slorek (approved):

PR #3143, opened 2024-09-03
Admin claims filter refactor

asmega:
# Context

- this refactors the admin claims filter into a form object
- although called a form object its not currently wired up to the form but has the same attribute names so could be wired up to do so later
- the object of this change is to extract excess controller code into correct objects to allow separation of responsibilities allow the code to be simplified more easily
- the original reason for this change was the attempt to introduce FE claims to admin areas which were slowed by the current form filters being coupled to the controller code


rjlynch (approved):
Nice 👍 will be easier to test too

alkesh (approved):

PR #3144, opened 2024-09-03
FE admin claims

asmega:
# Context

- This updates tests to including showing FE claims in back office

### DIFF
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -396,7 +396,7 @@ class Claim < ApplicationRecord
   end
 
   def award_amount_with_topups
-    topups.sum(:award_amount) + award_amount
+    topups.sum(:award_amount) + (award_amount || 0)
   end
 
   def must_manually_validate_bank_details?
--- a/app/models/payroll_run.rb
+++ b/app/models/payroll_run.rb
@@ -55,7 +55,7 @@ class PayrollRun < ApplicationRecord
           # associates the payment to the Topup, so we know it's payrolled
           group_topups = grouped_items.select { |i| i.is_a?(Topup) }
 
-          award_amount = grouped_items.sum(&:award_amount)
+          award_amount = grouped_items.map(&:award_amount).compact.sum(0)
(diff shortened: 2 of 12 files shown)


app/models/claim.rb
```diff
@@ -396,7 +396,7 @@ def important_notes
   end
 
   def award_amount_with_topups
-    topups.sum(:award_amount) + award_amount
+    topups.sum(:award_amount) + (award_amount || 0)
```
asmega, line 399: feeling ambivalent about this one

app/models/payroll_run.rb
```diff
@@ -55,7 +55,7 @@ def self.create_with_claims!(claims, topups, attrs = {})
           # associates the payment to the Topup, so we know it's payrolled
           group_topups = grouped_items.select { |i| i.is_a?(Topup) }
 
-          award_amount = grouped_items.sum(&:award_amount)
+          award_amount = grouped_items.map(&:award_amount).compact.sum(0)
```
asmega, line 58: looks ugly but fixes deprecation

config/locales/en.yml
```diff
@@ -827,6 +827,12 @@ en:
     feedback_email: "FE-Levellingup.PremiumPayments@education.gov.uk"
     support_email_address: "FE-Levellingup.PremiumPayments@education.gov.uk"
     claim_subject: "Further education payment"
+    policy_acronym: FE
```
rjlynch, line 830: I went for "TRIP" in my PR

config/locales/en.yml
```diff
@@ -827,6 +827,12 @@ en:
...
+    policy_short_name: Further education payments
```
rjlynch, line 831: In figma I think this is now "Targeted Retention Incentive Payment For FE Teachers"

rjlynch (approved):
Looks good, just left one comment about the policy short name

config/locales/en.yml
asmega: it's called `policy_short_name` 😂

config/locales/en.yml
asmega: i'll leave for now and sort it later

PR #3146, opened 2024-09-03
[LUPEYALPHA-977] Sign out link on mismatch organisation failure page

kenfodder:
* Actually calls the DSI logout endpoint
* Set a post logout redirect uri back to Claim
* Grab the claim off the session and clears the session
* Redirects to the verify url as if the user clicked on it again
* Added a new flash[:success] and a message to say signed out

UPDATE:
* Sign out button moved to the bottom
* Flash success design

Screenshot 2024-09-04 at 16 08 04

Screenshot 2024-09-04 at 16 08 12

### DIFF
--- a/app/controllers/omniauth_callbacks_controller.rb
+++ b/app/controllers/omniauth_callbacks_controller.rb
@@ -32,12 +32,12 @@ class OmniauthCallbacksController < ApplicationController
   def sign_out
     case current_journey_routing_name
     when "further-education-payments-provider"
+      claim = journey_session.answers.claim
+      clear_journey_sessions!
+
+      flash[:success] = "You have signed out of DfE Sign-in"
       redirect_to(
-        claim_path(
-          journey: current_journey_routing_name,
-          slug: "sign-in"
-        ),
-        notice: "You've been signed out"
+        Journeys::FurtherEducationPayments::Provider::SlugSequence.verify_claim_url(claim)
       )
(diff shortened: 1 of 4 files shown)


app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -29,6 +29,21 @@ def failure
...
 
+  def sign_out
+    case current_journey_routing_name
+    when "further-education-payments-provider"
+      redirect_to(
+        claim_path(
+          journey: current_journey_routing_name,
+          slug: "sign-in"
+        ),
+        notice: "You've been signed out"
```
asmega, line 40: smart apostrophe 😅

asmega (approved):

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -29,6 +29,21 @@ def failure
...
+      )
```
rjlynch, line 41: I think we probably want to clear the session here. We can do something like this

```suggestion
      claim = journey_sessions.answers.claim
      clear_journey_sessions!
      redirect_to(
        Journeys::FurtherEducationPayments::Provider::SlugSequence.verify_claim_url(claim),
        notice: "You've been signed out"
      )
```
So the claim they're trying to access is in the url and they'll be able to sign in again and start a new session

PR #3149, opened 2024-09-04
[LUPEYALPHA-993] Cache OL data

asmega:
# Context

- https://dfedigital.atlassian.net.mcas.ms/browse/LUPEYALPHA-993
- Store One Login IDV data and other OL related metadata

# Changes

- Moved a load of admin request specs into a folder aptly named `/admin`
- On the journey session we now store the following extra
  - `onelogin_auth_at` - when user authenticated with OL
  - `onelogin_idv_at` - when user passed IDV
  - `onelogin_idv_first_name`, `onelogin_idv_last_name`, `onelogin_idv_date_of_birth`
- All the above also get saved on the claim when claim is created at end of journey
- We also at the same time store `onelogin_uid` on the claim which was already present in journey session
- Bonus: we now pre-populate date of birth in personal details section with the value from one login

### DIFF
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -83,6 +83,12 @@ class Claim < ApplicationRecord
     logged_in_with_onelogin: false,
     onelogin_credentials: true,
     onelogin_user_info: true,
+    onelogin_uid: true,
+    onelogin_auth_at: false,
+    onelogin_idv_at: false,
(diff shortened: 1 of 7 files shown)


rjlynch (approved):
Looks good 👍 
Might be tempted to extract parsing the OL data into a form object but I it's fine as is 👍

PR #3150, opened 2024-09-04
[LUPEYALPHA-964] Claim Summary for FE

alkesh:
https://dfedigital.atlassian.net.mcas.ms/browse/LUPEYALPHA-964
Claim Summary section is different for FE claims

### DIFF
--- a/app/controllers/admin/base_admin_controller.rb
+++ b/app/controllers/admin/base_admin_controller.rb
@@ -58,10 +58,5 @@ module Admin
       session.delete(:role_codes)
       session.delete(:claims_backlink_path)
     end
-
-    def set_claim_summary_view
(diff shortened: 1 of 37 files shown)


app/controllers/admin/base_admin_controller.rb
```diff
@@ -58,5 +58,10 @@ def clear_session
       session.delete(:role_codes)
       session.delete(:claims_backlink_path)
     end
+
+    def set_claim_summary_view
+      policy_append = @claim.policy.to_s.underscore if @claim.policy.is_a?(Policies::FurtherEducationPayments)
+      @claim_summary_view = ["claim_summary", policy_append].compact.join("_")
+    end
```
asmega, line 65: can this be a view helper? that way there are zero controller changes

alkesh: 👍🏽 that is an excellent suggestion

PR #3157, opened 2024-09-06
[LUPEYALPHA-954] FE admin tasks

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-954
- FE admin tasks
- There are placeholder present to unblock areas of work for others

### DIFF
--- a/app/models/policies/further_education_payments/admin_tasks_presenter.rb
+++ b/app/models/policies/further_education_payments/admin_tasks_presenter.rb
@@ -26,6 +26,18 @@ module Policies
         []
       end
 
+      def employment
+        []
+      end
+
+      def identity_confirmation
+        []
+      end
+
+      def student_loan_plan
+        []
+      end
+
       private
 
       def verifier
--- a/app/models/policies/further_education_payments/claim_checking_tasks.rb
+++ b/app/models/policies/further_education_payments/claim_checking_tasks.rb
@@ -12,32 +12,20 @@ module Policies
       delegate :policy, to: :claim
 
       def applicable_task_names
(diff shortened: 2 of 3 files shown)


app/models/automated_checks/claim_verifiers/employment.rb
```diff
@@ -11,13 +11,18 @@ def initialize(claim:, admin_user: nil)
...
+        return unless required?
         return unless awaiting_task?
 
         no_data || no_match || matched
       end
 
       private
 
+      def required?
+        claim.eligibility.teacher_reference_number.present?
```
asmega, line 23: not sure if this negatively impacts other policies?

rjlynch: I think this should be ok, most of the other journeys require a TRN from the claimant to submit, IRP doesn't but it doesn't use this claim verifier

app/models/claim_checking_tasks.rb
```diff
@@ -12,26 +12,34 @@ def initialize(claim)
...
-        task_names.delete("arrival_date")
-        task_names.delete("employment_contract")
-        task_names.delete("employment_start")
-        task_names.delete("subject")
-        task_names.delete("teaching_hours")
-      end
-      unless claim.policy.further_education_payments?
+    case policy
+    when Policies::FurtherEducationPayments
+      Policies::FurtherEducationPayments::ClaimCheckingTasks
```
rjlynch, line 17: 👍 nice

rjlynch (approved):
Nice 👍
I think pulling out a class for the policies task lists is a great idea

PR #3158, opened 2024-09-06
Fix production One Login endpoint config

slorek:
`BYPASS_ONELOGIN_SIGN_IN` does not do anything except on review apps, and is misleadingly configured.


rjlynch (approved):
👍

PR #3164, opened 2024-09-06
LUPEYALPHA-955 - Add identity confirmation task to FE that auto passes

kenfodder:
* Show task outcome with OneLogin timestamp
* Tidy up status tags with tasks

Screenshot 2024-09-06 at 14 52 39

Screenshot 2024-09-06 at 14 52 45

Screenshot 2024-09-06 at 14 52 33

### DIFF
--- a/app/views/admin/tasks/_task_outcome.html.erb
+++ b/app/views/admin/tasks/_task_outcome.html.erb
@@ -10,7 +10,7 @@
       This task was performed by <%= user_details(task.created_by, include_line_break: false) %> on <%= l(task.updated_at) %>
     <% elsif task.created_by %>
       This task was performed by an automated check uploaded by <%= user_details(task.created_by, include_line_break: false) %> on <%= l(task.created_at) %>
-    <% elsif task.claim.identity_confirmed_with_onelogin? %>
+    <% elsif task.identity_confirmation? && task.claim.identity_confirmed_with_onelogin? %>
       This task was performed by GOV.UK One Login on <%= l(task.claim.onelogin_idv_at) %>
     <% else  %>
       This task was performed by an automated check on <%= l(task.created_at) %>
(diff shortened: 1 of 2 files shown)


rjlynch (approved):
LGTM, I think in future we'll want to change the auto pass method to something like

```
def auto_pass
  create_task(match: nil, passed: true) if claim.identity_confirmed_with_onelogin?
end
```

but I think checking the policy is fine for now

app/models/automated_checks/claim_verifiers/identity.rb
```diff
@@ -18,16 +18,18 @@ def perform
...
+        auto_pass || no_match || partial_match || complete_match
       end
 
       private
 
       attr_accessor :admin_user, :claim
       attr_reader :dqt_teacher_status
 
+      def auto_pass
+        create_task(match: nil, passed: true) if claim.policy.further_education_payments?
```
slorek, line 30: I think we should iterate the `Policies::POLICIES` and let them determine this. We have a few instances elsewhere were policy specific logic is hardcoded and has been missed.

app/views/admin/tasks/_task_outcome.html.erb
```diff
@@ -10,6 +10,8 @@
       This task was performed by <%= user_details(task.created_by, include_line_break: false) %> on <%= l(task.updated_at) %>
     <% elsif task.created_by %>
       This task was performed by an automated check uploaded by <%= user_details(task.created_by, include_line_break: false) %> on <%= l(task.created_at) %>
+    <% elsif onelogin_auth_at = task.claim.onelogin_auth_at %>
```
slorek, line 13: Is there a more explicit way of doing this?

slorek (changes requested):

slorek (approved):

PR #3167, opened 2024-09-09
Record who sent the email to the provider

rjlynch:
The figma designs require us to display the admin who resent the provider verification email. To capture this information we use the existing note mechanism on claims. This is a bit fragile as we assume the only "provider_verification" note is for emails being sent. It seems like the label on notes maps to a task name, so we went with that for the label. If in future we can take other provider verification notes we may want to consider renaming this email specific label.

Screenshot 2024-09-09 at 16 39 03
Screenshot 2024-09-09 at 16 41 38

### DIFF
--- a/app/controllers/admin/further_education_payments/provider_verification_emails_controller.rb
+++ b/app/controllers/admin/further_education_payments/provider_verification_emails_controller.rb
@@ -6,6 +6,12 @@ module Admin
       def create
         claim = Claim.find(params[:claim_id])
 
+        claim.notes.create!(
+          created_by: admin_user,
+          label: "provider_verification",
+          body: "Verification email sent to #{claim.school.name}"
+        )
+
         ClaimMailer.further_education_payment_provider_verification_email(claim).deliver_later
 
         flash[:notice] = "Verification email sent to #{claim.school.name}"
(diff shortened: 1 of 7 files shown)


slorek (commented):
If the email is resent multiple times it looks like we will only display the latest instance, is that correct?

rjlynch:
> If the email is resent multiple times it looks like we will only display the latest instance, is that correct?

Next to the button we'll only show the latest instance but we record each time it's sent in the case notes

slorek:
> > If the email is resent multiple times it looks like we will only display the latest instance, is that correct?
> 
> Next to the button we'll only show the latest instance but we record each time it's sent in the case notes

Rather than showing only the latest one, I don't see any reason why we can't just show all the relevant notes. I think this would be more helpful in the event of a support request.

rjlynch:
> > > If the email is resent multiple times it looks like we will only display the latest instance, is that correct?
> > 
> > 
> > Next to the button we'll only show the latest instance but we record each time it's sent in the case notes
> 
> Rather than showing only the latest one, I don't see any reason why we can't just show all the relevant notes. I think this would be more helpful in the event of a support request.

Have updated this to now show all the email sending events. I've also tacked on a bug fix to this pr to fix the copy when the email has been sent to the provider https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/3167/commits/c3c004c1a748df14cf3c70cb9298611b42edb6df

Screenshot 2024-09-12 at 17 03 08

slorek (approved):

PR #3173, opened 2024-09-10
Refactor: Move additional payments journey specific reminder code to journey module

slorek:
This code was quite misleading but only actually used [in one place](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/master/app/views/additional_payments/submissions/show.html.erb#L27) in the additional payments journey.

There's still a [bit of fudge](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/master/app/forms/reminders/confirmation_form.rb#L8) with the `itt_academic_year` being used as the basis for reminders, when this is not captured on some of the new journeys. We should consider renaming this column at least, but it's a separate decision point to this change.


rjlynch (approved):
Nice 👍

PR #3175, opened 2024-09-10
Track which journey a reminder is for and use correct support email

slorek:
https://dfedigital.atlassian.net/browse/LUPEYALPHA-1016

At the moment the codebase assumes that all reminders are for the additional payments journey.

I've added an extra column to `Reminder` to specify which journey the reminder is for, and removed the assumptions in the code in order to support reminders for other journeys.

At the moment the copy in the actual reminder template that is sent when the journey is opened only accommodates the additional payments journey. I've added a failsafe to prevent reminders being sent when a different journey is opened. Now only reminders for that journey which is being opened should be sent.

Additionally when the additional payments journey was opened it would send all pending reminders, including those for other journeys. This has been fixed.

Coming back to the goal of this ticket; these changes allow us to use a different email address depending on which journey the reminder is for.

### DIFF
--- a/app/forms/reminders/confirmation_form.rb
+++ b/app/forms/reminders/confirmation_form.rb
@@ -2,7 +2,7 @@ module Reminders
   class ConfirmationForm < Form
     def reminder
       @reminder ||= Reminder.find_by(
-        journey: Journeys.for_routing_name(journey_session.journey).to_s,
+        journey_class: Journeys.for_routing_name(journey_session.journey).to_s,
         full_name: journey_session.answers.reminder_full_name,
         email_address: journey_session.answers.reminder_email_address,
         email_verified: true,
--- a/app/forms/reminders/confirmation_form.rb
+++ b/app/forms/reminders/confirmation_form.rb
@@ -2,6 +2,7 @@ module Reminders
   class ConfirmationForm < Form
     def reminder
       @reminder ||= Reminder.find_by(
+        journey: Journeys.for_routing_name(journey_session.journey).to_s,
         full_name: journey_session.answers.reminder_full_name,
         email_address: journey_session.answers.reminder_email_address,
         email_verified: true,
--- a/app/models/reminder.rb
+++ b/app/models/reminder.rb
@@ -4,12 +4,16 @@ class Reminder < ApplicationRecord
 
   scope :email_verified, -> { where(email_verified: true) }
   scope :not_yet_sent, -> { where(email_sent_at: nil) }
-  scope :by_journey, ->(journey) { where(journey: journey.to_s) }
+  scope :by_journey, ->(journey) { where(journey_class: journey.to_s) }
   scope :inside_academic_year, -> { where(itt_academic_year: AcademicYear.current.to_s) }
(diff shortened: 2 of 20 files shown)


db/migrate/20240910135453_add_journey_to_reminders.rb
```diff
@@ -0,0 +1,19 @@
+class AddJourneyToReminders < ActiveRecord::Migration[7.0]
+  def up
+    add_column :reminders, :journey, :string
+
+    # Additional payments journey was not open when migration created, however
+    # the FurtherEducationPayments journey was opened for the first time on
+    # 2024/9/9 and so some reminders may also be set for the
+    # FurtherEducationPayments journey
+    Reminder.not_yet_sent.where("created_at < ?", Date.new(2024, 9, 9)).update_all(journey: Journeys::AdditionalPaymentsForTeaching)
```
rjlynch, line 9: Do we want to scope this by `not_yet_sent`? What about reminders in the DB that have been sent already?

rjlynch (approved):
Nice 👍 
Left one comment WRT migration, but have approved in case it's not actually an issue

app/models/reminder.rb
```diff
@@ -4,9 +4,14 @@ class Reminder < ApplicationRecord
 
   scope :email_verified, -> { where(email_verified: true) }
   scope :not_yet_sent, -> { where(email_sent_at: nil) }
+  scope :by_journey, ->(journey) { where(journey: journey.to_s) }
   scope :inside_academic_year, -> { where(itt_academic_year: AcademicYear.current.to_s) }
   scope :to_be_sent, -> { email_verified.not_yet_sent.inside_academic_year }
 
+  def journey
+    super.constantize
+  end
```
asmega, line 13: not sure how i feel about this and feel it might cause subtle issues. much like its common to have a column `user_id` which then backs a `record#user` method with an association. maybe the column should be `journey_class` and the method `#journey` rather an overloading the single method. 

happy to be overruled on this though

but yes definitely better to store type on the Reminder to deal with the different journeys

db/migrate/20240910135453_add_journey_to_reminders.rb
slorek: Good catch.

app/forms/reminders/confirmation_form.rb
```diff
@@ -2,6 +2,7 @@ module Reminders
   class ConfirmationForm < Form
     def reminder
       @reminder ||= Reminder.find_by(
+        journey: Journeys.for_routing_name(journey_session.journey).to_s,
```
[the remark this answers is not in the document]
slorek: Yes it complains about type casting otherwise.

app/models/reminder.rb
slorek: Good feedback, will implement your suggestions.

asmega:
I'm now closing this PR as the features in this PR should have been incorporated as part of https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/3436

PR #3176, opened 2024-09-10
LUPEYALPHA-1015 - FE claims with no TRN aren't considered matching other claims with no TRN

kenfodder:
* TRN is optional on FE claims.
* The matching logic didn't exclude fields if they are blank, so all FE claims with a blank TRN are considered a duplicate.
* This excludes querying attributes if the source claim value for that attribute is blank.

### DIFF
--- a/app/models/claim/matching_attribute_finder.rb
+++ b/app/models/claim/matching_attribute_finder.rb
@@ -27,7 +27,6 @@ class Claim
       # Claim attributes
       CLAIM_ATTRIBUTE_GROUPS_TO_MATCH.each do |attributes|
         vals = values_for_attributes(@source_claim, attributes)
-
         next if vals.blank?
 
         concatenated_columns = "CONCAT(#{attributes.join(",")})"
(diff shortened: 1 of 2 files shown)


app/models/claim/matching_attribute_finder.rb
```diff
@@ -38,13 +38,16 @@ def matching_claims
 
       # Eligibility attributes
       eligibility_ids = eligibility_attributes_groups_to_match.map { |attributes|
-        vals = values_for_attributes(@source_claim.eligibility, attributes)
-        concatenated_columns = "CONCAT(#{attributes.join(",")})"
+        present_attributes = attributes_for(@source_claim.eligibility, attributes).reject { |_, v| v.blank? }.map(&:first)
+        next if present_attributes.empty?
+
+        vals = values_for_attributes(@source_claim.eligibility, present_attributes)
+        concatenated_columns = "CONCAT(#{present_attributes.join(",")})"
```
alkesh, line 45: is this the same as having the original code, and just adding:

```
next if vals.empty?
```

alkesh (approved):

PR #3177, opened 2024-09-10
[LUPEYALPHA-1020] FE provision search errors on no results

asmega:
# Context 

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-1020
- Show error when no search results

# Changes

- There seem to be some bugs around going back changing answers and supporting both JS and non-JS seem to be quirky these should now be ironed out

# Screenshots

Screenshot 2024-09-10 at 16 22 27


rjlynch (approved):
Looks good 👍

PR #3180, opened 2024-09-11
[LUPEYALPHA-1036] Admin One Login IDV

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-1036
- Admin for OL identity confirmation

# Changes

- If user used OL, display their OL IDV data and opposing data from their claim in ID task
- If data matches it auto passes
- if data does not match it fails
- also on claim submission we do the same check and if it fails do not send the provider the verification email

# Screenshots

Screenshot 2024-09-18 at 15 39 09

---

Screenshot 2024-09-18 at 15 39 19

### DIFF
--- a/app/models/automated_checks/claim_verifiers/identity.rb
+++ b/app/models/automated_checks/claim_verifiers/identity.rb
@@ -17,8 +17,12 @@ module AutomatedChecks
       def perform
         return unless awaiting_task?(TASK_NAME)
 
-        # Order of matching matters so that subsequent conditions in methods fall through to execute the right thing
-        one_login || no_match || partial_match || complete_match
+        if claim.identity_confirmed_with_onelogin?
+          one_login
+        else
+          # Order of matching matters so that subsequent conditions in methods fall through to execute the right thing
+          no_match || partial_match || complete_match
+        end
       end
 
       private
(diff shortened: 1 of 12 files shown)


app/models/automated_checks/claim_verifiers/identity.rb
```diff
@@ -18,22 +18,21 @@ def perform
...
-        when :pass
-          create_task(match: nil, passed: true)
-        when :fail
-          create_task(match: nil, passed: false)
-        when :skip
-          nil
+      def one_login
+        if claim.identity_confirmed_with_onelogin?
+          if claim.onelogin_idv_full_name.downcase == claim.full_name.downcase && claim.onelogin_idv_date_of_birth == claim.date_of_birth
+            create_task(match: nil, passed: true)
```
asmega, line 32: not sure if task needs other data on it? eg should `match` be populated?

app/models/automated_checks/claim_verifiers/identity.rb
```diff
@@ -17,23 +17,24 @@ def initialize(
...
       private
 
       attr_accessor :admin_user, :claim
       attr_reader :dqt_teacher_status
 
-      def auto_pass
-        case claim.policy.auto_pass_identity_confirmation_task(claim)
-        when :pass
+      def one_login
+        if claim.onelogin_idv_full_name.downcase == claim.full_name.downcase && claim.onelogin_idv_date_of_birth == claim.date_of_birth
```
rjlynch, line 34: Probably one for product, but if one login only returns first + last name should this comparison be against `Claim#first_name` and `Claim#surname` rather than `Claim#full_name` which includes middle name?

rjlynch (approved):
Looks good 👍

rjlynch (approved):
Latest changes look good to me 👍

PR #3183, opened 2024-09-12
[LUPEYALPHA-1031] Cope with student loan data with "No data" rows. Remove invalid match logic

alkesh:
https://dfedigital.atlassian.net/browse/LUPEYALPHA-1031

No need for the old claim verifier logic - was used to match student loan answers in the journey, which have since been removed.
Also, the student loan data file can now include rows with "No data" values, which indicates that the SLC does not have information for the given claimants, which confirms they do not have a student loan.

Main changes:
* on the claim and answers, `has_student_loan?` is `nil` if there's no SLC data, `true` if there is data with a plan, `false` if there is data with "No data" for the plan
* `student_loan_plan` is `nil` if there's no SLC data, and if there is data, it is either the correct plan(s) or "not_applicable" if they don't have a student loan
* eligibility for the TSLR journey copes with both having a student loan with zero amount (the existing scenario), and having no student loan (the new scenario)
* The automated verification for the student loan plan task, will no be "No data" if there is no SLC data, or "Match" for when there is.

### DIFF
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -144,9 +144,9 @@ class Claim < ApplicationRecord
 
   validates :academic_year_before_type_cast, format: {with: AcademicYear::ACADEMIC_YEAR_REGEXP}
 
-  validates :has_student_loan, on: [:"student-loan"], inclusion: {in: [true, false]}
+  validates :has_student_loan, on: [:"student-loan"], inclusion: {in: [true, false]}, allow_nil: true
   validates :student_loan_plan, inclusion: {in: STUDENT_LOAN_PLAN_OPTIONS}, allow_nil: true
-  validates :student_loan_plan, on: [:"student-loan", :amendment], presence: {message: "Enter a valid student loan plan"}
+  validates :student_loan_plan, on: [:amendment], presence: {message: "Enter a valid student loan plan"}
 
   # TODO: remove when a form object is created for email-address
   validates :email_address, on: [:submit], presence: {message: "Enter an email address"}
--- a/app/models/claim_student_loan_details_updater.rb
+++ b/app/models/claim_student_loan_details_updater.rb
@@ -9,9 +9,13 @@ class ClaimStudentLoanDetailsUpdater
 
   def update_claim_with_latest_data
     claim.transaction do
-      eligibility.update!(eligibility_student_loan_attributes) if claim.has_tslr_policy?
+      if claim.has_tslr_policy?
+        eligibility.update!(eligibility_student_loan_attributes)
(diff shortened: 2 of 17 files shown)


app/models/claim_student_loan_details_updater.rb
```diff
@@ -32,11 +36,18 @@ def eligibility_student_loan_attributes
 
   def claim_student_loan_attributes
     {
-      has_student_loan: student_loans_data.found_data?,
+      has_student_loan: student_loans_data.has_student_loan?,
       student_loan_plan: student_loans_data.student_loan_plan
     }
   end
 
+  def claim_student_loans_policy_student_loan_attributes
```
alkesh, line 44: I didn't want to change these values for the TSLR journey - as it impacts how that journey works.

spec/features/admin/upload_slc_data_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+
+    # Student Loans
+
+    click_link sl_claim_with_slc_data_no_student_loan.reference
+    expect(page).to have_content "Student loan amount"
+    within "li.student_loan_amount" do
+      expect(page).to have_content "No match"
+    end
+    expect(sl_claim_with_slc_data_no_student_loan.reload.student_loan_plan).to eq "not_applicable"
+    expect(sl_claim_with_slc_data_no_student_loan.has_student_loan).to be true # is this correct?
```
alkesh, line 60: need to check if this has any impact on payroll

app/models/claim.rb
```diff
@@ -144,9 +144,9 @@ class Claim < ApplicationRecord
 
   validates :academic_year_before_type_cast, format: {with: AcademicYear::ACADEMIC_YEAR_REGEXP}
 
-  validates :has_student_loan, on: [:"student-loan"], inclusion: {in: [true, false]}
+  validates :has_student_loan, on: [:"student-loan"], inclusion: {in: [true, false]}, allow_nil: true
```
rjlynch, line 147: probably too much of a faff to change but this should eventually be converted to an `enum` or something if it's representing more than 2 states

app/models/policies/student_loans/policy_eligibility_checker.rb
```diff
@@ -67,7 +67,10 @@ def not_taught_enough?
         answers.mostly_performed_leadership_duties == true
       end
 
+      # checks two scenarios: (1) they do not have a student loan, (2) they have a student loan but the repayment amount is zero
```
rjlynch, line 70: 👍

rjlynch (approved):
Looks good 👍

PR #3188, opened 2024-09-13
[LUPEYALPHA-1002] Revamp cookies

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-1002
- Update cookies

# Changes

- Remove superfluous fluff from README
- Remove custom JS code that manages cookies client/browser side. Cookies are now all managed by rails side for the cookies journey
- Use GOVUK components for cookie banner, removing custom styling from our app
- fixed leakage of `noscript` script


app/views/layouts/application.html.erb
```diff
@@ -31,28 +33,22 @@
...
-    <div id="global-cookie-message" class="govuk-cookie-banner" role="region" aria-label="cookie banner">
-      <div class="govuk-width-container">
-        <p class="govuk-cookie-banner__message"><%= journey_service_name(current_journey_routing_name) %> uses cookies to make the site simpler.</p>
-        <div class="govuk-cookie-banner__buttons">
-          <button id="accept-cookies" class="govuk-button" type="submit" data-module="govuk-button">Accept cookies</button>
-          <%= link_to('Cookie information', cookies_path(current_journey_routing_name), class: "govuk-cookie-banner__link" ) %>
-        </div>
-      </div>
-    </div>
+    <a href="#main-content" class="govuk-skip-link">Skip to main content</a>
```
asmega, line 51: documentation [1] dictates cookies banner should appear before the main content

[1] https://design-system.service.gov.uk/components/cookie-banner/#how-it-works

app/controllers/static_pages_controller.rb
```diff
@@ -5,7 +5,7 @@ def accessibility_statement
   def contact_us
   end
 
-  def cookies
+  def cookies_page
```
asmega, line 8: this is so that calling `cookies` from the view calls the rails implementation rather than this controller action method

app/views/layouts/application.html.erb
```diff
@@ -31,28 +33,22 @@
...
-    <% if ENV["GOOGLE_ANALYTICS_ID"] %>
-      <noscript>
-        <iframe
-          src="https://www.googletagmanager.com/ns.html?<%= ENV["GTM_ANALYTICS"] %>"
-          height="0"
-          width="0"
-          style="display:none;visibility:hidden">
-        </iframe>
-      </noscript>
+    <% if cookies_accepted? %>
```
asmega, line 36: fixes leakage

spec/views/layouts/application.html.erb_spec.rb
```diff
@@ -0,0 +1,37 @@
+require "rails_helper"
+
+RSpec.describe "layouts/application.html.erb" do
```
asmega, line 3: first view spec, actually caught a bug with the `noscript` leakage

rjlynch (approved):
Nice 👍 Moving all the cookie setting logic server side is great 👍

app/controllers/cookies_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+  def hide
+    state = JSON.parse(cookies.encrypted[:accept_cookies])["state"]
+
+    cookies.encrypted[:accept_cookies] = {
+      value: {state:, message: false}.to_json,
+      expires: 90.days.from_now
+    }
+
+    redirect_to request.env["HTTP_REFERER"]
+  end
```
rjlynch, line 37: (not a blocker) could these actions just use `update` passing different params?

asmega: could be, i think it would shift more code to the front. i don't think it matters hugely and expect this code to be pretty stagnant

PR #3193, opened 2024-09-17
Capitalize trn and ni

rjlynch:
When rendering the teacher reference number or the national insurance
number in the matching details list we want to make sure they're
titlized


spec/helpers/admin/claims_helper_spec.rb
```diff
@@ -198,12 +198,12 @@
     subject { helper.matching_attributes(first_claim, second_claim) }
 
     it "returns the humanised names of the matching attributes" do
```
alkesh, line 200:
```suggestion
    it "returns the matching attributes" do
```

rjlynch:
This PR is no longer needed

"just checked with Ashleigh and she's confirmed that the convention for NiNo and TRN is correct as it is currently coded and should not be capitalised."

PR #3196, opened 2024-09-17
Fix find or create for provider

rjlynch:
There's a uniq index on the dfe sign in id, so we need to only find by
the dfe sign in id, not by the other attributes


rjlynch:
Fixes this issue https://app.rollbar.com/a/dfe-digital/fix/item/dfe-claims/1163?utm_campaign=item_velocity_message&utm_medium=slack&utm_source=rollbar-notification

alkesh (approved):

PR #3199, opened 2024-09-17
FE provider chaser email

kenfodder:
* Add a timestamp to when the last provider email was sent
* Checks for claims with a provider email last sent over 3 weeks ago and sends a chase email
* It updates that timestamp so it will send again in 3 weeks
* Daily job running at 8am

UPDATE

* It's updated to the send the chase email ONLY ONCE - clarified with Product
* Already verified, held and rejected claims are excluded from being send a chaser email.


slorek (changes requested):
I think we need some mechanism to cap the number of emails sent. I think the implementation here will send emails every 3 weeks forever, even if the claim is rejected. We should probably at least not include rejected claims, and consider a cap to the number of emails which are sent in total.

PR #3200, opened 2024-09-17
[LUPEYALPHA-790] Email timeout (magic link expiry)

vacabor:
https://dfedigital.atlassian.net/browse/LUPEYALPHA-790
https://dfedigital.atlassian.net/browse/LUPEYALPHA-1056

image

Adds a screen to help the user re-enter their email address if the magic link they clicked on has expired.

### DIFF
--- a/app/models/policies/international_relocation_payments.rb
+++ b/app/models/policies/international_relocation_payments.rb
@@ -54,10 +54,8 @@ module Policies
       start_of_academic_year - 2.years
     end
 
-    # NOTE RL: currently IRP only has a single reply to address, so notify
-    # doesn't show the address id
     def notify_reply_to_id
-      nil
+      "b11c0da5-f976-4cc2-8464-23e4dda63fc4"
     end
 
     def award_amount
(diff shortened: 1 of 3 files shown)


app/controllers/claims_controller.rb
```diff
@@ -121,6 +121,8 @@ def handle_magic_link
       journey_session.answers.assign_attributes(email_address: params[:email])
       journey_session.answers.assign_attributes(email_verified: true)
       journey_session.save!
+    else
```
alkesh, line 124: So the expired-link page will be shown if the OTP code is not expired, but just not valid - not sure if this is too much of an edge case, as it would require someone to construct their own magic link. It seems an unlikely scenario, and showing the expired-link page in this case seems fine.

alkesh (approved):

app/views/early_years_payment/provider/start/claims/expired_link.html.erb
```diff
@@ -0,0 +1,22 @@
...
+    <h1 class="govuk-heading-l">
+      This link has expired
+    </h1>
+
+    <p class="govuk-body">
+      Select the link below to send another link to the same email address.
+    </p>
+
+    <p class="govuk-body">
+      The link will expire in 15 minutes and can only be used once.
```
alkesh, line 15: technically, the link can be used multiple times within that 15 minutes.

PR #3205, opened 2024-09-19
[CAPT-1792] Admin IRP summary view tweak

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-1792
- In admin IRP summary view, show passport number instead of TRN

# Screenshots

Screenshot 2024-09-19 at 12 53 57


app/helpers/admin/claim_summary_helper.rb
```diff
@@ -1,8 +1,14 @@
 module Admin
   module ClaimSummaryHelper
     def claim_summary_view
-      policy_name_to_append = @claim.policy.to_s.underscore if @claim.policy.is_a?(Policies::FurtherEducationPayments)
-      ["claim_summary", policy_name_to_append].compact.join("_")
+      case @claim.policy
```
slorek, line 4: I think the previous implementation had merit so this doesn't have to be explicitly configured for every policy, but happy to go with it to speed this along.

slorek (approved):

app/helpers/admin/claim_summary_helper.rb
asmega: this implementation still holds that characteristic. any new or existing policy that is not FE or IRP will fall into the catchall therefore no additional configuration is required

asmega: if you want full zero configuration neither this implementation nor the old implementation has that characteristic and would be new feature. that is it would have to check if the partial exists otherwise fallback to the catchall

PR #3207, opened 2024-09-19
[LUPEYALPHA-1004] New status for FE claims: Awaiting provider verification

alkesh:
https://dfedigital.atlassian.net/browse/LUPEYALPHA-1004

### DIFF
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -218,6 +218,9 @@ class Claim < ApplicationRecord
         .or(Claim.where("further_education_payments_eligibilities.flagged_as_duplicate = TRUE").and(Claim.where(notes: {label: "provider_verification"})))
       )
   end
+  scope :not_awaiting_further_education_provider_verification, -> do
+    where.not(id: Claim.awaiting_further_education_provider_verification)
+  end
 
   def onelogin_idv_full_name
     "#{onelogin_idv_first_name} #{onelogin_idv_last_name}"
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -447,10 +447,7 @@ class Claim < ApplicationRecord
   def awaiting_provider_verification?
     return false unless has_further_education_policy?
 
-    return false if eligibility.verified?
-
-    # TODO - duplication with app/models/policies/further_education_payments/admin_provider_verification_task_presenter.rb
-    !eligibility.flagged_as_duplicate? || notes.where(label: "provider_verification").any?
+    eligibility.awaiting_provider_verification?
   end
 
   private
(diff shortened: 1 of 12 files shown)


app/models/claim.rb
```diff
@@ -431,8 +433,24 @@ def same_claimant?(other_claim)
     end
   end
 
+  def awaiting_provider_verification?
+    return false unless has_fe_policy?
+
+    return false if eligibility.verification.present?
```
rjlynch, line 439: Slight nitpick but could we use
```suggestion
    return false if eligibility.verified?
```
so we keep the definition of "being verified" in one place?

app/models/claim.rb
```diff
@@ -431,8 +433,24 @@ def same_claimant?(other_claim)
...
+
+    if tasks.where(name: "matching_details").any?
+      tasks.where(name: "matching_details", passed: true).any?
+    else
+      Claim::MatchingAttributeFinder.new(self).matching_claims.empty?
+    end
```
rjlynch, line 445: I know this is the logic described on the ticket but I wonder if what they _really_ want to know is if the provider verification email has been sent out and we're waiting on a response 🤔 if so we could move the logic from
`Policies::FurtherEducationPayments::AdminProviderVerificationTaskPresenter#verification_email_sent?` on to the eligibility then use that method here.

spec/support/admin_view_claim_feature_shared_examples.rb
```diff
@@ -42,7 +42,11 @@
   }
 
   let!(:approved_awaiting_payroll_claim) {
-    eligibility = create(:"#{policy.to_s.underscore}_eligibility", :eligible)
+    eligibility = if policy == Policies::FurtherEducationPayments
+      create(:"#{policy.to_s.underscore}_eligibility", :verified)
+    else
+      create(:"#{policy.to_s.underscore}_eligibility", :eligible)
```
rjlynch, line 48: Really hate shared example specs 🙃 
I was going to suggest aliasing `eligible` to `verified` in the FE eligibility factory but they kind of have different meanings

rjlynch (approved):
Looks good to me, left one comment that might be worth checking with the product team to see what they're trying to achieve with the new status

app/models/claim.rb
alkesh: I've asked on the ticket. If we change to using `verification_email_sent`, then the scope `awaiting_further_education_provider_verification` will have to change to check notes (I have a query ready that works).

rjlynch (approved):
Looks good 👍 
I might be tempted to put a scope on FE eligibility `awaiting_provider_verification` and then simplify the scope on the `Claim` to just merge the FE scope in, but I think it's fine as it

spec/support/admin_view_claim_feature_shared_examples.rb
alkesh: I think you're right.
The `eligible` trait should also be verified - I'm going to make a separate trait of `not_verified` for the other tests.

PR #3222, opened 2024-09-24
LUPEYALPHA 1096/fix claim started date 2

rjlynch:
Back fill started at

If the claim has a journey session we try and set the `started_at` to
that, if it doesn't have a journey session we'll just use the
`created_at` timestamp, which for claims pre the introduction of journey
sessions will be the started at time.

PR 2 of 3
1 https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/3221
3 https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/3224

### DIFF
--- a/db/migrate/20240924095435_back_fill_claims_started_at.rb
+++ b/db/migrate/20240924095435_back_fill_claims_started_at.rb
@@ -1,0 +1,22 @@
+class BackFillClaimsStartedAt < ActiveRecord::Migration[7.0]
+  def up
+    execute <<-SQL
+      UPDATE claims
+      SET started_at = journeys_sessions.created_at
+      FROM journeys_sessions
+      WHERE claims.journeys_session_id = journeys_sessions.id
+      AND claims.started_at IS NULL
+    SQL
+
+    execute <<-SQL
+      UPDATE claims
+      SET started_at = created_at
+      WHERE journeys_session_id IS NULL
+      AND claims.started_at IS NULL
+    SQL
+  end
+
+  def down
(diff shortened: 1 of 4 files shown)


db/migrate/20240924095435_back_fill_claims_started_at.rb
```diff
@@ -0,0 +1,24 @@
...
+    SQL
+
+    execute <<-SQL
+      UPDATE claims
+      SET started_at = created_at
+      WHERE journeys_session_id IS NULL
+      AND claims.started_at IS NULL
+    SQL
+
+    change_column_null :claims, :started_at, false
```
asmega, line 18: feels risky without a prior run? might save you more time if we do this change after confirmation? or happy to keep in if you're feeling confident, i don't think there's any material risk

asmega (approved):

db/migrate/20240924095435_back_fill_claims_started_at.rb
rjlynch: We're setting this on new claims in the previous PR and we're using the created_at timestamp which should be present on all records as rails is setting it. But yeah I think for avoiding risk we'll split this out into a separate pr (and also because I forgot to remove the not null constraint in the down migration 🙈)

rjlynch: Have moved the migration to make the column non nullable to this pr https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/3224/files

PR #3229, opened 2024-09-25
Fix flakey test

rjlynch:
Sometimes the string "FE" ends up in the generated claim reference, eg "XBCFEA", and so is visible in the table more than twice. To fix this we need to specifically look in the policy short names column.

### DIFF
--- a/spec/features/admin/admin_claims_filtering_spec.rb
+++ b/spec/features/admin/admin_claims_filtering_spec.rb
@@ -49,10 +49,10 @@ RSpec.feature "Admin claim filtering" do
 
     click_on "View claims"
 
-    expect(page.find("table")).to have_content("ECP").exactly(10).times
-    expect(page.find("table")).to have_content("TSLR").exactly(7).times
-    expect(page.find("table")).to have_content("STRI").exactly(2).times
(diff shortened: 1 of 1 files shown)


spec/features/admin/admin_claims_filtering_spec.rb
```diff
@@ -49,10 +49,12 @@
 
     click_on "View claims"
 
-    expect(page.find("table")).to have_content("ECP").exactly(10).times
-    expect(page.find("table")).to have_content("TSLR").exactly(7).times
-    expect(page.find("table")).to have_content("STRI").exactly(2).times
-    expect(page.find("table")).to have_content("FE").exactly(2).times
```
alkesh, line 55: I came across this recently too on my ticket.
My solution was:
```
expect(page).to have_selector("td[text()='ECP']", count: 10)
```
The only advantage to this is that it copes if columns are added/removed from this table, as it doesn't assume a column number.

alkesh (approved):

spec/features/admin/admin_claims_filtering_spec.rb
rjlynch: Yeah that's nicer, have changed

PR #3231, opened 2024-09-25
[LUPEYALPHA-1101] remove radio buttons from Student Loan Plan task

alkesh:
student loan plan task is always automated now, no option for manual answering.

Added missing Admin Checks feature, which we had for the other journeys.


rjlynch (approved):
LGTM 👍

PR #3234, opened 2024-09-26
[LUPEYALPHA-983] EY practitioner personal details

alkesh:


alkesh:
hmm - just noticed - the headings are one size too big compared to the prototype.
The main heading should large instead of xlarge, and the sub-headings should be medium.

rjlynch (approved):
👍

PR #3235, opened 2024-09-26
[LUPEYALPHA-984] EY practitioner enter home address

alkesh:
basing off `LUPEYALPHA-983-EY-practitioner-personal-details` branch


alkesh:
> LGTM, only question is how come we can't just render the address form for this journey? Is it because we don't want to include the "questions.personal_details" caption?

yeah - there are 2 changes required for this journey - no caption, and also the title is size 'l' instead of 'xl'.

PR #3237, opened 2024-09-26
[LUPEYALPHA-1045] EY practitioner find reference screen

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-1045
- Adds EY practitioner find reference screen

# Changes

- In order to support email in url query params and the way the current controller/journey concerns does a lot redirect, I've had to make tweaks to ensure these get passed around so they are not lost. This will affect all journeys
- I have concern where query params are stripped, thinking of mostly Safari where it can strip certain query params, may cause this feature to not work as we use it to pass state
- Updated email sent to practitioner to inject correct url to continue their journey

### DIFF
--- a/app/views/early_years_payment/practitioner/claims/find_reference.html.erb
+++ b/app/views/early_years_payment/practitioner/claims/find_reference.html.erb
@@ -1,2 +1,43 @@
-<p class="govuk-body">find reference page goes here</p>
-<%= govuk_button_to "Continue", claim_path(current_journey_routing_name), method: :patch %>
+<% content_for(:page_title, page_title(@form.t(:question), journey: current_journey_routing_name, show_error: @form.errors.any?)) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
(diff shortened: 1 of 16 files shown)


app/views/early_years_payment/practitioner/claims/find_reference.html.erb
```diff
@@ -1,2 +1,45 @@
...
+          If you can’t find the email:
+        </p>
+
+        <%= govuk_list [
+          "check your spam or junk folder",
+          "ask your employer to confirm they have sent your application. They will have received the reference number in their confirmation email."
+        ], type: :bullet %>
+
+        <p>
+          Contact us at <%= govuk_link_to "help@opsteam.com", "mailto:help@opsteam.com" %> for assistance.
```
alkesh, line 38:
```suggestion
          Contact us at <%= govuk_mail_to "help@opsteam.com" %> for assistance.
```
also, is it worth putting this email in config/locales? we have the emails for the provider journey in there.

asmega: yep agreed, i've stuck this on the last commit

alkesh (approved):

PR #3250, opened 2024-10-01
CAPT 983/slow loading payroll

rjlynch:
wip


app/models/payroll_run.rb
```diff
@@ -78,27 +80,72 @@ def download_triggered?
...
+        LEFT JOIN international_relocation_payments_eligibilities
+          ON claims.eligibility_id = international_relocation_payments_eligibilities.id
+          AND claims.eligibility_type = 'Policies::InternationalRelocationPayments::Eligibility'
+        LEFT JOIN levelling_up_premium_payments_eligibilities
+          ON claims.eligibility_id = levelling_up_premium_payments_eligibilities.id
+          AND claims.eligibility_type = 'Policies::LevellingUpPremiumPayments::Eligibility'
+        LEFT JOIN student_loans_eligibilities
+          ON claims.eligibility_id = student_loans_eligibilities.id
+          AND claims.eligibility_type = 'Policies::StudentLoans::Eligibility'
+      )
```
rjlynch, line 121: If we migrate `award_amount` to claim we can lose this CTE

app/controllers/admin/payroll_runs_controller.rb
```diff
@@ -33,7 +33,7 @@ def create
 
     # NOTE: Optimisation - preload payments, claims and eligibility
```
slorek, line 34:
```suggestion
```

app/models/base_policy.rb
```diff
@@ -58,4 +58,12 @@ def further_education_payments?
   def auto_check_student_loan_plan_task?
     false
   end
+
+  def award_amount_column
+    if to_s == "StudentLoans"
```
slorek, line 63: Perhaps the `StudentLoans` model should override this method instead

slorek (approved):

app/models/base_policy.rb
rjlynch: Yeah that's nicer, have fixed 👍

PR #3251, opened 2024-10-01
[LUPEYALPHA-1112] When no entry in SLC data, do not create task (shows as incomplete on admin view)

alkesh:
All journeys now behave the same (except TSLR which does not have a student loan plan task):
* when the claim is submitted, and there is no SLC data for the claimant, the task shows as 'Incomplete'
* when the SLC data is subsequently uploaded, the 'Incomplete' tasks will show as 'Passed' or 'Incomplete' depending if a SLC data entry has been found for that claimant
* when the claim is submitted, and there is SLC data for the claimant, the task does not show, but the plan is set correctly

For the 'Incomplete' student loan plan tasks, I've added the notes back, as that shows if the automated check was carried out (it used to show on the old 'No data' tasks).

We don't show 'No data' for student loan plan tasks any more - it's either 'Incomplete' or 'Passed'.

### DIFF
--- a/spec/jobs/student_loan_plan_check_job_spec.rb
+++ b/spec/jobs/student_loan_plan_check_job_spec.rb
@@ -7,39 +7,32 @@ RSpec.describe StudentLoanPlanCheckJob do
     create(:journey_configuration, :further_education_payments)
   end
 
-  let(:claim) { create(:claim, claim_status, academic_year:, policy: Policies::LevellingUpPremiumPayments) }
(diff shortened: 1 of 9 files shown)


spec/jobs/student_loan_plan_check_job_spec.rb
alkesh, line 1: I refactored this spec - it was testing too much, checking for tasks and notes, and the wording was confusing.
It now just checks if the `ClaimStudentLoanDetailsUpdater` and `AutomatedChecks::ClaimVerifiers::StudentLoanPlan` are called correctly.

PR #3252, opened 2024-10-01
Fix claims being flagged for qa

rjlynch:
When a new academic year has started, and we are yet to approved any claims from this academic year, all claims from the previous year are flagged for qa when approved. To avoid this we need to count the number of approved claims in the _claim's_ academic year rather than the current academic year.


cwadsworthdfe:
Thanks for jumping on this @rjlynch

PR #3253, opened 2024-10-01
[LUPEYALPHA-1109] Bug: Clear the employee-email slug on form failure so it is not bypassed

vacabor:
This is an edge case but if the user:
- Fills in a claim up until employee email
- Submits a blank employee email (causing a validation error)
- Clicks the back button to change the employee's contract type
- Selects any eligible value
- They were taken to the "check your answers" screen instead of to "employee email" as the email slug was considered already visited

Fixes the underlying cause of #3247.

### DIFF
--- a/app/controllers/admin/amendments_controller.rb
+++ b/app/controllers/admin/amendments_controller.rb
@@ -1,7 +1,7 @@
 class Admin::AmendmentsController < Admin::BaseAdminController
   before_action :load_claim
   before_action :ensure_service_operator
-  before_action :ensure_claim_is_amendable
+  before_action :ensure_claim_is_amendable, only: [:new, :create]
 
   def index
   end
--- a/spec/features/admin/admin_amend_claim_spec.rb
+++ b/spec/features/admin/admin_amend_claim_spec.rb
@@ -186,4 +186,29 @@ RSpec.feature "Admin amends a claim" do
       expect(page).to have_content("Claim has been amended successfully")
     end
   end
+
+  context "when claim is no longer amendable" do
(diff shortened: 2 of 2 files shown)


asmega (approved):
obviously outside the scope of this PR but think this whole mechanism needs an overhaul
- `session[:slugs]` should go
- these arbitrary controller callbacks
- the way we navigate forwards/backs in journeys

think it would really help prevent quirks like this and make life easier

app/controllers/claims_form_callbacks.rb
```diff
@@ -73,6 +73,11 @@ def check_your_answers_after_form_save_success
     create_and_save_claim_form
   end
 
+  def employee_email_after_form_save_failure
+    session[:slugs].delete("employee-email")
+    render_template_for_current_slug
+  end
```
alkesh, line 79: Are we not just missing a validation somewhere?
If the `practitioner_email_address` is blank, than the claim should not be valid, and the ClaimSubmissionForm should not be valid.
I can't see how to add this validation to just the EY Authenticated journey - so actually, this callback may be the best solution.

alkesh: Either way, I think this needs a feature spec in `spec/features/early_years_payment/provider/authenticated`

alkesh: 👍🏽 happy with it as it is for now.
My worry about a lack of feature spec, is when we refactor, we re-introduce this bug.

alkesh (approved):

PR #3254, opened 2024-10-02
dfe-analytics upgrade to v1.14.2

slorek:
https://github.com/DFE-Digital/dfe-analytics/blob/main/CHANGELOG.md

Introduces a new config file to support new PII handling in BigQuery. Until assurance has been completed we do not yet need to consider sending any PII, however.


asmega (approved):
Just to clear any confusion `config/analytics_hidden_pii.yml` looks to have been already been introduced prior to this version bump

PR #3256, opened 2024-10-02
[LUPEYALPHA-1121] Update find ref copy

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-1121
- Update copy on EY find reference screen
- Also fixed flakey JS test


rjlynch (approved):
👍 - Even though `find` supposedly waits for 2 seconds that never seems to work in my experience so I think the `sleep` is fine 👍

PR #3262, opened 2024-10-02
[CAPT-1832] enable reminders for LUP

alkesh:
need to use a separate `FINAL_LUP_POLICY_YEAR` for reminders instead of `FINAL_COMBINED_ECP_AND_LUP_POLICY_YEAR`


rjlynch (approved):
LGTM 👍 
This looks like a tricky one to have figured out!

PR #3269, opened 2024-10-03
CAPT 1834/claim allocations dropdown

rjlynch:
Updates the admin UI providing a select box for allocating a specific
number of claims.


asmega (approved):
I know you love form objects so think one would have worked well here

PR #3271, opened 2024-10-03
Admin stats - display FE provider verification auto-passed count

kenfodder:
Screenshot 2024-10-03 at 17 36 20


rjlynch (approved):
👍

alkesh (approved):
probably needs an entry in `spec/features/admin/admin_stats_spec.rb`, but that can be added later.

PR #3284, opened 2024-10-08
[CAPT-1797] Ops reports

alkesh:
* new nav bar entry on the admin site: Reports
* reports page shows previously ran reports
* reports stored in database
* ReportsJob set to run on second Tuesday of each month
* Report classes based on the existing `Claim::DataReportRequest` class

### DIFF
--- a/app/models/reports/duplicate_claims.rb
+++ b/app/models/reports/duplicate_claims.rb
@@ -24,24 +24,37 @@ module Reports
     def to_csv
       CSV.generate(write_headers: true, headers: HEADERS) do |csv|
         @claims.each do |claim|
-          csv << row(
-            claim.reference,
-            claim.eligibility.try(:teacher_reference_number),
-            claim.full_name,
-            claim.policy,
-            claim.award_amount,
-            status(claim),
-            claim.latest_decision&.created_at,
-            claim.latest_decision&.created_by&.full_name
-          )
+          csv << ClaimPresenter.new(claim).to_a
         end
       end
     end
 
     private
 
-    def row(*entries)
-      entries.map { |entry| ExcelUtils.escape_formulas(entry) }
+    class ClaimPresenter
+      include Admin::ClaimsHelper
+      include ActionView::Helpers::NumberHelper
+
+      def initialize(claim)
+        @claim = claim
+      end
+
+      def to_a
+        [
+          claim.reference,
+          claim.eligibility.try(:teacher_reference_number),
+          claim.full_name,
+          claim.policy,
+          claim.award_amount,
+          status(claim),
+          claim.latest_decision&.created_at,
+          claim.latest_decision&.created_by&.full_name
+        ]
+      end
+
+      private
+
+      attr_reader :claim
     end
   end
 end
--- a/app/models/reports/duplicate_claims.rb
+++ b/app/models/reports/duplicate_claims.rb
@@ -21,7 +21,6 @@ module Reports
       @claims = Claim
         .includes(:eligibility)
         .approved
-        .where.missing(:payments)
         .where(academic_year: AcademicYear.current)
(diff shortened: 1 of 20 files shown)


app/models/reports/duplicate_claims.rb
```diff
@@ -0,0 +1,47 @@
...
+      "Full name",
+      "Policy name",
+      "Claim amount",
+      "Claim status",
+      "Decision date",
+      "Decision agent"
+    ].freeze
+
+    def initialize
+      @claims = Claim.approved.select { |claim| Claim::MatchingAttributeFinder.new(claim).matching_claims.any? }
```
rjlynch, line 21:
```suggestion
      @claims = Claim.approved.includes(:eligibility).select { |claim| Claim::MatchingAttributeFinder.new(claim).matching_claims.any? }
```

Not sure if we want to include `decisions` too or if it's less work to filter first and make the additional trips to the db

app/models/reports/failed_provider_check_claims.rb
```diff
@@ -0,0 +1,72 @@
...
+      "Provider response: #{provider_verification_label("further_education_teaching_start_year")}",
+      "Provider response: #{provider_verification_label("teaching_hours_per_week")}",
+      "Provider response: #{provider_verification_label("half_teaching_hours")}",
+      "Provider response: #{provider_verification_label("subjects_taught")}",
+      "Provider response: #{provider_verification_label("taught_at_least_one_term")}",
+      "Provider response: #{provider_verification_label("teaching_hours_per_week_next_term")}"
+    ].freeze
+
+    def initialize
+      @claims = Claim.includes(:tasks)
```
rjlynch, line 33:
```suggestion
      @claims = Claim.includes(:tasks, :eligibility)
```

app/models/reports/failed_qualification_claims.rb
```diff
@@ -0,0 +1,77 @@
...
+      "Answered ITT start year",
+      "Answered ITT subject",
+      "DQT ITT subjects",
+      "DQT ITT start year",
+      "DQT QTS award date",
+      "DQT qualification name"
+    ].freeze
+
+    def initialize
+      @claims = Claim.includes(:tasks).where(tasks: {name: "qualifications", passed: false}).approved
```
rjlynch, line 23:
```suggestion
      @claims = Claim.includes(:tasks, :eligibility).where(tasks: {name: "qualifications", passed: false}).approved
```

db/migrate/20241007141638_create_reports.rb
```diff
@@ -0,0 +1,11 @@
+class CreateReports < ActiveRecord::Migration[7.0]
+  def change
+    create_table :reports, id: :uuid do |t|
+      t.string :name
+      t.text :csv
```
rjlynch, line 5: Nice approach 👍, very simple and gets the job done!

rjlynch (approved):
Looks good to me - would check with @slorek and see if this fits with his design for the app

rjlynch:
I've decided to use the background job based approach in this spike pr and have cleaned up the PR so it's good to go.
The reports don't need to be live so a background job generating them daily is fine (can be less frequent).
I think the Report model / storing the CSV in a column on the model taken in the spike is a good idea, v simple and does the job so kept that.

I've profiled the duplicate claim report against the production database.
Checking all approved claims for duplicates takes ~9minutes to run, which isn't bad for a background job.
Checking only approved unpaid claims takes ~11 seconds which is fine for a background job task.
Speaking with Dan, the main use of the duplicate claims report will be a final sense check before payroll so filtering out paid claims from this report seems fine. (The unpaid claims we're checking _are_ checked against paid claims for duplicates)

slorek (approved):

PR #3288, opened 2024-10-08
[CAPT-1614] Admin task pagination

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-1614
- Improve admins paginating thru tasks, so they can now paginate thru to `Decision` or `QA decision`
- The decision page now shows previous decisions rather than show full claim

# Changes

- There is a new admin `DecisonsController#index` action that displays all the previous decisions made on a claim in chronological order
- We now show this page when a decision has been made
- So when paginating thru tasks to `Decision` we now show the above page
- If QA is required we show next task as `QA decision` which displayed QA decision form
- if QA is completed we show all previous decisions

# Notes for review

- We have both a mixture of `tasks` and `decisions` from a code perspective though from a UI perspective these are all deemed as tasks. Why we have these 2 different concepts I do not know and is a decision made prior to my time.
- Given this future improvements could be to get rid of `decisions` altogether and convert these to tasks to homogenise everything
- OR write some kind of wrapper to do hide these concepts, which I feel is probably way too much effort for reward gained
- The concept of `decisions` are agnostic, there is no data so to signify if it was the original approval or a QA approval. as are result in this code and existing code there need to flags passed around to deal with this shortcoming. To remedy this we should probably attach the type of decision it was rather than guessing

### DIFF
--- a/app/controllers/admin/decisions_controller.rb
+++ b/app/controllers/admin/decisions_controller.rb
@@ -1,9 +1,14 @@
 class Admin::DecisionsController < Admin::BaseAdminController
   before_action :ensure_service_operator
   before_action :load_claim
-  before_action :reject_decided_claims, unless: -> { qa_decision_task? }
+  before_action :reject_decided_claims, only: [:new, :create], unless: -> { qa_decision_task? }
   before_action :reject_missing_payroll_gender, only: [:create]
   before_action :reject_if_claims_preventing_payment, only: [:create]
+
+  def index
+    @decisions = @claim.decisions.order(created_at: :asc)
+    @task_pagination = Admin::TaskPagination.new(claim: @claim, current_task_name:)
+  end
 
   def new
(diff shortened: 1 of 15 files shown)


rjlynch (approved):
I think this if fine and keeps all the weirdness well contained. I agree we need to do something to unify tasks and decisions.
Probably not the time to change how decisions and tasks work right before opening!

Could we introduce a "decide outcome" task that when submitted also creates the decision, then we wouldn't have to branch when deciding the next_task_path?

asmega:
> Could we introduce a "decide outcome" task that when submitted also creates the decision, then we wouldn't have to branch when deciding the next_task_path?

Not gonna do it here but yes that's probably a more elegant solution with a virtual/proxy task which passes onto a decision. it would defo solve the multi type issue but there still some other shortcomings too that would still need resolving to get this to a better place

PR #3292, opened 2024-10-09
[CAPT-1402] Back office can filter for auto approved claims

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-1402
- Back office can filter for auto approved claims


rjlynch (approved):
Looks good

PR #3293, opened 2024-10-09
[LUPEYALPHA-1075] Early Years Practitioner mobile number

alkesh:
* added mobile number page for EY practitioner journey
* fixed 'resend passcode' functionality for both email and mobile
* fixed bug - if claimant changed answer to "Would you like to provide your mobile number?" to "No" on the check-your-answsers screen, we would still incorrectly save the mobile number.


app/forms/mobile_number_form.rb
```diff
@@ -15,7 +15,6 @@ class MobileNumberForm < Form
 
   def save
     return false unless valid?
-    return true unless mobile_number_changed?
```
alkesh, line 18: I'm pretty sure this line meant that the 'resend passcode' functionality never worked.
Also, I can delete this line, and the spec still passes with no changes.

app/views/claims/_one_time_password.html.erb
```diff
@@ -27,9 +27,9 @@
 
       <div class="govuk-!-margin-bottom-6">
         <% if email_or_mobile == "email" %>
-          <%= govuk_link_to "Resend passcode (you will be sent back to the email address page)", claim_path(current_journey_routing_name, "email-address"), no_visited_state: true %>
+          <%= govuk_link_to "Resend passcode (you will be sent back to the email address page)", claim_path(current_journey_routing_name, "email-address", resend: true), no_visited_state: true %>
```
rjlynch, line 30:
```suggestion
          <%= govuk_link_to "Resend passcode (you will be sent back to the email address page)", claim_path(current_journey_routing_name, "email-address", claim: { resend: true }), no_visited_state: true %>
```

Does setting the resend params like this work and let us avoid having to set the `@resend` instance variable? If so we can lose the callbacks for setting `@resend` and not need to use `@resend` in the view.

rjlynch (approved):
LGTM but probably worth seeing if we can remove the `@resend` instance variable and lose the two callbacks - not a blocker but I think it should be pretty straightforward

app/views/claims/_one_time_password.html.erb
alkesh: Using `params[:resent]` in the view avoids the need for the callbacks. 
I couldn't get `claim: { resend: true }` to work here - but I'll try a couple of other things before giving up.

PR #3297, opened 2024-10-10
[CAPT-1854] Provider verification status

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-1854
- Fixes issue with provider verification status in back office
- If claim is rejected, this takes precedence over provider verification
- when filtering claims for `provider verification` it only includes claims awaiting a decision


rjlynch (approved):
Looks good 👍

PR #3301, opened 2024-10-10
CAPT-1840 Update payment breakdown template

kenfodder:
* Make single and multiple claim payments into one email template
* Add a PaymentMailerHelper to deal with conditional bullets

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -856,6 +856,7 @@ en:
       prepared_date: "24 September 2024"
     landing_page: Find out if you are eligible for a targeted retention incentive payment for further education teachers
     claim_description: for further education payments
+    claim_amount_description: Further education targeted retention incentive payment
     journey_name: Claim a targeted retention incentive payment for further education teachers
     feedback_email: "FE-targeted.retention-incentive@education.gov.uk"
     support_email_address: "FE-targeted.retention-incentive@education.gov.uk"


app/helpers/payment_mailer_helper.rb
```diff
@@ -0,0 +1,27 @@
+module PaymentMailerHelper
+  def breakdown_of_payment_bullets(payment)
+    bullets = []
+
+    !payment.claims.one? && payment.claims.each do |claim|
+      bullets << "#{I18n.t("#{claim.policy.locale_key}.claim_amount_description")}: #{number_to_currency(claim.award_amount)}"
```
alkesh, line 6: Do we need `claim_amount_description`s for IRP and FE?
And maybe a spec that checks `Policies.all.map{|p| I18n.t("#{p.locale_key}.claim_amount_description") }`

app/views/payment_mailer/payment_breakdown_confirmation.text.erb
```diff
@@ -0,0 +1,32 @@
+Dear <%= @display_name %>,
+
+You will receive <%=number_to_currency(@payment.net_pay) %> on <%= @payment_date.to_fs(:custom_ordinal) %>.
```
alkesh, line 3:
```suggestion
You will receive <%= number_to_currency(@payment.net_pay) %> on <%= @payment_date.to_fs(:custom_ordinal) %>.
```

spec/mailers/payment_mailer_spec.rb
alkesh, line 1: This spec checks all policies for the payments with a single claim,
but for payments with multiple claims, only TSLR is checked - probably should be expanded to check for all policies.

alkesh (approved):

PR #3310, opened 2024-10-14
[LUPEYALPHA-1158] Use govuk formbuilder for submissions

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-1158
- govuk formbuilder by default out of the box has double click prevention enabled which is desired


rjlynch (approved):
Nice 👍

alkesh (approved):

PR #3311, opened 2024-10-14
CAPT-1468 Add a no match when only the NiNo matches from the SLC data

kenfodder:


app/models/automated_checks/claim_verifiers/student_loan_plan.rb
```diff
@@ -31,18 +31,24 @@ def student_loans_data
         @student_loans_data ||= StudentLoansData.where(nino:, date_of_birth:)
       end
 
+      def student_loans_data_nino_only
+        @student_loans_data_nino_only ||= StudentLoansData.where(nino:)
```
rjlynch, line 35: Not sure if it's worth being a bit more explicit here
```suggestion
        @student_loans_data_nino_only ||= StudentLoansData.where(nino:).where.not(date_of_birth:)
```

rjlynch (approved):
Looks good 👍

PR #3314, opened 2024-10-15
LUPEYALPHA-1155 / Only update visited slugs if form submission succeeded

rjlynch:
Adjusts the completed slugs method to only update the list of visited slugs if the form is valid.

There's been some rollbar errors where the user _seems_ to have skipped ahead in the journey but hasn't be redirected back by the `check_page_is_in_sequence` callback.

All the tests pass but this seems like it could be risky. I'm not 100% sure that no code is subtly relying on the visited slugs being updated on failed form verifications.

We may want to consider a different approach to wizard state management, maybe passing the session answers to a state machine to determine which slugs are accessible.

## Notes for review
Please think of any scenarios where moving this call back to an after action might bite us!

## master
https://github.com/user-attachments/assets/d93d680e-1b31-4888-ba12-9d087c7f7e2a

## branch
https://github.com/user-attachments/assets/f4284cf6-53d5-4a1f-a706-b7d16f55592d

### DIFF
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -4,13 +4,13 @@ class ClaimsController < BasePublicController
   skip_before_action :send_unstarted_claimants_to_the_start, only: [:new, :create]
   before_action :initialize_session_slug_history
   before_action :check_page_is_in_sequence, only: [:show, :update]
-  before_action :update_session_with_current_slug, only: [:update]
   before_action :set_backlink_path, only: [:show, :update]
   before_action :check_claim_not_in_progress, only: [:new]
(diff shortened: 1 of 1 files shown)


app/controllers/claims_controller.rb
```diff
@@ -4,7 +4,7 @@ class ClaimsController < BasePublicController
   skip_before_action :send_unstarted_claimants_to_the_start, only: [:new, :create]
   before_action :initialize_session_slug_history
   before_action :check_page_is_in_sequence, only: [:show, :update]
-  before_action :update_session_with_current_slug, only: [:update]
+  after_action :update_session_with_current_slug, only: [:update]
```
asmega, line 7: i'd recommend moving this a few lines down so the `after_action` is not in a midst of `before_action`s

PR #3315, opened 2024-10-15
CAPT-1860 check submitted claim is a policy for current journey

kenfodder:
Review notes:

* Reverting the code in `app/controllers/concerns/part_of_claim_journey.rb` will cause the new test added to fail with exactly the same error as Rollbar error.
* Some minor typo tweaks


asmega (approved):

rjlynch (approved):
LGTM 👍

PR #3317, opened 2024-10-15
[LUPEYALPHA-1163] Migration so claim#reference cannot be NULL

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-1163
- Migration to add not NULL constraint so `Claim#reference` cannot be NULL
- DB level constraints prevents us getting into bad data integrity situations


rjlynch (approved):
Nice 👍 Now we'll know where these no ref claims are coming from

PR #3319, opened 2024-10-15
[LUPEYALPHA-989] EY Practitioner check-your-answers page

alkesh:
When the EY practitioner claim is submitted, it will update the existing claim (created by the EY provider) with the answers.

After discussion with @slorek, we decided that `claim.submitted_at` will only be populated when the practitioner submits the claim.
When the provider submits the claim, `claim.eligibility.provider_claim_submitted_at` will be populated.

### DIFF
--- a/app/forms/claim_submission_base_form.rb
+++ b/app/forms/claim_submission_base_form.rb
@@ -13,7 +13,7 @@ class ClaimSubmissionBaseForm
 
   def initialize(journey_session:)
     @journey_session = journey_session
-    @claim = build_or_find_claim
+    @claim = build_claim
   end
 
   def save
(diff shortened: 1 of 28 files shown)


app/forms/claim_submission_base_form.rb
```diff
@@ -13,15 +13,14 @@ class ClaimSubmissionBaseForm
...
-    @claim = build_claim
+    @claim = build_or_find_claim
   end
 
   def save
     return false unless valid?
 
     ApplicationRecord.transaction do
       set_attributes_for_claim_submission
-      main_eligibility.save!
```
alkesh, line 24: why did we ever save eligibility here?

app/forms/claim_submission_base_form.rb
```diff
@@ -35,20 +34,30 @@ def save
 
   delegate :answers, to: :journey_session
 
-  def build_claim
-    claim = Claim.new
+  def main_eligibility
+    @main_eligibility ||= eligibilities.first
+  end
```
alkesh, line 39: All the subclasses had this method, so I moved it here.
The only subclass with a different implementation is `Journeys::AdditionalPaymentsForTeachingClaimSubmissionForm`

PR #3322, opened 2024-10-16
[LUPEYALPHA-1168] Upgrade static error pages to govuk 5.7.1

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-1168
- Static error pages pointing to inexistent assets
- Upgrade govuk frontent 5.3.1 => 5.7.1 
- Add missing assets
- Also add tests to cover for missing assets

# Review notes

- `/assets` seems to be reserved, took way too long to figure out. haven't checked but assume rack middleware or engine or asset pipeline thing that takes over this subdirectory so static files now sit on root
- should greatly reduce error as our error pages no longer invoke further errors in the app


rjlynch (approved):
Nice 👍 Looks good

PR #3323, opened 2024-10-16
[LUPEYALPHA-990] EY practitioner confirmation page

alkesh:

### DIFF
--- a/app/views/early_years_payment/practitioner/submissions/show.html.erb
+++ b/app/views/early_years_payment/practitioner/submissions/show.html.erb
@@ -1,1 +1,41 @@
-placeholder
+<% content_for(:page_title, page_title("Claim submitted", journey: current_journey_routing_name)) %>
+
+<div class="govuk-grid-row">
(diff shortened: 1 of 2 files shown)


app/views/early_years_payment/practitioner/submissions/show.html.erb
```diff
@@ -1 +1,41 @@
...
+    <h2 class="govuk-heading-m">
+      What happens next
+    </h2>
+
+    <p class="govuk-body">
+      We’ll review your claim and let you know by email if it’s been accepted or rejected.
+    </p>
+
+    <p class="govuk-body">
+      After 6 months in your role, we’ll check with ? that you’re still working for them. If you are, we’ll:
```
asmega, line 34: is the `?` expected?

alkesh: ah yes - I need to put the nursery name in there 😅

asmega (approved):

PR #3325, opened 2024-10-17
LUPEYALPHA-1153 - add a generic error page if journey session missing after OL auth failure

kenfodder:
LUPEYALPHA-1153 - add a generic error page if journey session missing after OL auth failure

* Changed the failure template to use the application layout, add header_link content_for
* and renamed it to dfe_identity_failure so it's clear
* Added a session_missing_failure template if there is no journey_session

Content for the actual page needs Product review.

You can testing this by clearing cookies to ensure there isn't a claim session and hitting directly: https://localhost:3000/auth/failure?strategy=onelogin

Screenshot 2024-10-17 at 12 37 39


rjlynch (approved):
LGTM
Can we delete `app/views/layouts/omniauth_failure.html.erb` as it's empty?

asmega (approved):

PR #3327, opened 2024-10-17
[LUPEYALPHA-1201] Use visible legends for FE questions

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-1201
- Use visible legends for FE questions


rjlynch (approved):
👍

PR #3331, opened 2024-10-18
Add fraud check

rjlynch:
Add fraud check

Adds a mechanisim for admins to upload a csv of known risky trns and
ninos that will cause claims with these attributes to be flagged in the
admin ui.

Flagged values can be removed by uploading a CSV without those values.

We've added some new functionality to the task/index notification
banner. The banner now supports rendering multiple messages. If there is
a single banner notification is falls back to the original view and
renders a p tag, with multiple notifications it now renders a list.

## Notes for review
The main changes in this PR is the introduction of the RiskIndicator model, the new admin controllers for managing the fraud risk CSV, and the changes to the admin/tasks/index banner notification to support rendering multiple messages.
Probably best to start by reviewing the new `app/controllers/admin/fraud_risk_csv_uploads_controller.rb` as that's where the new flow kicks off.
Also of note is the new claim verifier `app/models/automated_checks/claim_verifiers/fraud_risk.rb`. This claim verifier doesn't create a task (not a requirement in the ticket but may be wanted in future), however a claim verifier seemed like the natural place to create a fraud risk note on the ticket.

## Acceptance criteria
* When a known TRN submits a claim, the claim is flagged in the main body of the claim. A note needs to be created in the notes section to explain this, similar to the hold/unhold feature.

* The claim cannot be approved, these feature is disabled

https://github.com/user-attachments/assets/02aa6c41-6ffd-4415-bd8d-35d3317158a0

### DIFF
--- a/app/forms/admin/fraud_risk_csv_upload_form.rb
+++ b/app/forms/admin/fraud_risk_csv_upload_form.rb
@@ -21,6 +21,10 @@ module Admin
         RiskIndicator.where.not(id: records.map(&:id)).destroy_all
 
         records.each(&:save!)
+
+        claims_to_note.each do |claim|
+          AutomatedChecks::ClaimVerifiers::FraudRisk.new(claim: claim).perform
+        end
       end
 
       true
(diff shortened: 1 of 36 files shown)


app/forms/admin/fraud_risk_csv_upload_form.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    def initialize(params = {})
+      super
+    end
+
+    def save
+      return false unless valid?
+
+      ApplicationRecord.transaction do
+        RiskIndicator.where.not(id: records.map(&:id)).destroy_all
```
asmega, line 21: wouldn't it have been easier to delete everything then just re-add the whatevers in the CSV?
not sure if this methodology has any benefits?

rjlynch: It would be a bit easier, just seemed weird to create new records when the data hasn't changed!

asmega (approved):

PR #3333, opened 2024-10-21
[CAPT-1781] Revisit OTP

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-1781
- Improves implementation of OTP

# Changes

- moved OTP classes from `/lib` => `app/models` to benefit from rails autoloading with zeitwerk
- use keyword arguments of OTP methods
- the concept of shared SECRET has now been obliterated. the caller is now responsible for setting a valid secret and managing their secret such as persisting
- there is no longer a default SECRET and a secret is now mandatory and must be passed to OTP classes otherwise will throw an error
- modified email and mobile verification journeys so they now generate and persist their own secrets which are passed to the OTP classes


app/mailers/claim_mailer.rb
```diff
@@ -69,6 +69,10 @@ def email_verification(claim, one_time_password)
       validity_duration: one_time_password_validity_duration
     }
 
+    if Rails.env.development?
+      Rails.logger.info("\n\nEmail verification code: #{@one_time_password}\n")
```
alkesh, line 73: this is super handy 👍🏽

alkesh (approved):

app/mailers/claim_mailer.rb
rjlynch: Yeah very nice addition 👍

rjlynch (approved):
LGTM 👍 
Do we need to find a quite time to deploy this so we have the minimum number of users midway through validating their emails / phone when this roles out?
Maybe not something to worry about as they can always request a new confirmation code.

PR #3334, opened 2024-10-21
Remove batching from payroll csv

rjlynch:
Requirement is to only produce a single payroll file with no limit on
the number of payments in the file.
This commit removes the zip file as we no longer need it and instead
serves the generated csv directly, this lets us simplify the code quite
a bit.


rjlynch:
closing to rebuild deploy app

PR #3338, opened 2024-10-23
[LUPEYALPHA-1170] EY full claim view

alkesh:
* save `provider_email_address`, `award_amount`, `returner_worked_with_children`, `returner_contract_type`, `award_amount` when the provider submits the claim
* admin full claim view works for both provider submitted claims, and practitioner submitted claims
* removed Claim.submitted scope (useful for EY, because the provider submitted claim has no `submitted_at` - this is populated when the practitioner submits the claim.
* `email_address` is not populated on the claim when the provider submits the claim, only when the practitioner does.


spec/mailers/previews/claim_mailer_preview.rb
```diff
@@ -31,10 +31,6 @@ def update_after_three_weeks_lup
     ClaimMailer.update_after_three_weeks(claim_for(Claim.approved.by_policy(Policies::LevellingUpPremiumPayments)))
   end
 
-  def email_verification
-    ClaimMailer.email_verification(claim_for(Claim.unsubmitted), OneTimePassword::Generator.new.code)
-  end
```
alkesh, line 36: do we use these mailer previews? none of them work for me locally.

asmega: i don't use em, happy for them to go

asmega (approved):

PR #3339, opened 2024-10-23
Don't show DQT match if DQT record not found

rjlynch:
Don't show DQT match if DQT record not found

If we don't have a DQT record for the claimant we don't want to show the
text "The claimant’s identity matches DQT records.".

The for the identity task `claim_verifier_match` can be in one of 4
states:
`nil` - the check hasn't run or identity was verified by one login¹
`none` - no dqt record was found
`any` - a partial match against the returned dqt record
`all` - a full match against the dqt record

¹ If the claim was verified by one login then the match on the task will
be `nil`, however the way the task form is set up it only renders the
body section if the policy has a corresponding body section in the
locales file. FurtherEducationPayments does not have a body section in
the locales file so we don't show the dqt related messages for these
claims.

The body portion of the form is currently only present in the locales
file for the `identity_confirmation` task and only for
additional_payments and student_loans claims.

We may want to revisit how we indicate claim's have had their identify
confirmed so it's easier to infer how this was done, eg something like
an `source` column on the `Task`, or update the claim with information
about identity confirmation when the idenity claim verifier runs.

### DIFF
--- a/spec/features/admin/admin_claim_without_verified_identity_spec.rb
+++ b/spec/features/admin/admin_claim_without_verified_identity_spec.rb
@@ -17,6 +17,7 @@ RSpec.feature "Admin checking a claim without a verified identity" do
     click_on I18n.t("admin.tasks.identity_confirmation.title")
 
     expect(page).to have_content("Did #{unverified_claim.full_name} submit the claim?")
+    expect(page).not_to have_content("The claimant’s identity matches DQT records.")
     expect(page).to have_content(unverified_claim.eligibility.current_school.name)
     expect(page).to have_content(unverified_claim.eligibility.current_school.phone_number)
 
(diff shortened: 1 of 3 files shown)


spec/features/admin/admin_claim_without_verified_identity_spec.rb
```diff
@@ -17,6 +17,7 @@
     click_on I18n.t("admin.tasks.identity_confirmation.title")
 
     expect(page).to have_content("Did #{unverified_claim.full_name} submit the claim?")
+    expect(page).not_to have_content("The claimant's identity matches the DQT record")
```
asmega, line 20: i don't think this test will ever go red due to the apostrophe not being a smart quote?

rjlynch: 👀 impressive spot. Have fixed!

asmega (approved):

PR #3348, opened 2024-10-28
[LUPEYALPHA-1172] EY admin matching details

alkesh:


app/mailers/claim_mailer.rb
```diff
@@ -84,6 +84,10 @@ def early_years_payment_provider_email(claim, one_time_password, email)
       magic_link: @magic_link
     }
 
+    if Rails.env.development?
+      Rails.logger.info("\n\nEmail verification code: #{one_time_password}\n")
+    end
+
```
slorek, line 90: Not sure this is related.

app/models/claim/matching_attribute_finder.rb
```diff
@@ -43,9 +43,9 @@ def matching_claims
 
         concatenated_columns = "CONCAT(#{attributes.join(",")})"
         policies_to_find_matches.map { |policy|
-          policy::Eligibility.where("LOWER(#{concatenated_columns}) = LOWER(?)", vals.join)
+          policy::Eligibility.where("LOWER(#{concatenated_columns}) = LOWER(?)", vals.join) if (attributes - policy::Eligibility.column_names).empty?
```
slorek, line 46: 🤔 could this result in a missed match? Perhaps it's better to strip out any columns which don't exist but still match those which do, rather than skipping entirely.

app/mailers/claim_mailer.rb
alkesh: it's not, no.
I'll put it in a separate commit.

app/models/claim/matching_attribute_finder.rb
alkesh: So right now, there is never more than one column, so we probably don't need to concatenate here.
I can refactor this code to remove the concat - not sure why it's here.
The column is based on the `ELIGIBILITY_MATCHING_ATTRIBUTES` for the policy,
which is `teacher_reference_number` for TSLR, LUPP, FE and ECP.
and `passport_number` for IRP.
I guess in theory, you could specify multiple columns per policy, but we haven't.

alkesh: To answer your question - I don't think this will result in a missed match.
That where clause is to check for eligibilities that match **all** of the columns in concatenated_columns.
If one of those columns does not exist on an eligibility, then there will never be a match with that eligibility.

app/models/claim/matching_attribute_finder.rb
```diff
@@ -41,11 +41,11 @@ def matching_claims
         vals = values_for_attributes(@source_claim.eligibility, attributes)
         next if vals.blank?
 
-        concatenated_columns = "CONCAT(#{attributes.join(",")})"
+        concatenated_columns = "CONCAT(#{attributes.join(",")})" # current policies will only have a single value in attributes
         policies_to_find_matches.map { |policy|
-          policy::Eligibility.where("LOWER(#{concatenated_columns}) = LOWER(?)", vals.join)
+          policy::Eligibility.where("LOWER(#{concatenated_columns}) = LOWER(?)", vals.join) if all_attributes_exist_on_the_eligibility(attributes, policy::Eligibility)
```
rjlynch, line 46: Why do we need the `all_attributes_exist_on_the_eligibility` check? Would we ever define `Policies::SomePolicy::ELIGIBILITY_MATCHING_ATTRIBUTES` with attributes that aren't present on the policy's eligibility?

app/models/claim/matching_attribute_finder.rb
rjlynch: I think [this commit](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/3348/commits/13e17f0553ee118307f00a5c08c5ab40e401ccc5) should let us still do the comparison if the eligibilities have some different attributes

alkesh: oh yeah - that is better 👍🏽

asmega (approved):

slorek (approved):

PR #3350, opened 2024-10-28
[LUPEYALPHA-1173][LUPEYALPHA-1176] EY rejection reasons

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-1173
- https://dfedigital.atlassian.net/browse/LUPEYALPHA-1176
- Add EY admin rejection reasons


app/jobs/claim_verifier_job.rb
```diff
@@ -1,14 +1,20 @@
...
+    dqt_teacher_status = if claim.policy == Policies::EarlyYearsPayments
+      nil
+    elsif claim.has_dqt_record?
+      Dqt::Teacher.new(claim.dqt_teacher_status)
+    else
+      Dqt::Client.new.teacher.find(
+        claim.eligibility.teacher_reference_number,
+        birthdate: claim.date_of_birth.to_s,
+        nino: claim.national_insurance_number
+      )
```
asmega, line 12: this isn't great but will do for now. ideally each automated check should be independent and all called async which is not the case now

config/locales/en.yml
```diff
@@ -1420,11 +1420,12 @@ en:
         By submitting this claim you’re confirming that, to the best of your knowledge, the details you’re providing are correct.
       btn_text: Accept and send
   early_years_payments:
-    <<: *early_years_payment_provider_authenticated
     claim_subject: "Early Years Payment"
     policy_acronym: EY
     policy_short_name: Early Years
     policy_full_name: Early years financial incentive payment
+    support_email_address: "earlycareerteacherpayments@digital.education.gov.uk"
```
asmega, line 1427: temporary placeholder till we know what these will be replaced with

alkesh (approved):

PR #3352, opened 2024-10-29
[LUPEYALPHA-1174] EY employment task

alkesh:


app/models/task.rb
```diff
@@ -53,4 +53,10 @@ def to_param
   def identity_confirmation?
     name == "identity_confirmation"
   end
+
+  def available?
```
asmega, line 57: i think i'd prefer to call this method `employment_task_available?`, even though it's a bit out of place as it's more specific than `available?` the next person is far less likely to hijack for some other use

alkesh: done

asmega (approved):

PR #3353, opened 2024-10-29
Able to manually delete specified review app

asmega:
# Context

- this allows us to specify review apps to delete on demand
- this is useful for when there are rogue review app instances
- this is untested but quickest way to test is probably to merge and give it a go


rjlynch (approved):
Nice 👍

PR #3357, opened 2024-10-30
LUPEYALPHA 1171/identity task

rjlynch:
Might be easier to review each commit separately.

The main change in this pr is adding the Identity task.
The EY identity task behaves slightly differently to the other tasks.
EY practitioners don't have DQT records so we're only using the information
from one login to check identity, additionally the rules around creating the
task are slightly different to other policies.

* If the one login check passes and the name returned from OL matches the name
  the provider entered the task is a pass

* If the one login check passes and the names are a partial match¹ then we
  create task with a partial match (see `Admin::ClaimsHelper#task_status_tag`)

* If the one login check passes and the name from OL is different to the
  provider supplied name we create a failed task.

* If the one login check fails we create a failed task.

If the task is a partial match we render the task form for admins to update the
task (either passing or failing it).
Another quirk is if the practitioner hasn't completed their portion of the
journey we don't want to render the task form, and instead we want to show some
copy informing admins the practitioner is yet to complete their journey.

As there's a few differences between how other identity tasks work we've
introduced a separate task view for EY identity tasks.

¹ - we'll be changing the rules around what counts as a partial match in a
separate ticket.

### DIFF
--- a/app/controllers/admin/tasks_controller.rb
+++ b/app/controllers/admin/tasks_controller.rb
@@ -17,7 +17,7 @@ class Admin::TasksController < Admin::BaseAdminController
     @notes = @claim.notes.automated.by_label(params[:name])
     @task_pagination = Admin::TaskPagination.new(claim: @claim, current_task_name:)
 
-    render @task.name
+    render task_view(@task)
   end
 
   def create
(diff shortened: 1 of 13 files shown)


rjlynch:
Closing to see if that updates the deployment

PR #3360, opened 2024-10-31
[LUPEYALPHA-1235] Rejecting EY claim also emails provider

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-1235
- When rejecting an EY claim, provider is also notified via email

# Review notes

- I'm aware that at this moment `ClaimMailer` is not in the best of shape and rather overloaded. I think at some point we need to add some kind of service class(es) to deal with business logic of when to send email leaving mail just to deal with mail


rjlynch (approved):
LGTM 👍 

> ClaimMailer is not in the best of shape and rather overloaded

Yeah I think we probably want some policy specific mailers

rjlynch (approved):
👍 LGTM

PR #3364, opened 2024-11-01
[LUPEYALPHA-1175] Can filter for EY awaiting states

asmega:
# Context

- https://dfedigital.atlassian.net/browse/LUPEYALPHA-1175
- Admin now able to filter by EY awaiting states

### DIFF
--- a/app/forms/admin/claims_filter_form.rb
+++ b/app/forms/admin/claims_filter_form.rb
@@ -74,6 +74,25 @@ class Admin::ClaimsFilterForm
         Claim.includes(:decisions).failed_bank_validation.awaiting_decision
       when "awaiting_provider_verification"
         Claim.by_policy(Policies::FurtherEducationPayments).awaiting_further_education_provider_verification.awaiting_decision
+      when "awaiting_claimant_data"
+        Claim
+          .by_policy(Policies::EarlyYearsPayments)
+          .where(submitted_at: nil)
+          .awaiting_decision
+      when "awaiting_retention_period_completion"
+        Claim
+          .by_policy(Policies::EarlyYearsPayments)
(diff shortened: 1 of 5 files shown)


app/forms/admin/claims_filter_form.rb
```diff
@@ -74,6 +74,25 @@ def claims
...
+        Claim
+          .by_policy(Policies::EarlyYearsPayments)
+          .where(submitted_at: nil)
+          .awaiting_decision
+      when "awaiting_retention_period_completion"
+        Claim
+          .by_policy(Policies::EarlyYearsPayments)
+          .joins(:early_years_payment_eligibility)
+          .where.not(submitted_at: nil)
+          .awaiting_decision
```
rjlynch, line 87: If we don't want to add a belongs_to to a specific policy to the claim we could do something like
```suggestion
Claim
  .where(
    id: Policies::EarlyYearsPayments
      .select("claims.id")
      .joins(:claim)
      .where("start_date > ?", Policies::EarlyYearsPayments::RETENTION_PERIOD.ago)
  )
  .where.not(submitted_at: nil)
  .awaiting_decision
```

Not sure what's better 🤷

rjlynch (approved):
LGTM 👍 
Left a suggestion but I'm not sure if it's better or not, so feel free to ignore!

app/forms/admin/claims_filter_form.rb
asmega: I get you and agree you could go either way as may be we don't want to clutter with subtype implementation in `Claim` on the other hand may need to duplicate elsewhere. I'm leaning towards leaving this for now then later if it gets worse extract somehow...

PR #3366, opened 2024-11-04
[CAPT-1921] EY claim index data

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-1921
- For back office admin viewing claims
- Tweak content for EY
- Show different decision deadline


rjlynch (approved):
LGTM - We should consider moving `status` to a method on the tag, and this helper should just set the colour

rjlynch (approved):
Looks good 👍

PR #3368, opened 2024-11-05
[CAPT-1937] EY approvable only after passed employment task

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-1937
- EY can only be approved after employment task has passed
- Also made employment task visible for EY task list which seemed to be missing


app/models/claim.rb
```diff
@@ -471,6 +477,14 @@ def attributes_flagged_by_risk_indicator
 
   private
 
+  def policy_specific_approvable?
+    return true if policy.nil?
+    return true if eligibility.nil?
```
rjlynch, line 482: Can these ever be `nil` or is this just for the test suite?

app/models/claim.rb
```diff
@@ -253,7 +253,13 @@ def submittable?
...
 
   def approvable?
-    submitted? && !held? && !payroll_gender_missing? && (!decision_made? || awaiting_qa?) && !payment_prevented_by_other_claims? && attributes_flagged_by_risk_indicator.none?
+    submitted? &&
+      !held? &&
+      !payroll_gender_missing? &&
+      (!decision_made? || awaiting_qa?) &&
+      !payment_prevented_by_other_claims? &&
+      attributes_flagged_by_risk_indicator.none? &&
+      policy_specific_approvable?
```
rjlynch, line 262: Not a blocker but `Claim#policy_specific_approvable?` and `Eligibility#approvable?` might be better as a method on the policy:

```rb
module Policies::EarlyYearsPayments
  def approvable?(claim)
    claim.tasks.find_or_initialize_by(name: "employment").passed?
  end
end

module BasePolicy
  def approvable?(claim)
    true
  end
end
```

that way the policy has all the policy specific rules.

```suggestion
    submitted? &&
      !held? &&
      !payroll_gender_missing? &&
      (!decision_made? || awaiting_qa?) &&
      !payment_prevented_by_other_claims? &&
      attributes_flagged_by_risk_indicator.none? &&
      policy.approvable?(self)
```

rjlynch (approved):
LGTM - left a suggestion about moving some methods around.

app/models/claim.rb
rjlynch: As part of the LUPEALPHA-1202 ticket I've added a method like this to the policy to determine if a task is available to admins, and seems to work quite well / keeps the policy specific rules in the policy class

app/models/claim.rb
asmega: theres actually no coverage for this, i was just erring on the side of caution cause i have no idea how upstream behaves and didn't want to cause an issue downstream

PR #3372, opened 2024-11-05
[CAPT-1938] Tweak EY journey name

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-1938
- Tweak EY journey name which affects the name of the service in the header


rjlynch (approved):
👍

PR #3378, opened 2024-11-07
CAPT-1962: Sync batch database operations with BigQuery

slorek:
https://dfedigital.atlassian.net/browse/CAPT-1962

We use [dfe-analytics](https://github.com/DFE-Digital/dfe-analytics) to sync database operations with the data warehouse held in Google BigQuery.

The library uses model callbacks to enable this, so in some instances where we use `insert_all`/`delete_all`/`update_all` these operations are being missed.

This change should rectify all of these sync issues.

In some cases I felt it was better, on balance, to trigger the model callbacks and sacrifice a little performance. This is because syncing large tables such as `claims` or `tasks` can take a very significant amount of time in the background jobs. 

In other instances it is less expensive to simply force a full import of the table, which I have done by invoking the supplied rake task rather than re-implementing it which risks becoming out of sync or incompatible with changes to the upstream library. Unfortunately there doesn't appear to be a convenient way to do this other than invoking the rake task.

### DIFF
--- a/lib/csv_importer/base.rb
+++ b/lib/csv_importer/base.rb
@@ -32,6 +32,8 @@ module CsvImporter
 
         target_data_model.insert_all(record_hashes) unless record_hashes.empty?
       end
+
+      Rake::Task["dfe:analytics:import_entity"].invoke(target_data_model.table_name) if DfE::Analytics.enabled?
     end
 
     def rows_with_data_count
--- a/lib/csv_importer/base.rb
+++ b/lib/csv_importer/base.rb
@@ -33,7 +33,7 @@ module CsvImporter
         target_data_model.insert_all(record_hashes) unless record_hashes.empty?
       end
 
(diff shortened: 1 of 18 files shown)


rjlynch (approved):
👍 I think the code changes look fine / safe

Not sure why those two tests are failing on CI though, I ran them on my machine and they passed 🙃

lib/csv_importer/base.rb
```diff
@@ -32,6 +32,8 @@ def run
 
         target_data_model.insert_all(record_hashes) unless record_hashes.empty?
       end
+
+      Rake::Task["dfe:analytics:import_entity"].invoke(target_data_model.table_name) if DfE::Analytics.enabled?
```
rjlynch, line 36: Would be tempted to wrap calling this rake task with our own model (`DfEAnalytics.sync(SomeClass)`) - we could then try stubbing the call to that class in the tests and potentially that would work on CI 🤷

PR #3381, opened 2024-11-08
[CAPT-1898] Practitioner claim start timestamp

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-1898
- Change definition of when practitioner starts their claim
- Was when journey session was created, now after they submit a reference


rjlynch (approved):
👍

PR #3388, opened 2024-11-11
Move payroll run to background job

rjlynch:
Creating the November payroll run timed out (October's too).
This commit moves creating the payment records into a background job.

We need to pass the claim ids and topup ids to the background job, this
is because once the admin has reviewed the claims that will be part of
this payroll run (shown in the `new` action) we can't include any claims
created in the interim. We may want to consider someother method of
indicating what claims should be part of this payroll run (create
pending payments? flag claims / topups as reviewed for payroll?) so we
don't have to pass such a large amount of information to the background
job. Have tested this with production data and it runs fine passing all the ids.

Not sure if the payroll numbers are public so have put the video walk through of testing this with the prod db on ira https://dfedigital.atlassian.net/browse/CAPT-1975?focusedCommentId=98115

### DIFF
--- a/app/views/admin/payroll_runs/_complete.html.erb
+++ b/app/views/admin/payroll_runs/_complete.html.erb
@@ -1,0 +1,158 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">
+      <%= @payroll_run.created_at.strftime("%B") %> payroll run
+    </h1>
+
+    <dl class="govuk-summary-list govuk-!-margin-bottom-9">
+      <div class="govuk-summary-list__row">
+        <dt class="govuk-summary-list__key">
+          Approved claims
+        </dt>
+
+        <dd class="govuk-summary-list__value">
+          <%= @payroll_run.number_of_claims_for_policy(:all, filter: :claims) %>
+        </dd>
+      </div>
+      <div class="govuk-summary-list__row">
+        <dt class="govuk-summary-list__key">
+          Top ups
+        </dt>
+
+        <dd class="govuk-summary-list__value">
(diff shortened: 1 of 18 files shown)


db/migrate/20241111114155_add_status_to_payroll_runs.rb
```diff
@@ -0,0 +1,7 @@
+class AddStatusToPayrollRuns < ActiveRecord::Migration[7.0]
+  def change
+    add_column :payroll_runs, :status, :string, default: "pending", null: false
+
+    PayrollRun.update_all(status: "complete")
```
rjlynch, line 5: There won't be any inflight payroll runs when this rolls out so it's safe to update all the existing ones to completed

app/views/admin/payroll_runs/_complete.html.erb
```diff
@@ -0,0 +1,158 @@
...
+          </tr>
+        <% end %>
+        <tr class="govuk-table__row">
+          <th scope="row" class="govuk-table__header"><%= I18n.t("levelling_up_premium_payments.policy_short_name") %> Top Ups</th>
+          <td class="govuk-table__cell"><%= @payroll_run.number_of_claims_for_policy(Policies::LevellingUpPremiumPayments, filter: :topups) %></td>
+          <td class="govuk-table__cell"><%= number_to_currency(@payroll_run.total_claim_amount_for_policy(Policies::LevellingUpPremiumPayments, filter: :topups)) %></td>
+          </tr>
+      </thead>
+      <tbody class="govuk-table__body">
+      </tbody>
```
asmega, line 91: think the rows need to go into here?

asmega: looks to be an existing bug i guess since its a copy/paste

PR #3389, opened 2024-11-12
Log emails in dev

rjlynch:
Useful for grabbing magic links etc

### DIFF
--- a/app/mailers/claim_mailer.rb
+++ b/app/mailers/claim_mailer.rb
@@ -89,10 +89,6 @@ class ClaimMailer < ApplicationMailer
       validity_duration: one_time_password_validity_duration
     }
 
-    if Rails.env.development?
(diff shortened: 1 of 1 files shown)


app/mailers/claim_mailer.rb
```diff
@@ -215,13 +211,25 @@ def template_ids(claim)
...
+      )
+    else
+      template_mail(
+        template_id,
+        to: @claim.email_address,
+        subject: @subject,
+        reply_to_id: @policy.notify_reply_to_id,
+        personalisation:
+      )
+    end
```
asmega, line 232: this will actually prevent the email from sending though, which is useful for debugging. do we not want to do this on top of the default behaviour?

rjlynch: I've never used that but will let it send emails through

asmega (approved):

PR #3391, opened 2024-11-12
CAPT-1934 Early Years data scrubbing

kenfodder:
* Override the query for rejected and paid claims for anything over a year old
* Early Years is not based around Academic Years


rjlynch (approved):
LGTM 👍

PR #3407, opened 2024-11-14
Copy update

rjlynch:


app/views/early_years_payment/practitioner/claims/how_we_use_your_information.html.erb
```diff
@@ -1,52 +1,123 @@
-<% title = "How we’ll use the information you provide" %>
+<% title = "How we'll process your claim" %>
```
asmega, line 1: smart quotes + was this not the one you reverted?

rjlynch: Sorted the smart quotes.
Yeah this was the one I reverted as it hadn't been business approved

asmega (approved):
there are more instance of the smart quotes though, i didn't feel the need indicate every single one though 😅

rjlynch:
> there are more instance of the smart quotes though, i didn't feel the need indicate every single one though 😅

I copied the copy straight from the prototype 😂 I'll fix up those quotes though

PR #3409, opened 2024-11-14
[CAPT-1984] EY clear dependent answers

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-1984
- The ticket is actually for removing a change link on EY check answers page which should not be present

# Changes

- For EY provider journey delete dependent answers
- Inject `session` into `Form` so it is available for reading or manipulation
- Now that `session` is available, delete from `session[:slugs]` visited slugs when clearing dependent answers for EY provider journey
- This solves the problem if:
  - a user goes down the extended branch of a journey
  - then changes an answer, to shorten the branch
  - then changes their answer again to extend the branch
  - they are now asked the missing questions rather than sent back to check answers page with incomplete questions
- This bug will likely be present for all journeys where there is possible branching, but the mechanism to at least be able ti fix it has been put in place 
- Fixed a typo

### DIFF
--- a/app/models/journeys/early_years_payment/provider/authenticated/answers_presenter.rb
+++ b/app/models/journeys/early_years_payment/provider/authenticated/answers_presenter.rb
@@ -79,7 +79,7 @@ module Journeys
 
           def returner_contract_type
             [
-              "Contact type for previous role in an early years setting",
+              "Contract type for previous role in an early years setting",
               answers.returner_contract_type,
               "returner-contract-type"
             ]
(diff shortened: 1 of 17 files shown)


app/models/journeys/early_years_payment/provider/authenticated/answers_presenter.rb
```diff
@@ -79,7 +79,7 @@ def returner_worked_with_children
 
           def returner_contract_type
             [
-              "Contact type for previous role in an early years setting",
+              "Contract type for previous role in an early years setting",
```
rjlynch, line 82: 👍

rjlynch (approved):
LGTM 👍 
Hopefully we'll have the time to removed the slugs from the session entirely!

PR #3410, opened 2024-11-14
Send emails when claim approved

rjlynch:
When an EY claim is approved we want to send an email to the
practitioner and the provider.
Mailers don't let us send two email from one action, so the approve
method enqueues jobs to send each of the emails.
I think it would probably be better for policies to be told a claim has
been approved and handle policy specific actions like sending emails in
the polciy, so rather than:
`@claimpolicy.mailer.approved(@claim).deliver_later`
we'd do
`@claim.policy.claim_approved(@claim)`
and let the policy determine the action to take (sending emails).
However this is a bit different to how we do things now, so for the time
being we'll just call the mailer, even it it's a bit odd and sends two
emails.

### DIFF
--- a/app/models/claim_auto_approval.rb
+++ b/app/models/claim_auto_approval.rb
@@ -25,7 +25,7 @@ class ClaimAutoApproval
         claim.notes.create!(body: "This claim has been marked for a quality assurance review")
       end
 
-      ClaimMailer.approved(claim).deliver_later unless claim.awaiting_qa?
+      claim.policy.mailer.approved(claim).deliver_later unless claim.awaiting_qa?
     rescue ActiveRecord::RecordInvalid => e
       raise AutoApprovalFailed, e
     end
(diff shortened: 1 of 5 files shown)


app/models/claim_auto_approval.rb
```diff
@@ -25,7 +25,7 @@ def auto_approve!
         claim.notes.create!(body: "This claim has been marked for a quality assurance review")
       end
 
-      ClaimMailer.approved(claim).deliver_later unless claim.awaiting_qa?
+      claim.policy.mailer.approved(claim).deliver_later unless claim.awaiting_qa?
```
asmega, line 28: so will not call `deliver_later` on
```
self.class.provider_approved(claim).deliver_later
```
does it work as expected? I feel like we still need something in between the mailer and whatever calls the mailer to handle the events lifecycle. I was able to get away with not doing it in a recent change. maybe this just crosses the line? tbh i'm not sure what this thing would look like though

rjlynch: The emails are sent to the provider and practitioner as expected.

I think what we really want to do is
```
claim.policy.claim_approved(claim)
```
(and `claim.policy.claim_rejected(claim)` etc)
and let the policy decide what emails to send / other actions to take when a claim is approved. Not sure if this should be part of this pr or in a separate pr.

PR #3411, opened 2024-11-15
[CAPT-1915] Improve admin allocations

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-1915
- Fix policy name in flash message

# Other changes

- use govuk formbuilder for allocations form
- some simple tidying up:
  - removing duplicate code
  - using multi line method calls over long single line method calls


rjlynch (approved):
LGTM 👍

PR #3413, opened 2024-11-19
Use new reminders controller + remove old version

asmega:
# Context

- Related to https://dfedigital.atlassian.net/browse/CAPT-1862
- I wrote a new controller to handle reminders for FE awhile back
- This newer reminder controller is free from `PartOfClaimJourney`, `FormSubmittable` and all the controller level callbacks that this mechanism depends upon in favour of more contained explicit form objects.
- The old controller had a feature that would submit the reminder immediately if the data was available in the submitted claim. this feature has been ported to the new controller
- Now there is only one controller handling reminders it should now be easier to handle un-subscriptions from this single point
- I'm aware of the existence of https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/3175 and will work to incorporate those changes once this has been merged


asmega:
> LGTM
> 
> * Should we rename `independent_reminders_path` to just `reminders_path` now?

Done

> One thing we might want to consider is if the claim already has a verified email, change the "Set reminder" link to be a `button_to` the update action and have update handle sending the reminder email, that way we can avoid sending emails in the show action and remove the `set_reminder_from_claim` method. Not a blocker though, just something we might want to consider in future.

Won't do this now but yes agreed that would be much nicer approach. I just kind of ported the old pattern without too much thought.

PR #3415, opened 2024-11-19
CAPT 1995/remove ecp

rjlynch:
This PR introduces some refactors to separate the LUP and ECP logic as part of the move to splitting LUP into it's own journey and deprecating ECP.
**This will be a lot easier to review looking at the individual commits!**

At a high level this PR
* Updates the LUP end date
* Moves policy specific logic into the respective policies
* Removes `lib/journey_subject_eligibility_checker.rb`
* Removes journey specific logic from the page sequence

We'll introduce the LUP only journey in a separate PR.


app/models/journeys/eligibility_checker.rb
```diff
@@ -63,6 +63,14 @@ def policies_eligible_now_and_sorted
       policies_eligible_now_with_award_amount_and_sorted.map { |policy_with_award_amount| policy_with_award_amount.policy }
     end
 
+    def potentially_still_eligible
```
asmega, line 66: maybe rename to `potentially_still_eligible_policies`?

asmega (approved):
Looks fine, nice one on removing more stuff! probably worth getting another review and pair of eyes tbh

rjlynch:
> Looks fine, nice one on removing more stuff! probably worth getting another review and pair of eyes tbh

Thanks, there's still one big piece I want to remove `lib/ineligibility_reason_checker` it reimplements all the ineligible logic, separate from the policy, to determine the reason a claim is ineligible for the ineligible screen. I'm surprised this doesn't get out of sync more. I think we want to just pass the ineligible reason as a param to the ineligible screen, something like `redirect_to ineligible_path(reason: :some_reason)`, rather than using the ineligiblity_reason_checker in the view

app/models/journeys/additional_payments_for_teaching.rb
```diff
@@ -30,24 +30,43 @@ module AdditionalPaymentsForTeaching
...
-      eligible_itt_years = JourneySubjectEligibilityChecker.selectable_itt_years_for_claim_year(next_year)
+      eligible_itt_years = selectable_itt_years_for_claim_year(next_year)
       eligible_itt_years.include?(itt_academic_year)
     end
 
     def requires_student_loan_details?
       true
     end
+
+    def selectable_subject_symbols(journey_session)
```
slorek, line 46: Suggestion: this could probably live in the `EligibleIttSubjectForm` class.

slorek (commented):
Nice work, and good refactoring to remove `JourneySubjectEligibilityChecker`.

app/models/journeys/additional_payments_for_teaching.rb
rjlynch: 👍 that's nicer, will do this as a separate pr though

rjlynch: https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/3461

PR #3421, opened 2024-11-21
[CAPT-1323] Upgrade rails 7.0.8.5 => 7.1.5

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-1323
- Upgrade to rails 7.1.x

# Notes

- Some weirdness with `travel_to`. `Date` never works and need to travel a bit forward or backwards beyond threshold for tests to pass


Dockerfile
```diff
@@ -74,12 +74,15 @@ COPY db ${APP_HOME}/db
...
 
+# Not actual SECRET_KEY_BASE and is randomly generated just for this task
+
 RUN DFE_SIGN_IN_API_CLIENT_ID= \
   DFE_SIGN_IN_API_SECRET= \
   DFE_SIGN_IN_API_ENDPOINT= \
   ADMIN_ALLOWED_IPS= \
   ENVIRONMENT_NAME= \
   SUPPRESS_DFE_ANALYTICS_INIT= \
+  SECRET_KEY_BASE="1d40d61e24db3cd4282716c45890c33b5be4ac99eee12ddfe758a01ad03d3206b5888911211ea9e29eee891c76e7011e8eae9f0c868d500575ea3b18749f31f1" \
```
rjlynch, line 85: I think we can also do `SECRET_KEY_BASE_DUMMY=1` but this is fine tbh

rjlynch (approved):
LGTM - nice thing about a JS light app is upgrades shouldn't break in subtle ways!

Dockerfile
asmega: never heard of this before and first time I've seen this. will give it a go

asmega: done

PR #3422, opened 2024-11-21
Implement EY provider guidance page

slorek:
https://dfedigital.atlassian.net/browse/CAPT-2009

Adds guidance page as a new static page in the application.

Although the ticket specifies it should be part of the journey, I have not tried to tie it into the page sequence since this would require substantial reworking of the way landing pages are tied into the page sequence (it currently makes a hard assumption the `landing-page` is the first page in each journey).

All we really need is a new page which links to the landing page and can be linked to externally.

I've had to add the [govuk_publishing_components](https://github.com/alphagov/govuk_publishing_components) library in order to add the contents list specified in the design. This component is unfortunately not implemented in [govuk-components](https://github.com/x-govuk/govuk-components).

The Word document uses a lot of bullet lists instead of the paragraphs shown in the prototype. I've used what was in the Word doc as this was pecified in the ticket.

I've also had to create a new component but I've used the responsive spacing helpers instead of fixed pixel sizing as used in the prototype.


rjlynch (approved):
Looks good 👍

PR #3428, opened 2024-11-22
[CAPT-2027] Fix EY practitioner answer

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2027
- EY practitioner is now formatted correctly rather than using the raw internal value for payroll gender


rjlynch (approved):
👍

PR #3436, opened 2024-11-26
Reminders store journey

asmega:
# Context

- This PR is based on PR #3175 and has been ported to the current version which has shifted since then
- Migration adds column on reminder to store the relevant journey
- It also populates this column based off the information we have
- Opening a journey still only sends for additional payments but with a small tweak can be made to send reminders for other journeys

### DIFF
--- a/db/migrate/20241126105650_add_journey_class_to_reminders.rb
+++ b/db/migrate/20241126105650_add_journey_class_to_reminders.rb
@@ -1,0 +1,49 @@
+class AddJourneyClassToReminders < ActiveRecord::Migration[7.2]
+  def up
+    add_column :reminders, :journey_class, :text
+    add_index :reminders, :journey_class
+
+    # old reminders are additional payments
+    Reminder
+      .where(itt_academic_year: ["2022/2023", "2023/2024", "2024/2025"])
+      .update_all(journey_class: Journeys::AdditionalPaymentsForTeaching.to_s)
+
+    # latest year
+    # if itt subject must be additional payments
+    Reminder
+      .where(itt_academic_year: "2025/2026")
+      .where("itt_subject IS NOT NULL")
+      .update_all(journey_class: Journeys::AdditionalPaymentsForTeaching.to_s)
+
+    # latest year
+    # if itt subject blank
+    # match emails to existing to additional payment claims
+    Reminder
+      .joins("JOIN claims on claims.email_address = reminders.email_address")
+      .where(itt_academic_year: "2025/2026")
+      .where(itt_subject: nil)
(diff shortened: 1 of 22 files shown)


spec/lib/tasks/fix_student_loan_repayment_amounts_spec.rb
```diff
@@ -3,6 +3,19 @@
...
+    original_stderr = $stderr
+
+    $stdout = File.open(File::NULL, "w")
+    $stderr = File.open(File::NULL, "w")
+
+    example.run
+
+    $stdout = original_stdout
+    $stderr = original_stderr
+  end
```
asmega, line 17: i want to slow tidy stdout when rspec runs

db/migrate/20241126105650_add_journey_class_to_reminders.rb
```diff
@@ -0,0 +1,39 @@
...
+      .update_all(journey_class: Journeys::AdditionalPaymentsForTeaching.to_s)
+
+    # latest year
+    # if itt subject blank
+    # match emails to existing to additional payment claims
+    Reminder
+      .joins("JOIN claims on claims.email_address = reminders.email_address")
+      .where(itt_academic_year: "2025/2026")
+      .where(itt_subject: nil)
+      .where(claims: {eligibility_type: ["Policies::LevellingUpPremiumPayments::Eligibility", "Policies::EarlyCareerPayments::Eligibility", "Policies::StudentLoans::Eligibility"]})
```
rjlynch, line 25: Should this include student loans as student loans is it's own journey? `Journeys::TeacherStudentLoanReimbursement`

spec/lib/tasks/fix_student_loan_repayment_amounts_spec.rb
rjlynch: Should probably just delete this spec and the rake task as it seems to be a one off task https://dfedigital.atlassian.net.mcas.ms/browse/CAPT-1877

rjlynch (commented):
Overall looks good, just left one comment about the migration that might need looking at, if not LMK and I'm happy to approve.
The other comment isn't a blocker

db/migrate/20241126105650_add_journey_class_to_reminders.rb
asmega: ah nice, good spot!

asmega: I've now tweaked this

rjlynch (approved):
Looks good 👍

PR #3437, opened 2024-11-27
[CAPT-2030] Fix EY started_at timestamp

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2030
- Fix EY claim started at timestamp which is being overwritten when the practitioner starts their half of the journey
- Once deployed will need to retro populate incorrect data which I'll do by hand on the box

### DIFF
--- a/app/forms/journeys/early_years_payment/practitioner/claim_submission_form.rb
+++ b/app/forms/journeys/early_years_payment/practitioner/claim_submission_form.rb
@@ -18,10 +18,18 @@ module Journeys
           []
         end
 
-        def new_or_find_claim
-          (Claim.find_by(reference: journey_session.answers.reference_number) || Claim.new).tap do |c|
-            if c.eligibility
-              c.eligibility.practitioner_claim_started_at = journey_session.answers.practitioner_claim_started_at
+        def existing_or_new_claim
+          Claim.find_by(reference: journey_session.answers.reference_number) || Claim.new
+        end
+
+        def build_claim
+          existing_or_new_claim.tap do |claim|
+            claim.eligibility ||= main_eligibility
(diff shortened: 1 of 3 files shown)


app/forms/journeys/early_years_payment/practitioner/claim_submission_form.rb
```diff
@@ -18,10 +18,18 @@ def generate_policy_options_provided
...
-            if c.eligibility
-              c.eligibility.practitioner_claim_started_at = journey_session.answers.practitioner_claim_started_at
+        def existing_or_new_claim
+          Claim.find_by(reference: journey_session.answers.reference_number) || Claim.new
+        end
+
+        def build_claim
+          existing_or_new_claim.tap do |claim|
+            claim.eligibility ||= main_eligibility
+            claim.eligibility.practitioner_claim_started_at = journey_session.answers.practitioner_claim_started_at
```
rjlynch, line 28:
```suggestion
```
Do we need this line? Is this not handled by the `set_eligibility_attributes` in the base class?

asmega: - I just tried and is needed and evident with failing test
- the provider has already created the claim and eligibility at this time
- unlike the other journeys where claim and eligibility are created once
- therefore it skips `main_eligibility` on the line above (27)
- which would have called `set_eligibility_attributes` which you mentioned

asmega: tbh the existing code its a little bit spaghetti and a bit of a roundabout way. could probably delete most of the stuff and be explicit on what is needed

PR #3438, opened 2024-11-27
[CAPT-1631] If one eligible_itt_subject show better error msg

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-1631
- When only one `eligible_it_subject` show a better error message


rjlynch (approved):
LGTM 👍

PR #3442, opened 2024-11-28
[CAPT-1999] Rails 8

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-1999
- Upgrade to rails 8

# Notes

- despite setting `config.active_support.to_time_preserves_timezone` the deprecation warning is still present. from what I can tell it seems to be issue with how the deprecation warning code implemented and should be fine to ignore


app/models/student_loans_data_importer.rb
```diff
@@ -28,12 +28,28 @@ def row_to_hash(row)
...
+  end
+
+  def calculate_no_of_plans_currently_repaying(value)
+    if value == "No data"
+      nil
+    else
+      value
+    end
+  end
+
```
asmega, line 52: I'm not familiar in this area. can some confirm if this is desired? are there other instances of this that need to be addressed that are potentially missing coverage

rjlynch: I'm not familiar on this bit of code, seems weird that a version upgrade would change this behaviour!

rjlynch (approved):
Upgrade looks good, maybe one to merge after the weekend!

app/models/student_loans_data_importer.rb
asmega: for posterity this change issue was introduced by https://github.com/rails/rails/commit/366909d658e7050d1aea42ab372e85da622ba19c#diff-3f89af7ff219385718b7c0e4635cc4190476349db18f8bf80fb84fa1e7f5e289L240

adding the casting our side is by far a better option and we simply got lucky before it was doing our desired behaviour by default

PR #3445, opened 2024-12-02
[CAPT-1862] Add unsubscribe functionality

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-1862
- Add unsubscribe link to reminder emails
- NOTE: notify templates must to be updated AFTER this change has been deployed

# Changes

- Send `unsubscribe_url` as part of reminder emails in the `personalisation` section so they can be appended to emails sent to users
- Should a user click on this link they are sent to the send to confirm if they wish to unsubscribe
- As recommended by notify https://docs.notifications.service.gov.uk/ruby.html#unsubscribe-link-at-the-bottom-of-the-email-recommended
- Upon success they get show the confirmation panel that transaction is complete
- Should user visit invalid link or if they are already unsubscribed they are show a page with copy explaining the situation
- `config.action_controller.action_on_unpermitted_parameters` has been changed back to the rails default setting which happens to also be the setting used in production. not sure why this was ever changed in the first place as it makes development and test behave differently to how production would behave

# Notes

- One click unsubscribe is not present in this change, however the server side mechanism has been added
- This is because the library we are using at the moment does not support this notify feature see https://github.com/dxw/mail-notify/issues/166
- We either need to switch to official library or add the needed functionality to support this feature

# Screenshots

## After user click unsubscribe link in email
Screenshot 2025-01-09 at 14 18 23

---

## Upon successfully unsubscribing
complete

---

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -65,6 +65,11 @@ Rails.application.routes.draw do
       end
     end
 
+    resources :unsubscribe,
+      path: "/unsubscribe/:resource_class",
+      constraints: {resource_class: /reminders/},
+      only: [:create, :show]
+
     scope constraints: {journey: /further-education-payments|additional-payments/} do
       resources :reminders,
         only: [:show, :update],
--- a/app/controllers/unsubscribe_controller.rb
+++ b/app/controllers/unsubscribe_controller.rb
@@ -1,0 +1,37 @@
+class UnsubscribeController < ApplicationController
+  skip_forgery_protection
+
+  def show
+    @form = Unsubscribe::ConfirmationForm.new(form_params)
+
+    if @form.reminder.nil?
+      render "subscription_not_found"
+    end
+  end
+
+  def create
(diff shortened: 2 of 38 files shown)


config/routes.rb
```diff
@@ -65,6 +65,11 @@ def matches?(request)
       end
     end
 
+    resources :unsubscribe,
+      path: "/unsubscribe/:resource_class",
+      constraints: {resource_class: /reminders/},
+      only: [:create, :show]
+
```
asmega, line 72: `resource_class` is not currently used. buy has been added so can the extended to support unsubscribing of other resource in the future easily if needed

slorek (dismissed):
Thanks for implmenting soft delete - could I ask that you instead re-use the `Deletable` concern to avoid duplication

rjlynch (approved):
This all looks really good. Though I do like the missing parameters raising in dev / test pretty useful for catching issues, still Rails defaults are usually best to stick to

PR #3454, opened 2024-12-05
[CAPT-2051] Better OL name

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2051
- We are only storing from OL the first and last name

# Changes

-  Persist all the names that OL returns
- When checking against OL name we match entered name matches the start and end of returned OL name
- `onelogin_idv_full_name` is no longer a virtual attribute and returns the full name OL returns
- We backfill `onelogin_idv_full_name` with data we have which is concatenation of first and last name

# Notes

- Once this is complete we can work to remove `onelogin_idv_first_name` and `onelogin_idv_last_name` as these will become redundant

### DIFF
--- a/app/models/one_login/core_identity_validator.rb
+++ b/app/models/one_login/core_identity_validator.rb
@@ -21,6 +21,10 @@ class OneLogin::CoreIdentityValidator
     Date.parse(decoded_jwt[0]["vc"]["credentialSubject"]["birthDate"][0]["value"])
   end
 
+  def full_name
+    name_parts.map { |hash| hash["value"] }.join(" ")
+  end
+
   private
 
   def name_parts
--- a/app/models/one_login/core_identity_validator.rb
+++ b/app/models/one_login/core_identity_validator.rb
@@ -22,7 +22,15 @@ class OneLogin::CoreIdentityValidator
   end
 
   def full_name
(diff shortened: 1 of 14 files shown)


app/models/one_login/core_identity_validator.rb
```diff
@@ -21,6 +21,10 @@ def date_of_birth
     Date.parse(decoded_jwt[0]["vc"]["credentialSubject"]["birthDate"][0]["value"])
   end
 
+  def full_name
+    name_parts.map { |hash| hash["value"] }.join(" ")
```
slorek, line 25: Should we ensure that the `FamilyName` part(s) are always last? I suspect there's no guarantee it will be ordered this way already.

asmega: I assume its in the correct order as theres no other data to signify the position of each part. regardless I've tweaked the implementation so `FamilyName` parts are always at the end

asmega:
> Alternatively we could consider keeping the separate first name and last name fields, but concatenating the name parts for `GivenName` and `FamilyName` types, rather than only selecting the first of each

Not sure I understand. Are you saying we should use all `GivenName` parts for the first name? If so we also collect `middle_name` in the app which would probably cause issues if all given names become the first name

slorek (approved):

PR #3457, opened 2024-12-06
Ops reports

rjlynch:
Ops reports

This PR introduces some new reporting functionality for the ops team.
We've added a new button on the admin claims index page that links to the
`/admin/reports` page where the ops team can download various reports.

Initially there are three reports:
1. Duplicate approved claims
2. Approved claims failing qualification task
3. FE approved claims failing provider verification

Reports 2 and 3 are fairly straight forward, we just need to join to the
appropriate tasks and present the data. Report 1 is involved! For report 1 we
need to duplicate the functionality of the `Claim::MatchingAttributeFinder` but
rather than checking for duplicates of a single claim, it needs to work across
all claims. Running the claim matching attribute finder in a loop over all
claims takes a _long_ time (6 minuets to check 1000 claims), even with moving
generating the report into a background job it would be tanning the DB and who
wants to wait for an 1hr 30 for a report to generate! To avoid this we've
re implemented all the duplicate checking in SQL.

Having the report in SQL feels a little less maintainable, so something we
might want to consider is:
* when a claim is created kick off a background job to check for duplicates
  using claim matching attribute finder
* if duplicates are found associate them with the "source" claim using some
  new join model
* duplicate claims are then claims which can be joined with the source model

However we'd need to back fill the duplicate records for existing claims on
production using something like the logic we've written here for finding the
duplicates.

If we do decide to keep the version of the claim duplicate checking from this
pr then we probably want to update the `Claim::MatchingAttributeFinder` to use
the logic introduced in this PR so we don't have two definitions of what
constitutes a duplicate claim.

Open for suggestions of the namespacing of these reports.

NOTE for reviewers: probably worth reviewing each commit separately.

### DIFF
--- a/app/models/admin/reports/duplicate_approved_claims.rb
+++ b/app/models/admin/reports/duplicate_approved_claims.rb
@@ -32,6 +32,8 @@ module Admin
 
       private
 
+      class DuplicateClaimRow < Struct.new(:id, :other_claim_id, keyword_init: true); end
+
       attr_reader :academic_year
 
       def rows
@@ -39,9 +41,12 @@ module Admin
       end
 
       def scope
-        Claim.where(
-          id: Set.new(duplicates_by_eligibility + duplicates_by_attributes)
-        ).includes(decisions: :created_by)
+        Claim.where(id: Set.new(duplicates.map(&:id)))
+          .includes(decisions: :created_by)
+      end
+
+      def duplicates
+        duplicates_by_eligibility + duplicates_by_attributes
       end
 
       def duplicates_by_eligibility
@@ -49,7 +54,7 @@ module Admin
           Policies::POLICIES.map do |policy|
             policy_with_claimable_policies(policy)
           end.compact.join("\nUNION\n")
-        ).map { |row| row["id"] }
+        ).map(&DuplicateClaimRow.method(:new))
       end
 
       def policy_with_claimable_policies(policy)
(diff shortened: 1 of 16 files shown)


app/models/admin/reports/duplicate_approved_claims.rb
```diff
@@ -0,0 +1,222 @@
...
+          ]
+        end
+
+        private
+
+        attr_reader :claim
+      end
+    end
+  end
+end
```
rjlynch, line 222: @kenfodder @slorek instead of doing this gnarly query we could:
* add a new claim verifier for detecting duplicates
* when the verifier runs use the existing matching attributes finder code to find any duplciates
* create a record associating the new claim with these duplicates (`Duplicate belongs_to :source_claim, belongs_to :duplicate_claim`)
* replace the use of matching attribute finder in the app with looking at the new duplicates record for the given claim.
* this report can be simplified by simply joining claims to the new duplicate record.

Might need this query logic for the initial back fill on production but it can then be deleted.

What do you guys think?

rjlynch:
Part of the work in CAPT-2065 (#3468) is to show the matching attribute alert for all
policy combinations (eg include IRP when checking other policies for dupes).
No longer needing to handle policy combinations (as we can compare them all)
will let us greatly simplify the `duplicates_by_eligibility` query.
I'll hold off merging this pr until I've refactored `duplicates_by_eligibility`
to remove the policy combination handling.

PR #3464, opened 2024-12-16
CAPT-1953 - Initial attempt to silence repeated unwanted requests

kenfodder:
Catch unwanted request patterns to stop to them alerting in Rollbar and force them to be a 404.

Tested in Review App seems to work.

Here are the ones that do come through that are deliberately incorrect:

Screenshot 2024-12-16 at 15 22 22


rjlynch (approved):
LGTM 👍

PR #3478, opened 2024-12-30
Log session answers to rollbar

rjlynch:
Some of the errors in rollbar are tricky to debug as we don't know what
answers the claimant provided. This commit adds a mechanism for flagging
attributes as PII and updates the ClaimsController to include the non
pii attributes in the Rollbar scope such that they are available in
Rollbar.


app/models/concerns/journeys/sessions/pii_attributes.rb
```diff
@@ -0,0 +1,28 @@
...
+      end
+
+      class_methods do
+        def attribute(name, type = nil, **options)
+          pii_attributes << name if options.delete(:pii)
+          super
+        end
+      end
+
+      def attributes_with_pii_redacted
```
asmega, line 17: can we add some basic coverage for this method?

rjlynch: Done 👍

slorek (approved):

PR #3485, opened 2025-01-03
[CAPT-2090] Add nil guard where eligible_itt_subject is nil

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2090
- Add nil guard where `eligible_itt_subject` is nil in admin area otherwise throws an error


rjlynch (approved):
LGTM, though I'm not sure if an ECP claim should be able to be submitted without an eligible_itt_subject. Looks like the claim that triggered this error was some sort of dupe and the other claim from the claimant was approved

PR #3490, opened 2025-01-03
[CAPT-649] Correct field types

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-649
- Use correct field types


rjlynch (approved):
LGTM 👍

PR #3500, opened 2025-01-07
CAPT-2104 Redirect to start page if no reminder in session

kenfodder:


rjlynch (approved):
LGTM - may want to show the user a flash message or something

PR #3516, opened 2025-01-15
[CAPT-2145] Refactor controller callbacks

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2145
- Refactor out controller callbacks to elsewhere, typically forms
- I've not been able to remove all of them, there are 3 left which should at least be manageable
- This means the "state machine" work should be much easier with fewer random state changes

### DIFF
--- a/app/controllers/claims_form_callbacks.rb
+++ b/app/controllers/claims_form_callbacks.rb
@@ -11,10 +11,6 @@ module ClaimsFormCallbacks
     redirect_to_slug("eligible-itt-subject") if no_eligible_itt_subject?
   end
 
-  def personal_bank_account_before_update
-    inject_hmrc_validation_attempt_count_into_the_form
-  end
-
   def information_provided_before_update
     return unless journey.requires_student_loan_details?
 
@@ -26,11 +22,6 @@ module ClaimsFormCallbacks
 
     retrieve_student_loan_details
     redirect_to_next_slug
-  end
-
-  def personal_bank_account_after_form_save_failure
-    increment_hmrc_validation_attempt_count if hmrc_api_validation_attempted?
-    render_template_for_current_slug
   end
 
   def check_your_email_after_form_save_success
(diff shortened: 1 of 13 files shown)


app/forms/bank_details_form.rb
```diff
@@ -48,6 +47,10 @@ def hmrc_api_validation_succeeded?
 
   private
 
+  def hmrc_validation_attempt_count
+    journey_session.answers.hmrc_validation_attempt_count
```
rjlynch, line 51: 🙏 so nice to get rid of `inject_hmrc_validation_attempt_count_into_the_form`!

app/forms/personal_details_form.rb
```diff
@@ -60,8 +60,11 @@ def save
     )
 
     reset_dependent_answers_attributes
-
     journey_session.save!
+
+    if journey.requires_student_loan_details?
+      journey::AnswersStudentLoansDetailsUpdater.call(journey_session)
+    end
```
rjlynch, line 67: 👍 makes so much more sense to do this in the form!

app/controllers/claims_form_callbacks.rb
```diff
@@ -1,50 +1,8 @@
 module ClaimsFormCallbacks
-  def current_school_before_show
-    set_backlink_override_to_current_slug if on_school_search_results?
-  end
-
-  def claim_school_before_show
-    set_backlink_override_to_current_slug if on_school_search_results?
-  end
-
   def teaching_subject_now_before_show
```
rjlynch, line 2: I think this one could be handled by the additional payments slug sequence. Not a blocker for this pr, have opened a separate ticket https://dfedigital.atlassian.net/browse/CAPT-2152

rjlynch (approved):
Really nice work, this makes things much more maintainable 👍

app/controllers/claims_form_callbacks.rb
asmega: i tried this one theres an explicit test for it which would need to be "modified". there isn't really a mechanism at the moment to allow for this kinda functionality without writing explicit controllers to bounce the use to correct place.

PR #3524, opened 2025-01-20
Move fields to the "clear" config

rjlynch:
Data and Insights have informed us that BigQuery is now allowed to
handle storing PII. This commit moves the requested fields from the
block list to the "clear" analytics config to allow them to be synced
over.

## Upon merging
- [ ] Inform Sophie the fields will be pulled over
- [ ] Kick off an analytics reimport - `bin/rake dfe:analytics:import_all_entities`

### DIFF
--- a/config/analytics.yml
+++ b/config/analytics.yml
@@ -80,6 +80,7 @@ shared:
     - logged_in_with_onelogin
     - onelogin_auth_at
     - onelogin_idv_at
+    - onelogin_uid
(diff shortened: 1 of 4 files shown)


slorek (changes requested):
The PII fields need to be added to `analytics_hidden_pii.yml` rather than `analytics.yml`.

slorek (approved):

PR #3527, opened 2025-01-22
CAPT 2121/expand topup functionality

rjlynch:
WIP


app/forms/admin/create_topup_form.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    validate :topup_is_valid
+    validate :note_is_valid
+
+    def initialize(claim:, created_by:, params: {})
+      @claim = claim
+      @created_by = created_by
+
+      super(params)
+    end
```
asmega, line 23: pretty sure you don't need to implement `initialize` and will work if you just delete this method?

asmega (approved):
looks fine but i'm not familiar with any of the payroll stuff tbh

rjlynch:
> looks fine but i'm not familiar with any of the payroll stuff tbh

Thanks, I checked that it doesn't change any of the old payrolls!

PR #3537, opened 2025-01-29
Make email address validation notify compliant

rjlynch:
Updates our email address validation to, mostly, match the validation
logic in notify. To avoid pulling in a gem to the project we're not
converting to puny code so some domains like `müller-büromöbel.de` won't
be permitted. If we want to support these domains we can convert them to
puny code and validate that
`https://github.com/alphagov/notifications-utils/blob/e89dcc53c7f88ef23053d1efb2c00b8ef2ea377e/notifications_utils/recipient_validation/email_address.py#L38`

The email validation code in notify seems pretty stable, having last
changed in 2022.

**Other option is to just rescue notify api errors and add a validation error** might be simpler.

### DIFF
--- a/spec/validators/email_address_format_validator_spec.rb
+++ b/spec/validators/email_address_format_validator_spec.rb
@@ -29,7 +29,7 @@ RSpec.describe EmailAddressFormatValidator do
         record = TestRecord.new
         record.email = string
 
-        expect(record.valid?).to be false
+        expect(record).not_to be_valid
 
         expect(record.errors[:email]).to include("no good")
       end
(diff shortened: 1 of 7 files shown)


spec/validators/email_address_format_validator_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      "test@example.com",
+      "first.o'surname+123@sub.domain.co.uk",
+      "abc@my-long-subdomain.gov.uk",
+      "someone@domain.org"
+    ].each do |string|
+      it "accepts #{string}" do
+        record = TestRecord.new
+        record.email = string
+
+        expect(record.valid?).to be true
```
asmega, line 50:
```suggestion
        expect(record).to be_valid
```

asmega (approved):

slorek (approved):
I'd suggest breaking down the `validate_each` method in the validator class into several private methods so it's more understandable. But otherwise I think the approach is sound - email address formats aren't changing any time soon.

PR #3538, opened 2025-01-29
[CAPT-2041] More FE provider verification chasers

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2041
- Now geared to be able to send infinite number of FE provider verification chasers
- Configurable with `CHASE_INTERVALS`
- `provider_verification_chase_email_last_sent_at` is no longer used and can be removed after this change is merged
- Instead `provider_verification_email_last_sent_at` simply stores last time we email the provider
- New column `provider_verification_email_count` now used to indicate how many times we have contacted provider


db/migrate/20250129115013_add_provider_verification_email_count.rb
```diff
@@ -0,0 +1,14 @@
...
+
+    Policies::FurtherEducationPayments::Eligibility
+      .where("provider_verification_email_last_sent_at IS NOT NULL")
+      .where("provider_verification_chase_email_last_sent_at IS NULL")
+      .update_all(provider_verification_email_count: 1)
+
+    Policies::FurtherEducationPayments::Eligibility
+      .where("provider_verification_chase_email_last_sent_at IS NOT NULL")
+      .update_all(provider_verification_email_count: 2)
+  end
```
rjlynch, line 13: 👍

rjlynch (approved):
Looks good 👍

db/migrate/20250129115013_add_provider_verification_email_count.rb
```diff
@@ -0,0 +1,14 @@
+class AddProviderVerificationEmailCount < ActiveRecord::Migration[8.0]
+  def change
+    add_column :further_education_payments_eligibilities, :provider_verification_email_count, :integer, default: 0
+
```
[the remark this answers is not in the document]
asmega: thanks good spot, i've added now

db/migrate/20250129115013_add_provider_verification_email_count.rb
asmega: now converted to use up and down migrations

PR #3550, opened 2025-02-04
Raise if student loan plan is blank

rjlynch:
We ran into an issue on production where this check was passing but the
claim had no student loan data. `ClaimStudentLoanDetailsUpdater` used to
rescue errors, so when `StudentLoanPlanCheckJob` ran and
`ClaimStudentLoanDetailsUpdater` errored this claim verifier was run for
a claim without a student loan set, causing much confusion. The
intention is for this claim verifier to always run after the
`ClaimStudentLoanDetailsUpdater` has run, so we're putting in this guard
in case a future code change accidentally breaks this assumption. We
should probably think about combining these two classes, or change how
we store SLC data on a claim so we don't need to find it twice.

There is a potential edge case here:
* Claimant completes the personal details step
* No student loan data is found
* Admin uploads student loan data that matches the claimant
* Claim is submitted
* Claim verifier runs
* We now throw an error (previously we pass the task with no slc data)

Raising is probably better in this scenario so we're alerted and know to
update the claim.


rjlynch:
After thinking about it, I think we just want to raise here rather than trying to automatically set the student loan. This will give us some visibility into the issue and let us investigate what scenarios this occurs under.

asmega (approved):

PR #3551, opened 2025-02-04
[CAPT-2124] Change test environment cookie key

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2124
- Prevent prod and test cookies from clashing


rjlynch (approved):
Looks good 👍

PR #3552, opened 2025-02-04
Payments show

rjlynch:
Add payments show resource

The claim payments index shows the payments belonging to a specific
claim, however a payment can contain multiple claims and topups, this
commit adds a payment show page where admins can view all the claims and
topups belonging to a payment along with a summary of the various
payment amounts.

### DIFF
--- a/app/controllers/admin/payments_controller.rb
+++ b/app/controllers/admin/payments_controller.rb
@@ -1,11 +1,18 @@
 module Admin
   class PaymentsController < BaseAdminController
     before_action :ensure_service_operator
-    before_action :find_payroll_run, except: [:index]
-    before_action :find_payment, except: [:index]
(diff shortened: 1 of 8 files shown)


app/controllers/admin/payments_controller.rb
```diff
@@ -1,13 +1,20 @@
 module Admin
   class PaymentsController < BaseAdminController
     before_action :ensure_service_operator
-    before_action :find_payroll_run, except: [:index]
-    before_action :find_payment, except: [:index]
+    before_action :find_payroll_run, except: [:index, :show]
+    before_action :find_payment, except: [:index, :show]
```
asmega, line 5: non blocker but think i would opt to whitelist this, or inline as extraction doesn't really have much benefit

asmega (approved):

PR #3554, opened 2025-02-05
FE address fix

asmega:
# Context

- As part of the navigator/state.machine work applied to FE journey there was a regression where for the FE journey searching for postcode then either choosing an address result or then opting to manually enter address was not working as desired. This PR should address these issues with extra coverage to ensure no future regressions


rjlynch (approved):
Looks good 👍

PR #3558, opened 2025-02-07
Rate limit file uploads

rjlynch:
Feedback from the ITHC.
We don't have a cache set so rate limiting will use the file cache,
storing the data in ./tmp on the containers. The rate limiting will be
per container, rather than app wide but given these uploads are
restricted to admins it wont be an issue. We don't do a lot of file
uploads so storing this info in the container file system shouldn't be
an issue.
If we start supporting public file uploads we'll want to investigate a
different cache store.


app/controllers/admin/eligible_ey_providers_controller.rb
```diff
@@ -4,6 +4,20 @@ class EligibleEyProvidersController < BaseAdminController
 
     helper_method :journey_configuration
 
+    rate_limit(
```
asmega, line 7: this uses `cache_store` which appears to not be set for production, which means this might be the null store? if thats the case this rate limit won't work?

rjlynch: It defaults to the file store, at least on review apps (which run in production mode)
```
dfe-teachers-payment-service(prod)> Rails.application.config.cache_store
=> [:file_store, "/app/tmp/cache/"]
```
This will be _per_ container so the rate limit is a bit higher than 1 request per IP per 30 seconds, but I don't think that should be an issue as it's only admins accessing this.
Nice spot though!

asmega (approved):

PR #3559, opened 2025-02-07
CAPT-2149 csv upload history and versioning

kenfodder:
Reviewing:
* Might be easier to review each commit to see the progression per importer.

FileUpload
* Repurpose FileUpload to retain uploads if required, adding `target_data_model`, `academic_year` and `completed_processing_at` attributes
* Using the FileUpload#id as a concept of a "version"

`EligibleEyProvider`, `EligibleFeProvider` and `Policies::LevellingUpPremiumPaymentsAward`
* Add `file_upload_id` foreign_key
* Data is not purged, set in the importer
* Added scopes handle fetching the correct latest version
* It finds the latest version by looking up the `FileUpload.latest_version_for(<target_data_model>, <academic_year>)` (academic_year is not require for EY), it uses the `completed_processed_at` and `created_at: :desc` to get the latest.  

Importer/CSV upload
* All uploads are stored
* Only uploads where the importing of data is completed is the FileUpload marked with a `completed_processing_at`, once that timestamp is set, it becomes the latest "version"

Basic File Upload history added to the journey upload admin pages to see a list of uploads, by AY, uploaded by and timestamp, 

Won't do in this PR:
  * Future addition to add downloading the CSV body that was uploaded to the upload history, fairly easy
  * Refactor STRI/LUPP CSV upload controller and importer to be the same pattern and the FE/EY importer (I tried to start this but backed out, too much for this PR)

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -21,15 +21,6 @@ if Rails.env.development? || ENV["ENVIRONMENT_NAME"].start_with?("review")
   Rake::Task["db:fixtures:load"].invoke
 end
 
-if ENV["ENVIRONMENT_NAME"].start_with?("review")
-  EligibleFeProvidersImporter.new(
-    File.new(Rails.root.join("spec/fixtures/files/eligible_fe_providers.csv")),
-    AcademicYear.current
-  ).run
-
-  SchoolDataImporterJob.perform_later if School.count < 100
-end
-
 if Rails.env.development?
   require "./lib/factory_helpers"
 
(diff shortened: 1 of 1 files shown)


app/models/eligible_ey_provider.rb
```diff
@@ -1,5 +1,8 @@
 class EligibleEyProvider < ApplicationRecord
   belongs_to :local_authority
+  belongs_to :file_upload
```
asmega, line 3: need `optional: true` on the end? unless at somehow existing records have this?

rjlynch (approved):
This looks good to me, just wish there was a way to avoid having to store academic year on the file upload model!

asmega (approved):

PR #3567, opened 2025-02-10
Replace enum with boolean #1

rjlynch:
First part of replacing the result enum with a boolean column.
This commit writes to the new boolean column when updating the result
enum.

### DIFF
--- a/db/migrate/20250210155819_back_fill_decisions_approved_column.rb
+++ b/db/migrate/20250210155819_back_fill_decisions_approved_column.rb
@@ -1,0 +1,18 @@
+class BackFillDecisionsApprovedColumn < ActiveRecord::Migration[8.0]
+  def up
+    execute <<-SQL
+      UPDATE decisions
+      SET approved = CASE result
+        WHEN 0 THEN TRUE
+        WHEN 1 THEN FALSE
+      END;
+    SQL
+  end
+
+  def down
+    execute <<-SQL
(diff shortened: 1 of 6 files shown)


db/migrate/20250210155819_back_fill_decisions_approved_column.rb
```diff
@@ -0,0 +1,18 @@
+class BackFillDecisionsApprovedColumn < ActiveRecord::Migration[8.0]
+  def up
+    execute <<-SQL
+      UPDATE decisions
+      SET approved = CASE result
+        WHEN 0 THEN TRUE
+        WHEN 1 THEN FALSE
+      END;
```
asmega, line 8: is result ever `nil`?

rjlynch: result _shouldn't_ ever be `null` as there's a presence validation on result.

> just ignore those and leave as null?
the next migration makes the `approved` column non nullable so can't leave them as null! I'm hoping there's no decisions with a `null` result on production as it doesn't make sense to create a decision without a result!

asmega: > I'm hoping there's no decisions with a `null` result on production
probs worth checking to prevent failed deploy

asmega (approved):

PR #3568, opened 2025-02-10
Navigator supports changing answers

asmega:
# Context

- This allows navigator to support changing answers in a journey
- When a user reaches a changing answers screen and clicks `change answer` link
- we append a param to the url `change=slug-of-check-answers`
- we use the `slug` of check answers page as some journeys have multiple check answer pages and this mechanic supports that
- the `change` param is continually passed forward so we know the user is in the process of changing answer
- if the change of answer does not affect their journey they are sent back the check your answers screen
- otherwise they fill in new forms they now must answer as a consequence of changing answer after which they will then be sent back to check answers page
- Also now using `govuk_summary_list` component on check answers page and removed custom version

### DIFF
--- a/app/forms/form.rb
+++ b/app/forms/form.rb
@@ -49,8 +49,8 @@ class Form
   end
 
   def url
-    if journey.use_navigator? && params[:change] == "true"
-      claim_path(journey::ROUTING_NAME, params[:slug], change: true)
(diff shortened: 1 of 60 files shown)


rjlynch (approved):
Really nice 👍 all makes sense

PR #3570, opened 2025-02-11
CAPT 1473/convert decision enum to boolean part 2

rjlynch:
Make the approved column non nullable

Now the backfill has run and we've checked the data on prod we can make
the approved column non nullable as it doesn't make sense to create a
decision without out a value for approved.

Now we've backfilled the existing records we can remove the enum
definition and reference the `approved` column.
We'll drop the `result` column in a future PR.

### DIFF
--- a/app/models/decision.rb
+++ b/app/models/decision.rb
@@ -13,7 +13,10 @@ class Decision < ApplicationRecord
   # NOTE: Don't store the rejected_reasons data params from the form when Approve is selected
   before_validation :clear_rejected_reasons, unless: :rejected?
 
-  validates :result, presence: {message: "Make a decision to approve or reject the claim"}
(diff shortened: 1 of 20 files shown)


app/models/decision.rb
```diff
@@ -23,29 +26,20 @@ class Decision < ApplicationRecord
...
-        value
-      end
+  # FIXME RL: Remove this once we've dropped the column
+  self.ignored_columns = %w[result]
 
-    self.approved = (normalized_value == self.class.results[:approved])
-
-    super(normalized_value)
+  def rejected?
+    !approved? && !approved.nil?
```
asmega, line 36: if you swap ordering i think you get tiny optimisation 😅

asmega (approved):

PR #3579, opened 2025-02-13
Add name / dob fields to analytics for FE only

rjlynch:
We want to send over name and date of birth over to DfE analytics for FE
claims only. These attributes are stored on the claim which is shared
between all policies. DfE analaytics gem only gives us course grain
control over which attributes to send to big query. This is a hack and
we're not stoked on having to modify the attributes hash like this but
it should do the trick until we think of a better solution.

This approach is _probably_ better https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/3581/files

### DIFF
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -411,6 +411,24 @@ class Claim < ApplicationRecord
     @attributes_flagged_by_risk_indicator ||= RiskIndicator.flagged_attributes(self)
   end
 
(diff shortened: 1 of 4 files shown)


rjlynch:
This is no longer required as we're going to send a CSV over to the analytics team, see discussion here https://dfedigital.atlassian.net.mcas.ms/browse/CAPT-2179

PR #3582, opened 2025-02-13
[CAPT-2195] OL return codes

asmega:
# IMPORTANT

- ~Prior to merge MUST enable return codes in OL first + enable claim `returnCode` otherwise this feature will not work~
- This has been completed

# Context

- https://dfedigital.atlassian.net/browse/CAPT-2195
- This adds the `returnCode` claim to the OL idv check
- Create a new db backed `Stat` model to store arbitrary stat data points
- Store any OL return codes into this `Stat` model

### DIFF
--- a/app/controllers/omniauth_callbacks_controller.rb
+++ b/app/controllers/omniauth_callbacks_controller.rb
@@ -136,7 +136,7 @@ class OmniauthCallbacksController < ApplicationController
     journey_session.save!
 
     one_login_return_codes.each do |code|
-      Stat.create!(one_login_return_code: code)
+      Stats::OneLogin.create!(one_login_return_code: code)
     end
 
     redirect_to(
--- a/app/controllers/omniauth_callbacks_controller.rb
+++ b/app/controllers/omniauth_callbacks_controller.rb
@@ -2,6 +2,7 @@ class OmniauthCallbacksController < ApplicationController
   include JourneyConcern
 
   ONELOGIN_JWT_CORE_IDENTITY_HASH_KEY = "https://vocab.account.gov.uk/v1/coreIdentityJWT".freeze
+  ONELOGIN_JWT_RETURN_CODE_HASH_KEY = "https://vocab.account.gov.uk/v1/returnCode".freeze
 
   def callback
     auth = request.env["omniauth.auth"]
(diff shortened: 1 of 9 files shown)


app/models/stat.rb
```diff
@@ -0,0 +1,2 @@
+class Stat < ApplicationRecord
```
rjlynch, line 1: Are we calling this a stat because OL return codes are only used for DfE analytics?

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -115,6 +128,25 @@ def process_one_login_identity_verification_callback(core_identity_jwt)
...
 
+  def process_one_login_return_codes_callback
+    journey_session.answers.assign_attributes(
+      identity_confirmed_with_onelogin: false,
+      onelogin_idv_at: Time.now
+    )
+    journey_session.save!
+
+    one_login_return_codes.each do |code|
+      Stat.create!(one_login_return_code: code)
```
rjlynch, line 139:
```suggestion
      Stat.create!(one_login_return_code: code, journey_session: journey_session)
```
Feel like we probably want to associate these with the session they came from, always end up needing this kind of info for the inevitable "what % of users with return code X went on to successfully complete a claim"

rjlynch (approved):
LGTM

Left one suggestion.
I'm not 100% sure on the name Stats, maybe `Stats::OneLoginReturnCode` but 🤷 not a blocker

app/models/stat.rb
asmega: Ticket says it needs to be a new thing, so I added this. I'm thinking `Stat` could be a generic thing and this may a useful and quick thing just to stick in data points that can consumed in big query?

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -44,14 +45,26 @@ def sign_out
     end
   end
 
+  # unfortunely this method has dual responsibilites
+  # handles both auth callback + idv callback
+  # logic must be included to handle this shortcoming
   def onelogin
-    core_identity_jwt = omniauth_hash.extra.raw_info[ONELOGIN_JWT_CORE_IDENTITY_HASH_KEY]
     return process_one_login_identity_verification_callback(core_identity_jwt) if core_identity_jwt
+    return process_one_login_return_codes_callback if one_login_return_codes.present?
```
slorek, line 53: Is the early return on the previous line going to stop this from ever running?

app/controllers/omniauth_callbacks_controller.rb
slorek: If it’s easy to associate then we can do it as may be useful, but in this case I expected the journey session would be purged as it will not relate to a submitted claim until we allow the user to continue.

app/models/stat.rb
slorek: Good thinking - as a suggestion we could have a type column and then make the value column a generic name so it’s more easily reusable.

asmega: done

slorek (approved):

PR #3587, opened 2025-02-17
LUP => TRI rename + data migration

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2128
- Mass rename LUP to TRI
- No changes have been made to FE and currently has no connection with TRI
- Expanded out acronyms from TRI to `targeted retention incentive`
- Unfortunately previous migration had incorrect primary key therefore this PR includes the table drop (of empty and currently unused table)

# Review notes

- Most important part is the data migration
- The rest is just variable and class renaming


rjlynch (approved):
Data migration looks good 👍 
Skimmed the rest of the changes and they look fine

PR #3598, opened 2025-02-20
[CAPT-2246] Data migration for policy_options_provided

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2246
- Data migration for `policy_options_provided`

### DIFF
--- a/db/data/20250220124403_policy_options_provided_lup_to_tri.rb
+++ b/db/data/20250220124403_policy_options_provided_lup_to_tri.rb
@@ -1,0 +1,13 @@
+# Run me with `rails runner db/data/20250220124403_policy_options_provided_lup_to_tri.rb`
+
+r = ActiveRecord::Base.connection.execute("SELECT id from (SELECT id, jsonb_path_query(policy_options_provided, '$.policy') AS options FROM \"claims\") as inner_table WHERE options = '\"LevellingUpPremiumPayments\"'")
+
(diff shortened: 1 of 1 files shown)


db/data/20250220124403_policy_options_provided_lup_to_tri.rb
```diff
@@ -0,0 +1,13 @@
+# Run me with `rails runner db/data/20250220124403_policy_options_provided_lup_to_tri.rb`
+
+r = ActiveRecord::Base.connection.execute("SELECT id from (SELECT id, jsonb_path_query(policy_options_provided, '$.policy') AS options FROM \"claims\") as inner_table WHERE options = '\"LevellingUpPremiumPayments\"'")
+
+ids = r.map { |e| e["id"] }
```
rjlynch, line 5: I _think_ we could do
```
ids = Claim.where(%{policy_options_provided @> '[{"policy": "LevellingUpPremiumPayments"}]'}).pluck(:id)
```
but not a blocker for a data migration 👍

rjlynch (approved):
LGTM 👍

db/data/20250220124403_policy_options_provided_lup_to_tri.rb
asmega: i always have to look up json syntax, don't use it enough to memorise, and ends up being different every time

PR #3600, opened 2025-02-20
Revert bugfix

rjlynch:
Now the data migration has been run we don't need this special case
handling.

Don't merge until data-migration-policy-options has been run on production


asmega (approved):
data migration is now complete

PR #3602, opened 2025-02-21
Add Form::Option

rjlynch:
As part of the LUP separate journey work I've found myself introducing a
Struct to represent form options in a couple of forms. This commit adds
an Option class to the base form and replaces usage of OpenStruct with
it. We use a class rather than a Struct as we have some optional key
words.
`value` and `label` would probably make more sense as attribute names
but the existing code, and the form builder docs, use `id` and `name` so
we've stuck with them.

### DIFF
--- a/app/forms/form.rb
+++ b/app/forms/form.rb
@@ -1,4 +1,22 @@
 class Form
+  class Option
+    attr_reader :id, :name, :description, :hint
+
+    def initialize(id:, name:, description: nil, hint: nil)
+      @id = id
+      @name = name
+      @description = description
+      @hint = hint
+    end
+
+    def ==(other)
+      id == other.id &&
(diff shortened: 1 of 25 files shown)


asmega (approved):

app/forms/form.rb
```diff
@@ -1,4 +1,22 @@
...
+      @description = description
+      @hint = hint
+    end
+
+    def ==(other)
+      id == other.id &&
+        name == other.name &&
+        description == other.description &&
+        hint == other.hint
+    end
```
asmega, line 17: is this just used for the spec? or is there some other use for it?

rjlynch: Just used to get the spec to pass. I think it's worth having rather than changing the spec

PR #3608, opened 2025-02-24
Make sure personal details form returns true

rjlynch:
Weird things can happen otherwise!


asmega (approved):

slorek (approved):
Perhaps we should have some specs (shared examples) to ensure all forms return a Boolean?

PR #3612, opened 2025-02-25
use navigator for FE provider journey

asmega:


app/forms/journeys/further_education_payments/provider/omniauth_callback_form.rb
```diff
@@ -19,7 +19,8 @@ def save!
             dfe_sign_in_role_codes: dfe_sign_in_role_codes,
             dfe_sign_in_first_name: dfe_sign_in_first_name,
             dfe_sign_in_last_name: dfe_sign_in_last_name,
-            dfe_sign_in_email: dfe_sign_in_email
+            dfe_sign_in_email: dfe_sign_in_email,
+            claim_started_verified: journey_session.answers.claim.eligibility.verified?
```
asmega, line 23: we cache state of whether claim is verified when journey starts. otherwise chicken and egg occurs after claim becomes verified we would send user to already verified page, this caching prevents that

app/models/journeys/further_education_payments/provider.rb
```diff
@@ -12,7 +12,11 @@ module Provider
 
       FORMS = {
         "claims" => {
-          "verify-claim" => VerifyClaimForm
+          "verify-claim" => VerifyClaimForm,
+          "complete" => CompleteForm,
+          "unauthorised" => UnauthorisedForm,
+          "expired-link" => ExpiredLinkForm,
+          "already-verified" => AlreadyVerifiedForm
```
asmega, line 19: all pages must now be forms

app/models/journeys/further_education_payments/provider/authorisation.rb
```diff
@@ -2,24 +2,22 @@ module Journeys
   module FurtherEducationPayments
     module Provider
       class Authorisation
-        def initialize(answers:, slug:)
+        def initialize(answers:)
```
asmega, line 5: auth no longer needs to know where the user is. it is now more about are they authorised for the whole journey

app/models/journeys/further_education_payments/provider/authorisation.rb
```diff
@@ -2,24 +2,22 @@ module Journeys
...
           @answers = answers
-          @slug = slug
         end
 
         def failure_reason
           return :organisation_mismatch if organisation_mismatch?
           return :no_service_access unless answers.dfe_sign_in_service_access?
           return :claim_admin if claim_admin?
           return :incorrect_role unless role_permits_access?
-          return :already_verified if already_verified? && slug != "complete"
```
asmega, line 15: I don't think this is auth related and has now been moved into the normal slug sequence flow

app/views/further_education_payments/provider/claims/unauthorised.html.erb
```diff
@@ -6,9 +6,17 @@
...
+    auth_checker = journey::Authorisation.new(
+      answers: journey_session.answers
+    )
+
+    failure_reason = auth_checker.failure_reason.to_s
+%>
+
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-  <% case params[:failure_reason] %>
```
asmega, line 11: this can now be determined so no longer need to pass params about

spec/models/journeys/further_education_payments/provider/authorisation_spec.rb
```diff
@@ -69,23 +69,5 @@
 
       it { is_expected.to eq(:claim_admin) }
     end
-
-    context "when the claim has already been verified" do
```
asmega, line 73: responsibility lies elsewhere, this is already covered by a feature spec

app/views/further_education_payments/provider/claims/unauthorised.html.erb
rjlynch: I think we can do this if the navigator exposes the failure reason
```
redirect_to claim_path(current_journey_routing_name, "unauthorised", navigator.unauthorised_reason)
```

rjlynch (approved):
Approved, though I really think we should keep passing the auth failure param, based on my experience with AP, calculating the ineligibility reason in two places makes it _very_ hard to change the code, if we keep passing the param we can just determine the auth reason in the slug sequence. + the more app state that can be determined from the url params the easier it is to reason about.

spec/features/further_education_payments/provider/provider_verifying_claims_spec.rb
```diff
@@ -56,11 +56,9 @@
     claim_link = Journeys::FurtherEducationPayments::Provider::SlugSequence.verify_claim_url(claim)
 
     visit claim_link
-
```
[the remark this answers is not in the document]
asmega: oh i don't use that shortcut. but regardless it should now make it easier as they are now logically group by page?

app/models/journeys/further_education_payments/provider/authorisation.rb
```diff
@@ -2,24 +2,22 @@ module Journeys
   module FurtherEducationPayments
     module Provider
       class Authorisation
-        def initialize(answers:, slug:)
```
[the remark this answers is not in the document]
asmega: sorry, my comment wasn't fully accurate. it's still very much possible to auth on selected pages. at the moment it's implemented like the following:
```
 79     def requires_authorisation?
 80       journey::SlugSequence::RESTRICTED_SLUGS.include?(current_slug)
 81     end
```
But what I mean was the `Authorisation` class should more be about are they allowed in as a whole not on what slug they are viewing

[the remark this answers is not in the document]
rjlynch: Ah the original plan was for Authorisation to determine if access to a specific slug was allowed, that way a journey could have some slugs that were unauthorised and some that required authorisation, just so happens that the FE provider journey required all slugs to be authorised. Given we never actually used that functionality I don't think it's an issue to remove it (YAGNI)

app/views/further_education_payments/provider/claims/unauthorised.html.erb
rjlynch: > this is only the auth for this section, i'm not familiar with how auth works for AP or if it's even needed at all?

There's no auth on AP, the issues on AP are to do with the eligibility checking where the eligibility check was performed in two places (one in the view to show the reason, one in the page sequence to kick you to ineligible) and it, looks like it, eventually became two separate classes for checking the eligibility (policy eligibility checker and lib/ineligibility_reason_checker). Passing the reason as a param solves this issue.

Ideally the app would do something like this:
```
redirect_to claim_path(current_journey_routing_name, "unauthorised", navigator.unauthorised_reason) if navigator.unauthorised_reason.present?

redirect_to claim_path(current_journey_routing_name, "ineligible", navigator.ineligible_reason) if navigator.ineligibile_reason.present?
```
but we're a way off that and it's just personal preference!

Not a blocker!

app/models/journeys/further_education_payments/provider/authorisation.rb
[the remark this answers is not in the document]
asmega: I'm gonna leave in for now as it works, but yes might want to re-visit

PR #3617, opened 2025-02-27
Add `try_t`

rjlynch:
This method doesn't raise when a translation is missing, which is handy
for optional text like hints, letting us avoid a conditional.
````
hint: { text: @form.try_t([:hint, @form.answers.qualification]) }
````

### DIFF
--- a/app/forms/form_helpers.rb
+++ b/app/forms/form_helpers.rb
@@ -22,6 +22,12 @@ module FormHelpers
     I18n.t("#{i18n_namespace}.#{base_key}", default: base_key, **args)
   end
 
+  def try_t(key, args = {})
+    t(key, args)
+  rescue I18n::MissingTranslationData
+    nil
+  end
+
   private
 
   def i18n_form_namespace

### CODE REVIEW

**Comment by asmega** [comment]
what would the use case be? if you didn't want the hint text then wouldn't you simply not add it in the first place? or is it due to some kind of sharing where some versions do different things?

**Comment by rjlynch** [comment]
> what would the use case be? if you didn't want the hint text then wouldn't you simply not add it in the first place? or is it due to some kind of sharing where some versions do different things?

It's for when some options have a hint and some don't and you want to keep that info in the locales
```
        QUALIFICATION_OPTIONS.map do |option|
          Option.new(
            id: option,
            name: t("options.#{option}"),
            hint: try_t("hints.#{option}")
          )
        end
```
or
```
      <%= f.govuk_radio_buttons_fieldset(
        :itt_academic_year,
        legend: {
          text: @form.t([:question, @form.answers.qualification]),
          tag: "h1",
          size: "l"
        },
        hint: { text: @form.try_t([:hint, @form.answers.qualification]) }
```
where the presence of the hint depends on the qualification chosen in a previous screen

I guess could always add an explicit conditional but it's kind of ugly!

asmega (approved):

slorek (changes requested):
I checked the source for `translate` here: https://github.com/rails/rails/blob/8030cff808657faa44828de001cd3b80364597de/actionview/lib/action_view/helpers/translation_helper.rb#L60

It seems if you pass `raise: false` as an option to `t` then it will not raise an exception.

Is this equivalent to what you were trying to do?

PR #3619, opened 2025-02-28
[CAPT-2249] OL without idv

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2249
- Allow users who cannot pass OL idv to continue their journey

# Changes

- For review app testing this always makes the user fail OL idv
- In order to achieve this more explicit mocking of OL has had to be added to tests
- We store their OL return codes onto their session and subsequently their claim when created


rjlynch (approved):
Code changes look good.
Do we need some sort of feature flag for this until the provider piece has been done?

asmega:
> Do we need some sort of feature flag for this until the provider piece has been done?

Wasn't on planning on merging for now. I don't think we have any kind of feature flag system in place at the moment but probably warrants a discussion first. happy to go either way on this.

asmega:
@rjlynch I've re-jigged this so it now fits in with new `FeatureFlag` feature

PR #3622, opened 2025-03-03
pass session into answers

rjlynch:
Pass reference to a session to answers

Sometimes we need a reference to the session in the answers for things
like timestamps.
Passing a reference to the session to the answers allows us to move some
attributes that are derived from the session answers off the journey session
model and onto the answers model.
Having access to the session on the answers will also allow us to implement
things like `answers.update!` and clean up the code where we have to do
`journey_session.assign_attributes(xxx); journey_session.save!`.

As we want to support subclasses of session answers for each journey we
unfortunately have to use a bit of Ruby magic to get this to work.

This code is extracted from the LUP only journey branch. If this approach is a
bit too esoteric we can leave some methods on the journey session and I'll
update the LUP only branch.

### DIFF
--- a/app/forms/journeys/additional_payments_for_teaching/correct_school_form.rb
+++ b/app/forms/journeys/additional_payments_for_teaching/correct_school_form.rb
@@ -19,7 +19,7 @@ module Journeys
       private
 
       def current_school
-        @current_school ||= journey_session.recent_tps_school || answers.current_school
+        @current_school ||= answers.recent_tps_school || answers.current_school
       end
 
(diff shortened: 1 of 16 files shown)


slorek (commented):
I'm not too concerned with the 'magic' here, but I'm split on whether it is better to be explicit about where the answer comes from: `journey_session.answers.recent_tps_school` tells me this but `answers.recent_tps_school` could be anything especially as it's not immediately clear where `answers` comes from when looking at the form class.

PR #3628, opened 2025-03-06
Fix admin amend claim

asmega:
# Context

- When admin amends a claim and does not choose a student loan plan
- Browser sends this as empty string
- On the backend this is stored as nil
- So on validation of form we convert empty string to nil
- This allows the form to be saved and not raise an error
- Also fixes regression and FE award amount + TRN is now amendable again


rjlynch (approved):
LGTM 👍

PR #3629, opened 2025-03-07
[CAPT-2180] Amend banking name

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2180
- All back office users can view banking name in the amend screen if the call to HMRC failed
- If HMRC check passed back office users do not see the banking name
- If back office user has admin role that can amend the banking name
- For other back office users they will have the field disabled
- Developer auth can now login as normal or admin back office user

### DIFF
--- a/app/controllers/admin/amendments_controller.rb
+++ b/app/controllers/admin/amendments_controller.rb
@@ -8,12 +8,12 @@ class Admin::AmendmentsController < Admin::BaseAdminController
 
   def new
     @amendment = @claim.amendments.build
-    @form = Admin::AmendmentForm.new(claim: @claim)
+    @form = Admin::AmendmentForm.new(claim: @claim, admin_user:)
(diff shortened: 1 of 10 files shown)


app/controllers/admin/amendments_controller.rb
```diff
@@ -35,6 +35,8 @@ def ensure_claim_is_amendable
   end
 
   def amendment_params
-    params.require(:amendment).permit(Claim::AMENDABLE_ATTRIBUTES + Policies::AMENDABLE_ELIGIBILITY_ATTRIBUTES + [:notes])
+    params
+      .require(:amendment)
+      .permit(Admin::AmendmentForm.amendable_attributes(claim: @claim, admin_user:))
```
rjlynch, line 40: 👍

rjlynch (approved):
LGTM 👍

PR #3631, opened 2025-03-07
add basic feature flag implementation

asmega:
# Context

- Adds a basic feature flag implementation that is database backed

### DIFF
--- a/app/models/feature_flag.rb
+++ b/app/models/feature_flag.rb
@@ -1,0 +1,5 @@
+class FeatureFlag < ApplicationRecord
+  def self.enabled?(name)
+    where(name: name, enabled: true).exists?
+  end
+end
--- a/db/migrate/20250307152200_create_feature_flags.rb
+++ b/db/migrate/20250307152200_create_feature_flags.rb
@@ -1,0 +1,12 @@
+class CreateFeatureFlags < ActiveRecord::Migration[8.0]
+  def change
+    create_table :feature_flags, id: :uuid do |t|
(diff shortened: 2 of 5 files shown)


db/migrate/20250307152200_create_feature_flags.rb
```diff
@@ -0,0 +1,12 @@
+class CreateFeatureFlags < ActiveRecord::Migration[8.0]
+  def change
+    create_table :feature_flags, id: :uuid do |t|
+      t.text :name, null: false
+      t.boolean :enabled, default: false, null: false
+
+      t.timestamps
+    end
+
+    add_index :feature_flags, :name, unique: true
```
rjlynch, line 10: 👍

app/models/feature_flag.rb
```diff
@@ -0,0 +1,9 @@
+class FeatureFlag < ApplicationRecord
+  def self.enabled?(name)
+    flag = FeatureFlag.find_by(name: name)
+
+    return false if flag.nil?
+
+    flag.enabled?
```
rjlynch, line 7:
```suggestion
  where(name: name, enabled: true).exists?
```
Minor but could save loading an object, not a blocker though!

rjlynch (approved):
Nice 👍 
I think this will be very useful!

app/models/feature_flag.rb
asmega: well tests pass so happy to include this change

PR #3634, opened 2025-03-10
CAPT 2269/provider verify id screen

rjlynch:
WIP
- **Extract postcode validator**
- **Extract NINO validator**
- **Use `have_summary_item` matcher**
- **Write answers to journey session first**
- **Reference answers**
- **Cache slugs**
- **Prefill form attributes from verification**
- **Adds claim identity verification**


spec/features/further_education_payments/provider/provider_verifying_claims_spec.rb
```diff
@@ -2,6 +2,11 @@
 
 RSpec.feature "Provider verifying claims" do
   before do
+    FeatureFlag.create!(
+      name: "fe_provider_identity_verification",
+      enabled: true
+    )
```
asmega, line 8: fyi I added this https://github.com/DFE-Digital/claim-additional-payments-for-teaching/blob/master/spec/support/feature_flags.rb so you can tag your test with
```
scenario "with flag", feature_flag: :fe_provider_identity_verification do
```

app/models/journeys/further_education_payments/provider/session_answers.rb
```diff
@@ -22,6 +47,29 @@ def claim
...
+        attribute :assertions_attributes, pii: false
+        def assertions_attributes
+          verification
+            .fetch("assertions", [])
+            .map(&:with_indifferent_access)
+            .index_by(&:itself)
+        end
+
+        def identity_verification_required?
+          unless FeatureFlag.enabled?(:fe_provider_identity_verification)
```
asmega, line 66: unfortunately not on master yet but somewhere I added `disabled?` to avoid double negations

asmega (approved):

slorek (approved):
Given that we want to eventually migrate the columns in the eligibility tables to a JSON column on `claims`, I think having a mix of JSON and text columns on the eligibility table might cause us a problem in future. I'm interested in what issues in particular did you find when trying to manage the data in a JSON column.

rjlynch:
> Given that we want to eventually migrate the columns in the eligibility tables to a JSON column on `claims`, I think having a mix of JSON and text columns on the eligibility table might cause us a problem in future. I'm interested in what issues in particular did you find when trying to manage the data in a JSON column.

The main issue is that it's different to how the other 

> Given that we want to eventually migrate the columns in the eligibility tables to a JSON column on `claims`, I think having a mix of JSON and text columns on the eligibility table might cause us a problem in future. I'm interested in what issues in particular did you find when trying to manage the data in a JSON column.

The main issue is that it's a different shape to the other data, so when we had to update this journey to write to a session, rather than directly to the eligibility, it was a bit tricky to get the form to populate it's attributes from the session data. I think storing the eligibility data as json on the claim will work well, but before we do that we should migrate the json on the fe eligibility to separate columns, so everything the journeys care about is the same shape. Probably a couple of days work.

PR #3638, opened 2025-03-11
swap out selenium for cuprite

asmega:


spec/support/capybara.rb
```diff
@@ -1,28 +1,19 @@
...
+Capybara.register_driver(:cuprite) do |app|
+  Capybara::Cuprite::Driver.new(
     app,
-    browser: :chrome,
-    options: Selenium::WebDriver::Options.chrome(
-      args: args
-    )
+    window_size: [1200, 800],
+    process_timeout: 20,
+    timeout: 10
```
rjlynch, line 9: Probably useful to keep supporting the `VIEW` setting. I find it's pretty handy for the odd bit of debugging
```suggestion
    timeout: 10,
    headless: ENV.fetch("VIEW", nil).blank?
```

asmega: cuprite supports screenshots but unfortunately does not seem to support non-headless mode

rjlynch (approved):
Nice one 👍

PR #3644, opened 2025-03-13
[CAPT-2279] Split out FE OL task

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2279
- FE OL task is now a separate task from identity confirmation
- This includes a data migration to acheive this
- EY should not be affected by these changes and remain the same
- We now have finer grain control of tasks name in admin area by using locales lookups rather than simply `humanize` the task name
- Finish refactoring of checking claim tasks by extracting to classes
- Rename task names on front end based on if feature flag is toggled
- The whole task thing and names is rather messy. i think this whole area needs some thinking on how we best make tasks a lot more flexible

### DIFF
--- a/app/models/automated_checks/claim_verifiers/one_login_identity.rb
+++ b/app/models/automated_checks/claim_verifiers/one_login_identity.rb
@@ -11,9 +11,10 @@ module AutomatedChecks
 
       def perform
         return unless awaiting_task?(TASK_NAME)
-        return unless claim.identity_confirmed_with_onelogin?
 
-        if claim.one_login_idv_mismatch?
+        if !claim.identity_confirmed_with_onelogin?
+          create_task(passed: false)
+        elsif claim.one_login_idv_mismatch?
           create_task(passed: false)
         else
           create_task(passed: true)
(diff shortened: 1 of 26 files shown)


app/models/automated_checks/claim_verifiers/one_login_identity.rb
```diff
@@ -0,0 +1,49 @@
+module AutomatedChecks
+  module ClaimVerifiers
+    class OneLoginIdentity
+      TASK_NAME = "one_login_identity".freeze
+      private_constant :TASK_NAME
+
+      def initialize(claim:, admin_user: nil)
+        self.admin_user = admin_user
```
slorek, line 8: I think for this task there will never be an `admin_user` so we should probably hardcode that.

app/models/claim_checking_tasks.rb
```diff
@@ -21,41 +21,29 @@ def initialize(claim)
...
+  def locale_key_for_task_name(name)
+    r = applicable_task_objects.find do |object|
+      object.name == name
+    end
+
+    if r.nil?
+      name
     else
-      Task::NAMES.dup.tap do |task_names|
-        task_names.delete("previous_payment")
```
slorek, line 39: 👏

slorek (approved):

PR #3648, opened 2025-03-18
[CAPT-2284] scrub DOB from rollbar

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2284
- Scrub DOB from rollbar


rjlynch (approved):
👍

PR #3653, opened 2025-03-18
CAPT 2280/alternative idv task

rjlynch:
If a claimant fails one login IDV we need to ask the provider to verify
the claimants identity. This commit adds the new task to the admin area
to track this alternative identity verification.

When a claim is considered to be a duplicate we don't automatically send
the verification email to the provider, it needs to be manually
triggered by an ops team member. The designs call for showing the resend
verification email on both the `alternative_identity_verification` and
`provider_verification` task views, both designs for this are the same,
so we've reused the parital from the `provider_verification` that
already handles this.

### DIFF
--- a/app/models/automated_checks/claim_verifiers/alternative_identity_verification.rb
+++ b/app/models/automated_checks/claim_verifiers/alternative_identity_verification.rb
@@ -10,6 +10,7 @@ module AutomatedChecks
       def perform
         return unless eligibility.claimant_identity_verified_at?
         return if task_exists?
+        return unless all_details_match?
 
         create_task!
       end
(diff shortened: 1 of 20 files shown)


app/models/automated_checks/claim_verifiers/alternative_identity_verification.rb
```diff
@@ -0,0 +1,93 @@
...
+      private
+
+      attr_reader :claim
+
+      delegate :eligibility, to: :claim
+
+      def create_task!
+        task = claim.tasks.build(
+          name: TASK_NAME,
+          manual: false,
```
asmega, line 27: was technically a manual test. does setting to `true` have any effect or meaning?

app/models/tasks.rb
```diff
@@ -0,0 +1,59 @@
+module Tasks
```
asmega, line 1: i think in the future we need to push these down even further. ie every task gets its own class (decorator thing?) we can delegate to. at the moment they kind of overlap and not so great to work with

config/locales/en.yml
```diff
@@ -133,6 +133,9 @@ en:
       one_login_identity:
         name: One Login identity check
         title: Confirm the claimant made the claim
+      alternative_identity_verification:
+        name: "Alternative identity verification"
+        title: "Confirm the provider has verified the claimant's identity"
```
asmega, line 138: smartquote

asmega (approved):

app/models/tasks.rb
rjlynch: Yeah I think tasks are pretty tricky to work with. We should probably have a wider discussion around how tasks work. Having incomplete tasks not be a thing and being represented by a string and then only having a record created for them when they're passed / failed isn't ideal, not to mention the whole `claim_verifier_match` stuff

config/locales/en.yml
rjlynch: Will fix this separately, got a few branches based off this one!

PR #3654, opened 2025-03-19
[CAPT-2192] Remove extra indexes

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2192
- Remove extra indexes that don't appear to be needed


rjlynch (approved):
LGTM 👍

PR #3659, opened 2025-03-19
[CAPT-2259] Change EY practitioner journey order

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2259
- For EY practitioner journey, move OL screen to after how we use your info screen


rjlynch (approved):
LGTM 👍

PR #3665, opened 2025-03-20
Update find reference to match designs

rjlynch:
Some tweaks to the find reference page to match designs

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1576,6 +1576,7 @@ en:
     forms:
       find_reference:
         question: Enter your claim reference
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -1576,6 +1576,7 @@ en:
     forms:
       find_reference:
         question: Enter your claim reference
+        hint: Your claim reference is in the claim invite email from your employer. It's a series of letters and numbers, like ABC1234E.
```
asmega, line 1579: smart quote

asmega (approved):

PR #3666, opened 2025-03-21
[CAPT-2304] Fix task tag for OL task

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2304
- Fix task tag because OL task was renamed for FE
- Long term I think we want to move away from these helper that do everything, ideally every task should have its own class that acts a decorator for figuring out the tag for a task


rjlynch (approved):
LGTM 👍

PR #3670, opened 2025-03-24
[CAPT-2302] Decision for high risk claims

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2302
- High risk claims can only be approved or rejected by service admin
- Return codes are only visible to service admins if claim has any return codes
- When undoing a decision this can only be done if admin has `service_admin` role
- Added some back links to some admin pages for better navigation
- Undo decision now uses a govuk form builder and now also backed by a form object, its not perfect but a step in the right direction

### DIFF
--- a/app/controllers/admin/decisions_undo_controller.rb
+++ b/app/controllers/admin/decisions_undo_controller.rb
@@ -19,7 +19,7 @@ class Admin::DecisionsUndoController < Admin::BaseAdminController
       params: amendment_params
     )
 
-    if @form.valid? && @form.save
+    if @form.save
       redirect_to admin_claim_tasks_path(@claim)
     else
       render :new
--- a/app/forms/admin/undo_decision_form.rb
+++ b/app/forms/admin/undo_decision_form.rb
@@ -24,6 +24,8 @@ module Admin
     end
 
     def save
+      return false if invalid?
+
       amendment = Amendment.undo_decision(decision, amendment_params.merge(created_by: current_admin))
       amendment.persisted?
     end


app/models/claim.rb
```diff
@@ -197,7 +197,11 @@ def submittable?
     valid?(:submit) && !submitted? && submittable_email_details? && submittable_mobile_details?
   end
 
-  def approvable?
+  def approvable?(current_admin: nil)
+    if current_admin && high_risk_ol_idv? && !current_admin.is_service_admin?
```
rjlynch, line 201: Fine for now, but I think all this logic (and the rest of the approvable? / rejectable?) logic should be in on the Decision / a DecisionForm.

app/forms/admin/undo_decision_form.rb
```diff
@@ -0,0 +1,53 @@
+module Admin
+  class UndoDecisionForm
```
rjlynch, line 2: 👍 Nice

rjlynch (approved):
LGTM - no of my comments are blockers!

app/controllers/admin/decisions_undo_controller.rb
```diff
@@ -1,17 +1,25 @@
...
-
-    if @amendment.persisted?
+    @form = Admin::UndoDecisionForm.new(
+      claim: @claim,
+      decision: @decision,
+      current_admin:,
+      params: amendment_params
+    )
+
+    if @form.valid? && @form.save
```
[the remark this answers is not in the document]
asmega: theres no reason, i can make that tweak. through the evolution i think i originally `@form.save` was not implemented and I had `@form.valid?` as a guard and it would then directly call `Amendment.undo_decision`

app/models/claim.rb
asmega: I agree that would be better. `Amendment` and those class methods too i think could do with some addressing

PR #3674, opened 2025-03-26
link to alt idv task from OL task

asmega:


app/views/admin/tasks/_task_outcome_one_login_identity.html.erb
```diff
@@ -21,9 +21,11 @@
         <% end %>
       </p>
 
-      <p>
-        You can confirm the provider has responded and verified the claimant’s identity.
-      </p>
+      <% if FeatureFlag.enabled?(:fe_provider_identity_verification) %>
```
rjlynch, line 24: Might be better to use
```
Policies::FurtherEducationPayments.alternative_identity_verification_required?(@claim)
```
As that determines whether we create the alt idv task.
If the claimant doesn't fail one login IDV we don't create the ALT IDV task - probably fine to use the feature flag like you have, as we're only sending the links to those who failed IDV, but on the off chance someone passes IDV we won't create the alt idv task

rjlynch (approved):
LGTM - left a comment but it's probably edge case

PR #3675, opened 2025-03-26
CAPT 2313/ol review app functionality

rjlynch:
Add form for test users to set return codes

In environments where we're bypassing one login we now present a form to
testers where they can enter details that would be returned from one
login, along with setting the desired one login return codes.
Now, rather than failing IDV in test based on the `alternative_idv`
feature flag we fail/pass IDV based on the presence of return codes
entered in the form.

Due to some quirks of how the form builder works, we've had to reuse the
`PersonalDetailsForm` to support this functionality. We subclass the
`PersonalDetailsForm` in the controller as `OneLoginTestUser` to make
bypassing it in tests more straightforward.

I think a slightly nicer approach would be to update the
`test_user_auth_hash` method to return a JWT with the content we want,
so we only have to branch on `OneLoginSignIn.bypass?` in one place. This
however would require moving quite a bit more code (injecting
`OneLogin::DidCache` as an are to the `CoreIdentityValidator`
constructor). Something to consider if we decide to refactor this
controller when we're a bit less pressed for time.

### DIFF
--- a/app/controllers/omniauth_callbacks_controller.rb
+++ b/app/controllers/omniauth_callbacks_controller.rb
@@ -148,18 +148,20 @@ class OmniauthCallbacksController < ApplicationController
     )
   end
 
-  ONE_LOGIN_TEST_USER = {
-    first_name: "TEST",
-    last_name: "USER",
-    date_of_birth: Date.new(1970, 1, 1)
-  }
+  class OneLoginTestUser < PersonalDetailsForm; end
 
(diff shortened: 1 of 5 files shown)


rjlynch:
We don't have to merge this one if it's a bit too hacky, it's only really needed for CAPT-2313 but might be useful for future testing scenarios

asmega (approved):
I think if this works I'm happy. can re-visit another time

PR #3677, opened 2025-03-28
Slack analytics change notification

rjlynch:
Send a notification when analytics file changes

We want to notify the data insights team when there's changes to the dfe
analytics files so they can update big query.

### DIFF
--- a/.github/workflows/notify-schema-change.yml
+++ b/.github/workflows/notify-schema-change.yml
@@ -1,0 +1,60 @@
+name: Notify Analytics Config Changes
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'config/analytics_hidden_pii.yml'
+      - 'config/analytics_blocklist.yml'
+      - 'config/analytics.yml'
+      - 'config/analytics_pii.yml'
+  workflow_dispatch:
+
+jobs:
+  notify-slack:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Fetch remote master for comparison
+        run: git fetch origin master --depth=2
+
+      - name: Get analytics config changes
+        id: analytics-diff
+        run: |
+          echo "ANALYTICS_DIFF<<EOF" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


asmega:
do we want to observe schema changes? or do we want observe the analytics yaml manifests?

rjlynch:
> do we want to observe schema changes? or do we want observe the analytics yaml manifests?

I'm not sure tbh I considered the analytics file first but thought I'd get it working with schema first.
Can observe both though

rjlynch:
> do we want to observe schema changes? or do we want observe the analytics yaml manifests?

Yeah actually I think just checking the analytics files will make the diff a bit more readable

.github/workflows/notify-schema-change.yml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 2
+
+      - name: Get analytics config changes
+        id: analytics-diff
+        run: |
+          echo "ANALYTICS_DIFF<<EOF" >> $GITHUB_ENV
+          git diff HEAD~1 HEAD -- config/analytics_hidden_pii.yml config/analytics_blocklist.yml config/analytics.yml config/analytics_pii.yml >> $GITHUB_ENV
```
asmega, line 27:
```
git diff HEAD~1 HEAD
```
not sure this is going to work? if the branch is over several commits and the change is not in the last one don't think it'll pick up on the change?

rjlynch: Good spot! Have updated it to compare against the state of master before the merge

.github/workflows/notify-schema-change.yml
```diff
@@ -0,0 +1,60 @@
...
+        run: |
+          COMMIT_MSG=$(git log -1 --pretty=%B)
+          COMMIT_AUTHOR=$(git log -1 --pretty=%an)
+          COMMIT_LINK="https://github.com/${{ github.repository }}/commit/${{ github.sha }}"
+          echo "COMMIT_MSG=$COMMIT_MSG" >> $GITHUB_ENV
+          echo "COMMIT_AUTHOR=$COMMIT_AUTHOR" >> $GITHUB_ENV
+          echo "COMMIT_LINK=$COMMIT_LINK" >> $GITHUB_ENV
+
+      - name: Notify Slack
+        uses: rtCamp/action-slack-notify@master
```
asmega, line 44: fine, but next time might be better to use official version https://github.com/marketplace/actions/slack-send-to-slack

PR #3679, opened 2025-03-31
[CAPT-2015] EY consent form is now HTML

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2015
- EY consent form is now HTML based


rjlynch (approved):
Looks good 👍

PR #3680, opened 2025-04-01
[CAPT-2324] EY tweak email form for better a11y

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2324
- EY tweak email form for better a11y


rjlynch (approved):
LGTM 👍

PR #3682, opened 2025-04-01
[CAPT-2092] fix admin accepting cookies before auth

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2092
- fix admin accepting cookies before auth


rjlynch (approved):
Nice 🥳 this gets me so often on review apps!

PR #3686, opened 2025-04-01
[CAPT-1458] Set security headers

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-1458
- Set security headers


rjlynch (approved):
👍

PR #3701, opened 2025-04-07
Fix pending database migrations preventing dfe-analytics initialisation

slorek:
We have noticed that when deploying changes including database migrations, dfe-analytics stops sending ActiveRecord-related events. Web request events still work. This is because dfe-analytics is [rescuing the exception and failing to initialise](https://github.com/DFE-Digital/dfe-analytics/blob/ad9e5a24296aa4bf755d04ceb96ad21f7f8c76d8/lib/dfe/analytics.rb#L134).

We are advised that moving the `prepare-database` step to before `terraform-apply` should resolve this.

### DIFF
--- a/.github/actions/deploy-environment/action.yml
+++ b/.github/actions/deploy-environment/action.yml
@@ -51,6 +51,15 @@ runs:
         project_id: claim-additional-payments
         workload_identity_provider: projects/638192024625/locations/global/workloadIdentityPools/claim-additional-payments-for-te/providers/claim-additional-payments-for-te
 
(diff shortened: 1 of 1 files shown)


rjlynch (approved):
LGTM, should be safe to move this before the terraform apply step.🤞 this does the trick

asmega:
I think this has broken review apps, can't seem to get a review app to deploy now. Also not sure how this would work, how can you run migrations before terraform apply has happened? surely there would be no migrations to apply at that moment in time?

PR #3705, opened 2025-04-08
Remove LUP#subject_symbols

rjlynch:
In practice LUP only ever returns the fixed subject symbols. While removing the `subject_symbols` method creates a bit of noise, as we can't duck type the two policies when checking subject symbols, it will make it easier to create an LUP only journey.


app/forms/journeys/additional_payments_for_teaching/eligible_itt_subject_form.rb
```diff
@@ -31,7 +31,7 @@ def self.qualification_to_substring(qualification_symbol)
       end
 
       def available_subjects
-        @available_subjects ||= subject_symbols.map(&:to_s)
+        @available_subjects ||= subject_symbols.map(&:to_s).sort
```
asmega, line 34: i don't think you need the `sort` on the end as method already sorts?

asmega (approved):

app/forms/journeys/additional_payments_for_teaching/eligible_itt_subject_form.rb
rjlynch: For some reason one of the tests was failing without it 🤷 I couldn't work out why

PR #3715, opened 2025-04-11
[CAPT-2411] Better a11y for address related forms

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2411
- Mark optional address fields with `optional`
- Give example of valid postcode in error message


rjlynch (approved):
LGTM 👍

PR #3718, opened 2025-04-14
capt 2054/lup only journey

rjlynch:
# Implement TRI only journey.
This PR implements a standalone journey for the targeted retention incentive policy.
The relevant forms from the additional payments journey have a been reimplemented to support only the TRI journey.
The relevant views from additional payments have been reimplemented to support TRI only and also migrated to use the govuk form builder and govuk components.

## Review notes
The main files to pay attention to are
Session answers - `app/models/journeys/targeted_retention_incentive_payments/session_answers.rb`
Slug sequence - `app/models/journeys/targeted_retention_incentive_payments/slug_sequence.rb`

The views and forms are all standard and follow the general conventions in the app, can probably just skim these.
The slug sequence is a little different to other journeys in that we conditionally build up the list of slugs rather than conditionally remove slugs, this makes it much easier to reason about, at least to me. The constants in the slug sequence are now redundant but are required for compatibility with the rest of the app.

We can't run both additional payments and TRI journeys simultaneously, as the code expects a policy to only existing in one journey. To deal with this we've added a feature flag for enabling TRI only. When this feature flag is enabled additional payments only contains the ECP policy. We'll be removing additional payments soon so we don't need to support policies working for multiple journeys.

I've extracted the locales for TRI to their own file as 1.) the en.yml file is getting unwieldy, and 2.) it will make it easier to remove the additional payments journey as we won't need to audit as many keys. If we want to keep a single locales file it's not much effort to put the TRI keys back in there.

Most scenarios are covered by feature specs (eg ineligible paths), a bit slower but we get better coverage. Setup is repeated in each of the it blocks as it makes a single spec much easier to read / modify.


app/controllers/claims_controller.rb
```diff
@@ -96,7 +96,9 @@ def redirect_to_existing_claim_journey
   end
 
   def set_backlink_path
-    @backlink_path = claim_path(current_journey_routing_name, previous_slug) if previous_slug.present?
+    if previous_slug.present? && Journeys::PageSequence::DEAD_END_SLUGS.exclude?(current_slug)
```
rjlynch, line 99: We do this as we don't want to show backlinks on the ineligible pages on the LUP only journey.
Backlinks only render if `@backlink_path` has been set.

We should consider moving `DEAD_END_SLUGS` into the `SlugSequence` once everything is moved over to use the navigator.

app/models/journeys/additional_payments_for_teaching.rb
```diff
@@ -30,6 +30,14 @@ module AdditionalPaymentsForTeaching
       }
     }.freeze
 
+    def policies
+      if FeatureFlag.enabled?(:tri_only_journey)
+        [Policies::EarlyCareerPayments]
+      else
+        POLICIES
+      end
+    end
```
rjlynch, line 39: We have to do this as the code doesn't support a Policy (`TargetedRetentionIncentivePayments`) being present in multiple journeys, there's various places where we look up the journey for a given policy. When this feature flag is enabled the additional payments journey only allows claiming for EY. This isn't an issue as we're not accepting AP claims, and in the PRs to removed additional payments we remove this journey and the usage of the feature flag.
We've introduced the feature flag to avoid doing too much in this already large PR!

app/models/journeys/targeted_retention_incentive_payments/slug_sequence.rb
```diff
@@ -0,0 +1,164 @@
...
+
+      delegate :answers, to: :journey_session
+
+      def initialize(journey_session)
+        @journey_session = journey_session
+      end
+
+      def self.start_page_url
+        Rails.application.routes.url_helpers.landing_page_path(
+          "targeted-retention-incentive-payments"
```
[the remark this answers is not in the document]
rjlynch: Yeah Laura's been testing the prototype with that path and it's been sent to policy for testing.

> i'm guessing there will need to be changes outside the app that will now need pointing here?
Good point 👍 will need to sort that before we flip the feature flag, will open a ticket to keep track.

asmega (approved):
nice one! had a look through, skimmed some bits. probably want more eyes for a big change

> I've extracted the locales for TRI to their own file

this is way better. hoping at some point we can get some defaults and some hierarchy structure in place should be able to remove lots of duplication and dead stuff

> The slug sequence is a little different to other journeys in that we conditionally build up the list of slugs rather than conditionally remove slugs

agreed, way easier to comprehend. hoping we can switchover the others at some point to do the same thing.

app/models/journeys/additional_payments_for_teaching.rb
rjlynch: Yeah, TMA (too many acronyms)

PR #3723, opened 2025-04-15
CAPT-1475 - STEP 2 - Rename the string enum column to the original column

kenfodder:
This relies on STEP 1 deployment and instructions being run first: https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/3722

This drops the old column and renames the new columns back to original names.

I've tested this on a [Review App](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/3738) doing the all the steps.

### DIFF
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -15,7 +15,7 @@ class School < ApplicationRecord
   validates :school_type_group, presence: true
   validates :school_type, presence: true
 
-  scope :fe_only, -> { where(phase: 6) } # PhaseOfEducation == 16 plus
+  scope :fe_only, -> { where(phase: "sixteen_plus") }
 
   PHASES = {
     not_applicable: 0,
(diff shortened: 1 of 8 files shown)


app/models/school.rb
```diff
@@ -28,6 +28,8 @@ class School < ApplicationRecord
     all_through: 7
   }.freeze
 
+  PHASE_STRINGS = PHASES.keys.map(&:to_s).index_with(&:itself).freeze
```
rjlynch, line 31: Not a blocker but would probably rename this to `PHASES` (or inline it directly in the enum definition) and move the existing `PHASES` and `self.phase_code_to_enum` to `app/models/school_data_importer.rb` as that seems to be the only place it's used.
Same with the other X_STRINGS constants

rjlynch (approved):
Looks good, but would def run this on a local copy of production first!

PR #3727, opened 2025-04-16
[CAPT-2112] Remove delayed job code

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2112
- Remove delayed job queue code as now using solid queue
- Does not remove database tables, these can be removed once this change is deployed
- Maybe run solid queue for a bit before we merge this one


rjlynch (approved):
LGTM 👍

PR #3742, opened 2025-04-23
Remove a spec helper

rjlynch:
This isn't used anywhere and I'm not sure what it would be used for.


asmega:
given the sea of red i'm guessing it's still used somewhere 😂

rjlynch:
> given the sea of red i'm guessing it's still used somewhere 😂

😂 the module is included in spec helper, forgot to commit that change 😂

asmega (approved):

PR #3746, opened 2025-04-24
[CAPT-2417] Refactor TSLR subjects taught page

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2417

# Notes

- Changes TSLR journey subjects taught form
- This is to aid in migrating TSLR journey to use navigator
- `subjects_taught` is now the public api of the form as this in the input from the HTML form
- `taught_eligible_subjects` and all the `_taught` fields are now part of the `private` api as these are part of backend data structures
- Also now use govuk form builder for this form


rjlynch (approved):
Much nicer 👍

PR #3753, opened 2025-04-28
TSLR uses 2 pages for school search

asmega:
# Context

- This intermediary change should aid with migration of TSLR journey to use navigator
- this uses an adapted version from FE provision search
- in the future hoping to streamline this into something re-usable cross-journey
- this means the we can store state on which page user is on
- therefore back link will work correctly when navigator is used
- previously this used a GET reuqest to the same page therefore state would be stored in the broswer and never server side


spec/features/backlink_spec.rb
```diff
@@ -64,21 +64,33 @@
 
   scenario "Student Loans journey" do
     create(:journey_configuration, :student_loans)
+
     school = create(:school, :student_loans_eligible)
+
```
rjlynch, line 69: 🥳

rjlynch (approved):
Looks good 👍

PR #3760, opened 2025-05-01
Migrate the database before app starts

slorek:
https://dfedigital.atlassian.net/browse/CAPT-1990

Currently the worker and web instances are started before the database migrations have run. This change ensures database migrations run beforehand, accounting for the likelihood of multiple instances starting concurrently.

For review apps we know only one instance of each type is created but we also will need to create the database on the first deployment, so the behaviour is a little different for review apps.


rjlynch (approved):
LGTM - this seems to be the approach taken by a few other govuk services

PR #3767, opened 2025-05-02
CAPT 2055/remove additional payments part 2

rjlynch:
At high level this PR removes the forms and tests associated with the additional payments journey and fixes any specs that fail as a result.


app/helpers/journey_helper.rb
```diff
@@ -0,0 +1,7 @@
+module JourneyHelper
```
rjlynch, line 1: This was originally `AdditionalPaymentsHelper` but the remaining code isn't specific to AP

PR #3768, opened 2025-05-02
CAPT 2055/remove additional payments part 3

rjlynch:
Removes additional payments related classes, such as the session and slug sequence, and EarlyCareerPayments classes used for calculating parts of the journey, such as award amount calculator, school eligibility.
As we're removed award amount calculator we've removed the ability to amend the award amount on ECP claims in the admin area, this should be a non issue as no new ECP claims are being paid.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -16,6 +16,8 @@ Rails.application.routes.draw do
   # /early-career-payments is now /additional-payments - redirect old urls to a gov page
   get "early-career-payments(/*anything)", to: redirect("https://www.gov.uk/government/collections/additional-payments-for-teaching-eligibility-and-payment-details")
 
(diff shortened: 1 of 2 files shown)


spec/factories/policies/early_career_payments/eligibilities.rb
```diff
@@ -158,10 +146,10 @@
...
     end
 
-    trait :ineligible do
-      eligible_now
-      ineligible_school
-    end
+    # trait :ineligible do
+    #   eligible_now
+    #   ineligible_school
+    # end
```
rjlynch, line 152: This commented out trait gets removed in a future pr where we strip out most of the methods in this factory

asmega (approved):
Looks good at a glance. bravo on the huge negative diff!

PR #3789, opened 2025-05-08
CAPT-1474 - step1 ecp eligibility, copy enum columns

kenfodder:
Early Career Payments Eligibilities table, copy `qualification` and `eligible_itt_subject` to a temporary string column.

Best reviewed by commit.

* Two migrations to add columns
* Two manual data migrations to copy the data across

Next [PR](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/3790) will be to remove the old column and rename the string column to the old column name, along with the model handling and expecting string backed enums.


rjlynch (approved):
Looks good 👍

PR #3791, opened 2025-05-09
[CAPT-2418] EY practitioner journey supports navigator

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2418
- EY practitioner journey supports navigator
- After this change all EY journey will be migrated to use navigator

### DIFF
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -3,6 +3,14 @@ class ClaimsController < BasePublicController
 
   skip_before_action :send_unstarted_claimants_to_the_start, only: [:new, :create, :signed_out]
 
+  before_action :create_session_if_skip_landing_page, if: :skip_landing_page?
+
+  def create_session_if_skip_landing_page
+    return unless journey.use_navigator?
+
+    find_journey_session || create_journey_session!
+  end
+
   before_action :initialize_session_slug_history
   before_action :check_page_is_in_sequence, only: [:show, :update]
   before_action :check_page_is_authorised, only: [:show]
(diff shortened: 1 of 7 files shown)


app/controllers/claims_controller.rb
```diff
@@ -3,6 +3,14 @@ class ClaimsController < BasePublicController
 
   skip_before_action :send_unstarted_claimants_to_the_start, only: [:new, :create, :signed_out]
 
+  before_action :create_session_if_skip_landing_page, if: :skip_landing_page?
+
+  def create_session_if_skip_landing_page
+    return unless journey.use_navigator?
+
+    find_journey_session || create_journey_session!
```
rjlynch, line 11:
```suggestion
    journey_session || create_journey_session!
```
not a blocker, but might be better to use `journey_session` as it calls `find_journey_session` and sets the `@journey_session` ivar, so saves a second call to the db if some other callback needs to use the session in this request.

spec/features/early_years_payment/practitioner/email_address_spec.rb
```diff
@@ -26,24 +26,6 @@
     expect(ActionMailer::Base.deliveries.count).to eq 2
   end
 
-  scenario "Resend passcode for verified email" do
```
rjlynch, line 29: How come we're deleting this spec?

rjlynch (approved):
LGTM - just left one question WRT a spec being removed

spec/features/early_years_payment/practitioner/email_address_spec.rb
asmega: i actually remove the page from the journey if the email has been verified, i don't see why one would verify the email again to an email they've already verified? if you hit the back link on next page it now takes them back to email page where they can change their email if they want

PR #3792, opened 2025-05-09
Use new GOV.UK file upload component

slorek:
https://dfedigital.atlassian.net/browse/CAPT-2301

Implements the new [file upload component](https://design-system.service.gov.uk/components/file-upload/#using-the-improved-file-upload-component) which brings drag and drop and accessibility improvements.

Most of the forms in the admin are not using the form builder library so these need to be re-implemented alongside this change.


rjlynch (approved):
Looks good, much nicer wrapping these uploaders in forms 👍

PR #3799, opened 2025-05-13
[CAPT-2446] failing EY OL idv causes a deadend

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2446
- The feature flag used for bypassing OL idv for FE has affected EY journey
- This remedies that


rjlynch (approved):
Looks good 👍

PR #3801, opened 2025-05-13
remove page sequence

rjlynch:
Now all journeys have been migrated to use the navigator we no longer require the `PageSequence` class.

### DIFF
--- a/app/controllers/base_public_controller.rb
+++ b/app/controllers/base_public_controller.rb
@@ -9,7 +9,6 @@ class BasePublicController < ApplicationController
   private
 
   def clear_claim_session
-    session.delete(:slugs)
     clear_journey_sessions!
   end
 
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -9,7 +9,6 @@ class ClaimsController < BasePublicController
     journey_session || create_journey_session!
   end
 
-  before_action :initialize_session_slug_history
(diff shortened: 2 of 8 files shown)


rjlynch:
Note will need to handle the changes to the TRI form from this PR https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/3792

app/controllers/base_public_controller.rb
```diff
@@ -9,7 +9,6 @@ class BasePublicController < ApplicationController
   private
 
   def clear_claim_session
-    session.delete(:slugs)
```
asmega, line 12: 👍

app/models/journeys/early_years_payment/provider/authenticated/slug_sequence.rb
```diff
@@ -26,6 +26,11 @@ class SlugSequence
 
           RESTRICTED_SLUGS = [].freeze
 
+          DEAD_END_SLUGS = %w[
```
asmega, line 29: I'm hoping at some point in the future we can remove the `DEAD_END_SLUGS` constants now that we have a state machine-esque setup and once the positive slugs sequences are in place

rjlynch: We still need DEAD_END_SLUGS to determine when to render a back link or not, though there's probably a better approach for that

PR #3808, opened 2025-05-14
CAPT-1476 - step1 tslr eligibility

kenfodder:
Student Loans Eligibilities table, copy qts_award_year and employment_status to a temporary string column.

Best reviewed by commit.

Two migrations to add columns
Two manual data migrations to copy the data across
Next [PR](https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/3809) will be to remove the old column and rename the string column to the old column name, along with the model handling and expecting string backed enums.

### DIFF
--- a/app/models/policies/student_loans/eligibility.rb
+++ b/app/models/policies/student_loans/eligibility.rb
@@ -98,6 +98,17 @@ module Policies
         super
       end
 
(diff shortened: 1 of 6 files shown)


rjlynch (approved):
LGTM - we should probably have a rails generator to do this!

PR #3813, opened 2025-05-15
[CAPT-2177] Move policy from eligibility to claim

asmega:
# Context

- Pre-requisite for https://dfedigital.atlassian.net/browse/CAPT-2177
- this change is required as concept of eligibilities are to be removed
- as policy is inferred from eligibility
- this moves policy directly onto claim as I feel we will want this as first class citizen on the claim itself


rjlynch (approved):
Looks good. Probably want a follow up migration to make the policy column non nullable.

PR #3823, opened 2025-05-21
CAPT 2474/first year payment task

rjlynch:
Adds previous year check for IRP

A requirement for IRP is to flag to admins if a new claim matches a
claim in a previous year. We match on national insurance number which,
for IRP, is retained for 2 years. Once the 2 year retention period for
NINO is passed we'll be scrubbing this data and so we need a more
permanent association between claims, hence the new join model
`Claims::Match` which associates two claims together.


app/models/claims/match.rb
```diff
@@ -0,0 +1,6 @@
+module Claims
+  class Match < ApplicationRecord
+    belongs_to :source_claim, class_name: "Claim"
+    belongs_to :matching_claim, class_name: "Claim"
+  end
+end
```
rjlynch, line 6: Initially had a scope on here for explicitly finding matches in the previous year
```
scope :previous_year, -> do
  joins(
    <<~SQL
      JOIN claims
        AS source_claims
        ON source_claims.id = claims_matches.source_claim_id
      JOIN claims
        AS matching_claims
        ON matching_claims.id = claims_matches.matching_claim_id
    SQL
  ).where(
    <<~SQL
      SPLIT_PART(source_claims.academic_year, '/', 1) =
      SPLIT_PART(matching_claims.academic_year, '/', 2)
    SQL
  )
end
```
and also had some associations on the claim, `claim_matches`, `previous_year_claims` but I think that was too much surface area for this change.
Also considered having this join model use the IRP eligibility table, rather than the claim, but that would make moving eligibility attributes on to claims trickier.

Perhaps this model should be `Claims::PreviousYearMatch` and have a validation that the matching claim is from the previous AY to the source claim, as that's currently implicit. Thoughts?

rjlynch: I've added a validation that the match is from the previous year to protect the assumptions made in the task view.

rjlynch: Other option if we don't want to introduce a new model for this is store the matching claim ids on the IRP eligibility 
```
  previous_year_claim_ids: ["abcdef"]
```

app/models/claims.rb
```diff
@@ -0,0 +1,5 @@
+module Claims
+  def self.table_name_prefix
+    "claims_"
+  end
```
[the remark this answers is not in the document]
rjlynch: Rails uses it to get the table name for `Claims::Match` - file was added by the rails generator. Could explicitly set the table name in the `Claims::Match` model and lose this but figured best to stick with the defaults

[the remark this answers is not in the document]
asmega: just read the docs and yes you are correct. just would not have thought this would be in the correct place in the hierarchy

asmega (approved):

PR #3824, opened 2025-05-23
[CAPT-2540] student_loan_repayment_amount to award_amount

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2540
- Rename `Policies::StudentLoans::Eligibility#student_loan_repayment_amount` to use `Policies::StudentLoans::Eligibility#award_amount`
- This gets us a step closer to having all eligibilities aligned so they can be duck typed
- Next step after this is to then move `#award_amount` from the eligibility to the claim
- Includes migration to populate the new column. The old column will essentially be orphaned and we can delete at a later date

# Feedback notes

- Not hugely familiar with this side of the app, any possible potential pitfalls I've missed?


rjlynch (approved):
Overall looks good. Left one comment don't think it's a blocker.
I'd recommend getting a copy of prod db and checking payroll totals on this branch vs master to make sure everything works as expected.

PR #3831, opened 2025-05-30
Capt 2478/employment history

rjlynch:
Adds the employment history task.

This task is a bit tricky as it needs to support the ability to add and remove "employments"

https://github.com/user-attachments/assets/03c199bd-3186-4674-8b51-e63427dc71f7

Employments are stored as JSON data structures in the `Polices::InternationalRelocationPayments::Eligibility#employment_history` array.
We considered introducing a new ActiveRecord model to model the Employments, however as these are only used for one task on one policy and that policy is in it's last year, we decided to use a JSON data structure.
The Employment JSON objects are wrapped in an ActiveModel to give them a Rails feel.

We've introduced a new controller `admin/claims/employment_history/employments` to manage creating / destroying employments.

The designs required a "school search". The current "school search" JS assumes that there is a single field form for school searches, and takes over the whole form. The design had a "school search" field along with other form fields so we've needed to implement a new school search JS to support this. The existing school search pages could be updated to use this new field based school search JS.


app/forms/admin/claims/employment_history/destroy_employment_form.rb
```diff
@@ -0,0 +1,39 @@
...
+        def initialize(claim, params: {})
+          @claim = claim
+
+          super(params)
+        end
+
+        def save!
+          eligibility = claim.eligibility
+
+          employment = eligibility.employment_history.find { |e| e.id == employment_id }
```
asmega, line 25: can use not use an activerecord `find` here?

app/forms/admin/claims/employment_history/destroy_employment_form.rb
```diff
@@ -0,0 +1,39 @@
...
+
+          raise ActiveRecord::RecordInvalid unless employment
```
asmega, line 27: activerecord `find` will raise `ActiveRecord::RecordNotFound` which may be sufficient for your needs?

PR #3846, opened 2025-06-06
Removing remaining additional payments journey references

slorek:
https://dfedigital.atlassian.net/browse/CAPT-2503

### DIFF
--- a/public/404.html
+++ b/public/404.html
@@ -41,7 +41,7 @@
           </a>
         </div>
         <div class="govuk-header__content">
-          <a href="/additional-payments/landing-page" class="govuk-header__link govuk-header__service-name"
+          <a href="#" class="govuk-header__link govuk-header__service-name"
             >Claim additional payments for teaching</a
           >
         </div>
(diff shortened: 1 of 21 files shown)


public/404.html
```diff
@@ -41,7 +41,7 @@
           </a>
         </div>
         <div class="govuk-header__content">
-          <a href="/additional-payments/landing-page" class="govuk-header__link govuk-header__service-name"
+          <a href="#" class="govuk-header__link govuk-header__service-name"
```
rjlynch, line 44: Should we remove this a tag? or change it to a `https://www.gov.uk/government/collections/additional-payments-for-teaching-eligibility-and-payment-details`?

terraform/application/config/production_app_env.yml
```diff
@@ -17,7 +17,7 @@ HMRC_API_BANK_VALIDATION_ENABLED: true
 
 ORDNANCE_SURVEY_API_BASE_URL: https://api.os.uk
 
-TID_BASE_URL: https://claim-additional-teaching-payment.service.gov.uk/additional-payments/claim
+TID_BASE_URL: https://claim-additional-teaching-payment.service.gov.uk/targeted-retention-incentive-payments/claim
```
rjlynch, line 20: 👍

rjlynch (approved):
👍 thanks for sorting the remaining AP bits

public/404.html
slorek: I think it needs an anchor tag for the styling, but we can address this in a follow up.

PR #3857, opened 2025-06-12
FE move optional OL sign in to start of journey

asmega:


app/models/journeys/further_education_payments/slug_sequence.rb
```diff
@@ -165,7 +183,10 @@ def slugs
         if poor_performance_form.completed_or_valid? && !answers.subject_to_problematic_actions?
           array << SLUGS_HASH["check-your-answers-part-one"]
           array << SLUGS_HASH["eligible"]
-          array << SLUGS_HASH["sign-in"]
+          if !answers.previously_claimed? && answers.have_one_login_account == "false" ||
+              !answers.previously_claimed? && answers.have_one_login_account == "i_dont_know"
```
rjlynch, line 187: I'd be tempted to add a helper method to the answers
```
def has_one_login_account?
  have_one_login_account == "true"
end
``` 

so this could be `if !answers.previouusly_claimed? || !answers.has_one_login_account?`

app/forms/journeys/further_education_payments/have_one_login_account_form.rb
```diff
@@ -0,0 +1,37 @@
+module Journeys
+  module FurtherEducationPayments
+    class HaveOneLoginAccountForm < Form
+      attribute :have_one_login_account, :string
+
+      validates :have_one_login_account,
+        inclusion: {
+          in: ["true", "false", "i_dont_know"],
```
rjlynch, line 8: Would be tempted to change this to 
```suggestion
          in: ["yes", "no", "i_dont_know"],
```
"true" and "false" as strings seem like they might trip us up in future

rjlynch (approved):
LGTM, left some minor suggestions

PR #3866, opened 2025-06-17
Form builder migration

rjlynch:
Migrates some more forms over to use govuk form builder

### DIFF
--- a/app/views/claims/select_email.html.erb
+++ b/app/views/claims/select_email.html.erb
@@ -1,45 +1,49 @@
-<% content_for(:page_title, page_title(t("forms.select_email.questions.select_email"), journey: current_journey_routing_name, show_error: @form.errors.any?)) %>
+<% content_for(
+  :page_title,
+  page_title(
+    t("forms.select_email.questions.select_email"),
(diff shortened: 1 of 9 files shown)


app/views/claims/select_email.html.erb
```diff
@@ -1,45 +1,50 @@
...
+            link_errors: true
+          ) %>
+
+          <%= f.govuk_radio_divider %>
+
+          <%= f.govuk_radio_button(
+            :email_address_check,
+            false,
+            label: { text: "A different email address" },
+            link_errors: true
```
asmega, line 42: same as above

app/views/claims/select_mobile.html.erb
```diff
@@ -3,62 +3,55 @@
...
+          ) %>
+
+          <%= f.govuk_radio_divider %>
+
+          <%= f.govuk_radio_button(
+            :mobile_check,
+            :declined,
+            label: { text: t("forms.select_mobile_form.answers.decline") },
+            link_errors: true
+          ) %>
```
asmega, line 50: same as above

PR #3881, opened 2025-06-20
[CAPT-2609] Remove submissions controller

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2609
- Remove submissions controller

# Review notes

- Main goal of this PR is to remove the `SubmissionsController`
- This controller has a shared module with `ClaimsController`. This removal allows for the module to be inlined simplifying things
- With `SubmissionsController` gone, the behaviour now lives inside the `CheckYourAnswersForm` for each journey
- That way it is treated like any other `Form#save`, the only quirk is the last page, the `confirmation` page behaves slightly differently. as the session loses account of the `journey session` we bypass any checks that user must have a `journey session` otherwise they will be bounced to start of the journey
- Code duplication has increased in this PR, as each form for each journey duplicates the behaviour to submit a claim
- I'm aware we wanted some kind of delegate submission service, but think this PR is hefty so have left out. This change however should now enable that change to now be very easy.
- The controller has now become greatly simplified and i think now in a much more tenable state


app/controllers/claims_controller.rb
```diff
@@ -89,6 +88,8 @@ def redirect_to_existing_claim_journey
   end
 
   def set_backlink_path
+    return if navigator.current_slug == "confirmation"
```
rjlynch, line 91: Might be one to throw in `DEAD_END_SLUGS` so we keep all the show / don't show back link configuration in one place

app/controllers/claims_controller.rb
```diff
@@ -114,6 +115,8 @@ def check_page_is_authorised
   end
 
   def check_page_is_permissible
+    return if navigator.current_slug == "confirmation"
```
rjlynch, line 118: I think this guard should be in the `navigator.permissible_slug?` method

app/controllers/claims_controller.rb
```diff
@@ -164,4 +167,51 @@ def add_answers_to_rollbar_context
       )
     end
   end
+
+  def check_whether_closed_for_submissions
+    return if session[:submitted_claim_id].present?
```
rjlynch, line 172: Do we need this guard? Seems unlikely that someone would submit a claim right as the journey is closed

PR #3891, opened 2025-06-24
migrate sign in continue form to use form builder

rjlynch:


app/views/claims/_confirmation_form.html.erb
```diff
@@ -1,38 +1,33 @@
...
-  <p class="govuk-body">By selecting yes you are confirming that, to the best of your knowledge, the details above are correct.</p>
+  <%= f.govuk_collection_radio_buttons(
+    :details_check,
+    f.object.radio_options,
+    :id,
+    :name,
+    legend: {
+      text: "Are these details correct?",
+      size: "m",
+    }
```
asmega, line 24: do we need to add `inline: true`? have we moved from inline to non-inline?

slorek (approved):

PR #3902, opened 2025-06-30
[CAPT-2580] DfE sign in supports multiple clients

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2580
- This change enables us to connect to multiple DfE sign in accounts compared to just one before this change

# IMPORTANT

- ~requires new secrets into production vault before merging~ DONE
- ~also toggle review apps bypass off before merging~ DONE

# Changes

- I've moved all the DfE sign in code from `lib` to the `models`, this means everything is auto loaded by rails and no need for manual loading
- `DfeSignIn.configure` can be called multiple times for multiple clients
- `DfeSignIn::User#user_type` has been added to differentiate between admin and provider users
- I've move some code around so hopefully `DfeSignIn::Api::Client` can house more of the client code
- Changed a docker package due to complaining about a vulnerability, couldn't seem to recreate on `master` though

# Notes

- We seem to have some secrets that should not be secrets present in our vault and these seem to be overriding those set by terraform at least for review apps, this can be fixed afterwards. but took some time to diagnose


config/initializers/dfe_sign_in.rb
```diff
@@ -1,5 +1,13 @@
...
 DfeSignIn.configure do |config|
   config.client_id = ENV.fetch("DFE_SIGN_IN_API_CLIENT_ID")
   config.secret = Rails.env.production? ? ENV.fetch("DFE_SIGN_IN_API_SECRET") : ENV.fetch("DFE_SIGN_IN_API_SECRET", "secret")
   config.base_url = ENV.fetch("DFE_SIGN_IN_API_ENDPOINT")
 end
+
+DfeSignIn.configure do |config|
+  config.client_id = ENV.fetch("DFE_SIGN_IN_INTERNAL_CLIENT_ID", "foo")
+  config.secret = ENV.fetch("DFE_SIGN_IN_INTERNAL_API_SECRET", "bar")
+  config.base_url = ENV.fetch("DFE_SIGN_IN_API_ENDPOINT")
```
rjlynch, line 12: Should we provide a fallback here? Might be better to just blow up early if these are not set

app/models/dfe_sign_in/user_data_importer.rb
```diff
@@ -13,9 +15,17 @@ def run
 
     private
 
+    def client_id
+      @client ||= DfeSignIn::User.client_id_for_user_type(user_type)
```
rjlynch, line 19:
```suggestion
      @client_id ||= DfeSignIn::User.client_id_for_user_type(user_type)
```
Not a blocker!

rjlynch (approved):
LGTM 👍 
Neither of  my comments is a blocker!

PR #3914, opened 2025-07-03
[CAPT-2668] New rebranded header and footer

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2668
- Upgrades header and footer to new GOVUK rebranding
- Service name in header is deprecated and now moved to new service nav component
- One Login header remains the same as waiting on OL team to release their rebrand
- Also applied to error pages
- Introduces new `DfeSign::NullUser` to handle when no user logged in with null object rather than nil guards. should mean we are able to simplify some code down the road


app/views/admin/shared/_signed_in_header.html.erb
```diff
@@ -0,0 +1,16 @@
...
+  current_path: request.fullpath
+) %>
+
+<%
+  nav_document = Nokogiri::HTML.fragment(nav_html)
+  sign_out_link = nav_document.css("a").last
+  sign_out_link["data-method"] = "delete"
+%>
+
+<%= nav_document.to_s.html_safe %>
```
asmega, line 16: service nav component does not support additional html attributes or data into link, this is the cleanest approach I could think of to remedy that

PR #3924, opened 2025-07-07
[CAPT-2662] Add provider nav

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2662
- Add provider nav
- Switched to us newly introduced `DfeSignIn::NullUser`
- Added sign out capability
- Added new controller for verified claims, we may want to rename other controller to `UnverifiedClaims`

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,3 +1,5 @@
+require "pathname"
+
 module ApplicationHelper
   def page_title(title, journey: nil, show_error: false)
     [].tap do |a|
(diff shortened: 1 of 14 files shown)


config/routes.rb
```diff
@@ -188,11 +188,12 @@ def matches?(request)
 
   namespace :further_education_payments do
     namespace :providers do
-      resources :sessions, only: %i[new]
+      resource :sessions, only: %i[new destroy]
```
rjlynch, line 191: Should we rename this resource to `session` if it's now singular?

app/views/further_education_payments/provider/_header.html.erb
```diff
@@ -0,0 +1,14 @@
...
+
+<% nav_html = govuk_service_navigation(
+  navigation_items: provider_nav_items,
+  current_path: request.fullpath
+) %>
+
+<%
+  nav_document = Nokogiri::HTML.fragment(nav_html)
+  sign_out_link = nav_document.css("a").last
+  sign_out_link["data-method"] = "delete" if sign_out_link
```
rjlynch, line 11: I think this is probably fine but we may want other consider linking to a page with a sign out button that can post (`_method=delete`) to the sessions#destroy end point to support signing out without JS. Probably one for a separate ticket

app/helpers/application_helper.rb
```diff
@@ -140,4 +140,27 @@ def admin_footer_links
       }
     ]
   end
+
+  def provider_nav_items
```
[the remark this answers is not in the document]
asmega: how do they differ based on role? or are you referring to admins?

config/routes.rb
asmega: done

app/views/further_education_payments/provider/_header.html.erb
asmega: agreed one for next time

PR #3925, opened 2025-07-08
Remove admin back links on root pages

asmega:
# Context

- Remove superfluous back links on admin root pages. they don't really make sense since they are linked to a main navigation item


rjlynch (approved):
👍

PR #3926, opened 2025-07-08
Add performance and discipline screen

rjlynch:
Not much more to say!


spec/features/further_education_payments/providers/provider_verifying_claims_spec.rb
```diff
@@ -63,6 +63,21 @@
 
       click_on "Save and continue"
 
+      # Performance and discipline
```
asmega, line 66: not an issue but i find checking the `h1` makes for debugging issues quicker in future where you may be on a wrong page for whatever reason

PR #3935, opened 2025-07-12
CAPT-2430 - Use internal HE data and not DQT API

kenfodder:


rjlynch (approved):
LGTM - I don't expect the where clause to return many records but if you think it will might be worth replacing the `map` with a `pluck`

PR #3939, opened 2025-07-14
Do not run DfE sign in importer for review apps

asmega:
- this because it is bypassed for review apps
- therefore do not want to import any data as any mocked users present in the review app will be deleted


rjlynch (approved):
LGTM 👍

rjlynch:
Ah sorry clicked close rather than comment 😂 I was just going to add this is a bit quality of life improvement!

PR #3940, opened 2025-07-14
CAPT 2683/dont allow modifying verified claims

rjlynch:
Handle attempted edits of verified claims

If a claim has been verified we don't want to allow verifiers to modify
answers. We need somewhere to redirect them to, as such we've
implemented an initial version of the read only claim view. We've put
this in verification/show for the time being, could also go in
verified_claims/show


app/controllers/further_education_payments/providers/claims/verifications_controller.rb
```diff
@@ -3,7 +3,15 @@ module Providers
     module Claims
       class VerificationsController < BaseController
         before_action :authorise_claim!
-        before_action :set_form
+        before_action :set_form, except: %i[show]
+        before_action :redirect_if_verified, except: %i[show]
```
asmega, line 7: not a blocker but i think filters are easier to comprehend when inverted to `only` as opposed to `except`

db/seeds.rb
```diff
@@ -16,7 +16,7 @@
   Journeys::Configuration.create!(routing_name: Journeys::EarlyYearsPayment::Provider::Authenticated::ROUTING_NAME, current_academic_year: AcademicYear.current)
   Journeys::Configuration.create!(routing_name: Journeys::EarlyYearsPayment::Practitioner::ROUTING_NAME, current_academic_year: AcademicYear.current)
 
-  FeatureFlag.enable!(:tri_only_journey)
+  FeatureFlag.enable!(:provider_dashboard)
```
asmega, line 19: fyi this should already be merged, possible conflict

slorek (approved):

PR #3941, opened 2025-07-14
Remove check required tile

asmega:
# Context

- Remove check required tile from provider unverified claims dashboard


rjlynch (approved):
LGTM 👍

PR #3944, opened 2025-07-15
Provider dashboard QOL updates

asmega:
# Context

- Some QOL updates around FE provider dashboard
- Update deploy comment to include link to provider dashboard (pushed after deployment which is why you don't see it below)
- change underscore URLs to hyphens, see https://insidegovuk.blog.gov.uk/url-standards-for-gov-uk/
- enable `provider_dashboard` feature flag in seeds so review apps always get this feature


rjlynch (approved):
LGTM 👍

PR #3964, opened 2025-07-18
[CAPT-2723] Add readonly view for provider verified claim

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2723
- Adds readonly view for verified claims

### DIFF
--- a/app/controllers/further_education_payments/providers/claims/verifications_controller.rb
+++ b/app/controllers/further_education_payments/providers/claims/verifications_controller.rb
@@ -32,7 +32,7 @@ module FurtherEducationPayments
             elsif wizard.completed?
               flash[:success] = "Claim Verified for #{@form.claimant_name}"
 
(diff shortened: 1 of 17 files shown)


app/controllers/further_education_payments/providers/claims/verifications_controller.rb
```diff
@@ -32,7 +32,7 @@ def update
             elsif wizard.completed?
               flash[:success] = "Claim Verified for #{@form.claimant_name}"
 
-              redirect_to further_education_payments_providers_claims_path
+              redirect_to further_education_payments_providers_verified_claims_path
```
rjlynch, line 35: Do we want to update the `redirect_if_verified` callback to send them to the verified claims show action?

rjlynch (approved):
Looks good 👍

PR #3979, opened 2025-07-28
FE taught_at_least_one_term earlier in sequence

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2751
- In FE claimant journey move page `taught_at_least_one_term` to earlier in the sequence
- Add missing answers to check answers page
- Answer copy has been simplified


rjlynch (approved):
Looks good

PR #4005, opened 2025-08-13
Split personal details

asmega:


app/forms/date_of_birth_form.rb
```diff
@@ -0,0 +1,81 @@
+class DateOfBirthForm < Form
```
rjlynch, line 1: If we merge https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/4009
Most of this could be replaced with

```
include DateOfBirth
self.date_of_birth_field = :date_of_birth
```

app/forms/journeys/early_years_payment/practitioner/check_your_answers_form.rb
```diff
@@ -7,6 +7,10 @@ class CheckYourAnswersForm < Form
         def save
           return false if invalid?
 
+          if journey.requires_student_loan_details?
+            journey::AnswersStudentLoansDetailsUpdater.call(journey_session)
+          end
```
rjlynch, line 12: Not one for this PR but makes sense to do this for all journeys, probably for when we split personal details for the other journeys

rjlynch (commented):
Looks good 👍

PR #4006, opened 2025-08-13
CAPT-2778/ey-provider-alt-idv

rjlynch:
# EY provider alternative IDV journey

If the practitioner's identity can't be verified by One Login we need to get the
provider to do the identity checks (the "alt idv" journey).

At the end of the practitioner journey the CheckYourAnswersForm
(`app/forms/journeys/early_years_payment/practitioner/check_your_answers_form.rb`)
sends and email to the provider if the claimant has failed OL IDV.
The provider clicks the magic link in the email and is taken to the new alt idv
journey. We're currently using the claim reference in the magic link to find the
claim.

The magic link takes the provider to the journey landing page, where we use the
`Journeys::Session` form to set the claim on the session. If the claim can't be
found from the params, or it has already been reviewed the eligibility checker
kicks provider out of the journey.

There's two exits to the alt idv journey
* Claimant works at the nursery - in which case the provider sees all the
  screens
* Claimant does not work at the nursery - in which case we end the journey
  early.

In both cases we set the `alternative_idv_completed_at` time stamp on the
eligibility. We can use this in the admin area to determine if alt idv has been
completed on a claim. Whether or not the alternative idv check is a pass will be
handled in a future ticket but we just need to compare the claimant's answers
with the `alternative_idv_claimant_*` attributes on the eligibility.
We also store the `alternative_idv_completed_at` on the journey session so the
eligibility checker doesn't kick the provider out when they're viewing one of
the post completion pages.

## Order to review pr
The flow through the app is
* `app/forms/journeys/early_years_payment/practitioner/check_your_answers_form.rb`
* `app/mailers/early_years_payments_mailer.rb`
* `app/views/early_years_payment/provider/alternative_idv/landing_page.html.erb`
* `app/models/journeys/early_years_payment/provider/alternative_idv/session_answers.rb`
* `app/models/journeys/early_years_payment/provider/alternative_idv/slug_sequence.rb`
* `app/models/journeys/early_years_payment/provider/alternative_idv/eligibility_checker.rb`
* wizard forms

so it probably makes sense to review the pr in that order.

### DIFF
--- a/app/forms/journeys/early_years_payment/provider/alternative_idv/ineligible_form.rb
+++ b/app/forms/journeys/early_years_payment/provider/alternative_idv/ineligible_form.rb
@@ -10,7 +10,7 @@ module Journeys
           def message
             case params[:ineligible_reason]
             when "claim_not_found"
-              "We can't find this claim"
+              "We can’t find this claim"
             when "alternative_idv_already_completed"
               "An employment check has already been completed for this claim"
             end
--- a/app/views/early_years_payment/provider/alternative_idv/claims/claimant_not_employed_by_nursery.html.erb
+++ b/app/views/early_years_payment/provider/alternative_idv/claims/claimant_not_employed_by_nursery.html.erb
@@ -1,7 +1,7 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-l">
-      You've told us this applicant does not work at <%= @form.nursery_name %>.
+      You’ve told us this applicant does not work at <%= @form.nursery_name %>.
     </h1>
 
     <h2 class="govuk-heading-m">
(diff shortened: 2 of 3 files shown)


app/forms/journeys/early_years_payment/provider/alternative_idv/ineligible_form.rb
```diff
@@ -0,0 +1,22 @@
...
+      module AlternativeIdv
+        class IneligibleForm < Form
+          def completed?
+            false
+          end
+
+          def message
+            case params[:ineligible_reason]
+            when "claim_not_found"
+              "We can't find this claim"
```
asmega, line 13: smartquote 🙃

app/views/early_years_payment/provider/alternative_idv/claims/claimant_not_employed_by_nursery.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l">
+      You've told us this applicant does not work at <%= @form.nursery_name %>.
```
asmega, line 4: smartquote 🙃

asmega (approved):

PR #4014, opened 2025-08-20
CAPT 2801/EY additional confirmation code

rjlynch:
# Adds secondary email verification to alt idv

When the claimant complete their EY journey, we send an email to the
provider to verify the claimant's email address. We don't know when the
provider will access their emails so we can't make the link to start the
alt idv journey expire. In order to ensure that the person verifying the
id still has access to the mail box, upon starting the alt idv journey
we send a one time passcode the provider needs to enter before they can
continue the alt idv journey.


raul-gracia (approved):
👍🏻 LGTM

PR #4029, opened 2025-09-01
fix time sensitive tests

asmega:


spec/features/admin/admin_claim_tasks_update_with_dqt_api_spec.rb
```diff
@@ -729,7 +729,7 @@ def task_outcome
         let(:data) do
           {
             qts_award_date: Date.new( # 1st September is start of academic year
-              policy.first_eligible_qts_award_year(claim.academic_year).start_year,
+              2014,
```
asmega, line 732: hardcoded as i think this is the first year we pass the 11 year mark which means this start to fail. but would be good if someone could double check my reasoning

rjlynch (approved):
LGTM - when we've got a bit of down time we really need to audit the test suite, some of these specs are really hairy. Would be good to use `travel_to` rather than hardcode dates but appreciate that the tests are hard to update!

PR #4030, opened 2025-09-01
[CAPT-2836] EY practitioner landing page

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2836


rjlynch (approved):
LGTM - one for tech debt but I think we can probably remove the skip_landing_page functionality now, as it was introduced due to EY practitioner not having a landing page

asmega:
> LGTM - one for tech debt but I think we can probably remove the skip_landing_page functionality now, as it was introduced due to EY practitioner not having a landing page

Oh I thought it was a much earlier feature. But yes another feature that I'd love to remove. I'd also like the 2 separated EY provider journeys to be combined into a single journey

PR #4033, opened 2025-09-02
[CAPT-2833] EY tweak radio option copy

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2833
- Tweak radio option copy


rjlynch (approved):
👍

PR #4036, opened 2025-09-03
CAPT-2818: Add practitioner reminder emails for Early Years Payments

raul-gracia:
Implement a new email reminder system to send chase emails to practitioners
who haven't completed their Early Years Payment claims after provider submission.

Changes:
- Add new email template ID (cf03a3c7-587a-48c4-83b9-0cd762d103f6) for practitioner reminders
- Create PractitionerReminderEmailJob to send emails at 1, 2, and 4 week intervals
- Add database fields to track reminder email state (count and last sent timestamp)
- Update EarlyYearsPaymentsMailer with new practitioner_claim_reminder method
- Add complete_claim_url to existing practitioner email template (f97480c8-7869-4af6-b50c-413929b8cc88)
- Configure daily job execution at 8am in recurring.yml
- Add analytics tracking for reminder email fields
- Include comprehensive test coverage for all reminder scenarios

The system automatically sends up to 3 reminder emails:
- First reminder: 1 week after provider submission
- Second reminder: 2 weeks after first reminder
- Third reminder: 4 weeks after second reminder


app/models/journeys/early_years_payment.rb
```diff
@@ -0,0 +1,11 @@
+module Journeys
+  module EarlyYearsPayment
+    extend Base
+    extend self
+
+    ROUTING_NAME = "early-years-payment"
+    VIEW_PATH = "early_years_payment"
+    I18N_NAMESPACE = "early_years_payment"
+    POLICIES = [Policies::EarlyYearsPayments]
+  end
```
rjlynch, line 10: Do we need this?
I don't think we're using it?

spec/jobs/early_years_payments/practitioner_reminder_email_job_spec.rb
```diff
@@ -0,0 +1,241 @@
...
+          )
+          claim
+        end
+
+        it "sends first reminder email after 1 week" do
+          expect {
+            perform_job
+          }.to have_enqueued_mail(EarlyYearsPaymentsMailer, :practitioner_claim_reminder)
+            .once
+        end
```
rjlynch, line 40: I'd be tempted to change this to

```suggestion
  perform_enqueued_jobs { perform_job }
  expect(practitioner_email).to have_received_email("notify-template-id", complete_claim_url: "xxx")
```

that way the mailer logic runs so you can test that everything is wired up as expected. If you search the specs for `have_received_email` you'll see a bunch of example uses of this matcher. It's defined in `spec/support/have_received_email_matcher.rb`

rjlynch (changes requested):
I think the email job looks fine, only concern is the `Journeys::EarlyYearsPayment` module as that module doesn't map to a journey

PR #4045, opened 2025-09-04
Fix multiple issues in Early Years practitioner reminder emails

raul-gracia:
This commit addresses several critical issues in the practitioner reminder email system:

1. Historical Data Safeguard
   - Added 6-month cutoff to prevent old claims from receiving emails
   - Only claims where provider submitted within last 6 months are eligible
   - Prevents unexpected emails to practitioners from historical data when deployed

2. Performance Optimization
   - Moved held/rejected status filtering from Ruby to SQL queries
   - Added proper LEFT JOIN with decisions table for rejection checks
   - Reduces database load by filtering at query level

3. Race Condition Prevention
   - Update reminder count BEFORE sending email (not after)
   - Prevents duplicate emails if job fails between sending and updating
   - Ensures data consistency even during partial failures

4. Explicit Reminder Limit
   - Added guard clause to prevent more than 3 reminder attempts
   - Returns empty collection for any reminder number > 3
   - Makes the 3-reminder limit explicit rather than implicit

5. Improved Test Coverage
   - Added tests for historical data safeguard
   - Added tests for maximum reminder limit
   - Added tests verifying update-before-send behavior
   - All tests passing with proper isolation

These changes ensure safer deployment to production environments with existing data
and more reliable email delivery without duplicates or unexpected notifications.


app/jobs/early_years_payments/practitioner_reminder_email_job.rb
```diff
@@ -6,13 +6,23 @@ class PractitionerReminderEmailJob < ApplicationJob
       3 => 4.weeks
     }.freeze
 
+    # Safeguard: Only send reminders for claims submitted by providers within the last 6 months
+    # This prevents historical claims from suddenly receiving emails when deployed to production
+    CUTOFF_DATE = 6.months.ago.freeze
```
rjlynch, line 11: Minor but this will be set when the class loads rather than when the job runs

```suggestion
    CUTOFF_DATE = 6.months
```
Then use `CUTOFF_DATE.ago`

Probably not an issue in practice as we deploy every day

rjlynch (approved):
👍 - left a minor comment, but I don't think it's a blocker

PR #4059, opened 2025-09-09
[CAPT-2743] FE resume existing session

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2743
- Allow users to resume a previous session based on their OL uid
- We ignore any previous sessions that are in ineligible state

# Other changes

- We now purge FE journey sessions only when the session has not been touched for a year
- Introduced concept of expiring journey sessions. this means journey sessions remain persisted but no longer available for lookup
- These journey remain persisted till they are purged at a later date. this allows users to resume a previous journey session we have persisted
- This is done via an hourly job. This means max theoretical session time is now 25 hours (24.hours - 1.second + 1.hour) down from the current time of 48 hours (24.hours - 1.second + 24.hours)

### DIFF
--- a/app/models/one_login_account.rb
+++ b/app/models/one_login_account.rb
@@ -1,0 +1,21 @@
+class OneLoginAccount
+  attr_reader :uid
+
+  def initialize(uid:)
+    @uid = uid
+  end
+
+  def journey_sessions(journey:, academic_year: nil)
+    academic_year ||= journey.configuration.current_academic_year
+
+    journey::Session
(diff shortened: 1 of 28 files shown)


app/models/one_login_account.rb
```diff
@@ -0,0 +1,34 @@
+class OneLoginAccount
```
rjlynch, line 1: Nice, like this abstraction 👍

rjlynch (approved):
Looks good 👍

PR #4064, opened 2025-09-11
CAPT-2793 implement Year 2 enhancements for FE provider verification

raul-gracia:
- Add 4-option teaching qualification field display
  - Show actual contract types and teaching hours values in admin
  - Add conditional fields for variable hours and fixed-term contracts
  - Enhance auto-pass logic to accept multiple teaching qualification values
  - Fix form validation patterns for proper save and exit handling
  - Update admin presenter to display provider responses with formatting


app/models/policies/further_education_payments/admin_provider_verification_task_presenter.rb
```diff
@@ -115,7 +293,89 @@ def claimant_answer(assertion)
...
+      def provider_performance_disciplinary_answer(field)
+        eligibility = claim.eligibility
+        case field
+        when "performance_measures"
+          # Provider field is "subject to performance measures" - show Yes if true, No if false
+          eligibility.provider_verification_performance_measures ? "Yes" : "No"
+        when "disciplinary_action"
+          # Provider field is "subject to disciplinary action" - show Yes if true, No if false
+          eligibility.provider_verification_disciplinary_action ? "Yes" : "No"
+        end
```
rjlynch, line 378: I think we want to use a separate task for year 2 FE claims so we only need to do the branching in one place (the decision on which task to render)

config/analytics_blocklist.yml
```diff
@@ -232,6 +165,73 @@
...
+  - provider_contact_name
+  - onelogin_idv_first_name
+  - onelogin_idv_last_name
+  - onelogin_idv_date_of_birth
+  - onelogin_idv_full_name
+  - date_of_birth_day
+  - date_of_birth_month
+  - date_of_birth_year
+  :journeys_sessions:
+  - answers
```
rjlynch, line 237: Can we lose this reorder from the diff and just add the new line changed?

app/models/automated_checks/claim_verifiers/provider_verification.rb
```diff
@@ -36,9 +36,89 @@ def create_task!
...
+        performance_disciplinary_pass = check_performance_and_disciplinary_responses
+
+        # Check matching responses between claimant and provider
+        matching_responses_pass = check_matching_responses
+
+        # Check teaching qualification acceptance
+        teaching_qualification_pass = check_teaching_qualification
+
+        provider_yes_pass && performance_disciplinary_pass && matching_responses_pass && teaching_qualification_pass
+      end
```
rjlynch, line 64: I don't think this code will be run. The claim verifier isn't being called after the provider verification form is submitted.
On master we've removed this claim verifier (can always add it back), but the claim verifiers run on initial claim submission so would need to explicitly call this verifier after the provider has submitted the verification.

rjlynch (commented):
Have left a couple of comments

PR #4075, opened 2025-09-16
[CAPT-2870] Copy update FE claimant eligibility intro page

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2870


rjlynch (approved):
👍

PR #4112, opened 2025-09-26
[CAPT-2905] Annotate analytics manifest

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2905


rjlynch (approved):
Haven't checked them all but really good.
Thoughts on adding a spec to check all lines have an associated comment?

PR #4126, opened 2025-10-02
[CAPT-2909] Remove unused govuk verify code

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2909
- Try to remove `govuk_verify` code
- there still remains the data and one method call which is wired
- the remainder is removal of dead code


rjlynch (approved):
Nice :thumbnail

PR #4159, opened 2025-10-17
Move slug sequence use into navigator

rjlynch:
Slug sequence is very closely tied to the navigator. Rather than passing
in an instantiated slug sequence and fetching the journey session from
that, pass in the journey session and build the slug sequence in the
navigator.
Also navigator is now responsible for checking for dead end slugs when
determining if there is a previous slug.

### DIFF
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -55,14 +55,10 @@ class ClaimsController < BasePublicController
     @journey_session = nil
   end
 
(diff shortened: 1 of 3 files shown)


rjlynch:
I think this is the better approach but there's some difficulty in how we clear the session upon successful claim submission. We implicitly rely on being able to construct a slug sequence without having a journey session to render the confirmation page. Closing for now

PR #4161, opened 2025-10-17
[CAPT-2929] More govuk form builder

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2929
- This change further utilises govuk formbuilder and removes use of custom code
- There is still more work to remove custom form code but these involve more heavy handed changes so have decided to cut a PR here


rjlynch (approved):
Nice, looks good

PR #4163, opened 2025-10-17
Move slug sequence use into navigator

rjlynch:
Slug sequence is very closely tied to the navigator. Rather than passing
in an instantiated slug sequence and fetching the journey session from
that, pass in the journey session and build the slug sequence in the
navigator.
Also navigator is now responsible for checking for dead end slugs when
determining if there is a previous slug.

Not happy about having to introduce the struct to trick the navigator
into thinking there's a journey session to pull the journey from (the
only bit of info it needs to render the completed page)

### DIFF
--- a/app/models/journeys/navigator.rb
+++ b/app/models/journeys/navigator.rb
@@ -197,7 +197,6 @@ module Journeys
 
     def set_backlink?
       return false if slug_sequence.class::DEAD_END_SLUGS.include?(current_slug)
-      return false if current_slug == "confirmation"
       return false if previous_slug.nil?
       true
     end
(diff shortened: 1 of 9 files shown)


app/models/journeys/navigator.rb
```diff
@@ -195,6 +195,13 @@ def changing_answer?
       params[:change].present?
     end
 
+    def set_backlink?
+      return false if slug_sequence.class::DEAD_END_SLUGS.include?(current_slug)
+      return false if current_slug == "confirmation"
```
asmega: where is `confirmation` used? can it be added to `DEAD_END_SLUGS`?

PR #4172, opened 2025-10-22
CAPT-2930 Remove current AY scope on awaiting payroll filters

kenfodder:
Issue: Currently when a claim is ready for payroll and it's for the previous academic year and we have just started a new academic year the claim isn't visible in admin.

Fix: Relax the current academic year scope on the awaiting payroll listings.


rjlynch (approved):
LGTM

PR #4181, opened 2025-10-28
[CAPT-2929] One fewer autocomplete

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2929
- Replace school search with newer mechanism used by FE
- Old mechanism used a single page that was bridged to the same page via a GET request
- New mechanism now split over 2 pages bridged by a PATCH request. Where first page deals with search and second page deals with selection.
- Should we want to get rid of second page due to JS autocomplete selecting school, this should be a trivial change if needed
- Unfortunately some of the names are still FE specific, however should not affect functionality and can be made generic at a later date
- Delete old schools version JS
- Delete helper from custom form module
- There is still another autocomplete version that exists solely for the admin area. This will need to be tackled if we want to rid of the all the custom form helper code


asmega:
> I think we probably want to switch to using the version used in the admin area for the journeys as it only hijacks a single field, whereas the other versions hijack the whole form, plus it's configurable entirely by data attributes.

I've not taken a look at the admin version, just acknowledge its existence. But yeah if it's better we can make it the sole version that supports everything

PR #4186, opened 2025-10-29
CAPT 2793/fe admin

rjlynch:
Adds auto passing the FE provider verification task.
Adds some more auto failing scenarios.

One thing to highlight, on the admin task view we always want to show "N/A" for certain claimant answers, even if we have their information - this is a business requirement, I've added the reasoning to the "Always show N/A" commit.

### DIFF
--- a/app/models/automated_checks/claim_verifiers/fe_provider_verification_v2.rb
+++ b/app/models/automated_checks/claim_verifiers/fe_provider_verification_v2.rb
@@ -92,10 +92,6 @@ module AutomatedChecks
           @failed_checks << "disciplinary_action"
         end
 
-        if eligibility.provider_verification_contract_type != eligibility.contract_type
-          @failed_checks << "contract_of_employment_mismatch"
-        end
-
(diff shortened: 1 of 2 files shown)


app/models/automated_checks/claim_verifiers/fe_provider_verification_v2.rb
```diff
@@ -53,6 +38,91 @@ def create_task(failed_checks)
...
+        eligibility.provider_verification_performance_measures == false &&
+          eligibility.provider_verification_disciplinary_action == false
+      end
+
+      def provider_indicates_claimant_is_interested_in_becoming_qualified?
+        eligibility.provider_verification_teaching_qualification != "no_not_planned"
+      end
+
+      def failed_checks
+        return @failed_checks if defined?(@failed_checks)
```
asmega: fyi, you can do the following as instance variable will return nil if not set
```ruby
return @failed_checks if @failed_checks
```

asmega (approved):

PR #4188, opened 2025-10-30
[CAPT-2965] Rails 8.1

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2965
- Upgrade rails from 8.0.x to 8.1.x


rjlynch (approved):
Nice, took a while to track down exactly what that preserve timezone initializer was doing but can confirm it's safe to remove.

PR #4197, opened 2025-11-03
[CAPT-2300] Refine NI number format validation

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2300
- Update NI validation rules according to https://www.gov.uk/hmrc-internal-manuals/national-insurance-manual/nim39110
- Does not include temporary or legacy numbers
- Have checked existing production `Claim#national_insurance_number` and all appear to be valid with implemented rules

### DIFF
--- a/app/validators/national_insurance_number_format_validator.rb
+++ b/app/validators/national_insurance_number_format_validator.rb
@@ -2,12 +2,10 @@ class NationalInsuranceNumberFormatValidator < ActiveModel::EachValidator
   NINO_REGEX_FILTER = /\A[A-Z]{2}[0-9]{6}[A-D]{1}\Z/
 
   def validate_each(record, attribute, value)
-    normalised_value = value.gsub(/\s/, "").upcase
-
-    validate_regex(record, attribute, normalised_value)
-    validate_first_character(record, attribute, normalised_value)
-    validate_second_character(record, attribute, normalised_value)
-    validate_prefixes(record, attribute, normalised_value)
+    validate_regex(record, attribute, value)
+    validate_first_character(record, attribute, value)
+    validate_second_character(record, attribute, value)
+    validate_prefixes(record, attribute, value)
   end
 
   private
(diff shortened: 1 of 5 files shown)


app/validators/national_insurance_number_format_validator.rb
```diff
@@ -2,8 +2,50 @@ class NationalInsuranceNumberFormatValidator < ActiveModel::EachValidator
...
+    return if record.errors[attribute].any?
+    return if value.match?(NINO_REGEX_FILTER)
 
     record.errors.add(attribute, options[:message] || :format)
   end
+
+  def validate_first_character(record, attribute, value)
+    return if record.errors[attribute].any?
+
+    invalid_first_charaters = %w[D F I Q U V]
```
rjlynch: I think this is probably cleaner than updating the regex 👍

app/validators/national_insurance_number_format_validator.rb
```diff
@@ -2,8 +2,50 @@ class NationalInsuranceNumberFormatValidator < ActiveModel::EachValidator
   NINO_REGEX_FILTER = /\A[A-Z]{2}[0-9]{6}[A-D]{1}\Z/
 
   def validate_each(record, attribute, value)
-    return if value.gsub(/\s/, "").upcase.match?(NINO_REGEX_FILTER)
+    normalised_value = value.gsub(/\s/, "").upcase
```
rjlynch: Should we do this in the validation? Seems like something the form/model should do before calling the validation

rjlynch (approved):
Looks good, left one comment about trimming spaces but I don't think it's a blocker

app/validators/national_insurance_number_format_validator.rb
asmega: after having a think and looking around, i think the line isn't needed and I was just assuming it would come in badly
```
NINO_REGEX_FILTER = /\A[A-Z]{2}[0-9]{6}[A-D]{1}\Z/
```
The regex should cater for whitespace and casing so I'm thinking I'm just gonna remove highlighted line and use `value`

PR #4202, opened 2025-11-05
[CAPT-2970] Journey config tweaks

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-2970
- `ROUTING_NAME`, `VIEW_PATH` and `I18N_NAMESPACE` are now inferred if constants not provided, otherwise can be specified if present
- Form class names are now directly coupled with their slugs on 1:1 basis
- If another form is needed with the same slug, simply namespace the form
- Therefore we now just need to hardcode an array of forms rather than a hash with the slug
- A possible future enhancement would be to load forms from conventionalised location and no need to hard code the forms and simply load everything from the relevant directory


rjlynch (approved):
Nice, left one comment but it's unrelated to these changed

PR #4204, opened 2025-11-06
CAPT-2966 - Namespace EligibleFeProvider

kenfodder:
This is to move `EligibleFeProvider` to the right namespace.

FileUploads will use the model name as a string in the database, so a migration is needed to include the namespace update.

`EligibleEyProvider` will be addressed in a separate PR.


rjlynch (approved):
Looks good 👍

PR #4221, opened 2025-11-13
CAPT-2995 - migration to fix the target-data-model typo

kenfodder:


rjlynch (approved):
This should do the trick 👍

PR #4224, opened 2025-11-18
Capt 2959 fe previous claim provider verified no 5yr

kenfodder:


app/models/policies/further_education_payments.rb
```diff
@@ -121,6 +121,17 @@ def teaching_start_year_mismatch?(claim)
       end
     end
 
+    def previously_start_year_matches_claim_false?(claim)
```
rjlynch:
```suggestion
    def last_years_claim_rejected_due_to_start_year_mismatch?(claim)
```

rjlynch (approved):
Code looks fine but probably worth adding some comments to the fe_repeat_applicant check as it's pretty tricky to infer the business rules here.
Also do we need to handle claimants who went through ALT IDV?

PR #4236, opened 2025-11-20
move claim access methods onto fe provider

rjlynch:
As part of sending the provider emails we need to recalculate the various
in progress and overdue claim counts. Rather than duplicate that logic in the
mailer and the controller we've moved some methods on to the eligible fe
provider.
The first commit in this pr uses a pretty standard approach of defining a bunch
of regular methods, however these methods are awkward to call as they're not
compose-able and passing in the academic year to each method is noisy. The
second commit switches to using associations which may or may not be overkill.
To avoid polluting the Claim model with policy specific scopes we've used an
extension on the EligibleFeProvider#claims association.
We then update the provider verification controllers to use these new 
associations removing the need for the `claim_scope` controller method.

### DIFF
--- a/app/controllers/further_education_payments/providers/base_controller.rb
+++ b/app/controllers/further_education_payments/providers/base_controller.rb
@@ -39,12 +39,6 @@ module FurtherEducationPayments
       end
       helper_method :current_user
 
-      def current_school
-        return if current_user.current_organisation.ukprn.nil?
-
-        School.find_by(ukprn: current_user.current_organisation.ukprn)
-      end
-
       def current_provider
         @current_provider ||= Policies::FurtherEducationPayments::EligibleFeProvider.find_by(
           ukprn: current_user.current_organisation.ukprn


app/models/policies/further_education_payments/eligible_fe_provider.rb
```diff
@@ -1,13 +1,57 @@
...
+
+        def verified
+          merge(
+            Policies::FurtherEducationPayments::Eligibility
+              .where.not(provider_verification_completed_at: nil)
+          )
+        end
+
+        def verification_overdue
+          where(claims: {created_at: ..2.weeks.ago})
```
asmega: not 100% sure but does this need to be a proc thing? otherwise in 2 weeks time this will be `Time.now`? I don't know if these things are scopes or dynamically evaluated every time

app/controllers/further_education_payments/providers/claims/information_controller.rb
```diff
@@ -3,7 +3,7 @@ module Providers
     module Claims
       class InformationController < BaseController
         def show
-          @claim = claim_scope.find(params[:claim_id])
+          @claim = current_provider.claims.find(params[:claim_id])
```
asmega: we do lose the academic year I believe here so in future may show more claims than desired?

PR #4238, opened 2025-11-21
[CAPT-3009] Single allocation

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-3009
- Able to bulk allocate a single claim
- Also moved interactor classes from `/lib` to `/app` so these classes are autoloaded
- Fixed issue where admins to allocate to would only be present if they had exact roles rather than doing an inclusive query


rjlynch (approved):
LGTM - though I don't understand why allocating 1 random claim would be useful 😂

PR #4244, opened 2025-11-25
[CAPT-1942] OL logout tweaks

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-1942
- Fixes for OL logout token. Token uses a different endpoint to fetch public keys for verification


rjlynch (approved):
Looks good.
I'd be tempted to have this code in `lib` even though it doesn't auto load, it's nice, non domain specific, functionality other apps might want

PR #4262, opened 2025-12-09
[CAPT-3013] Reduce number of http clients

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-3013
- this removes Typhoeus http client gem
- this was used for Ordanance suvey and HMRC api calls and have now been switched over to used Faraday which is defaulting to use `net/http`
- as these are now using faraday it should be relatively trivial to swap to another http client as Faraday act as an abstraction layer and allows for easy switching of drivers/adapters should the need arise
- going forward I recommend always using faraday as an abstraction layer and using default `net/http` unless we need to swap over to something else which can be done on a per api basis or global

### DIFF
--- a/lib/hmrc/bank_account_verification_response.rb
+++ b/lib/hmrc/bank_account_verification_response.rb
@@ -12,7 +12,7 @@ module Hmrc
 
     def code
       if payload.respond_to?(:status)
-        payload.status
+        payload.status.to_i
       else
         payload.code
       end
--- a/lib/ordnance_survey/client/response.rb
+++ b/lib/ordnance_survey/client/response.rb
@@ -12,7 +12,7 @@ module OrdnanceSurvey
     end
 
     def code
-      response.status
+      response.status.to_i
     end
 
     private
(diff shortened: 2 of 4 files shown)


lib/hmrc/bank_account_verification_response.rb
```diff
@@ -12,6 +10,14 @@ def body
       @body ||= JSON.parse(payload.body)
     end
 
+    def code
+      if payload.respond_to?(:status)
+        payload.status
```
rjlynch: Do we need to call `to_i` on this? IIRC code is an int?

lib/ordnance_survey/client/response.rb
```diff
@@ -13,6 +11,10 @@ def body
       @body = JSON.parse(response.body, symbolize_names: true)
     end
 
+    def code
+      response.status
```
rjlynch: do we need `to_i` here too?

rjlynch (approved):
Looks good, left a couple of comments but they're not blockers

PR #4283, opened 2026-01-05
persisting matching claims checks

rjlynch:
Persist claim match data

First step in moving towards persisted claim tasks.

Previously we checked for any matching claims on page load. This worked
OK when viewing individual claims, however when needing to view match
status of multiple claims we had to make a DB trip for each claim.
Additionally as matches were determined on the fly we couldn't report on
duplicates in big query.

We only need to check for claim matches when a claim is created (journey
completion) or when a claim is updated (amendment / pii scrubbing).

To persist matches we've introduced a new model `Claims::Match`. This
model joins two claims and stores the attributes they match on. To find
the matches (and determine the matching attributes) we reuse the
existing `Claim::MatchingAttributeFinder` class.
There should only be one match record per pair of matching claims, we
achieve this by enforcing that the left claim is older than the right
claim and have a unique index on the claim ids.
`Claims::Match.update_matching_claims!` handles sorting the claims,
creating, updating, and removing matches and should be the only
interface used for managing matches, just call this class method when
data on a claim changes.

To avoid matching attribute down time we're using two shim methods
`Claims::Match.matching_claims_shim` and `Claims::Match.matches_shim`
when displaying match information in the UI. If match records for the
claim haven't been generated these shims will generate them before
returning the match data to display. In a future PR we'll add a
background job to generate matches for all claims, and once
`Claims::Match.update_matching_claims!` has been run for all claims we
can remove these shims.

We'd considered using a call back on the claim to update the matching claims,
however we decided against this as when we switch to persisting tasks we're
going to need to regenerate the task list for any new duplicates which seems a
lot of work for a callback.

- [x] Check results against current match approach for a sample of claims
- [x] Write follow up PR with background job to generate matches
- [ ] Remove shim methods

### DIFF
--- a/app/models/amendment.rb
+++ b/app/models/amendment.rb
@@ -48,6 +48,8 @@ class Amendment < ApplicationRecord
       amendment.claim_changes = changes_hash
 
       raise ActiveRecord::Rollback unless amendment.save
+
+      Claims::Match.update_matching_claims!(claim)
     end
 
     amendment
--- a/app/models/amendment.rb
+++ b/app/models/amendment.rb
@@ -49,7 +49,7 @@ class Amendment < ApplicationRecord
 
       raise ActiveRecord::Rollback unless amendment.save
 
-      Claims::Match.update_matching_claims!(claim)
+      AutomatedChecks::ClaimVerifiers::MatchingClaims.new(claim: claim).perform
     end
 
     amendment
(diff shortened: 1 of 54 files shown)


app/models/amendment.rb
```diff
@@ -48,6 +48,8 @@ def self.amend_claim(claim, claim_attributes, amendment_attributes)
       amendment.claim_changes = changes_hash
 
       raise ActiveRecord::Rollback unless amendment.save
+
+      AutomatedChecks::ClaimVerifiers::MatchingClaims.new(claim: claim).perform
```
asmega, line 52: for another time but looks like this method is only used in `ClaimStudentLoanDetailsUpdater` so might be able to inline?

asmega (approved):

app/models/amendment.rb
rjlynch: In the follow up PR this claim verifier takes on some more functionality, it creates the claim match records and creates the matching tasks

PR #4426, opened 2026-03-04
Data retention refactor

rjlynch:
Refactor data retention.

Adds a new declarative approach, where each policy defines it's own
`<PolicyName>::DataRetention::Policy` which documents how long attributes are
retained for.

To start reviewing this I'd recommend looking at
`app/models/policies/targeted_retention_incentive_payments/data_retention/policy.rb`
first to get an idea of what the data retention "dsl" looks like, then review
`app/jobs/policies/data_retention/policies_job.rb`, as that's where the action
kicks off.

PoliciesJob loops through all policies, finds the claims that need scrubbing
according to the policy's data retention policy, then calls passes each of the
claims to the `ApplyPolicyJob` which handles the scrubbing. Then take a look at
`DataRetention::Policy.apply`, this handles returning new scrubbed attributes
for a claim, but doesn't modify the claim.

The scrubbing works as follows
We create a `ChangeSet` from the `DataRetentionPolicy` (what should be deleted
when) and an `EvaluationContext` (determines if the "when" conditions in the
data retention policy have been met).
The `ChangeSet` returns a hash of claim, eligibility, and amendment attributes
with the new scrubbed values. The main methods of interest are
`expired_claim_attributes` and the similar `expired_eligibility_attributes`.
These methods generate an array of attribute names that should be scrubbed.
To determine if a given attribute should be scrubbed, we look the attribute up
in the `DataRetentionPolicy.claim_attributes` hash to get the name of the method
we need to check against the evaluation context for a given claim.
The evaluation context powers the data retention policy's  mini "dsl" that lets
us, and stakeholders, quickly see how long we retain a given attribute.

We'll add in the data retention rules for the other policies in follow up PRs
and wire up the job. We probably will hide the bulk of the claim attributes as
we don't want all policies to have to define things like `Claim#updated_at` is
retained etc.


app/jobs/policies/data_retention/apply_policy_job.rb
```diff
@@ -0,0 +1,24 @@
+module Policies
+  module DataRetention
+    class ApplyPolicyJob < ApplicationJob
+      def perform(claim)
+        change_set = claim.policy::DataRetention::Policy.apply(claim)
+
+        ApplicationRecord.transaction do
+          claim.journey_session&.destroy!
```
asmega: at the moment I don't think we destroy the session? I'm happy to destroy but if you've not already I think it's worth double checking all the data that we need has been copied over to the claim/eligibility models

app/models/policies/data_retention/base_policy.rb
```diff
@@ -0,0 +1,66 @@
...
+              JOIN #{policy::Eligibility.table_name}
+              ON #{policy::Eligibility.table_name}.id = claims.eligibility_id
+              AND claims.eligibility_type = '#{policy::Eligibility}'
+            SQL
+          )
+
+        claim_scope = claim_redacted_attributes.map(&:to_s).map do |attr|
+          # Can't use `Claim.column_defaults` due to typecasting around array
+          # columns
+          default_value = Claim.new.send(attr)
```
asmega, line 42: non-blocker: can change later but can this line be done outside of the map? by that i mean the `new` part

PR #4480, opened 2026-04-02
Fix employment check

rjlynch:
Fix employment check

The employment check wasn't scoping TPS records correctly, it was
checking if the claimant had an employment start date before the end of
the previous financial year OR they had an employment end date after the
end of the previous financial year, not that they were employed during
the previous financial year.

The tests in `spec/requests/admin/admin_tps_data_upload_spec.rb` were
incorrectly passing and have been updated. Previously the specs set up a note
which was `Ineligibile` and had both a "Current school" and two "Claim school"
rows. If there is a "Current school" and a "Claim school" in the notes then the
note should not be "Ineligibile", to create an "Ineligibile" note for TSLR there
should be either only a claim school or a current school. Also the notes in the
specs had a "Current school" row and _two_ "Claim school" rows, as due to the
scoping bug, where any TPS entry with an end date after the previous financial
year was returned, the "Current school" was being counted as one of the "Claim
schools".

### DIFF
--- a/app/models/automated_checks/claim_verifiers/employment.rb
+++ b/app/models/automated_checks/claim_verifiers/employment.rb
@@ -7,11 +7,9 @@ module AutomatedChecks
       def initialize(claim:, admin_user: nil)
         self.admin_user = admin_user
         self.claim = claim
-        self.teachers_pensions_service = TeachersPensionsService.by_teacher_reference_number(claim.eligibility.teacher_reference_number)
       end
 
       def perform
-        return if claim.policy == Policies::StudentLoans # disabled as algorithm has errors and should be redesigned and mapped to current service logic
         return unless required?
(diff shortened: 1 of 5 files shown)


app/models/automated_checks/claim_verifiers/employment.rb
```diff
@@ -68,35 +38,74 @@ def end_of_previous_financial_year
...
       end
 
       def no_match
-        return unless teachers_pensions_service.empty? || !eligible?
+        if claimant_tps_records.empty?
+          # call no_data first
+          raise "Attempting to create a no match when no tps records exist"
+        end
+
+        return unless !eligible?
```
1993jamesgrant: Better as "if eligible?"

1993jamesgrant (approved):

PR #4481, opened 2026-04-02
Remove Student loan policy automatic employment check

1993jamesgrant:
Update remove_approved_student_loan_employments_job.rb

### DIFF
--- a/app/jobs/remove_approved_student_loan_employments_job.rb
+++ b/app/jobs/remove_approved_student_loan_employments_job.rb
@@ -1,0 +1,15 @@
+class RemoveApprovedStudentLoanEmploymentsJob < ApplicationJob
+  def perform
+    claims = Claim
+      .includes(:tasks)
+      .where(tasks: {name: "employment", passed: true})
+      .where("tasks.created_at > ?", Date.new(2026, 3, 1))
+      .by_academic_year(Policies::StudentLoans.current_academic_year)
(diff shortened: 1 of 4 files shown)


app/jobs/remove_approved_student_loan_employments_job.rb
```diff
@@ -0,0 +1,14 @@
+class RemoveApprovedStudentLoanEmploymentsJob < ApplicationJob
+  def perform
+    claims = Claim
+      .includes(:tasks)
+      .where(tasks: {name: "employment", passed: true})
+      .where("tasks.created_at > ?", Date.new(2026, 3, 1))
+      .by_academic_year(Policies::StudentLoans.current_academic_year)
+      .by_policy(Policies::StudentLoans)
```
rjlynch, line 8:
```suggestion
      .includes(:tasks)
      .awaiting_decision
      .where(tasks: {name: "employment", passed: true})
      .where("tasks.created_at > ?", Date.new(2026, 3, 1)) # do we want to retry tasks for all unapproved claims or just those from the last upload?
      .by_academic_year(Policies::StudentLoans.current_academic_year)
      .by_policy(Policies::StudentLoans)
```

I think we'll want the `awaiting_decision` scope so we don't change anything for approved or rejected claims.

rjlynch (approved):
I think this looks good, probably want to scope the job to only undecided claims.

PR #4492, opened 2026-04-15
[CAPT-3558] user_type required for DfE sign in

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-3558
- `user_type` is now required as part of DfE sign in flow


rjlynch (approved):
LGTM 👍

PR #4498, opened 2026-04-20
[CAPT-3567] Fix FE weekly email job

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-3567
- `DISTINCT` previously was not working as expected
- now emails used are based on latest file upload

### DIFF
--- a/app/jobs/further_education_payments/providers/weekly_update_email_job.rb
+++ b/app/jobs/further_education_payments/providers/weekly_update_email_job.rb
@@ -4,17 +4,23 @@ module FurtherEducationPayments
       def perform
         return unless FeatureFlag.enabled?("fe_provider_dashboard")
 
+        relevant_academic_year = Journeys::FurtherEducationPayments
+          .configuration
+          .current_academic_year
+
         providers_with_unverified_claims =
-          Policies::FurtherEducationPayments::EligibleFeProvider.joins(
-            "JOIN schools ON " \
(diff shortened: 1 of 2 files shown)


rjlynch (approved):
LGTM - we should speak with product and see if we actually need to have an audit of file uploads or if it's only really required for payroll. Might be able to simplify things a bit

app/jobs/further_education_payments/providers/weekly_update_email_job.rb
```diff
@@ -4,17 +4,30 @@ class WeeklyUpdateEmailJob < ApplicationJob
...
+              "JOIN schools ON " \
+              "schools.ukprn::integer = eligible_fe_providers.ukprn::integer"
+            ).joins(
+              "JOIN further_education_payments_eligibilities " \
+              "ON further_education_payments_eligibilities.school_id = schools.id"
+            ).merge(
+              Policies::FurtherEducationPayments::Eligibility
+                .awaiting_provider_verification_year_2
+            ).where(
+              file_upload: latest_file_upload
```
kenfodder: Can we just use`.by_academic_year(relevant_academic_year)` ?

There is a scope in `app/models/policies/further_education_payments/eligible_fe_provider.rb` that does the same thing you just did I think?

asmega: done. wasn't aware that existed

PR #4500, opened 2026-04-20
[CAPT-3564] Remove superfluous task controller

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-3564
- Remove duplicated controller
- this controller also had a bug whereby `created_by` was not set


rjlynch (approved):
👍

PR #4501, opened 2026-04-20
[CAPT-3481] Add teacher auth bypass

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-3481
- Add ability to be able to bypass teacher auth. therefore third party integration not required
- By default bypass teacher auth on local dev environment and review apps

# Screenshots

Screenshot 2026-04-22 at 12 26 28

### DIFF
--- a/app/controllers/journeys/early_years_teachers_financial_incentive_payments/auth_controller.rb
+++ b/app/controllers/journeys/early_years_teachers_financial_incentive_payments/auth_controller.rb
@@ -19,24 +19,28 @@ module Journeys
         @omniauth_hash ||= if !TeacherAuth::Config.instance.bypass?
           request.env["omniauth.auth"]
         else
-          form = Debug::TeacherAuth::SignInForm.new(
-            journey_session: nil,
-            journey: nil,
-            params:
-          )
+          omniauth_bypass_hash
+        end
+      end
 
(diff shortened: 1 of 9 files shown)


app/controllers/journeys/early_years_teachers_financial_incentive_payments/auth_controller.rb
```diff
@@ -7,10 +7,36 @@ def callback
...
+        persist_callback_to_session
+
+        redirect_to claim_path(current_journey_routing_name, "trn-found")
+      end
+
       private
 
       def omniauth_hash
-        @omniauth_hash ||= request.env["omniauth.auth"]
+        @omniauth_hash ||= if !TeacherAuth::Config.instance.bypass?
```
1993jamesgrant, line 19: I think this would read better as something like

hash ||= instance_bypass_logic

def instance_bypass_logic
if else
end

rjlynch (approved):
LGTM 👍

app/controllers/journeys/early_years_teachers_financial_incentive_payments/auth_controller.rb
asmega: yep, just extracted a chunk to a private method

PR #4502, opened 2026-04-21
Improve the ability to delete claims and add missing logic into specs

1993jamesgrant:
We need to delete a claim however reviewing the codebase we have some missing test coverage.

I've also added a job to delete the claim. Rather than calling destroy manually.

### DIFF
--- a/app/jobs/destroy_claim_job.rb
+++ b/app/jobs/destroy_claim_job.rb
@@ -1,0 +1,8 @@
+# Destroys a single claim and all its dependent records.
+# Enqueue via: DestroyClaimJob.perform_later(claim.id)
+class DestroyClaimJob < ApplicationJob
+  def perform(claim_id)
+    claim = Claim.find(claim_id)
+    claim.destroy!
+  end
+end
--- a/spec/models/journeys/session_spec.rb
+++ b/spec/models/journeys/session_spec.rb
@@ -2,6 +2,7 @@ require "rails_helper"
 
 RSpec.shared_examples_for "a journey session" do |journey|
   let(:factory_name) { :"#{journey.i18n_namespace}_session" }
+  let(:policy) { journey.policies.first }
 
   describe "validations" do
     describe "journey" do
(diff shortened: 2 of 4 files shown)


app/jobs/destroy_claim_job.rb
```diff
@@ -0,0 +1,8 @@
+# Destroys a single claim and all its dependent records.
+# Enqueue via: DestroyClaimJob.perform_later(claim.id)
+class DestroyClaimJob < ApplicationJob
+  def perform(claim_id)
+    claim = Claim.find(claim_id)
+    claim.destroy!
```
rjlynch, line 6: I think we should do something like

```
fail "Claim not destroyable" if claim.decisions.any?
```

I don't think we want to support deleting claims that have been rejected / approved, otherwise we'll start breaking things in the admin site (eg payments)

app/jobs/destroy_claim_job.rb
```diff
@@ -0,0 +1,8 @@
+# Destroys a single claim and all its dependent records.
+# Enqueue via: DestroyClaimJob.perform_later(claim.id)
+class DestroyClaimJob < ApplicationJob
+  def perform(claim_id)
```
rjlynch, line 4: Might be better to pass in the claim reference here, as the request from ops will be "delete claim reference ABC123X". Saves the person who's running this the step of finding the claim id first!

spec/models/journeys/session_spec.rb
```diff
@@ -60,6 +61,21 @@
...
+    let(:session) { create(factory_name) }
+    let!(:claim) { create(:claim, :submitted, policy: policy, journey_session: session) }
+
+    it "keeps the session and nullifies the claim association" do
+      claim.destroy!
+
+      aggregate_failures do
+        expect(described_class.where(id: session.id)).to exist
+        expect(session.reload.claim).to be_nil
+        expect(session).not_to be_submitted
```
rjlynch, line 75:
```suggestion
        expect(session).to be_destroyed
```
I think we probably want to delete the session with the claim as the session will have a bunch of PII. Note the claim may not have a session so we'll need to do `claim.session&.destroy!`

(we just keep the session around for a while for debugging purposes)

rjlynch (commented):
Looks good, though I don't think we want to support deleting claims that have a decision. If the claim has a decision we should scrub the PII but keep the record around.

app/jobs/destroy_claim_job.rb
```diff
@@ -0,0 +1,8 @@
...
+    claim = Claim.find(claim_id)
```
kenfodder, line 5: If the claim is already gone when the job runs (e.g. double-enqueue, manual retry), Claim.find raises and the job will retry until exhausted. Won't raise a Sentry error either.

PR #4507, opened 2026-04-23
Stagger wait time of payment confirmation emails

1993jamesgrant:
API limit on Notify is 3000 per 60 seconds. This will stagger the emails under the limit

### DIFF
--- a/app/models/payment_confirmation_upload.rb
+++ b/app/models/payment_confirmation_upload.rb
@@ -42,8 +42,8 @@ class PaymentConfirmationUpload
 
       raise ActiveRecord::Rollback if errors.any?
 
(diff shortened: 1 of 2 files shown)


asmega (approved):

kenfodder (approved):
Nitpicks: 

Sub-second delays might/will be rounded to nearest second I think.

Also maybe adding the comment in the code about the rate limit.

1993jamesgrant:
> Nitpicks:
> 
> Sub-second delays might/will be rounded to nearest second I think.

Good point, I believe solid queue only polls every second. So although they won't run in millisecond time the 1 second batches would spread the load

PR #4511, opened 2026-04-24
Update template to generate based on payment tax year automatically

1993jamesgrant:
The template needs updating manually currently, this will general the correct year range based on the payment tax year.

### DIFF
--- a/app/helpers/payment_mailer_helper.rb
+++ b/app/helpers/payment_mailer_helper.rb
@@ -1,4 +1,9 @@
 module PaymentMailerHelper
+  def tax_year_for(date)
+    year = (date.month < 4 || (date.month == 4 && date.day < 6)) ? date.year - 1 : date.year
+    "#{year} to #{year + 1}"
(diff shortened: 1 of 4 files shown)


asmega (approved):

rjlynch (approved):

app/helpers/payment_mailer_helper.rb
```diff
@@ -1,4 +1,9 @@
 module PaymentMailerHelper
+  def tax_year_for(date)
+    year = (date.month < 4 || (date.month == 4 && date.day < 6)) ? date.year - 1 : date.year
+    "#{year} to #{year + 1}"
+  end
+
```
kenfodder, line 6: Maybe add some edge case testing for this, and would we ever pass in `nil` into this method?

I'm not fussed possibly move this as a utility method somewhere, but no need if not used elsewhere.

kenfodder (approved):

PR #4516, opened 2026-04-27
Move notify stagger to common method

1993jamesgrant:
There is likely to be more areas the deliver_later has to be removed in favour of a custom staggered release, as there is already 2 I've moved it to a shared method

### DIFF
--- a/app/mailers/application_mailer.rb
+++ b/app/mailers/application_mailer.rb
@@ -58,4 +58,9 @@ class ApplicationMailer < Mail::Notify::Mailer
     PROVIDER_SIX_MONTH_EMPLOYMENT_REMINDER_TEMPLATE_ID: "bc7faa96-8a19-4765-9d7a-6a6fd02aee9e".freeze,
     PRACTITIONER_CLAIM_REMINDER_TEMPLATE_ID: "cf03a3c7-587a-48c4-83b9-0cd762d103f6".freeze
   }
+
+  def self.deliver_later_with_throttling(mail_delivery, index:, **options)
+    options[:wait] ||= (index / 10.0).seconds
+    mail_delivery.deliver_later(**options)
+  end
 end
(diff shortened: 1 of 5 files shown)


rjlynch (approved):

app/mailers/application_mailer.rb
```diff
@@ -58,4 +58,9 @@ class ApplicationMailer < Mail::Notify::Mailer
     PROVIDER_SIX_MONTH_EMPLOYMENT_REMINDER_TEMPLATE_ID: "bc7faa96-8a19-4765-9d7a-6a6fd02aee9e".freeze,
     PRACTITIONER_CLAIM_REMINDER_TEMPLATE_ID: "cf03a3c7-587a-48c4-83b9-0cd762d103f6".freeze
   }
+
+  def self.deliver_later_with_throttling(mail_delivery, index:, **options)
+    options[:wait] ||= (index / 10.0).seconds
+    mail_delivery.deliver_later(**options)
+  end
```
rjlynch, line 65: I wonder if there's someway to get application mailer to do the throttling for us. With solid queue we can use limits_concurrency

```
limits_concurrency to: 300, key: -> { "emails" }, duration: 1.minute
```

rjlynch (approved):

app/mailers/application_mailer.rb
```diff
@@ -58,4 +58,9 @@ class ApplicationMailer < Mail::Notify::Mailer
     PROVIDER_SIX_MONTH_EMPLOYMENT_REMINDER_TEMPLATE_ID: "bc7faa96-8a19-4765-9d7a-6a6fd02aee9e".freeze,
     PRACTITIONER_CLAIM_REMINDER_TEMPLATE_ID: "cf03a3c7-587a-48c4-83b9-0cd762d103f6".freeze
   }
+
+  def self.deliver_later_with_throttling(mail_delivery, index:, **options)
+    options[:wait] ||= (index / 10.0).seconds
```
kenfodder, line 63: nitpick: if the caller explicitly passes `wait: nil` it will do default throttling. But I guess the caller wouldn't use this method anyway.

kenfodder (approved):

app/mailers/application_mailer.rb
1993jamesgrant: Both work - can discuss it in standup tomorrow

PR #4517, opened 2026-04-27
Move claims to controller load rather than inpage

1993jamesgrant:
This page hit “If you are using more than 4096 parameters, you need to configure a higher limit. the easiest way to do so is by setting the RACK_QUERY_PARSER_PARAMS_LIMIT environment variable to a higher value before loading Rack.”

Moving the claims to load ID's in the controller logic rather than browser passing the ID's


rjlynch (approved):
lgtm

kenfodder (approved):

PR #4521, opened 2026-04-30
[CAPT-3606] EYTFI provider upload

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-3606
- Admins can upload CSV of EYTFI providers
- Added a new mechanism/pattern to do file uploads for admins
- Can now view a file upload to view its state
- I have not done so in this PR but we should store some kind of success/fail state against a file upload
- Can now persist errors of a file upload so issues about an upload can be played back to admin
- This upload process is now async so it runs as a background job rather than blocking a web thread

### DIFF
--- a/app/forms/admin/eligible_eytfi_providers_form.rb
+++ b/app/forms/admin/eligible_eytfi_providers_form.rb
@@ -33,9 +33,9 @@ class Admin::EligibleEytfiProvidersForm
   end
 
   def save
-    file_upload.save
-  end
-
-  def run_import!
+    ApplicationRecord.transaction do
+      file_upload.save
+      EarlyYearsTeachersFinancialIncentivePayments::ImportEligibleEytfiProvidersJob.perform_later(file_upload)
+    end
   end
 end
--- a/app/forms/admin/eligible_eytfi_providers_form.rb
+++ b/app/forms/admin/eligible_eytfi_providers_form.rb
@@ -34,7 +34,7 @@ class Admin::EligibleEytfiProvidersForm
 
   def save
     ApplicationRecord.transaction do
-      file_upload.save
+      file_upload.save!
       EarlyYearsTeachersFinancialIncentivePayments::ImportEligibleEytfiProvidersJob.perform_later(file_upload)
     end
   end
--- a/app/forms/admin/eligible_eytfi_providers_form.rb
+++ b/app/forms/admin/eligible_eytfi_providers_form.rb
@@ -1,0 +1,41 @@
+class Admin::EligibleEytfiProvidersForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  attribute :academic_year, AcademicYear::Type.new
+  attribute :file
+
+  validates :file,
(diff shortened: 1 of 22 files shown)


app/forms/admin/eligible_eytfi_providers_form.rb
```diff
@@ -0,0 +1,41 @@
...
+      uploaded_by: admin_user,
+      body: File.read(file),
+      target_data_model: Policies::EarlyYearsTeachersFinancialIncentivePayments::EligibleEytfiProvider.to_s,
+      academic_year: academic_year.to_s
+    )
+  end
+
+  def save
+    ApplicationRecord.transaction do
+      file_upload.save
```
rjlynch:
```suggestion
      file_upload.save!
```

rjlynch (approved):
Looks good 👍

app/models/policies/early_years_teachers_financial_incentive_payments/eligible_eytfi_provider.rb
```diff
@@ -0,0 +1,11 @@
+module Policies
+  module EarlyYearsTeachersFinancialIncentivePayments
+    class EligibleEytfiProvider < ApplicationRecord
+      belongs_to :file_upload
+
+      scope :by_academic_year, ->(academic_year) {
+        where(file_upload: FileUpload.latest_version_for(EligibleFeProvider, academic_year))
```
rjlynch:
```suggestion
      scope :by_academic_year, ->(academic_year) do
        where(
          file_upload: FileUpload.latest_version_for(
            Policies::EarlyYearsTeachersFinancialIncentivePayments::EligibleEytfiProvider,
            academic_year
          )
        )
      end

```

rjlynch (changes requested):
Just noticed this

app/models/policies/early_years_teachers_financial_incentive_payments/eligible_eytfi_provider.rb
asmega: oh yeah, good spot

app/jobs/early_years_teachers_financial_incentive_payments/import_eligible_eytfi_providers_job.rb
```diff
@@ -0,0 +1,128 @@
...
+      private
+
+      def cast_bool(value)
+        case value
+        when "TRUE"
+          true
+        when "FALSE"
+          false
+        else
+          raise "#{value} could not be cast to a boolean"
```
rjlynch, line 74: This should probably set a validation error on the parser to give feedback to the uploader rather than throw

rjlynch (approved):
Looks good, left a minor comment but not a blocker

app/jobs/early_years_teachers_financial_incentive_payments/import_eligible_eytfi_providers_job.rb
asmega: theres already a validation present for this. I think i wrote this chunk prior to any validation existing. in theory this code should never be hit in normal circumstances

PR #4526, opened 2026-05-05
[CAPT-3583] EYTFI wire qualification confirmation screen

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-3583


rjlynch (approved):
👍

PR #4533, opened 2026-05-11
Capt 3573 eytfi upload employment proof

kenfodder:
### Summary

- **Azure Blob Storage** — wires up Active Storage with the Azure Blob adapter, including the storage account configuration, environment variables, and Active Storage migration
- **File upload journey steps** — adds five new slug steps to the EYTFI journey after `eligible-qualification-confirmed`: upload, review, uploaded, delete, and success. Each step has a form class, view, and is registered in the slug sequence and journey module
- **Upload/review/delete flow** — supports uploading multiple files, reviewing each before confirming, deleting individual files (with a flash notice on deletion), and a skip link once at least one file is confirmed. Answering "no" on review always returns to the upload form
- **Flash message hook** — adds a generic `flash_message` hook to `Form` base and `FormSubmittable` so individual forms can surface notices on redirect, used here by `DeleteEmploymentProofForm`
- **Test coverage** — unit specs for all five new forms, a `:with_employment_proof` factory trait for the EYTFI session, an `upload_employment_proof` feature helper for the happy path, and a `upload_employment_proof_multiple_with_delete` helper exercising the full multi-upload, reject, delete, and re-upload flow

### Testing

- Run EYTFI form specs: `bundle exec rspec spec/forms/journeys/early_years_teachers_financial_incentive_payments/`
- Run EYTFI feature specs: `bundle exec rspec spec/features/early_years_teachers_financial_incentive_payments/`
- Walk the happy path in browser through to `upload-employment-proof-success`
- Verify file deletion shows "File deleted" flash and redirects correctly based on whether files remain


app/forms/journeys/early_years_teachers_financial_incentive_payments/upload_employment_proof_form.rb
```diff
@@ -0,0 +1,32 @@
...
+          content_type: file.content_type,
+          key: "#{journey_session.id}/employment_proof/#{SecureRandom.base58(24)}"
+        )
+
+        true
+      end
+
+      def completed?
+        journey_session.employment_proofs.attached?
+      end
```
rjlynch, line 25: I wonder if we can remove some of the `redirect?` methods and have more of the redirect logic in the slug sequence / `completed?` methods. Eg if this method was 

```suggestion
      def completed?
        journey_session.employment_proofs.attached? && answers.upload_another !== "yes"
      end
```

could we lose the `redirect?` and `redirect_to` from `UploadedEmploymentProofForm`?

app/models/journeys/early_years_teachers_financial_incentive_payments/session.rb
```diff
@@ -1,6 +1,7 @@
 module Journeys
   module EarlyYearsTeachersFinancialIncentivePayments
     class Session < Journeys::Session
+      has_many_attached :employment_proofs
```
rjlynch, line 4: 👍 - wish this could go on the `answers` but can't see how that would work.

rjlynch (approved):
LGTM - left some comments on leaning more on the navigator for redirecting but don't think they're blockers.

(unrelated but having to put has_one_attached on the session rather than the answers makes me think our data model is a bit off!)

PR #4534, opened 2026-05-11
[CAPT-3512] Async EYTFI TRS api call for qualifications

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-3512
- Adds API call to TRS to fetch qualifications
- This is done async so there is a holding page added
- This is bypass-able when setting auth bypass is enabled

### DIFF
--- a/app/controllers/concerns/form_submittable.rb
+++ b/app/controllers/concerns/form_submittable.rb
@@ -105,7 +105,7 @@ module FormSubmittable
       return if @form.blank?
       return if @form.auto_refresh.blank?
 
-      response.headers["Refresh"] = "5"
+      response.headers["Refresh"] = @form.auto_refresh
     end
   end
 end
--- a/app/controllers/concerns/form_submittable.rb
+++ b/app/controllers/concerns/form_submittable.rb
@@ -3,6 +3,7 @@ module FormSubmittable
 
   included do
     before_action :load_form_if_exists, only: [:show, :update, :create]
+    before_action :set_auto_refresh_headers
 
     def new
       redirect_to_next_slug
@@ -99,5 +100,12 @@ module FormSubmittable
       DfeSignIn::NullUser.new
     end
     helper_method :current_user
+
+    def set_auto_refresh_headers
+      return if @form.blank?
+      return if @form.auto_refresh.blank?
+
+      response.headers["Refresh"] = "5"
+    end
   end
 end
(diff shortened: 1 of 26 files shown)


app/controllers/concerns/form_submittable.rb
```diff
@@ -99,5 +100,12 @@ def current_user
       DfeSignIn::NullUser.new
     end
     helper_method :current_user
+
+    def set_auto_refresh_headers
+      return if @form.blank?
+      return if @form.auto_refresh.blank?
+
+      response.headers["Refresh"] = "5"
```
rjlynch:
```suggestion
      response.headers["Refresh"] = @form.auto_refresh
```

app/forms/journeys/early_years_teachers_financial_incentive_payments/qualifications_check_form.rb
```diff
@@ -0,0 +1,17 @@
...
+      def save
+        true
+      end
+
+      def redirect_to_next_slug?
+        journey_session.answers.trs_data_fetched_at.present?
+      end
+
+      def auto_refresh
+        5
```
rjlynch: This is fine, but I think we can get away with lower, probably `1`

app/jobs/early_years_teachers_financial_incentive_payments/fetch_qualifications_job.rb
```diff
@@ -0,0 +1,22 @@
...
+module EarlyYearsTeachersFinancialIncentivePayments
+  class FetchQualificationsJob < ApplicationJob
+    def perform(journey_session)
+      return if journey_session.answers.trs_data_fetched_at.present?
+
+      client = Dqt::Client.new
+      trs_data = client
+        .teacher
+        .find(journey_session.answers.teacher_auth_teacher_reference_number)
+
```
rjlynch, line 12: Do we need to handle failed api calls here?

lib/core_ext/open_struct.rb
```diff
@@ -0,0 +1,15 @@
...
+
+module CoreExt
+  module OpenStruct
+    def as_json(options = nil)
+      if (options || {})[:without_table]
+        @table.as_json(options)
+      else
+        super
+      end
+    end
```
rjlynch, line 11: Seems fine, though may want to consider whacking this overwrite on Dqt::Object

rjlynch (approved):
Looks good. Left some minor comments. Think we probably want to consider doing something if the api call to TRS fails, maybe updating the holding page to say "we're unable to check your qualifications, please try again later" - something for product to decide

lib/core_ext/open_struct.rb
asmega: I tried that and unfortunately doesn't work. the nested objects are openstructs and not `Dqt::Object`

PR #4536, opened 2026-05-12
Add eytfi employment task

rjlynch:
Handling attachments that are pending a virus scan or have failed a virus scan
will be addressed in CAPT-3674. This pr assumes all files displayed have passed
virus checks.


app/models/policies/early_years_teachers_financial_incentive_payments/eligibility.rb
```diff
@@ -4,6 +4,8 @@ class Eligibility < ApplicationRecord
       self.table_name = "early_years_teachers_financial_incentive_payments_eligibilities"
 
       has_one :claim, as: :eligibility, inverse_of: :eligibility
+      belongs_to :eligible_eytfi_provider
```
asmega: not sure we want this? the `eligible_eytfi_provider` table to work with the file upload stuff does not use `upsert` as i think is assumed here but is append only. therefore if someone uploaded a new list the reference would not be correct and all the pointers would need updating. i think a more loose lookup is required i know for example for FE we used `UKPRN`

rjlynch: sigh - yeah you're right, will get that sorted

kenfodder:
@rjlynch where's `eligible_eytfi_provider_urn` set or populated?

UPDATE...

Ah after we submit the claim I guess?

kenfodder (approved):

rjlynch:
> @rjlynch where's `eligible_eytfi_provider_urn` set or populated?
> 
> UPDATE...
> 
> Ah after we submit the claim I guess?

Yeah we'll need to match up the journey with what the admin site expects in the check answers form

PR #4548, opened 2026-05-13
CAPT 3609/admin task overview page

rjlynch:
Update claim summary view for EYTFI

### DIFF
--- a/db/migrate/20260513152936_add_award_and_reference_to_eligible_eytfi_provider.rb
+++ b/db/migrate/20260513152936_add_award_and_reference_to_eligible_eytfi_provider.rb
@@ -1,0 +1,9 @@
+class AddAwardAndReferenceToEligibleEytfiProvider < ActiveRecord::Migration[8.1]
+  def change
+    add_column(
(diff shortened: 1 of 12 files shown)


db/migrate/20260513152936_add_award_and_reference_to_eligible_eytfi_provider.rb
```diff
@@ -0,0 +1,9 @@
+class AddAwardAndReferenceToEligibleEytfiProvider < ActiveRecord::Migration[8.1]
```
kenfodder, line 1: The filename/class mentions `Award` I presume the migration changed and just needs renaming?

kenfodder (approved):
Just that migration naming, otherwise LGTM.

db/migrate/20260513152936_add_award_and_reference_to_eligible_eytfi_provider.rb
rjlynch: Yeah originally was going to add the award but noticed it was already there - don't think any one will look at the migration class name!

PR #4549, opened 2026-05-14
CAPT-3575 - Fetch and persist malware scan result for EYTFI employment proofs

kenfodder:
After a claimant completes the EYTFI journey, enqueue a background job for each confirmed employment proof blob that fetches the Azure Defender for Storage scan result and persists it to two new columns on active_storage_blobs: malware_scan_result (passed/failed/skipped) and malware_scanned_at.

Both columns nil means the scan is still pending. The job retries up to 10 times at 5-minute intervals; if all retries are exhausted the blob is marked as skipped with the current timestamp.

### DIFF
--- a/app/forms/journeys/early_years_teachers_financial_incentive_payments/information_provided_form.rb
+++ b/app/forms/journeys/early_years_teachers_financial_incentive_payments/information_provided_form.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module Journeys
+  module EarlyYearsTeachersFinancialIncentivePayments
+    class InformationProvidedForm < Form
+      def save
+        # TODO: move job trigger to the check your answers / submission form once built
+        blob_ids = journey_session.answers.confirmed_employment_proof_blob_ids
+
+        blob_ids.each do |blob_id|
+          ::EarlyYearsTeachersFinancialIncentivePayments::FetchEmploymentProofMalwareScanResultJob.perform_later(blob_id)
+        end
+
+        true
+      end
+    end
+  end
+end
--- a/app/jobs/early_years_teachers_financial_incentive_payments/fetch_employment_proof_malware_scan_result_job.rb
+++ b/app/jobs/early_years_teachers_financial_incentive_payments/fetch_employment_proof_malware_scan_result_job.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+
+module EarlyYearsTeachersFinancialIncentivePayments
+  class FetchEmploymentProofMalwareScanResultJob < ApplicationJob
+    MAX_ATTEMPTS = 10
+
+    DEFENDER_RESULT_KEY = "Malware Scanning scan result"
+    DEFENDER_TIME_KEY = "Malware Scanning scan time UTC"
+
+    DEFENDER_RESULT_MAPPING = {
+      "No threats found" => ActiveStorage::Blob::MALWARE_SCAN_RESULT_PASSED,
+      "Malicious" => ActiveStorage::Blob::MALWARE_SCAN_RESULT_FAILED
+    }.freeze
+
+    def perform(blob_id)
+      blob = ActiveStorage::Blob.find(blob_id)
+
+      if Rails.env.development?
+        blob.update!(
+          malware_scan_result: ActiveStorage::Blob::MALWARE_SCAN_RESULT_PASSED,
+          malware_scanned_at: Time.zone.now
+        )
+
+        return
+      end
+
+      service = ActiveStorage::Blob.service
+      tags = service.client.get_blob_tags(blob.key)
+      defender_result = tags[DEFENDER_RESULT_KEY]
+      scan_time = tags[DEFENDER_TIME_KEY]
+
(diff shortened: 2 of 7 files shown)


app/jobs/early_years_teachers_financial_incentive_payments/fetch_employment_proof_malware_scan_result_job.rb
```diff
@@ -0,0 +1,48 @@
...
+      defender_result = tags[DEFENDER_RESULT_TAG]
+
+      raise PendingScanError, "Malware scan pending for blob #{blob_id}" if defender_result.blank?
+
+      scan_time = tags[DEFENDER_TIME_TAG].then { |t| Time.zone.parse(t) if t.present? }
+
+      blob.update!(
+        malware_scan_result: RESULT_MAPPING.fetch(defender_result, defender_result),
+        malware_scanned_at: scan_time
+      )
```
rjlynch: Might be better to re-enqueue the job ourselves rather than throw as it's easier to follow what's going on rather than having jumps.

```suggestion
      service = ActiveStorage::Blob.service
      tags = service.client.get_blob_tags(blob.key)
      defender_result = tags[DEFENDER_RESULT_TAG]
      scan_time = tags[DEFENDER_TIME_TAG]
      
      blob.malware_scan_attemps += 1
      
      if defender_result.present?
        blob.update!(
          malware_scan_result: defender_result,
          malware_scanned_at: scan_time
        )
      elsif blob.malware_scan_attempts > 10
        blob.update!(
          malware_scan_result: skipped,
          malware_scanned_at: scan_time
        )
      else
        blob.save!
        self.class.set(wait: 1.minute).perform_later(blob.id)
      end
```

also I wonder if the job should just write the results from defender directly to the blob and then have the mapping from result to domain meaning on the blob? eg

```
def passed?
  malware_scan_result == "No threats found"
end

def scanned_at
  Time.zone.parse(malware_scanned_at)
end
```

This gives us more flexibility if defender changes their responses, rather than using `RESULT_MAPPING.fetch(defender_result, defender_result)`.

Haven't looked at this too closely so may be missing some reason we can't do this

app/forms/journeys/early_years_teachers_financial_incentive_payments/information_provided_form.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Journeys
+  module EarlyYearsTeachersFinancialIncentivePayments
+    class InformationProvidedForm < Form
+      def save
+        # TODO: move job trigger to the check your answers / submission form once built
```
rjlynch, line 7: Yeah in check answers is better otherwise we'll need to updated the job to check the blob still exists, as the user may remove the upload

rjlynch (approved):
LGTM - not a blocker but I think we could make the job a bit more linear

app/jobs/early_years_teachers_financial_incentive_payments/fetch_employment_proof_malware_scan_result_job.rb
kenfodder: @rjlynch 

Retrying:
- Good shout, I adopted the linear approach, that's updated to do that
- Stores the attempt count
- Enqueue with a wait as you suggested

Results from defender:
- No, we don't want to store the defender result strings in the DB, that would be a mess if they changed them.
- However you make a good point that made me think...
- I've added a value called `defender_result_unrecognised` so if we see a new value, we flag that
- Also I've added a Sentry call to shoot off the defender result so we can see what was returned so we be notified and update it.
- Then we can easily see which blobs have `defender_result_unrecognised` and retry those after updating the code to handle new defender values (if it ever happens)

Let me know what you think now.

app/jobs/early_years_teachers_financial_incentive_payments/fetch_employment_proof_malware_scan_result_job.rb
```diff
@@ -0,0 +1,58 @@
...
+
+        blob.update!(
+          malware_scan_result: mapped_result,
+          malware_scanned_at: scan_time
+        )
+      elsif blob.malware_scan_attempts >= MAX_ATTEMPTS
+        blob.update!(
+          malware_scan_result: ActiveStorage::Blob::MALWARE_SCAN_RESULT_SKIPPED,
+          malware_scanned_at: Time.zone.now
+        )
```
rjlynch, line 55:
```suggestion
        )
        Sentry.capture_message(
          "Malware scan attempts exceeded for blob #{blob.id}",
          level: :error
        )
```

probably worth alerting if we can't scan a blob after 10 goes, likely indicates an issue we need to look at

rjlynch (approved):
Nice 👍 

Let one comment, again not a blocker

app/jobs/early_years_teachers_financial_incentive_payments/fetch_employment_proof_malware_scan_result_job.rb
kenfodder: Done

PR #4553, opened 2026-05-15
Provider email doesn't include rejected count - CAPT-3698

1993jamesgrant:
The provider email count included the rejected emails

### DIFF
--- a/app/mailers/further_education_payments_mailer.rb
+++ b/app/mailers/further_education_payments_mailer.rb
@@ -33,10 +33,10 @@ class FurtherEducationPaymentsMailer < ApplicationMailer
       Journeys::FurtherEducationPayments.configuration.current_academic_year
     )
 
-    number_overdue = claims.unverified.verification_overdue.count
-    number_in_progress = claims.unverified.verification_in_progress.count
-    number_not_started = claims.unverified.verification_not_started.count
-    number_overall = claims.unverified.count
+    number_overdue = claims.unverified.not_rejected.verification_overdue.count
+    number_in_progress = claims.unverified.not_rejected.verification_in_progress.count
+    number_not_started = claims.unverified.not_rejected.verification_not_started.count
+    number_overall = claims.unverified.not_rejected.count
 
     return if number_overall.zero?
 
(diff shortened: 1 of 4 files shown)


rjlynch (approved):
looks good 👍

app/mailers/further_education_payments_mailer.rb
```diff
@@ -33,10 +33,10 @@ def provider_weekly_update_email
...
     )
 
-    number_overdue = claims.unverified.verification_overdue.count
-    number_in_progress = claims.unverified.verification_in_progress.count
-    number_not_started = claims.unverified.verification_not_started.count
-    number_overall = claims.unverified.count
+    number_overdue = claims.unverified.not_rejected.verification_overdue.count
+    number_in_progress = claims.unverified.not_rejected.verification_in_progress.count
+    number_not_started = claims.unverified.not_rejected.verification_not_started.count
+    number_overall = claims.unverified.not_rejected.count
```
rjlynch: Not a blocker but I wonder if putting these in `FurtherEducation::Providers::Claims::Stats` and updating the controller to use that too would prevent these getting out of sync

kenfodder (approved):

app/mailers/further_education_payments_mailer.rb
1993jamesgrant: Good idea, I've moved this around, want to double check before thumbing it up?

spec/models/further_education_payments/providers/claims/stats_spec.rb
```diff
@@ -284,6 +284,88 @@
...
+    it "counts not started unverified, not rejected claims" do
+      expect(subject.unverified_not_started_count).to eq(2)
+    end
+
+    it "counts all unverified, not rejected claims" do
+      expect(subject.unverified_overall_count).to eq(3)
+    end
+  end
+
+  describe "weekly email counts" do
```
kenfodder, line 287: Duplicate `describe "weekly email counts" `

kenfodder (changes requested):

kenfodder:
Question will...

A claim with a past `provider_verification_deadline` but `nil` `provider_verification_started_at` might appear in both `unverified_overdue_count` and `unverified_not_started_count`? 🤔

1993jamesgrant:
> A claim with a past `provider_verification_deadline` but `nil` `provider_verification_started_at` might appear in both `unverified_overdue_count` and `unverified_not_started_count`

I think it's okay - as a claim can be both overdue and not started?

kenfodder (approved):

PR #4556, opened 2026-05-18
CAPT-3571 - Restrict EYTFI employment proof uploads to PDF, JPG, PNG and HEIC under 20 MB

kenfodder:
Adds server-side content type and file size validations to the upload form, with an accept attribute on the file input as a browser-level hint. Covers both validations with unit and feature specs.

### DIFF
--- a/app/views/early_years_teachers_financial_incentive_payments/claims/upload_employment_proof.html.erb
+++ b/app/views/early_years_teachers_financial_incentive_payments/claims/upload_employment_proof.html.erb
@@ -29,7 +29,8 @@
 
       <p class="govuk-body">You upload a clear photo or a scanned copy of the document.</p>
 
-      <%= f.govuk_file_field :file, label: { hidden: true } %>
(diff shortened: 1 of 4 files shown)


app/views/early_years_teachers_financial_incentive_payments/claims/upload_employment_proof.html.erb
```diff
@@ -29,7 +29,8 @@
 
       <p class="govuk-body">You upload a clear photo or a scanned copy of the document.</p>
 
-      <%= f.govuk_file_field :file, label: { hidden: true } %>
+      <%= f.govuk_file_field :file, label: { hidden: true },
+            accept: "application/pdf,image/jpeg,image/png,image/heic,image/heif" %>
```
rjlynch:
```suggestion
            accept: f.object.class::ALLOWED_CONTENT_TYPES.join(",") %>
```

rjlynch (approved):
LGTM

PR #4560, opened 2026-05-19
Add qualifications task

rjlynch:
This task will always be passed as only claimants holding valid
qualifications will be allowed to complete the journey.

### DIFF
--- a/app/models/policies/early_years_teachers_financial_incentive_payments/admin_tasks_presenter.rb
+++ b/app/models/policies/early_years_teachers_financial_incentive_payments/admin_tasks_presenter.rb
@@ -14,6 +14,15 @@ module Policies
           ["Student loan plan", claim.student_loan_plan&.humanize]
         ]
       end
+
+      def qualifications
+        [
+          [
+            "Qualification",
+            "Qualification verified as #{claim.eligibility.ey_qualification} on #{l(claim.eligibility.trs_data_fetched_at.to_date)}"
+          ]
+        ]
+      end
     end
   end
 end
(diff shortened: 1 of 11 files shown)


app/models/policies/early_years_teachers_financial_incentive_payments/eligibility.rb
```diff
@@ -17,6 +17,24 @@ def eligible_eytfi_provider
...
           .find_by!(urn: eligible_eytfi_provider_urn)
       end
+
+      def ey_qualification
+        if dqt_teacher.has_valid_qts?
+          "Qualified Teacher Status"
+        elsif dqt_teacher.has_valid_eyts?
+          "Early Years Teacher Status"
+        elsif dqt_teacher.has_valid_eyps?
+          "Early Years Provider Status"
```
kenfodder: Typo? `Professional` not `Provider`?

app/models/policies/early_years_teachers_financial_incentive_payments/admin_tasks_presenter.rb
```diff
@@ -14,6 +14,15 @@ def student_loan_plan
           ["Student loan plan", claim.student_loan_plan&.humanize]
         ]
       end
+
+      def qualifications
+        [
+          [
+            "Qualification",
+            "Qualification verified as #{claim.eligibility.ey_qualification} on #{l(claim.eligibility.trs_data_fetched_at.to_date)}"
```
kenfodder, line 22: Do we need a nil guard/presence on this, is it possible to have `nil` `trs_data_fetched_at` and it get a nil error?

kenfodder (approved):
1 typo, approved anyway for when that is fixed.

app/models/policies/early_years_teachers_financial_incentive_payments/admin_tasks_presenter.rb
rjlynch: I don't think so - we'll always be setting this as all claimants will have to have had TRS data fetched so we can check their quals before they can submit a claim. I think if this is nil we want to throw an error so we investigate what's gone wrong

PR #4562, opened 2026-05-21
[CAPT-3599] EYTFI claim submission

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-3599


app/forms/journeys/early_years_teachers_financial_incentive_payments/check_your_answers_form.rb
```diff
@@ -0,0 +1,122 @@
...
+          "your knowledge"
+        }
+
+      def save
+        return false if invalid?
+
+        journey_session.answers.assign_attributes(claimant_declaration:)
+        journey_session.save!
+
+        journey::AnswersStudentLoansDetailsUpdater.call(journey_session)
```
rjlynch, line 23: 👍

app/models/policies/early_years_teachers_financial_incentive_payments.rb
```diff
@@ -32,5 +32,13 @@ def auto_check_student_loan_plan_task?
     def decision_deadline_in_weeks
       10.weeks
     end
+
+    def duplicate_claim?(claim)
+      Claim::MatchingAttributeFinder.new(claim).matching_claims.exists?
```
rjlynch: Do we want to use matching attribute finder here or do we want to use the OLID like?

```
Eligibility.where(teacher_auth_one_login_uid: claim.eligibility.teacher_auth_one_login_uid).where.not(id: claim.eligibility_id)
```

Matching attribute finder will already flag the claims in the admin site (think it checks other policies too).

app/forms/journeys/early_years_teachers_financial_incentive_payments/check_your_answers_form.rb
```diff
@@ -0,0 +1,122 @@
...
+          claim.save!
+          mark_service_access_code_as_used!
+          Event.create(claim:, name: "claim_submitted")
+        end
+
+        claim.policy.mailer.submitted(claim).deliver_later
+        ClaimVerifierJob.perform_later(claim)
+
+        if Policies::EarlyYearsTeachersFinancialIncentivePayments.duplicate_claim?(claim)
+          claim.eligibility.update!(flagged_as_duplicate: true)
```
rjlynch: Do we need to do this?
We're not sending the claims to a provider to verify and if we want to stop dupes probably best to do it when they return from T auth 🤷

db/migrate/20260519114611_fields_for_eytfi_eligiblities.rb
```diff
@@ -0,0 +1,39 @@
...
+      :teacher_auth_verified_date_of_birth, :date
+
+    add_column :early_years_teachers_financial_incentive_payments_eligibilities,
+      :teacher_auth_one_login_uid, :text
+
+    add_column :early_years_teachers_financial_incentive_payments_eligibilities,
+      :teacher_auth_completed_at, :datetime
+
+    add_column :early_years_teachers_financial_incentive_payments_eligibilities,
+      :nursery_id, :text
```
rjlynch, line 22: Not a blocker but I don't know if we want to set this. The eligibility uses `eligible_eytfi_provider_urn` so I don't think nursery_id is set. Might be better to update the check answers form to not set this and instead set `eligible_eytfi_provider_urn` 🤷 - can always tweak this after it's merged.

rjlynch (approved):
Looks good, left some comments but we can sort them after merging if needs be

db/migrate/20260519114611_fields_for_eytfi_eligiblities.rb
asmega: ~this is probably due to merge/rebase issue. will double check~

asmega: @rjlynch i don't see `eligible_eytfi_provider_urn` being set any where though? how does it ever get persisted?

app/models/policies/early_years_teachers_financial_incentive_payments.rb
asmega: I've now removed

app/forms/journeys/early_years_teachers_financial_incentive_payments/check_your_answers_form.rb
asmega: I've now removed

db/migrate/20260519114611_fields_for_eytfi_eligiblities.rb
asmega: I've now set these on check answers form

app/forms/journeys/early_years_teachers_financial_incentive_payments/check_your_answers_form.rb
```diff
@@ -0,0 +1,131 @@
...
+
+      def set_eligibility_attributes(eligibility)
+        answers.attributes.each do |name, value|
+          if eligibility.respond_to?(:"#{name}=")
+            eligibility.public_send(:"#{name}=", value)
+          end
+        end
+
+        eligibility.award_amount = Policies::EarlyYearsTeachersFinancialIncentivePayments.award_amount
+        eligibility.eligible_eytfi_provider_urn = answers.nursery.urn
```
rjlynch, line 100: 👍

PR #4565, opened 2026-05-21
CAPT 3658/claims per setting

rjlynch:
Add claim limit check task

If there are more than <some number> of claims at a provider we want to
add a task to any undecided claims for that provider for the ops team to
review. The limit will be set as a column in the eytfi provider csv
that's uploaded to the admin site.

### DIFF
--- a/app/jobs/early_years_teachers_financial_incentive_payments/import_eligible_eytfi_providers_job.rb
+++ b/app/jobs/early_years_teachers_financial_incentive_payments/import_eligible_eytfi_providers_job.rb
@@ -41,7 +41,7 @@ module EarlyYearsTeachersFinancialIncentivePayments
 
       validates "Max claims",
         presence: true,
-        numericality: {only_integer: true, greater_than: 0}
+        numericality: {only_integer: true, greater_than_or_equal_to: 0}
 
       def initialize(row:)
         @row = row
--- a/app/jobs/early_years_teachers_financial_incentive_payments/import_eligible_eytfi_providers_job.rb
+++ b/app/jobs/early_years_teachers_financial_incentive_payments/import_eligible_eytfi_providers_job.rb
@@ -40,8 +40,8 @@ module EarlyYearsTeachersFinancialIncentivePayments
         inclusion: {in: %w[TRUE FALSE], message: "must be TRUE or FALSE"}
 
       validates "Max claims",
-        presence: true,
-        numericality: {only_integer: true, greater_than_or_equal_to: 0}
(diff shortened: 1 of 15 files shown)


app/models/policies/early_years_teachers_financial_incentive_payments/claim_checking_tasks.rb
```diff
@@ -4,6 +4,7 @@ class ClaimCheckingTasks < Policies::ClaimCheckingTasks
       def applicable_task_names
         tasks = []
 
+        tasks << "provider_claim_count" if add_provider_claim_count_task? || task_exists?("provider_claim_count")
```
asmega, line 7: should it not be the other way around for `task_exists?`. ie when the task does not exist? and we should add it? or maybe i've misunderstood the task?

app/jobs/early_years_teachers_financial_incentive_payments/import_eligible_eytfi_providers_job.rb
```diff
@@ -38,6 +39,10 @@ class RowParser
         presence: true,
         inclusion: {in: %w[TRUE FALSE], message: "must be TRUE or FALSE"}
 
+      validates "Max claims",
+        presence: true,
+        numericality: {only_integer: true, greater_than: 0}
```
asmega: should this be `greater_than_or_equal_to`? as there will be ineligible providers in the list and we probably want to set those to zero even though there is another column to signify they are ineligible.

app/models/policies/early_years_teachers_financial_incentive_payments/claim_checking_tasks.rb
rjlynch: I've put the `task_exists?` check in as we always want to show completed tasks on the claim to handle the following scenario:
* Some nursery has a claim limit of 5
* Six claims are completed so we show the task
* The task is passed / failed
* The claim limit is updated to 10

Without the `task_exists?` when re-viewing the claim we would no longer show the task as the number of claims is now below the limit.

Once we switch to always permitting tasks, rather than only permitting on completion we can remove the `task_exists?` check.

app/jobs/early_years_teachers_financial_incentive_payments/import_eligible_eytfi_providers_job.rb
rjlynch: Sorted 👍  - have also set the column to default to `5` so whoever is building the spreadsheet doesn't have to set a value in every row, but can overwrite the default value if they need to.

asmega (approved):

PR #4572, opened 2026-05-26
[CAPT-3745] EYTFI double submission is ineligible

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-3745


rjlynch (approved):
Looks good

PR #4573, opened 2026-05-26
CAPT 3747/rename journey strings

rjlynch:
Rename EYTTFI -> EYTRP

I've left the TRS callback routes as
"early-years-teachers-financial-incentive-payments" so we don't have to get them
re-whitelisted. I've also left the bypass routes as
"early-years-teachers-financial-incentive-payments" for consistency. When
private beta closes we should look to rename these.


config/routes.rb
```diff
@@ -98,7 +98,9 @@ def matches?(request)
       end
 
       get "auth/teacher/callback", to: "journeys/early_years_teachers_financial_incentive_payments/auth#callback"
+    end
```
rjlynch, line 101: @asmega - I've renamed the route to `early-years-teachers-recognition-payments`, however I've left the existing `early-years-teachers-financial-incentive-payments/auth/teacher/callback` route as is to avoid any issues with TRS callback white listing. Do you for see any other issues from TRS due to renaming the journey?

asmega: resolved offline

asmega (approved):

PR #4577, opened 2026-05-26
Purge Azure blobs when unsubmitted journey sessions are deleted

kenfodder:
`PurgeUnsubmittedClaimsJob` used `destroy_all` which bypasses ActiveRecord callbacks, leaving Active Storage blobs orphaned in Azure. Now enqueues `ActiveStorage::PurgeJob` for all attachments before destroying the session records.

### DIFF
--- a/app/jobs/purge_unsubmitted_claims_job.rb
+++ b/app/jobs/purge_unsubmitted_claims_job.rb
@@ -4,8 +4,10 @@
 class PurgeUnsubmittedClaimsJob < ApplicationJob
   def perform
     Journeys::JOURNEYS.each do |journey|
-      Rails.logger.info "Purging #{journey::Session.purgeable.count} old and unsubmitted #{journey::Session} journeys from the database"
-      journey::Session.where(journey: journey.routing_name).purgeable.destroy_all
+      purgeable = journey::Session.where(journey: journey.routing_name).purgeable
+      Rails.logger.info "Purging #{purgeable.count} old and unsubmitted #{journey::Session} journeys from the database"
+
+      purgeable.in_batches(&:destroy_all)
     end
   end
 end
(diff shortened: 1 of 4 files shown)


asmega:
> `destroy_all` which bypasses ActiveRecord callbacks

if you do not want to fire callbacks you'll want to use `delete_all`. i think firing callbacks is desired here though

rjlynch (commented):
Do we need to do anything here?
`destroy_all` does `each(&:destroy)` under the hood, so should clean up the attachments.

I think we should keep the spec, which passes against `app/jobs/purge_unsubmitted_claims_job.rb` on master

app/jobs/purge_unsubmitted_claims_job.rb
```diff
@@ -4,8 +4,18 @@
 class PurgeUnsubmittedClaimsJob < ApplicationJob
   def perform
     Journeys::JOURNEYS.each do |journey|
-      Rails.logger.info "Purging #{journey::Session.purgeable.count} old and unsubmitted #{journey::Session} journeys from the database"
-      journey::Session.where(journey: journey.routing_name).purgeable.destroy_all
+      purgeable = journey::Session.where(journey: journey.routing_name).purgeable
+      Rails.logger.info "Purging #{purgeable.count} old and unsubmitted #{journey::Session} journeys from the database"
+
+      purgeable.in_batches do |batch|
+        ids = batch.pluck(:id)
```
asmega:
```suggestion
        ids = batch.select(:id)
```
otherwise @rjlynch will complain about extra query

kenfodder: @asmega @rjlynch - Yeah I spotted that and tried that already doesn't work...

`batch.select(:id)` produces a subquery rather than a fixed list — it carries all the conditions from the `purgeable` scope, including `updated_at: ..24.hours.ago.` When `purge_later` calls `record.touch` internally, it updates the session's `updated_at` to now. The subquery is then re-evaluated when `destroy_all` runs, the `updated_at` condition no longer matches, and nothing gets deleted.

`pluck(:id)` materialises the IDs into a Ruby array at that moment, so the subsequent queries use a fixed list that isn't affected by `record.touch`.

Hence also keeping it in batches, so the array isn't too big.

Also the spec in this PR would also fail proving that.

asmega (approved):

rjlynch (approved):
LGTM 👍

PR #4587, opened 2026-05-28
CAPT-3776 - Fix image preview size outside development

kenfodder:
The inline style attribute on the employment proof image preview was blocked by the CSP (config/initializers/content_security_policy.rb) in non-development environments. Development permits :unsafe_inline but other environments do not, causing large images to render at full native resolution. Moved the styles to a stylesheet class instead.

```ruby
# config/initializers/content_security_policy.rb
# ...
if Rails.env.development?
  policy.style_src :self, :unsafe_inline
else
  policy.style_src :self
end
# ...
```

## TESTING

- Use `Busy Bees Preschool` if you want to test quickly.


kenfodder:
CI is having issues with the build, closing to open a new PR.

PR #4593, opened 2026-05-29
Fix claim cancelled

rjlynch:
This was a tricky one.
When the claimant chooses "No" on the continue claim screen we need to
clear their session and show them the claim cancelled page. This is
tricky as once their session is cleared the claims controller kicks the
claimant back to the landing page before the navigator can determine the
slug to use. To work round this we've added a new route to the static
pages controller and moved the EYTRP static pages to come before the
journey route definitions, that way requests for `claim-cancelled` will
be picked up by the statis pages controller, which doesn't check for
active sessions.
This seemed to be the lightest touch approach as we keep the
claim-cancelled slug in the slug sequence and the navigator remains
responsible for picking the next slug. Though it feels like a bit of a
hack.

Now that we need a page that sits outside the claim controller's
control, along with forms specifying their own redirects makes me think
we need to think a bit more about we handle redirects in the app.
Ideally we'd specify all redirection rules in the slug sequence, and the
navigator would look only at that to decide what to do.

### DIFF
--- a/app/controllers/static_pages_controller.rb
+++ b/app/controllers/static_pages_controller.rb
@@ -60,6 +60,20 @@ class StaticPagesController < BasePublicController
     end
   end
 
+  def claim_cancelled
+    claim_cancelled_page = "#{journey.view_path}/claims/claim_cancelled"
+
+    if lookup_context.template_exists?(claim_cancelled_page, [], false)
+      render claim_cancelled_page
+    else
+      render(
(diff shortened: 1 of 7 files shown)


asmega:
Yeah this is kind of going beyond the limits of what it was intended or its capabilities. I think I would plumb in an after hook that gets fired after the request/reponse has been served that clears the session as the cleanest approach

asmega (approved):

PR #4595, opened 2026-06-01
CAPT-3771 - Add EYTRP check eligibility form

kenfodder:
- Two checkboxes
- Both required to be checked to be eligible

## Testing

- Use `Busy Bees Preschool` nursery

### DIFF
--- a/app/models/policies/early_years_teachers_financial_incentive_payments/policy_eligibility_checker.rb
+++ b/app/models/policies/early_years_teachers_financial_incentive_payments/policy_eligibility_checker.rb
@@ -18,6 +18,12 @@ module Policies
           :ineligible_provider
         elsif answers.teaching_qualification_confirmation == false
           :teaching_qualification_not_confirmed
+        elsif answers.fifty_percent_time_as_eyt == false && answers.not_subject_to_performance_and_disciplinary == false
+          :check_eligibility_both_not_confirmed
+        elsif answers.fifty_percent_time_as_eyt == false
+          :fifty_percent_time_as_eyt_not_confirmed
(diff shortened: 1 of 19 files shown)


app/models/policies/early_years_teachers_financial_incentive_payments/policy_eligibility_checker.rb
```diff
@@ -18,6 +18,8 @@ def ineligibility_reason
           :ineligible_provider
         elsif answers.teaching_qualification_confirmation == false
           :teaching_qualification_not_confirmed
+        elsif answers.fifty_percent_time_as_eyt == false || answers.not_subject_to_performance_and_disciplinary == false
+          :teaching_qualification_not_confirmed
```
rjlynch: May want to consider splitting this out in to 3 separate reasons
* `fifty_percent_time_as_eyt == false && not_subject_to_performance_and_disciplinary == false`
* `fifty_percent_time_as_eyt == false`
* `not_subject_to_performance_and_disciplinary == false`

that way the ineligible reason for each scenario will be reportable by analytics. May be redundant and they'll look at the sessions table, but not sure how they're reporting on ineligible reasons.

rjlynch (approved):
LGTM - left one comment about splitting out the ineligible reasons

app/models/policies/early_years_teachers_financial_incentive_payments/policy_eligibility_checker.rb
kenfodder: @rjlynch interesting - so 3 separate failure pages?

rjlynch (approved):

PR #4596, opened 2026-06-01
Add rejection / approval constants

rjlynch:
Not sure what values these should be set to but easy enough to change.
Need these in place to unblock approving / rejecting claims.


asmega (approved):

asmega:
maybe set to 100 so everything goes thru QA process?

rjlynch:
> maybe set to 100 so everything goes thru QA process?

Will check with Claire what she's wants this setting to

PR #4597, opened 2026-06-01
[CAPT-3735] EYTFI in app feedback survey

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-3735
- Adds a new basic journey to collect feedback
- Some parts are hardcoded but with some minor changes should not be difficult to use for other journeys as well

### DIFF
--- a/app/forms/feedbacks/details_form.rb
+++ b/app/forms/feedbacks/details_form.rb
@@ -1,0 +1,135 @@
+module Feedbacks
+  class DetailsForm < Form
+    attribute :rating, :string
+    attribute :area, :string
+    attribute :specific_page, :string
+    attribute :comment, :string
+    attribute :research_participation, :boolean
+    attribute :email_address, :string
+    attribute :occupation, :string
+    attribute :origin, :string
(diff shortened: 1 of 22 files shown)


app/forms/feedbacks/details_form.rb
```diff
@@ -0,0 +1,135 @@
+module Feedbacks
+  class DetailsForm < Form
```
rjlynch: I'd be tempted not to subclass `Form` here and just pull in the `ActiveModel` dependencies directly, that way we don't have to create a throw away journey session in the controller, and keeps all the feedback stuff decoupled from journey specific weirdness as much as possible.

rjlynch (approved):
Overall looks good, but I'd be tempted to remove the dependency on a journey session existing as we don't really need it.

app/forms/feedbacks/details_form.rb
asmega: done

rjlynch (approved):
Nice 👍

PR #4604, opened 2026-06-02
clean update task status helper

rjlynch:
Refactor admin helper

This helper was doing a lot of work.
I've moved returning the status for a claim into claim checking tasks.
Mostly it's a copy paste of what the helper was doing and could be
cleaned up further but moving it into an object is enough for now.

What colour the ui should be seems like a decent responsibility for a
helper so have left that there.

### DIFF
--- a/app/models/policies/claim_checking_tasks.rb
+++ b/app/models/policies/claim_checking_tasks.rb
@@ -21,6 +21,24 @@ module Policies
         .select(&:not_passed?)
     end
 
+    def identity_status
+      task = claim.tasks.detect { |t| t.name == "identity_confirmation" }
+
+      if task.nil?
+        "Unverified"
+      elsif task.passed?
+        "Passed"
+      elsif task.passed == false
+        "Failed"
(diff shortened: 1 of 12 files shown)


app/models/policies/claim_checking_tasks.rb
```diff
@@ -21,6 +21,24 @@ def blocking_approval
...
+        "Failed"
+      elsif task.claim_verifier_match_all?
+        "Full match"
+      elsif task.claim_verifier_match_any?
+        "Partial match"
+      elsif task.claim_verifier_match_none?
+        "No match"
+      end
+    end
+
```
kenfodder, line 41: @rjlynch Does EYTFI/EYTRP need to override this? It won't have a `identity_confirmation` right?

rjlynch: Good spot. EYTFI overrides this already but I've left off EYTRP. Pushing that now!

kenfodder (approved):

PR #4611, opened 2026-06-03
[CAPT-3748][CAPT-3749] Remove EYTFI back links

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-3748
- https://dfedigital.atlassian.net/browse/CAPT-3749


kenfodder (approved):
I'm ok approving this, I'm sensing it smells with a partial having a state and affects the pages behaviour outside the partial.

asmega:
> I'm ok approving this, I'm sensing it smells with a partial having a state and affects the pages behaviour outside the partial.

yea kind of acts as a global variable for the request. probably bending the rules a bit

PR #4618, opened 2026-06-04
[CAPT-3800] Handle teacher auth failure

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-3800


rjlynch (approved):
LGTM

Thought it should really read

```
Sorry, there is a problem with their service
```
🙈

PR #4619, opened 2026-06-04
CAPT-3831 - Add claim_mail_to helper with screen reader context for EYTFI email links

kenfodder:
Introduces a claim_mail_to helper that wraps govuk_mail_to with visually-hidden spans so screen readers announce context (e.g. "email ... if you need support") when tabbing to an email link. Replaces all govuk_mail_to calls across the EYTFI journey with the new helper using screen_reader: :default.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -22,6 +22,17 @@ module ApplicationHelper
     number_to_currency(value, delimiter: "", unit: "")
   end
 
+  # Wraps govuk_mail_to with visually-hidden spans so screen readers announce context
+  # when tabbing to an email link and handle all mail links in one place
+  # TODO: all govuk_mail_to should use claim_mail_to as it delegates to govuk_mail_to
+  #       if screen_reader params is not used
+  def claim_mail_to(email, text = nil, screen_reader: {}, **options)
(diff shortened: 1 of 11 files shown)


app/helpers/application_helper.rb
```diff
@@ -22,6 +22,17 @@ def currency_value_for_number_field(value)
...
+  # when tabbing to an email link and handle all mail links in one place
+  # TODO: all govuk_mail_to should use claim_mail_to as it delegates to govuk_mail_to
+  #       if screen_reader params is not used
+  def claim_mail_to(email, text = nil, screen_reader: {}, **options)
+    screen_reader = {before: "email", after: "if you need support"} if screen_reader == :default
+    before = screen_reader[:before] ? content_tag(:span, "#{screen_reader[:before]} ", class: "govuk-visually-hidden") : ""
+    after = screen_reader[:after] ? content_tag(:span, " #{screen_reader[:after]}", class: "govuk-visually-hidden") : ""
+    govuk_mail_to(email, "#{before}#{text || email}#{after}".html_safe, **options)
+  end
+
```
kenfodder, line 35: Initially thinking `:default` doesn't need passing in, but I wanted it to fallback if we did a global replace of all `govuk_mail_to` it would just work as it did.

kenfodder:
Not the right approach and no longer required.

PR #4621, opened 2026-06-04
only show link if journey open

rjlynch:


1993jamesgrant (approved):
Approved, but need to fix the unit test

PR #4623, opened 2026-06-05
Update postcode_search.html.erb

1993jamesgrant:


app/forms/postcode_search_form.rb
```diff
@@ -44,6 +44,7 @@ def address_data
     return nil if postcode.blank?
 
     @address_data ||= Rails.cache.fetch("address_data/#{postcode}", expires_in: 1.hour) do
+      journey_session.answers.assign_attributes(ordnance_survey_error: false)
```
rjlynch: Not sure about updating the journey session here, feel like this method should just be responsible for returning the address. Might be better to move this into `postcode_has_address` as that method is already setting `ordnance_survey_error` on the answers.

(this whole form needs refactoring imo, quite hard to follow the flow, though it's been working for a while, one for the back log!)

kenfodder:
@1993jamesgrant - I tested this...

Steps:
- Land on post code search
- Click link `Enter your address manually`
- Click `back`
- Enter a post code
- Takes you to Manual address search.

Is it correct behaviour that if you ever skipped the search you can never search again?

app/forms/select_home_address_form.rb
```diff
@@ -29,9 +29,15 @@ def save
 
     if skip_postcode_search?
       journey_session.answers.assign_attributes(
-        skip_postcode_search:
+        skip_postcode_search:,
+        postcode:
       )
     else
+      if address.blank?
+        errors.add(:address, "Select an address")
```
rjlynch: Could this use a macro?

```
validates(
  :address,
  presence: {
    message: "Select an address"
  },
  unless: :skip_postcode_search?
)
```

Appreciate the existing validation flow is quite confusing, as it's validating the session ??? rather than the form attributes so using a macro here might not work!

1993jamesgrant:
> @1993jamesgrant - I tested this...
> 
> Steps:
> 
> * Land on post code search
> * Click link `Enter your address manually`
> * Click `back`
> * Enter a post code
> * Takes you to Manual address search.
> 
> Is it correct behaviour that if you ever skipped the search you can never search again?

Looks like OS is down so I'll add a new error message

app/forms/select_home_address_form.rb
1993jamesgrant: I've put the test coverage in, can refactor the whole thing at a later date

app/forms/postcode_search_form.rb
rjlynch:
```suggestion
```
Still think this needs to go!

app/forms/select_home_address_form.rb
```diff
@@ -29,9 +29,19 @@ def save
 
     if skip_postcode_search?
       journey_session.answers.assign_attributes(
-        skip_postcode_search:
+        skip_postcode_search:,
+        address_line_1: nil,
+        address_line_2: nil,
+        address_line_3: nil,
+        address_line_4: nil,
+        postcode:
```
rjlynch: I think this will solve issue 3 on the jira ticket 👍

app/views/claims/postcode_search.html.erb
```diff
@@ -6,56 +6,31 @@
...
 
-      <% if @form.errors.any? %>
-        <%= f.govuk_fieldset legend: {
-          text: I18n.t("questions.address.home.no_address_found"),
-          tag: "h1",
-          size: "l"
-        } do %>
-          <p class="govuk-body">
-            We have not been able to find your address. Check the details you have
-            provided, or enter your address manually.
```
rjlynch, line 18: Are we no longer showing this message if we can't find any addresses for a postcode? Eg if someone enters `TE57 1NG`, which is a valid non real postcode

rjlynch (commented):
Over all seems fine.
Only real concern is that we seem to be no longer showing the "We have not been able to find your address." message. I could be wrong though, the way these forms interact is complicated!

app/forms/select_home_address_form.rb
rjlynch: For reference issue 3 is only an issue if you search for a postcode, select and address, go back and the pick "I can't find my address in the list"

app/views/claims/postcode_search.html.erb
1993jamesgrant: Alex wants us to stay on the main postcode page and select manual enter if there is an issue

app/forms/postcode_search_form.rb
1993jamesgrant: I can remove it yep

1993jamesgrant:
> Over all seems fine. Only real concern is that we seem to be no longer showing the "We have not been able to find your address." message. I could be wrong though, the way these forms interact is complicated!

Yeah, the ask is to remove it and just keep the person on the enter postcode form

rjlynch:
> > Over all seems fine. Only real concern is that we seem to be no longer showing the "We have not been able to find your address." message. I could be wrong though, the way these forms interact is complicated!
> 
> Yeah, the ask is to remove it and just keep the person on the enter postcode form

🤔 - I guess they'll see the validation error "address not found" which is probably good enough. I do think this needs some more thought as this code is used across all journeys. Changing the validation error message to be more useful might be enough.

rjlynch (approved):

1993jamesgrant:
> > > Over all seems fine. Only real concern is that we seem to be no longer showing the "We have not been able to find your address." message. I could be wrong though, the way these forms interact is complicated!
> > 
> > 
> > Yeah, the ask is to remove it and just keep the person on the enter postcode form
> 
> 🤔 - I guess they'll see the validation error "address not found" which is probably good enough. I do think this needs some more thought as this code is used across all journeys. Changing the validation error message to be more useful might be enough.

Good thought, I'll check if Alex is still ok with this on Monday

PR #4637, opened 2026-06-08
Capt3803

1993jamesgrant:
Fix bug on back button manual address input

The manual address page wasn't coming up if you complete the address section through to the gender page then go back and request to add it manually.

### DIFF
--- a/app/forms/postcode_search_form.rb
+++ b/app/forms/postcode_search_form.rb
@@ -21,10 +21,22 @@ class PostcodeSearchForm < Form
     return false if invalid?
 
     if skip_postcode_search
-      journey_session.answers.assign_attributes(skip_postcode_search:)
+      journey_session.answers.assign_attributes(
+        skip_postcode_search:,
+        address_line_1: nil,
+        address_line_2: nil,
+        address_line_3: nil,
+        address_line_4: nil
+      )
     else
-      journey_session.answers.assign_attributes(skip_postcode_search: false)
-      journey_session.answers.assign_attributes(postcode:)
+      journey_session.answers.assign_attributes(
+        skip_postcode_search: false,
+        address_line_1: nil,
+        address_line_2: nil,
+        address_line_3: nil,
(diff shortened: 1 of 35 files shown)


rjlynch (commented):
Would it be easier to not set the params on the submit buttons, and change the deleted form to

```
   <%= form_with(
      model: @form,
      url: @form.url,
      builder: GOVUKDesignSystemFormBuilder::FormBuilder,
      html: { id: "skip-postcode-search" }
    ) do |f| %>
      <%= f.hidden_field :skip_postcode_search, value: true %>
    <% end %>
```
?

app/forms/postcode_search_form.rb
```diff
@@ -21,7 +21,13 @@ def save
     return false if invalid?
 
     if skip_postcode_search
-      journey_session.answers.assign_attributes(skip_postcode_search:)
+      journey_session.answers.assign_attributes(
+        skip_postcode_search:,
+        address_line_1: nil,
+        address_line_2: nil,
+        address_line_3: nil,
+        address_line_4: nil
```
rjlynch, line 29: 👍 
Also think we should do this anytime a new postcode is entered.

rjlynch (commented):
I think the changes to the postcode search form should do the trick to clear the address, not sure why we need to set params on the submit button

1993jamesgrant:
> <%= form_with(
>       model: @form,
>       url: @form.url,
>       builder: GOVUKDesignSystemFormBuilder::FormBuilder,
>       html: { id: "skip-postcode-search" }
>     ) do |f| %>
>       <%= f.hidden_field :skip_postcode_search, value: true %>
>     <% end %>

I prefer keeping the logic within the form rather than referencing it elsewhere. Both achieve the same thing.

rjlynch (approved):
lgtm - only concern is having the skip post code search button submitting the same form. It's probably fine but will send over the other params which we don't care about if skip postcode search is clicked.

PR #4689, opened 2026-06-16
[CAPT-3871] Compatibility with dart sass

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-3871
- Govuk 6 has changed tooling to use dart sass
- There is a pure javascript implementation of dart sass which we use here for ease and reduce dependencies
- JS implementation is slower but this should not be important as only needed once at build time plus we have a very small amount of css

# Other changes

- Removed govuk publishing components. we only used one component `contents_list` which I have copied css and re-implemented component partial. this saves us pulling in a whole dependency for something very small
- new sass api means that globals are deprecated. css dependencies in files should be declared and calls must be namespaced
- removed or inlined some css which did not appear to be in-use or re-used
- some govuk colours have been deprecated and updated here. there maybe some minor changes in colours throughout the app
- still a couple of minor deprecation warning these will be fixed on next change
- added new `yarn` task. can be called via `yarn build:css` which will build css
- added `Procfile.dev` which can be used via foreman. call via `foreman start -f Procfile.dev`. this will run rails server and watch css for changes and automatically build changes for development


app/assets/stylesheets/components/contents_list.scss
```diff
@@ -0,0 +1,125 @@
...
+    padding: spacing.govuk-spacing(2) 0;
+    margin: 0;
+  }
+
+  .gem-c-contents-list__list-item-dash::before {
+    display: none;
+  }
+
+  .gem-c-contents-list__list-item--active {
+    padding: spacing.govuk-spacing(1) govuk-spacing(2);
```
kenfodder: Should be `spacing.govuk-spacing(2)`

rjlynch (approved):
Looks good, had a poke around locally, looks good

kenfodder (changes requested):

kenfodder (approved):

PR #4732, opened 2026-06-18
[CAPT-3871] upgrade rails_semantic_logger 4.18.0 => 4.20.0

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-3871

# Review notes

- Tested by signing into logit and viewing logs from both web and workers which seem to be displayed and reporting correctly


kenfodder (approved):
Approved for when the conflict is resolved.

PR #4734, opened 2026-06-18
Capt 3874 - add preview components to root page

1993jamesgrant:
This adds:

- Change to the root url on !prod containing
-- Journey components
-- Admin links
-- Journey links


app/controllers/claims_controller.rb
```diff
@@ -89,12 +89,16 @@ def persist_claim
   end
 
   def check_page_is_authorised
+    return if admin_component_preview_enabled_for_current_journey?
+
```
rjlynch, line 93: One thing we might want to consider is using module Prepend to overwrite these methods, and the ones in static pages controller, something like:

```
def check_page_is_authorised
  return if admin_component_preview_enabled_for_current_journey?
  
  super
end
```

most of the time when reading the claims controller code we wont care about preview functionality and we'll probably be reading it in a hurry while trying to fix some bug, so the simpler we can keep it the better.

Can also make the prepend conditional on the rails environment.

Other option would be to sub class the controller and send the routes to that if we're dealing with a preview but I think that would be higher impact.

app/controllers/claims_controller.rb
```diff
@@ -196,6 +201,21 @@ def set_cache_headers
     response.headers["Expires"] = "Fri, 01 Jan 1990 00:00:00 GMT"
   end
 
+  def admin_component_preview_enabled_for_current_journey?
+    return false if Rails.env.production?
```
rjlynch: Pretty sure review apps run in production mode.
Ideally we'd add a review_app? mode to the `env` so we can do:

return false if Rails.env.production? && !Rails.env.review_app?

app/models/journeys.rb
```diff
@@ -21,6 +21,12 @@ def all
     JOURNEYS
   end
 
+  def customer_journeys
+    all
+      .reject(&:start_with_magic_link?)
+      .sort_by(&:full_name)
+  end
```
rjlynch: Would be tempted to inline this in `app/views/static_pages/customer_journeys.html.erb` as it's only used for the preview functionality, and I could see a future dev relying on this method.

Would be tempted to name this "landing_page_journeys" as it's all journeys where the claimant / provider click "Start now"

config/routes.rb
```diff
@@ -1,10 +1,17 @@
...
+    root to: redirect(Rails.application.config.guidance_url)
+  else
+    root to: "static_pages#home"
+  end
 
   get "/claim/auth/tid/callback", to: "omniauth_callbacks#callback"
   get "/auth/failure", to: "omniauth_callbacks#failure"
   get "/auth/onelogin", to: "omniauth_callbacks#onelogin"
+  get "/journey-components", to: "static_pages#journey_components", as: :journey_components
+  get "/customer-journeys", to: "static_pages#customer_journeys", as: :customer_journeys
```
rjlynch: I think, unlike the claims controller, it wouldn't be too difficult to hang these off the `admin/components` controller so the static pages controller doesn't need to know about the previews use case. Would also let us inline the code in the `JourneyComponentsPage` concern.

app/controllers/admin/components_controller.rb
```diff
@@ -0,0 +1,140 @@
...
+
+      allowed_attributes = journey::SessionAnswers.attribute_names
+
+      default_answers
+        .merge(ineligible_preview_answers_for(journey, slug))
+        .slice(*allowed_attributes.map(&:to_sym))
+    end
+
+    def current_academic_year_for(journey)
+      Journeys::Configuration.unscoped.find_by(routing_name: journey.routing_name)&.current_academic_year || AcademicYear.current
```
rjlynch: How come we're unscoping this?
I _think_ we probably don't want to show the additional payments journey or the year one fe provide journey

app/controllers/admin/components_controller.rb
```diff
@@ -0,0 +1,140 @@
+module Admin
+  class ComponentsController < BaseAdminController
+    before_action :ensure_service_operator
```
rjlynch, line 3: `before_action -> { raise ActiveRecord::RecordNotFound if Rails.env.production? }`

I think we probably want a guard here so this can't run on prod

app/controllers/admin/components_controller.rb
1993jamesgrant: Ok, good to know, are they particularly secure?

app/controllers/admin/components_controller.rb
1993jamesgrant: Implemented

app/controllers/claims_controller.rb
1993jamesgrant: Good idea, implemented

app/models/journeys.rb
1993jamesgrant: I've kept this for future use. But renamed it for clarity across the PR

config/routes.rb
1993jamesgrant: Yep!

app/controllers/admin/components_controller.rb
rjlynch: They're old journeys that have been removed!

app/controllers/claims_controller.rb
1993jamesgrant: I haven't specifically done this change, might be worth doing via a future review if this tool becomes more popular

rjlynch: 👍 I think it's well worth doing in a follow up PR to keep the surface area of these changes small

app/controllers/static_pages_controller.rb
```diff
@@ -99,5 +112,8 @@ def journey_available?
   def current_user
     DfeSignIn::NullUser.new
   end
-  helper_method :current_user
+
+  def set_home_journey
+    @journey = Journeys::TargetedRetentionIncentivePayments
+  end
```
rjlynch: I think referencing a specific journey in this controller is a sign we should have the routes file's `root to:` point at the components controller, and move the `home` method and these changes into the components controller.

config/routes.rb
```diff
@@ -1,6 +1,10 @@
 Rails.application.routes.draw do
   # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
-  root to: redirect(Rails.application.config.guidance_url)
+  if Rails.env.production? && !Rails.env.review_app?
+    root to: redirect(Rails.application.config.guidance_url)
+  else
+    root to: "static_pages#home"
```
rjlynch:
```suggestion
    root to: "components#home"
```

rjlynch (approved):
LGTM - left a comment about pointing the app's root at the components controller, which I think is well worth doing before merging.

config/routes.rb
1993jamesgrant: done, good suggestion

PR #4740, opened 2026-06-19
[CAPT-3929] Claims without SLC data considers nil or false

asmega:
# Context

- https://dfedigital.atlassian.net/browse/CAPT-3929
- Claims with `submitted_using_slc_data` set to either `nil` or `false` are considered as claims submitted without SLC data


rjlynch (approved):
LGTM - my question is why can this be `nil` 😂 either the claim was submitted with SLC data or it wasn't!

PR #4746, opened 2026-06-23
Remove spring

kenfodder:
Remove Spring

Spring is a process preloader that hasn't been included in Rails defaults since Rails 7. On Rails 8.1 it provides no meaningful boot-time benefit and causes macOS fork-safety crashes (`+[NSCharacterSet initialize] may have been in progress in another thread when fork() was called`).

Changes:

- Remove `spring` and `spring-watcher-listen` gems from `Gemfile`
- Remove Spring loader from `bin/rails` and `bin/rake`
- Delete `bin/spring`
- Remove a now-exposed dead code block in `bin/rails` that attempted to start SimpleCov — it was never reachable before because Spring intercepted first; coverage tracking was already correctly handled by `spec/spec_helper.rb`


rjlynch (approved):
👍

PR #4750, opened 2026-06-23
persist check matching claims task

rjlynch:
Persist matching details tasks

When a claim is changed we need to update the matching details tasks.
This is a bit tricky now we're persisting tasks rather than calculating
if the task is needed on the fly.

When a claim is changed we need to recalculate if the matching details
task is now required, or if it needs to be removed. If the claim is now
a duplicate we set the matching details task on it and any of it's
duplicate claims. If the claim is no longer a duplicate we remove the
matching details task from the claim and any of it's duplicates unless:
* the former duplicate has other duplicates
* the claim is persisted
* the matching details task has been completed

We don't "reset" a completed matching details task if a new duplicate is
found, this matches the existing behaviour. Whether or not we want to
undo the matching details task if a new duplicate is submitted, which
would be a change in behaviour, is a decision for the product team.

There's two scenarios where the behaviour on this branch is different from
master.

1.) Duplicate comes in for an already decided claim.
On master we add the matching details task to the task list on the decided
claim. On this branch we do not (though we still show the warning).
I think the behaviour on master is incorrect. For claims that have been decided
the task list shouldn't change, as it's a snap shot of what was done to approve
/ reject the claim.

2.) Matching details task was decided (approved / rejected) and the duplicate
was resolved by amending one of the claims.
On master the task is removed from the task list. On this branch the task
remains in the task list.
I think the behaviour on master is also incorrect. If the task has been
completed then it should remain in the task list. It's not much of a problem on
master, affecting only 5 out of the 15K claims we've tested against.

### DIFF
--- a/app/models/automated_checks/claim_verifiers/matching_claims.rb
+++ b/app/models/automated_checks/claim_verifiers/matching_claims.rb
@@ -9,35 +9,16 @@ module AutomatedChecks
         ApplicationRecord.transaction do
           result = Claims::Match.update_matching_claims!(source_claim)
 
-          if result.removed_matches.any?
-            # remove the matching details task from the other claims only if
-            # * the task is not completed
-            # * the other claim is not decided
-            # * there are no _other_ match pairs on that claim
-            result.removed_matches
-              .select { |claim| Claims::Match.matching_claims(claim).none? }
-              .select { |claim| task_updateable?(claim) }
-              .each { |claim| remove_matching_details_task!(claim) }
-
-            if task_updateable?(source_claim) && Claims::Match.matching_claims(source_claim).none?
-              remove_matching_details_task!(source_claim)
-            end
+          result.removed_matches.each do |removed_match|
+            remove_match!(source_claim, removed_match)
+            remove_match!(removed_match, source_claim)
           end
 
           current_matches = result.new_matches + result.existing_matches
 
-          if current_matches.any?
-            # add the matching details task to matching claims only if
-            # * the other claim doesn't already have the task
-            # * the other claim is not decided
-            current_matches
-              .select { |claim| task_updateable?(claim) }
-              .select { |claim| !already_has_task?(claim) }
-              .each { |claim| add_matching_details_task!(claim) }
-
-            if task_updateable?(source_claim) && !already_has_task?(source_claim)
-              add_matching_details_task!(source_claim)
-            end
+          current_matches.each do |matching_claim|
+            record_match!(source_claim, matching_claim)
+            record_match!(matching_claim, source_claim)
           end
         end
       end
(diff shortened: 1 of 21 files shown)


app/models/policies/early_career_payments/claim_checking_tasks.rb
```diff
@@ -4,6 +4,8 @@ module Policies
   module EarlyCareerPayments
     class ClaimCheckingTasks < Policies::ClaimCheckingTasks
       def applicable_task_names
+        persisting_tasks_shim("matching_details")
+
```
kenfodder, line 8: So this actually triggers a job to create a task, I see it tries to guard idempotency, but will it handle it getting hit in quick succession?

app/models/automated_checks/claim_verifiers/matching_claims.rb
```diff
@@ -2,16 +2,84 @@ module AutomatedChecks
...
     class MatchingClaims
       def initialize(claim:)
-        @claim = claim
+        @source_claim = claim
       end
 
       def perform
-        Claims::Match.update_matching_claims!(claim)
+        ApplicationRecord.transaction do
+          result = Claims::Match.update_matching_claims!(source_claim)
```
kenfodder, line 10: If this raises and exception, do we need to catch it in `AutoApproveClaimJob`?

```
AutoApproveClaimsJob#perform
  ClaimAutoApproval#eligible?
    ClaimAutoApproval#auto_approvable?
      applicable_task_names   ← delegated to claim_checking_tasks (line 39)
        Policies::StudentLoans::ClaimCheckingTasks#applicable_task_names  (line 6)
          persisting_tasks_shim("matching_details")   ← base class, line 73
            task_exists?("matching_details") → false
              MatchingClaims.new(claim:).perform
                ApplicationRecord.transaction
                  Claims::Match.update_matching_claims!(source_claim)
                  task.save!(context: :claim_verifier)  ← raises RecordInvalid

```

kenfodder (commented):
I might need to study this more, but initial things I think might be issues.

app/models/policies/early_career_payments/claim_checking_tasks.rb
rjlynch: It's calling a claim verifier, which is run inline, so page load is a tiny bit slower first time it's hit, but then way faster as it doesn't need to run the matching claims finder again. Hopefully it won't ever be hit as we've got a job to create all the tasks. I just like having compatibility shims!

app/models/automated_checks/claim_verifiers/matching_claims.rb
rjlynch: I don't _think_ so. It will throw in two scenarios
1.) Two claims with the exact same timestamp, which is an indication we have a bug (also will throw before auto approval runs)
2.) `Claim::MatchingAttributeFinder` returns a matching claim who's attributes don't match, which again indicates a bug (don't think it's currently possible but may introduce a regression)

either way I think this should fail loudly.

Let me know if I'm missing something here 👍

kenfodder: Should be ok 👍

app/models/policies/early_career_payments/claim_checking_tasks.rb
kenfodder: Ah yes inline, should be fine then 👍

app/models/automated_checks/claim_verifiers/matching_claims.rb
```diff
@@ -2,16 +2,84 @@ module AutomatedChecks
...
 
-      attr_reader :claim
+      attr_reader :source_claim
+
+      def record_match!(target_claim, duplicate_claim)
+        return unless task_updateable?(target_claim)
+
+        task = target_claim.tasks.matching_details.last || target_claim.tasks.matching_details.new
+
+        data = task.data || {}
```
asmega, line 37: cyclical reference is a bit weird? can this not be `task.data ||= {}`, delete later assignment and use `task` rather than local variable

asmega (approved):

PR #4763, opened 2026-06-29
Fix task list not showing all claims

rjlynch:
Some tasks have additional statuses we weren't handling, eg
"partial_match", so claims with these statuses for some tasks were being
filtered out.

One thing to note is EYTRP is overloading the existing `employment` task
for it's employment check which is different to how other policy's
`employment` tasks works, so we're showing some extra filter options on
the EYTRP employment task which don't really apply. Didn't think it was
worth adding additional complexity to deal with these extra options
(they're harmless and will just show no claims if filtered). I think we
may want to consider renaming the `employment` task for EYTRP as it's
doing something different to the original `employment` task.

### DIFF
--- a/app/models/admin/task_list_form.rb
+++ b/app/models/admin/task_list_form.rb
@@ -19,15 +19,12 @@ module Admin
         end
 
         def filter_status
-          if name == "employment"
-            case status
-            when "na" then "incomplete"
-            else status.tr(" ", "_")
-            end
-          elsif status == "no data" || status == "na"
+          if status == "no data" && TASK_STATUSES.fetch(name, {}).include?("no_data")
+            "no_data"
+          elsif status == "na" || status == "no data"
             "incomplete"
           else
-            status
+            status.tr(" ", "_")
           end
         end
       end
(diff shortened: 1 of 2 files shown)


rjlynch:
Fixes #178

asmega (approved):
I think all the tasks need to be way more granular. and more of throw away pieces of code. having all this re-usability is simply terrible if we simply stuff them full of conditionals. I'd rather have 5 employment tasks, one for each policy.

rjlynch:
> I think all the tasks need to be way more granular. and more of throw away pieces of code. having all this re-usability is simply terrible if we simply stuff them full of conditionals. I'd rather have 5 employment tasks, one for each policy.

Yeah couldn't agree more.
I also think we need a `status` field on the Task, rather than it being a combination of
1.) Checking persistence
2.) Checking `claim_verifier_match`
3.) Checking if the task has passed or failed.

asmega:
> I also think we need a `status` field on the Task, rather than it being a combination of

I think once the persistence piece is in. this will make far more sense

PR #4771, opened 2026-07-01
170/self serve service access links

rjlynch:
# Make generating service access links self serve

Previously the ops team would have to reach out to us devs to generate these
service access links. The requests are frequent enough that we want to make this
self serve.


1993jamesgrant (approved):
LGTM

PR #4772, opened 2026-07-02
Forward service access code to old session

rjlynch:
If the journey is closed and the claimant is using a magic link, when
they go through one login and decide to resume from a previous session
we need to make sure we keep the service access code when overwriting
the session attributes with the old session's values, otherwise the
journey availability checking logic will kick them out.


kenfodder (approved):
Approved, linting error though needs sorting.

rjlynch:
> Approved, linting error though needs sorting.

Yeah it's just an issue with the terraform lint job falling over, rerun should sort it

PR #4777, opened 2026-07-03
[#141] Tidy up seeding code

asmega:
# Context

- https://github.com/DFE-Digital/claim-issues/issues/141
- Convert existing seeding script to use classes and methods
- No longer uses manually crafted claims, instead goes through `FactoryBot`

# Next steps

- Now that we are using classes, these can be isolated and tested
- Make seeding process idempotent so process can be run on every review app deployment
- Address data issue to cover more scenarios out of the box


db/seeders/base.rb
```diff
@@ -0,0 +1,125 @@
...
+
+    private
+
+    def load_factory_bot_definitions
+      FactoryBot.find_definitions
+    rescue FactoryBot::DuplicateDefinitionError
+      nil
+    end
+
+    def create_journey_configurations
```
1993jamesgrant, line 38: Could you do this automatically rather than needing to add a new one every time we create a journey?

asmega: We can change them to be in-memory ruby objects instead of database records if that's what you are suggesting? If they remain as database records they have to be added some point. If they are in memory objects, the config still needs to be persisted somewhere though. Not sure how you are suggesting they get added automatically?

db/seeds.rb
```diff
@@ -1,286 +1,17 @@
...
+require_relative "seeders/review"
 
-  Policies::TargetedRetentionIncentivePayments::Test::SchoolImporter.run
+seeder = if Rails.env.development?
+  Seeders::Development.new
+elsif Rails.env.test?
+  Seeders::Null.new
+else
+  Seeders::Review.new
 end
```
rjlynch, line 15: Would be tempted to switch this so Null is the default just on the very rare chance this gets accidentally run on prod some how!

```suggestion
seeder = if Rails.env.development?
  Seeders::Development.new
elsif ENV["ENVIRONMENT_NAME"].start_with?("review")
  Seeders::Review.new
else
  Seeders::Null.new
end
```

rjlynch (approved):
Much nicer.

Would be tempted to make the seeds idempotent but prob not worth it as realistically we only run them once.

db/seeders/base.rb
1993jamesgrant: I was thinking you can loop through the built journeys using the actual codebase as the reference point rather than a db object.

1993jamesgrant: We're maintaining a list in app/models/journeys

asmega: ok I see, minor thing but yea I can do that later on

db/seeds.rb
asmega: I did have that originally but didn't work. Just tested again and does work, so must have been something else. But yes will make this change

asmega:
> Would be tempted to make the seeds idempotent but prob not worth it as realistically we only run them once.

I too would like but yes might not be worth it. I will have a play. At least these structural changes should at least make that change far more manageable

PR #4787, opened 2026-07-07
127 - Refactor - FE - use journey_session.answers.update

kenfodder:
- Update FE specific journey only
- Other journeys to be updated in separate PRs

### DIFF
--- a/app/forms/journeys/further_education_payments/further_education_provision_search_form.rb
+++ b/app/forms/journeys/further_education_payments/further_education_provision_search_form.rb
@@ -17,7 +17,7 @@ module Journeys
         return if invalid? || no_results?
 
         if changed_query?
-          journey_session.answers.assign_attributes(
+          journey_session.answers.update!(
             possible_school_id: nil,
             provision_search:
           )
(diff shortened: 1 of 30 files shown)


app/forms/journeys/further_education_payments/further_education_provision_search_form.rb
```diff
@@ -17,30 +17,28 @@ def save
         return if invalid? || no_results?
 
         if changed_query?
-          journey_session.answers.assign_attributes(
+          journey_session.answers.update!(
```
asmega: i don't think file should be changed. we don't have to use `update!` everywhere just because it exists. i don't think the multiple saves here is what we want?

rjlynch (commented):
Looks good but agree Phil's comment needs addressing.
Next up it would be good to see if we could lose all the references to journey session, so it's just answers.update!, but that's for a future PR!

app/forms/journeys/further_education_payments/further_education_provision_search_form.rb
kenfodder: @rjlynch @asmega yep - good spot...

Reverted the problem bits in a second commit for ease of review: 727047c7f7176930b688568cc7f2f4114edb435b

kenfodder:
Created another PR for bundle all the changes...

https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/4793

PR #161, opened 2019-07-11
Set up Azure Release to deploy the app

erbridge:
Most of the release configuration is specified in the web UI, as it doesn't appear to be possible to specify it in YAML, yet.

Releases are done by pushing tags to DockerHub and restarting the app service.

PR #4789, opened 2026-07-07
[#161] Remove faraday_middleware gem

asmega:
# Context

- https://github.com/DFE-Digital/claim-issues/issues/161
- gem has been deprecated
- gem does not appear to be used anywhere

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -70,8 +70,6 @@ gem "rotp"
 
 gem "uk_postcode"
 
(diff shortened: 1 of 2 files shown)


rjlynch (approved):
I think this should be fine, we do have some specs that should exercise this. If there's an issue we can just bump faraday to 2.0 as that version has the stuff we used middleware for (json, authentication)

PR #4792, opened 2026-07-08
Show decisions have been undone

rjlynch:
If a decision on a claim has been undone we need to display this on the
decisions page to avoid confusion, otherwise we receive bug reports that
a claim has been approved twice.


1993jamesgrant (approved):
LGTM

PR #4810, opened 2026-07-14
Allow scoping search to current AY

rjlynch:
Request from the ops team to have the option to scope search results to
only the current academic year's claims.

### DIFF
--- a/app/controllers/admin/claims_controller.rb
+++ b/app/controllers/admin/claims_controller.rb
@@ -38,9 +38,14 @@ class Admin::ClaimsController < Admin::BaseAdminController
   end
 
   def search
+    @search = Claim::Search.new(
+      params[:query],
+      current_year_only: params[:current_year_only]
+    )
+
     return unless params[:query].present?
 
-    @claims = Claim::Search.new(params[:query]).claims
+    @claims = @search.claims
 
     if @claims.none?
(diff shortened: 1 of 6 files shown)


kenfodder:
If you click search with no search param you end up in the get params:

```
https://localhost:3000/admin/claims/search?query=&current_year_only=false&current_year_only=true&commit=Search
```
Screenshot 2026-07-20 at 11 23 43

rjlynch:
> If you click search with no search param you end up in the get params:

Kind of tricky to avoid that when we have a check box we want to default to checked. Unchecked checkboxes don't send up a param, so we use a hidden field tag to set the false value (so we can tell `false` a part from the default page load, where we want `current_year_only` to be `true`), which if the checkbox is checked gets superseded by param set by the checkbox. Does make the url a bit ugly, other option would be doing something in the controller like
```
current_year_only = if params[:current_year_only].nil? 
  true
else
  params[:current_year_only]
end
```

or pull active model into the search form.
Will take a look at pulling active model into the search form

kenfodder (approved):
Approved as it's functional and get it delivered, can make the fix in a separate PR.

rjlynch:
tl``

> Approved as it's functional and get it delivered, can make the fix in a separate PR.

Switched this all over to use active model though the double param issue is still there when the check box is set to true, just kind of how check boxes work!

PR #4812, opened 2026-07-15
2856 - Make staging more like review rather than test

hopep-mobilise:
### Context

Making the new staging environment more like review rather than test. 

### Changes proposed in this pull request

- global config and application config changed to be more like review rather than test.

### Guidance to review

Run command `make staging terraform-plan`. This should produce a valid set of required resources.

### Link to Trello card

https://trello.com/c/bA2sFyil/2856-capt-add-staging-env

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


terraform/application/config/staging.tfvars.json
```diff
@@ -3,30 +3,24 @@
     "namespace": "srtl-test",
     "config": "staging",
     "environment": "staging",
```
kenfodder: I think we need this to be `review` as in our code we look up the environment variable `ENVIRONMENT_NAME` to be `review` in order to run the seed rake task and other features to act like a review app.

kenfodder (changes requested):

terraform/application/config/staging.tfvars.json
hopep-mobilise: Updated

kenfodder (approved):

PR #4813, opened 2026-07-15
[#224] Persist task census subjects taught

asmega:
# Context

- Future `CensusSubjectsTaught` tasks are persisted to db when verifier called
- Add data migration to persist missing tasks to database

### DIFF
--- a/app/models/automated_checks/claim_verifiers/census_subjects_taught.rb
+++ b/app/models/automated_checks/claim_verifiers/census_subjects_taught.rb
@@ -64,6 +64,8 @@ module AutomatedChecks
           SchoolWorkforceCensus::ECP_ELIGIBLE_SUBJECTS
         ].reduce({}, :update)
 
+        return [] if claim.eligibility.eligible_itt_subject.nil?
+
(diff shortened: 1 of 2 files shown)


app/models/automated_checks/claim_verifiers/census_subjects_taught.rb
```diff
@@ -76,6 +64,8 @@ def early_career_payments_policy_subjects
           SchoolWorkforceCensus::ECP_ELIGIBLE_SUBJECTS
         ].reduce({}, :update)
 
+        return [] if claim.eligibility.eligible_itt_subject.nil?
```
rjlynch, line 67: Seems fine but I think we can delete this method and the line where it's referenced in the case statement.

rjlynch (approved):
LGTM

app/models/automated_checks/claim_verifiers/census_subjects_taught.rb
asmega: i had to add this line as there was an issue when testing. but if i remove does that not mean ECP will never be eligible which is not desired?

PR #4816, opened 2026-07-15
2856 - Revert staging infra to permanent resources

hopep-mobilise:
### Context

Making the new staging environment more like review rather than test. 

### Changes proposed in this pull request

- Added back in the resources required for a permanent environment rather than the review app.

### Guidance to review

Run command `make staging terraform-plan`. This should produce a valid set of required resources.

### Link to Trello card

https://trello.com/c/bA2sFyil/2856-capt-add-staging-env

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


kenfodder (approved):

global_config/staging.sh
```diff
@@ -1,6 +1,7 @@
 CONFIG=staging
 CONFIG_SHORT=st
+ENVIRONMENT=review
```
RMcVelia: staging not review

terraform/application/config/staging.tfvars.json
```diff
@@ -3,24 +3,33 @@
...
+    "enable_monitoring": true,
+    "statuscake_contact_groups":
+    [
+        195955,
+        282453
+    ],
+    "external_url": "https://staging.claim-additional-teaching-payment.service.gov.uk/healthcheck",
     "enable_logit": true,
+    "enable_postgres_backup_storage": true,
     "dataset_name": "claim_events_test",
```
RMcVelia, line 29: this is sending events to the same dataset as used by the test env. Is that what is wanted?

hopep-mobilise: There is no claim_events_staging table. Kenneth has requested that it go to claim_events_test the same as review and test.

global_config/staging.sh
hopep-mobilise: Sorted

terraform/application/config/staging.tfvars.json
kenfodder: We don't use it, we'll disable emitting there anyway, but just needs to point somewhere

global_config/staging.sh
kenfodder: ~Review on purpose, as `ENVIRONMENT_NAME` env var is used everywhere to check for `review` and `staging` in our app will act like that, but we have a list item to resolve that and flip it back to `staging` once we handle that in multiple places in our codebase. I believe it's not use infra wise just our codebase.~

Ignore that I was referring to `terraform/application/config/staging.tfvars.json`

PR #4819, opened 2026-07-16
[#224] Persist employment task

asmega:
# Context

### DIFF
--- a/app/models/automated_checks/claim_verifiers/employment.rb
+++ b/app/models/automated_checks/claim_verifiers/employment.rb
@@ -13,16 +13,22 @@ module AutomatedChecks
         return unless required?
         return unless awaiting_task?
 
-        no_data || no_match || matched
+        if claimant_tps_records.empty?
+          create_task(match: nil)
+        elsif claimant_tps_records.empty? || !eligible?
+          create_task(match: :none)
+        else
+          create_task(match: :all, passed: true)
+        end
       end
 
       private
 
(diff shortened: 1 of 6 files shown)


rjlynch:
@asmega apologies I rebased this PR thinking I was looking at my PR! Hasn't changed the actual files!

rjlynch (approved):
LGTM

app/models/automated_checks/claim_verifiers/employment.rb
```diff
@@ -11,19 +11,25 @@ def initialize(claim:, admin_user: nil)
 
       def perform
         return unless required?
-        return unless awaiting_task?
+        return if task.has_result?
```
rjlynch, line 14: should this be `return if task.passed?` as it looks like the employment check job passes old failed tasks to the verifier

asmega: but that would mean if a task is failed (eg via manual check) this automatic check could run again changing the state which would not be desired? i think if an outcome has been determined regardless of outcome we want to leave it alone?

rjlynch: Agreed. Not a blocker but might be worth amending that query so it's not passing claims with failed employment tasks to the verifier just to avoid confusion. One for a follow up pr

rjlynch (approved):
Approved again

PR #4821, opened 2026-07-16
223/projected payroll

rjlynch:
Add payroll projection

Shows an estimate of next month's payroll.
Unpaid approved claims always go into the next month's payroll.
Unpaid topups always go into the next month's payroll.
We then assume all undecided claims with a decision deadline before the
next month's payroll will end up being approved before payroll runs and
thus fall into the payroll.

### DIFF
--- a/app/models/payroll/projection.rb
+++ b/app/models/payroll/projection.rb
@@ -1,0 +1,57 @@
+module Payroll
+  class Projection
+    def total_award_amount
+      claim_scope.sum(:award_amount)
+    end
+
+    def number_of_claims_for_policy(policy)
+      claim_scope.by_policy(policy).count
+    end
+
+    def total_claim_amount_for_policy(policy)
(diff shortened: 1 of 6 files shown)


app/models/payroll/projection.rb
```diff
@@ -0,0 +1,61 @@
...
+      PayrollRun.maximum(:created_at)
+    end
+
+    def next_payroll_run_date
+      previous_payroll_run_date + 1.month
+    end
+
+    def date_range
+      previous_payroll_run_date..next_payroll_run_date
+    end
```
asmega: do we want the date range? why not just everything that is payrollable and approved?

rjlynch: Agreed, no date range is better. Originally I had this work with date ranges but all we really care about is the next payroll run, can't say much useful about other future payroll runs. Have pushed up a fix

asmega (approved):

PR #4823, opened 2026-07-17
215/deprecate-irp-journey

rjlynch:
Deprecates IRP journey as we're no longer accepting IRP applications.
Removes all the journey related code. Leaves the policy code required to view
IRP claims in the admin site.


rjlynch:
Don't merge until ops give us the go ahead.

PR #4824, opened 2026-07-17
[#224] Persist fe provider v2 task

asmega:
# Context

- Task is now idempotent and can be run as many times as needed
- Will update `passed` depending on state
- Amendment form no longer destroys task to recreate. task is simply re-ran which will update the value
- Added data migration that will persist all non-existent tasks

### DIFF
--- a/app/models/automated_checks/claim_verifiers/fe_provider_verification_v2.rb
+++ b/app/models/automated_checks/claim_verifiers/fe_provider_verification_v2.rb
@@ -11,16 +11,30 @@ module AutomatedChecks
       end
 
       def perform
-        return if task_already_persisted?
-
         if passable?
-          create_task(passed: true)
-        elsif failed_checks.any?
-          create_task(passed: false, data: {failed_checks: failed_checks})
+          if existing_task
+            existing_task.update(passed: true)
+          else
+            create_task(passed: true)
+          end
+        elsif provider_verification_completed? && failed_checks.any?
+          if existing_task
+            existing_task.update(passed: false, data: {failed_checks: failed_checks})
+          else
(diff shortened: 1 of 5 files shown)


app/models/automated_checks/claim_verifiers/fe_provider_verification_v2.rb
```diff
@@ -11,17 +11,31 @@ def initialize(claim)
...
+          if existing_task
+            existing_task.update(passed: false, data: {failed_checks: failed_checks})
+          else
+            create_task(passed: false, data: {failed_checks: failed_checks})
+          end
+        elsif existing_task
+          existing_task.update(passed: nil)
+        else
+          create_task(passed: nil)
         end
```
rjlynch, line 30:
```suggestion
      task = existing_task || claim.tasks.fe_provider_verification_v2.new
      
      if passable?
        task.update!(passed: true)
      elsif  provider_verification_completed? && failed_checks.any?
        task.update!(passed: false, data: {failed_checks: failed_checks})
      else
        task.update!(passed: nil)
      end
     end
```
I _think_ that works

rjlynch (approved):
Nice, left a suggestion but it's pretty minor

app/models/automated_checks/claim_verifiers/fe_provider_verification_v2.rb
asmega: yeah i agree it's not great but it works. the issue with your version is that is doesn't consider the `save` context. there are various save callbacks which need to fire in different contexts. I've not bothered touching those hence why i left the many branches

PR #4827, opened 2026-07-20
dont create journey config in claim factory

rjlynch:
Remove dependence on journey config from claim

Claim shouldn't need to know about journey configurations.

Bit of a noisy PR as there's quite a lot of implicit assumptions in the test suite about what year we're running in which we've had to make explicit.


kenfodder (approved):
👍

PR #4835, opened 2026-07-22
128 - Treat staging like a review app for auth bypasses and debug tooling

kenfodder:
Add Rails.env.review_app_like? (review_app? || staging_app?) and use it everywhere a review-only check previously gated debug sign-in bypasses, the admin component preview, the /testdata route, and session cookie domain handling, so the new staging environment gets the same capabilities as review apps.

config/initializers/dfe_analytics.rb keeps an inline check since it loads before rails_env_extensions.rb alphabetically. db/seeds.rb is left on review_app? only, so seed data still only runs for review apps.

### DIFF
--- a/app/controllers/admin/components_controller.rb
+++ b/app/controllers/admin/components_controller.rb
@@ -1,7 +1,7 @@
 module Admin
   class ComponentsController < BaseAdminController
     before_action :ensure_service_operator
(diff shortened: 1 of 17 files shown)


rjlynch (approved):
LGTM - though I think at some point we'll want to change `Rails.env.production?` to only return `true` in production!

PR #4841, opened 2026-07-24
Different approach to persisting matching claims task

rjlynch:
Seeing if this is simpler / will do what we need.
tldr keep using the live query for the warning in the views and just focus on
persisting the matching task.
Can then drop the whole Claims::Match stuff

## Notes for reviewer
There's a fair bit of noise in this pr where we remove the `Claims::Match` stuff
and replace it with the original MatchingAttributeFinder based approach.
The main file to pay attention to is
`app/models/automated_checks/claim_verifiers/matching_claims.rb` which handles
creating / removing the task if there's no longer a match.

I'll send up a follow up pr to drop the claims_match table.


app/models/automated_checks/claim_verifiers/matching_claims.rb
```diff
@@ -2,16 +2,98 @@ module AutomatedChecks
...
+      attr_reader :source_claim
+
+      def existing_matches
+        claim_ids = Task
+          .matching_details
+          .joins(:claim)
+          .merge(Claim.by_academic_year(source_claim.academic_year))
+          .where(
+            "data @> ?",
+            {matching_claims: [source_claim.reference]}.to_json
```
asmega, line 39: for another day but pretty confident there would be an operator that lets you check if a string exists in an array?

app/models/task.rb
```diff
@@ -88,8 +88,16 @@ def identity_confirmation?
...
+  # the task being persisted meant it was completed. Now incomplete tasks can be
+  # persisted sso we need to know if the task has an outcome to determine
+  # completeness.
   def completed?
-    persisted?
+    if name == "matching_details"
+      !passed.nil?
+    else
+      persisted?
+    end
```
asmega, line 100: this works for now as it keeps the behaviour the same. but i think we need to circle back at some point. surely for a task to be completed there should be some kind of outcome ie pass/fail?

db/data/20260727133142_persist_matching_attributes_task.rb
```diff
@@ -0,0 +1,17 @@
+# Run me with `rails runner db/data/20260727133142_persist_matching_attributes_task.rb`
+
+# Put your Ruby code here
```
asmega, line 3: i generally delete this line 😂

db/data/20260727133142_persist_matching_attributes_task.rb
```diff
@@ -0,0 +1,17 @@
...
+
+claims_with_persisted_task = Task
+  .matching_details
+  .joins(:claim)
+  .merge(Claim.by_academic_year(AcademicYear.new(2025)))
+  .select(:claim_id)
+
+claims_without_persisted_task = Claim
+  .by_academic_year(AcademicYear.new(2025))
+  .where.not(id: claims_with_persisted_task)
```
asmega, line 13: just an fyi theres a scope `awaiting_task` on claim. you can use for things without a certain task that might meet your needs

asmega (approved):

app/models/task.rb
rjlynch: Agreed. I the current approach is for `persisted?` to mean the task is complete, unless there's `claim_verifier_match`, so was trying to have one method handling both old and new tasks. Once you're parts are done we'll come back to this 👍

app/models/automated_checks/claim_verifiers/matching_claims.rb
rjlynch: I think that's what that operator does! Checks if any element in the `matching_claims` array has that value, eg:

```
Task.matching_details.where("data @> ?", { matching_claims: ["4LCYUGCX"] }.to_json).last.data #=> {"matching_claims" => ["4LCYUGCX", "abc"]}
```
so we've matched one of the entries in the array

PR #4843, opened 2026-07-27
Upgrades review and test postgres to v17

mob-galtryn:
This PR upgrades the postgres version to v17 for the test and review apps

### DIFF
--- a/terraform/application/variables.tf
+++ b/terraform/application/variables.tf
@@ -89,6 +89,12 @@ variable "postgres_flexible_server_sku" {
 variable "postgres_enable_high_availability" {
   default = false
 }
+
+variable "server_version" {
+  type = number
+  default = 16
+  description = "Version number of Postgres to use"
+}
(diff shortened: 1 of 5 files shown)


terraform/application/variables.tf
```diff
@@ -89,6 +89,12 @@ variable "postgres_flexible_server_sku" {
 variable "postgres_enable_high_availability" {
   default = false
 }
+
+variable "server_version" {
+  type = number
+  default = 16
```
kenfodder, line 95: @mob-galtryn - Is supposed to be 16?

mob-galtryn: Yes that is the default value, currently applies to production and staging. test.tfvars.json overrides that to 17 with server_version variable. That will be added to the other envs as we go and the default changed 17 once all envs are complete.

kenfodder (approved):

PR #4845, opened 2026-07-28
[#224] Persist student loan plan task

asmega:
# Context


app/jobs/student_loan_plan_check_job.rb
```diff
@@ -1,35 +1,51 @@
...
   def perform(admin)
-    delete_no_data_student_loan_plan_tasks
-    claims = current_year_claims_awaiting_decision.awaiting_task("student_loan_plan")
+    claims = current_year_claims_awaiting_decision
+      .joins(:tasks)
+      .where(
+        tasks: {
+          name: "student_loan_plan"
+        }
+      ).merge(Task.awaiting_passed_result)
```
asmega, line 9: we assume any record where `passed` is `nil` still requires processing

db/data/20260727142406_persist_sl_plan_task.rb
```diff
@@ -0,0 +1,26 @@
+# Run me with `rails runner db/data/20260727142406_persist_sl_plan_task.rb`
+
+admin = DfeSignIn::User.find("6b15ba5b-8f3a-48ed-bb7f-721630ffdf6a")
```
asmega, line 3: me

app/jobs/student_loan_plan_check_job.rb
```diff
@@ -1,35 +1,51 @@
...
-    current_year_claims_awaiting_decision.joins(:tasks).where(tasks: {name: "student_loan_plan", claim_verifier_match: nil, manual: false})
-  end
+  private
 
   def applicable_policies
     Policies.all.select(&:auto_check_student_loan_plan_task?)
   end
 
   def current_year_claims_awaiting_decision
-    Claim::ClaimsAwaitingDecisionFinder.new(policies: applicable_policies).claims_submitted_without_slc_data
```
asmega, line 33: this class was only used here so has been inlined

app/jobs/student_loan_plan_check_job.rb
kenfodder: Would a duplicate note be created on every SLC upload? 

In `app/models/automated_checks/claim_verifiers/student_loan_plan.rb`

app/models/automated_checks/claim_verifiers/student_loan_plan.rb
```diff
@@ -20,10 +11,26 @@ def initialize(claim:, admin_user: nil)
 
       def perform
         return unless claim.policy.auto_check_student_loan_plan_task?
-        return unless claim.submitted_without_slc_data?
-        return unless awaiting_task?
+        return if task.has_result?
+
+        if student_loans_data.any?
+          if claim.student_loan_plan.blank?
+            task.assign_attributes(claim_verifier_match: nil, passed: nil, reason: nil)
```
kenfodder, line 20: Two claims come through the same SLC upload:

Claim A — SLC has a row matching their NINO and date of birth, but claim.student_loan_plan is blank (the claim never captured a plan).
Claim B — SLC has no row for them at all.

What each ends up with:

Claim A (data found, no plan)
- attrs `match: nil, passed: nil, reason: nil`
- status: "No data" / red
- note: "SLC data checked, no matching entry found""

Claim B (genuinely no data)
- attrs: `match: nil, passed: nil, reason: "incomplete"`
- status: "Incomplete" / grey
- note: "SLC data checked, no matching entry found"

Before this branch, Claim A raised MissingClaimPlanError — the job rescued it into Sentry, so someone got told about the data anomaly. Now it produces a task and a note that both point at the wrong cause.

Do we need a `missing_claim_plan` or something?

app/jobs/student_loan_plan_check_job.rb
asmega: Yes. I've made some changes so we detect any changes and only update the task and create a note if there are any changes detected. otherwise task it not updated and no note is created

app/models/automated_checks/claim_verifiers/student_loan_plan.rb
asmega: Made a change so where previously an exception was raised, the content for the note in this scenario is the same message as the exception message. do you think that is enough to cover the issue?

kenfodder (approved):

PR #4849, opened 2026-07-29
280 - Extract award amount validation into policy rules objects

kenfodder:
Preparation for moving award_amount onto claims. SL and TRI are the only policies constraining award size and their rules differ in shape, so each now owns an AwardAmountRules object reached via BasePolicy#award_amount_rules, following the existing ClaimCheckingTasks/AdminTasksPresenter dispatch idiom. The objects are duck typed on whatever record owns the award amount, so they move with the column unchanged.

No behaviour change; all existing specs pass unedited. TRI's current_academic_year bug is preserved deliberately and fixed when the rules move onto Claim.

Adds a characterisation spec documenting that on: :amendment rules do not fire through the admin amendment UI, and pinning a pre-existing 500 when a Student Loans award is amended above £99,999.

### DIFF
--- a/app/models/base_policy.rb
+++ b/app/models/base_policy.rb
@@ -94,6 +94,12 @@ module BasePolicy
     10_000.00
   end
 
+  # Policies that constrain the size of an award define an AwardAmountRules and
+  # overwrite this. `record` is whichever model owns the award amount.
+  def award_amount_rules(record)
+    nil
+  end
+
   def current_academic_year
     Journeys.for_policy(self).configuration.current_academic_year
   end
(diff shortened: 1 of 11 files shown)


rjlynch (commented):
Given we only change award amounts via the amendments form, wondering if it would be simpler to have the validations on there and drop them from the eligibility model?
For loans and TRI we'd just introduce subclasses of the amendments form (mechanism for this is already in place) and overwrite the award amount validation from the base class. Save introducing a new concept into the code base. Thoughts?

asmega:
closing as the contents of this commit are present in https://github.com/DFE-Digital/claim-additional-payments-for-teaching/pull/4905

PR #4855, opened 2026-07-30
[#224] Persist qualifications task

asmega:
# Context

### DIFF
--- a/app/models/automated_checks/claim_verifiers/qualifications.rb
+++ b/app/models/automated_checks/claim_verifiers/qualifications.rb
@@ -15,9 +15,13 @@ module AutomatedChecks
       end
 
       def perform
-        return unless awaiting_task?(TASK_NAME)
+        return if task_exists?
 
-        no_match || complete_match
+        if dqt_teacher_status.nil? || !dqt_teacher_status.eligible?
+          create_task(match: :none)
+        else
+          create_task(match: :all, passed: true)
+        end
       end
 
       private
@@ -25,14 +29,8 @@ module AutomatedChecks
       attr_accessor :admin_user, :claim
       attr_reader :dqt_teacher_status
 
-      def awaiting_task?(task_name)
-        claim.tasks.none? { |task| task.name == task_name }
-      end
-
-      def complete_match
-        return unless dqt_teacher_status.eligible?
-
-        create_task(match: :all, passed: true)
+      def task_exists?
+        claim.tasks.where(name: TASK_NAME).exists?
       end
 
       def create_note(match:)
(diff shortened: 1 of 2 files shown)


app/models/automated_checks/claim_verifiers/qualifications.rb
```diff
@@ -15,32 +15,30 @@ def initialize(
...
+        claim.tasks.where(name: TASK_NAME).exists?
       end
 
       def create_note(match:)
         body = if dqt_teacher_status.nil?
           "[DQT Qualification] - Not eligible"
         else
           <<~HTML
-            [DQT Qualification] - #{(match == :none) ? "Ine" : "E"}ligible:
+            [DQT Qualification] - #{(match == :none) ? "Ineligible:" : "Eligible:"}
```
rjlynch, line 41: I'm not sure about this! The old approach was very optimised 😂 😂

db/data/20260730135238_persist_task_qualifications.rb
```diff
@@ -0,0 +1,38 @@
...
+
+relevant_policies = [
+  Policies::StudentLoans,
+  Policies::EarlyCareerPayments,
+  Policies::TargetedRetentionIncentivePayments
+]
+
+relevant_policies.each do |policy|
+  puts "policy: #{policy}"
+  claims = Claim.where(policy:)
```
rjlynch, line 11: Would be tempted to scope this just to the current academic year as the api calls could get quite slow

db/data/20260730135238_persist_task_qualifications.rb
```diff
@@ -0,0 +1,38 @@
...
+  claims_without_task = Claim.where(id: delta)
+
+  claims_without_task.find_each do |claim|
+    dqt_teacher_status = if claim.has_dqt_record?
+      Dqt::Teacher.new(claim.dqt_teacher_status)
+    elsif claim.eligibility.teacher_reference_number.present?
+      Dqt::Client.new.teacher.find(
+        claim.eligibility.teacher_reference_number,
+        include: "alerts,induction,routesToProfessionalStatuses"
+      )
```
rjlynch, line 27: We should probably open a ticket to make `app/jobs/claim_verifier_job.rb` store the dqt_teacher_status on the claim record if it's missing, rather than pass it in and then just lose the data once the checks have finished.

rjlynch (approved):
LGTM - might be worth checking how many claims we're going to have to make the api call for and if it's a lot have the runner kick off some jobs to persist the task

db/data/20260730135238_persist_task_qualifications.rb
rjlynch: Have put a ticket in the backlog to have claim verifier job persist the dqt data it fetches.

db/data/20260730135238_persist_task_qualifications.rb
asmega: i've checked the numbers and don't seem overly excessive so think it should be fine in this instace

PR #4856, opened 2026-07-30
Automatically close journeys 254

1993jamesgrant:
Adds a "close at" field to journeys to enable windows to be scheduled for closure rather than manually.

Also send's notification reminders to all admins so no one forgets

### DIFF
--- a/app/controllers/admin/components_controller.rb
+++ b/app/controllers/admin/components_controller.rb
@@ -180,10 +180,16 @@ module Admin
     end
 
     def ensure_preview_journey_configuration!(journey)
-      Journeys::Configuration.find_or_create_by!(routing_name: journey.routing_name) do |configuration|
-        configuration.current_academic_year = AcademicYear.current
+      configuration = Journeys::Configuration.find_or_initialize_by(routing_name: journey.routing_name)
+      configuration.current_academic_year ||= AcademicYear.current
+
+      if configuration.new_record?
         configuration.open_for_submissions = true
+        configuration.close_at = 1.year.from_now.in_time_zone("London")
       end
+
+      configuration.save!
+      configuration
     end
 
     def ineligible_preview_answers_for(journey, slug)
--- a/app/controllers/admin/components_controller.rb
+++ b/app/controllers/admin/components_controller.rb
@@ -140,8 +140,6 @@ module Admin
     end
 
     def create_prepopulated_preview_session!(journey, slug)
-      ensure_preview_journey_configuration!(journey)
-
       session_key = :"#{journey.routing_name}_journeys_session_id"
       session.delete(session_key)
 
@@ -177,13 +175,6 @@ module Admin
 
     def current_academic_year_for(journey)
       Journeys::Configuration.find_by(routing_name: journey.routing_name)&.current_academic_year || AcademicYear.current
-    end
-
-    def ensure_preview_journey_configuration!(journey)
-      Journeys::Configuration.find_or_create_by!(routing_name: journey.routing_name) do |configuration|
-        configuration.current_academic_year = AcademicYear.current
-        configuration.open_for_submissions = true
-      end
     end
 
     def ineligible_preview_answers_for(journey, slug)
--- a/app/jobs/close_expired_journeys_job.rb
+++ b/app/jobs/close_expired_journeys_job.rb
@@ -5,9 +5,28 @@ class CloseExpiredJourneysJob < ApplicationJob
     Journeys::Configuration
       .where(open_for_submissions: true)
       .where.not(close_at: nil)
-      .where("close_at <= ?", now)
       .find_each do |journey_configuration|
+        send_admin_notification_if_due(journey_configuration, now)
+
+        next unless journey_configuration.close_at <= now
+
         journey_configuration.update!(open_for_submissions: false, close_at: nil)
       end
+  end
+
+  private
+
+  def send_admin_notification_if_due(journey_configuration, now)
+    return unless journey_configuration.close_at.present?
+
+    due_for_notification_at = [journey_configuration.close_at - 7.days, journey_configuration.close_at - 2.days]
+
+    due_for_notification_at.each do |notification_time|
+      next unless notification_time <= now && notification_time > now - 1.minute
+
+      DfeSignIn::User.admin.not_deleted.find_each do |admin_user|
+        AdminMailer.service_closing_soon(admin_user.email, journey_name: journey_configuration.journey.journey_name, close_at: journey_configuration.close_at).deliver_now
+      end
+    end
   end
 end
--- a/app/jobs/close_expired_journeys_job.rb
+++ b/app/jobs/close_expired_journeys_job.rb
@@ -1,0 +1,13 @@
+class CloseExpiredJourneysJob < ApplicationJob
+  def perform
+    now = Time.current.in_time_zone("London")
+
+    Journeys::Configuration
+      .where(open_for_submissions: true)
+      .where.not(close_at: nil)
+      .where("close_at <= ?", now)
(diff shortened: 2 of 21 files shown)


app/controllers/admin/components_controller.rb
```diff
@@ -180,10 +180,16 @@ def current_academic_year_for(journey)
...
-        configuration.current_academic_year = AcademicYear.current
+      configuration = Journeys::Configuration.find_or_initialize_by(routing_name: journey.routing_name)
+      configuration.current_academic_year ||= AcademicYear.current
+
+      if configuration.new_record?
         configuration.open_for_submissions = true
+        configuration.close_at = 1.year.from_now.in_time_zone("London")
       end
+
+      configuration.save!
```
asmega: don't think you ever want to save? i think you can delete this line any everything else should just work without need to persist anything?

config/recurring.yml
```diff
@@ -55,3 +58,6 @@ production: &production
     schedule: "0 * * * *" # hourly on the hour
 development:
   <<: *production
+  close_expired_journeys:
+    class: CloseExpiredJourneysJob
+    schedule: "* * * * *"
```
asmega: think you can delete these 3 lines? if you've already included it in the production anchor the alias on the line above should load it

config/recurring.yml
```diff
@@ -29,6 +29,9 @@ production: &production
   heartbeat:
     class: HeartbeatJob
     schedule: "* * * * *"
+  close_expired_journeys:
+    class: CloseExpiredJourneysJob
+    schedule: "* * * * *"
```
asmega, line 34: fyi solid queue now supports dynamic scheduled tasks, see https://github.com/rails/solid_queue#scheduling-and-unscheduling-recurring-tasks-dynamically might be easier than a cron that runs every minute?

app/views/admin/journey_configurations/index.html.erb
```diff
@@ -27,12 +27,12 @@
...
         <% @journey_configurations.each do |journey_configuration| %>
           <tr class="govuk-table__row" data-policy-configuration-routing-name="<%= journey_configuration.routing_name %>">
-            <th scope="row" class="govuk-table__header"><%= journey_configuration.journey.full_name %></th>
+            <th scope="row" class="govuk-table__header"><%= journey_configuration.journey&.full_name %></th>
             <td class="govuk-table__cell"><%= journey_configuration.current_academic_year %></td>
             <td class="govuk-table__cell"><%= journey_configuration.open_for_submissions? ? "Open" : "Closed" %></td>
             <td class="govuk-table__cell">
               <%= govuk_link_to edit_admin_journey_configuration_path(journey_configuration) do %>
-                Change <span class="govuk-visually-hidden"><%= journey_configuration.journey.full_name %></span>
+                Change <span class="govuk-visually-hidden"><%= journey_configuration.journey&.full_name %></span>
```
asmega: how come we now need a `nil` guard? if this was due to accidentally persisting journey configs which hopefully will be resolved from my other comment, can this now be reverted?

app/views/admin/journey_configurations/edit.html.erb
```diff
@@ -57,26 +57,28 @@
...
+                  Close at (Europe/London)
+                </label>
+                <% if close_at_error %>
+                  <p class="govuk-error-message">
+                    <span class="govuk-visually-hidden">Error:</span>
+                    <%= journey_configuration.errors.full_messages_for(:close_at).first %>
                   </p>
-                </div>
-                <%= f.text_area :availability_message, class: "govuk-textarea", "aria-describedby" => "availability_message-hint", rows: 5 %>
+                <% end %>
```
asmega: it does work but if we use a form object at the controller similar to the other form objects there it would have saved you from needing to manually handle all this

app/jobs/close_expired_journeys_job.rb
```diff
@@ -0,0 +1,35 @@
...
+  def perform
+    now = Time.current.in_time_zone("London").to_i
+
+    Journeys::Configuration
+      .where(open_for_submissions: true)
+      .where.not(close_at: nil)
+      .find_each do |journey_configuration|
+        send_admin_notification_if_due(journey_configuration, now)
+
+        close_at = journey_configuration.close_at&.in_time_zone("London")&.to_i
```
asmega: theres no need to convert dates to integers. you are able to compare 2 date objects natively without needing to do any translating first. unless i'm misunderstanding something?

asmega: the nil guard i think is also not needed as this should have been guarded against in the SQL query

app/models/journeys/configuration.rb
```diff
@@ -26,6 +26,7 @@ class Configuration < ApplicationRecord
     attribute :current_academic_year, AcademicYear::Type.new
 
     validates :current_academic_year_before_type_cast, format: {with: AcademicYear::ACADEMIC_YEAR_REGEXP}
+    validate :close_at_validations, if: -> { open_for_submissions == true }
```
asmega: are we sure we want to force this feature on users? will we ever not know when a window will close in advance? ie do we want to make this field optional?

config/recurring.yml
1993jamesgrant: I'll raise this as a piece of tech debt to migrate the entire cron system so we're not using both approaches simultaneously

app/views/admin/journey_configurations/edit.html.erb
1993jamesgrant: Switched back to form helpers

app/models/journeys/configuration.rb
1993jamesgrant: Yes, Product wanted it mandatory with notifications 👍

app/jobs/close_expired_journeys_job.rb
```diff
@@ -0,0 +1,35 @@
+class CloseExpiredJourneysJob < ApplicationJob
+  def perform
+    now = Time.current.in_time_zone("London").round(0)
+
+    Journeys::Configuration
+      .where(open_for_submissions: true)
+      .where.not(close_at: nil)
```
asmega, line 7: can we change this so the `close_at` is part of the query rather than having to check each record? so something like
```suggestion
      .where(close_at: ..Time.zone.now)
```
then you can remove the extra checks in the block

asmega: may be ignore me, i guess you are wanting to send out notifications prior to closing which i missed

asmega: you will have potential bugs though if your jobs are executing sub-second and the scheduler decides to enqueue the next job

db/seeders/base.rb
```diff
@@ -37,7 +37,9 @@ def create_journey_configurations
         Journeys::Configuration
           .create!(
             routing_name: journey.routing_name,
-            current_academic_year: AcademicYear.current
+            current_academic_year: AcademicYear.current,
+            open_for_submissions: true,
+            close_at: 1.year.from_now.in_time_zone("London")
```
asmega: think i'd set this to 100 years in the futures to prevent journeys from closing?

app/jobs/close_expired_journeys_job.rb
asmega: i think conflating service closing and the notifications makes this code more difficult than needs to be. if they were separated it naturally be simplified.

asmega (approved):

app/jobs/close_expired_journeys_job.rb
1993jamesgrant: I was thinking to implement and split it out but seemed like excess at this stage.

PR #4857, opened 2026-07-31
287 - Disable Active Storage variant processing

kenfodder:
Silences the boot warning about the missing image_processing gem, and makes explicit that Claim never transforms uploaded images.


rjlynch (approved):
👍

PR #4862, opened 2026-08-04
Ci improvements

rjlynch:
# Some tweaks to speed up CI

Switches to parallel_tests in ci as turbo_tests don't let us provide a runtime
log (used for balancing nodes) and we don't care about incremental output in ci.

Disables simple cov in CI - test suite spread across nodes so we can't get
accurate coverage anyway

Switches back to cached classes - we're not using spring so we can go back to
the default.

Fixes some flakes where the school select drop down isn't being dismissed.


kenfodder:
@rjlynch - I don't see any test output in the logs, might not need it, what does it look like if there are failures?

asmega:
@rjlynch the failed test doesn't output anything useful to identify what failed? you'll have to run locally to find them which would defeat the point!

rjlynch:
> @rjlynch the failed test doesn't output anything useful to identify what failed? you'll have to run locally to find them which would defeat the point!

Have sorted that now - here's output from a failure https://github.com/DFE-Digital/claim-additional-payments-for-teaching/actions/runs/31707154896/job/94470477702?pr=4862

config/environments/test.rb
```diff
@@ -8,7 +8,7 @@
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
-  config.cache_classes = false
+  config.cache_classes = true
```
asmega, line 11: 👍

asmega (approved):

PR #4863, opened 2026-08-04
[#42] hide fe_repeat_applicant_check task if auto passed

asmega:
# Context

- https://github.com/DFE-Digital/claim-issues/issues/42


rjlynch (approved):
LGTM

I think eventually we'll probably want to move to something like a `Task#shown?` method to control visibility, so we can keep the task in the task list. I guess it depends on how the task list evolves

asmega:
> LGTM
> 
> I think eventually we'll probably want to move to something like a `Task#shown?` method to control visibility, so we can keep the task in the task list. I guess it depends on how the task list evolves

yeah not sure what the mechanism will be for show/hiding, potentially quite a few options seem available

PR #4869, opened 2026-08-06
Update personal_bank_account.html.erb

1993jamesgrant:
Changes to bank details form to bring inline with Gov style guidelines

### DIFF
--- a/app/views/claims/personal_bank_account.html.erb
+++ b/app/views/claims/personal_bank_account.html.erb
@@ -15,12 +15,11 @@
           spellcheck: "false",
           autocomplete: "name",
           label: {
-            text: @form.t("banking_name.label", default: "Name on your account")
+            text: @form.t("banking_name.label", default: "Name on the account")
           },
           hint: {
             text: @form.t(
-              "banking_name.hint",
-              default: t("questions.account_hint")
+              "banking_name.hint"
             )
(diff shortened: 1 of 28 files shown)


app/views/claims/personal_bank_account.html.erb
```diff
@@ -15,12 +15,12 @@
           spellcheck: "false",
           autocomplete: "name",
           label: {
-            text: @form.t("banking_name.label", default: "Name on your account")
+            text: @form.t("banking_name.label", default: "Name on the account")
           },
           hint: {
             text: @form.t(
               "banking_name.hint",
-              default: t("questions.account_hint")
```
rjlynch, line 23: can probably drop this line from the locales now as it's now not used anywhere

rjlynch (approved):
LGTM - would be tempted to drop the `hint` section from the view as it's now only used by EYFI, which is going to be deprecated anyway.

kenfodder:
@1993jamesgrant 

app/forms/personal_bank_account_form.rb:49 requires it to be 8 digits, but the message says 6-8 digits?

kenfodder:
@1993jamesgrant 

app/models/journeys/base_answers_presenter.rb:29 might need a slight tweak to make it match.

PR #4870, opened 2026-08-06
[#291] Role based admins

asmega:
# Context

- https://github.com/DFE-Digital/claim-issues/issues/291
- Introduced pundit, although probably bent the rules on how it was intended to be used
- Instead policies are applied at controller level this way fewer policies are required as for the moment we only have 2 which are used globablly
- However should we wish to change tack having pundit in place should make this change much easier
- At the moment attribute based permissions does not look like a direction we will be heading down so have left the current single form where it is applied for banking details. Pundit does support attributes should we use to use this in the future


app/forms/admin/amendment_form.rb
```diff
@@ -109,7 +109,21 @@ class Admin::AmendmentForm
       equal_to: ->(form) { form.claim.banking_name },
       message: "You do not have permission to change the banking name"
     },
-    if: :banking_name_disabled?
+    if: proc { |form| form.field_disabled?(:banking_name) }
+
+  validates :bank_sort_code,
+    comparison: {
```
rjlynch, line 133: Nice 👍

app/forms/admin/amendment_form.rb
```diff
@@ -135,13 +149,30 @@ def self.form_for_claim(claim)
...
+    instance = new(claim:, admin_user:)
+    instance.amendable_attributes
+  end
+
+  def field_disabled?(field)
+    amendable_attributes.exclude?(field)
+  end
+
+  def amendable_attributes
     array = Claim::AMENDABLE_ATTRIBUTES + claim.policy::Eligibility::AMENDABLE_ATTRIBUTES + [:notes]
```
rjlynch: Unrelated, but AMENDABLE_ATTRIBUTES should probably be defined in the amendment forms, with policy specific amendment forms handling the different policy amendable attributes

rjlynch (approved):
Looks good.

Do feel like we're going to want subclasses of DSI User soon, ServiceAdmin, ServiceOperator, and FeProvider

asmega:
> Do feel like we're going to want subclasses of DSI User soon, ServiceAdmin, ServiceOperator, and FeProvider

Not sure. If we're sticking with roles then maybe we won't head that way. We will see

app/forms/admin/amendment_form.rb
asmega: You're right just had a quick search and seems to be localised to "amending" only. I thought these were more widespread concept used in jobs and scrubbing but does not seem to be the case. I'll have a quick play now and see if I can pull this off the claim

asmega: I've now moved this from the `Claim` model to the form

PR #4871, opened 2026-08-07
Add multiuse to service access codes

1993jamesgrant:
Adds a multiple use option for generated magic links when we don't know the number of claimants a school needs to distribute the link to.

### DIFF
--- a/app/views/admin/service_access_links/show.html.erb
+++ b/app/views/admin/service_access_links/show.html.erb
@@ -23,17 +23,16 @@
     </p>
 
     <p class="govuk-body">
-      This link will expire
-      on <strong><%= l(@service_access_code.expires_at.to_date) %></strong>
+      This link will expire on <strong><%= l(@service_access_code.expires_at.to_date) %></strong>, 
+      the claimant can restart the journey as many times as they like.
     </p>
 
     <p class="govuk-body">
-      The link can only be used to submit one claim.
-    </p>
-
-    <p class="govuk-body">
-      Until the link expires, the claimant can restart the journey as many times
-      as they like.
+      <% if @service_access_code.multiuse? %>
+        This link can be used multiple times.
+      <% else %>
+        The link can only be used to submit one claim.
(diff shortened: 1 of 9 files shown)


kenfodder:
- Should we have a way to deactivate a multi-use link?
- A multi-use link still flips `used=true` on the first use? A boolean for this doesn't make sense more of a side-effect.
- I noticed we use `multiuse` and `multi_use`, might be worth making that consistent.

db/migrate/20260805120000_add_multiuse_to_journeys_service_access_codes.rb
```diff
@@ -0,0 +1,5 @@
+class AddMultiuseToJourneysServiceAccessCodes < ActiveRecord::Migration[8.0]
```
kenfodder: Should be `8.1`?

app/views/admin/service_access_links/show.html.erb
```diff
@@ -27,14 +27,20 @@
...
+      <p class="govuk-body">
+        The link can only be used to submit one claim.
+      </p>
 
-    <p class="govuk-body">
-      Until the link expires, the claimant can restart the journey as many times
-      as they like.
-    </p>
+      <p class="govuk-body">
+        The link expires on <strong><%= l(@service_access_code.expires_at.to_date) %></strong>, the claimant can restart the journey as many times
```
kenfodder: The page states the expiry date twice on the page now, is this ok with product/content?

1993jamesgrant:
> * Should we have a way to deactivate a multi-use link?
Don't think this is needed
> * A multi-use link still flips `used=true` on the first use? A boolean for this doesn't make sense more of a side-effect.
That's fine - the logic is that it has been used, which still applies even on multiuse
> * I noticed we use `multiuse` and `multi_use`, might be worth making that consistent.
Good point, resolved

asmega (approved):

PR #292, opened 2019-08-15
Enable increased logging for app and VSP

erbridge:

PR #4874, opened 2026-08-12
[#292] Further back office role restrictions

asmega:
# Context

- https://github.com/DFE-Digital/claim-issues/issues/292
- Adds restrictions to role service operators


rjlynch (approved):
Looks good to me.
`policy(:DqtHigherEducationQualificationUpload)` IDK why but this looks off to me, even though I checked and it works fine!

asmega:
> Looks good to me. `policy(:DqtHigherEducationQualificationUpload)` IDK why but this looks off to me, even though I checked and it works fine!

its whatever you name the class and its a new class. touch points are very minimal so easy to rename if we want to

PR #4877, opened 2026-08-13
45/verify ammended bank details with hmrc

rjlynch:
# Verify bank detail amendments with hmrc.

When the ops team amend the bank details on a claim we now check the bank
details against the hmrc api and display the result as a notification banner.

The first commit does some faffing to make it easier to setup notification
banners. The second commit makes the api call to hmrc and shows the result in a
banner.

### DIFF
--- a/app/controllers/admin/amendments_controller.rb
+++ b/app/controllers/admin/amendments_controller.rb
@@ -19,6 +19,12 @@ class Admin::AmendmentsController < Admin::BaseAdminController
     )
 
     if @form.valid? && @form.save
+      if @form.amendment.bank_details_changed?
+        Admin::Claims::HmrcValidationJob.perform_now(@claim)
+
+        flash[:hmrc_bank_verification_claim_id] = @claim.id
+      end
+
       redirect_to admin_claim_tasks_url(@claim), notice: "Claim has been amended successfully"
     else
       render "new"
--- a/app/views/admin/tasks/index.html.erb
+++ b/app/views/admin/tasks/index.html.erb
@@ -4,9 +4,7 @@
   <%= govuk_back_link href: claims_backlink_path %>
 <% end %>
 
-<% if @banner_messages.any? %>
-  <%= render partial: "banner", locals: { messages: @banner_messages } %>
-<% end %>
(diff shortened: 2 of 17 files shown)


app/controllers/admin/amendments_controller.rb
```diff
@@ -19,6 +19,12 @@ def create
     )
 
     if @form.valid? && @form.save
+      if @form.amendment.bank_details_changed?
+        Admin::Claims::HmrcValidationJob.perform_now(@claim)
```
1993jamesgrant, line 23: This ties the success of the external action to the controller. Perhaps consider a way to decouple? HMRC failure/delay would lead to strange user experiences

app/views/admin/tasks/index.html.erb
```diff
@@ -4,8 +4,8 @@
   <%= govuk_back_link href: claims_backlink_path %>
 <% end %>
 
-<% if @banner_messages.any? %>
-  <%= render partial: "banner", locals: { messages: @banner_messages } %>
+<% @banners.each do |banner| %>
```
1993jamesgrant: Can do 
<%= render partial: "banner", collection: @banners, as: :banner %>

app/controllers/admin/amendments_controller.rb
rjlynch: `Hmrc::Client` handles any non success responses from the api. Agree that it would be better to enqueue the job and use something like broadcast replace to show the banner but admin site isn't really set up for that.

rjlynch: I guess the other option is to redirect to a waiting page if the amendment triggers a hmrc api call, poll on there and redirect when the job has finished running. Not sure if it's worth doing though as amending bank details isn't a frequent activity.

asmega (approved):

app/controllers/admin/amendments_controller.rb
1993jamesgrant: We could probably put a default timeout? E.g. 5 seconds, want to avoid an admin thrashing refresh button

PR #4880, opened 2026-08-13
Block auto automatic_approvals on demand

1993jamesgrant:
Journey configurations are always created

block auto approvals on the whole policy

### DIFF
--- a/app/controllers/admin/journey_configurations_controller.rb
+++ b/app/controllers/admin/journey_configurations_controller.rb
@@ -1,7 +1,8 @@
 module Admin
   class JourneyConfigurationsController < BaseAdminController
     helper_method :journey_configuration
-    before_action :ensure_service_operator, :journey_configuration
+    before_action :journey_configuration
+    before_action :ensure_service_admin, only: [:update], if: -> { automatic_approvals_update_requested? }
     after_action :send_reminders, only: [:update]
 
     FILE_UPLOAD_TARGET_DATA_MODELS = {
@@ -20,6 +21,7 @@ module Admin
 
     def update
       if journey_configuration.update(journey_configuration_params)
+        flash[:success] = automatic_approvals_flash_message if journey_configuration.saved_change_to_automatic_approvals?
         redirect_to edit_admin_journey_configuration_path(journey_configuration)
       else
         load_edit_data
@@ -67,7 +69,8 @@ module Admin
           :close_at,
           :open_for_submissions,
           :current_academic_year,
-          :teacher_id_enabled
+          :teacher_id_enabled,
+          :automatic_approvals
         )
 
       permitted[:close_at] = Time.zone.parse(permitted[:close_at]).in_time_zone("London") if permitted[:close_at].present?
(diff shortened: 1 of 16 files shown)


app/forms/journeys/teacher_student_loan_reimbursement/claim_school_form.rb
```diff
@@ -16,15 +16,15 @@ class ClaimSchoolForm < Form
...
-
-        if changed_query?
+        if possible_claim_school_id.present?
+          if changed_possible_school? || changed_query?
+            journey_session.answers.assign_attributes(
+              claim_school_id: possible_claim_school_id,
+              possible_claim_school_id: possible_claim_school_id
+            )
+            reset_dependent_answers
+          end
```
rjlynch: Should this be in the commit?

app/views/student_loans/claims/claim_school.html.erb
```diff
@@ -14,7 +14,7 @@
       <%= f.govuk_school_search \
         :provision_search,
         search_path: school_search_index_path,
-        result_attribute: :possible_school_id,
+        result_attribute: :possible_claim_school_id,
```
rjlynch: Do we need this in the commit?

app/models/journeys/base.rb
```diff
@@ -1,7 +1,7 @@
 module Journeys
   module Base
     def configuration
-      Configuration.find(routing_name)
+      Configuration.find_or_create_by!(routing_name: routing_name)
```
rjlynch: Do we need to add this for automatic approvals? Not sure if we want a new API for creating journey configs? Currently we assume all configs are setup in the seeds

app/controllers/admin/journey_configurations_controller.rb
```diff
@@ -1,7 +1,8 @@
 module Admin
   class JourneyConfigurationsController < BaseAdminController
     helper_method :journey_configuration
-    before_action :ensure_service_operator, :journey_configuration
+    before_action :journey_configuration
+    before_action :ensure_service_admin, only: [:update], if: -> { automatic_approvals_update_requested? }
```
rjlynch, line 5:
```suggestion
    before_action :ensure_service_admin
```
Would be tempted to make this whole end point admin only.

spec/features/student_loans_claim_spec.rb
```diff
@@ -7,7 +7,7 @@
   let!(:school) { create(:school, :student_loans_eligible) }
   let(:imported_slc_data) { create(:student_loans_data, nino: "PX321499A", date_of_birth: "28/2/1988", plan_type_of_deduction: 1, amount: 1_100) }
 
-  def answer_eligibility_questions_and_fill_in_personal_details
+  def answer_eligibility_questions_and_fill_in_personal_details(javascript_enabled: false)
```
rjlynch: Should these changes be in a separate pr?

rjlynch (changes requested):
Over all looks fine. Seems like some school search changes have made their way into this pr, probably want to drop those.

Do we want `Journey.configuration` to create a new record or should it continue to throw if the configuration isn't set up correctly? I'm leaning towards the latter, we don't want to accidentally persist an journey config we didn't intend to setup.

app/controllers/admin/journey_configurations_controller.rb
1993jamesgrant: Good shout - I'll get this in a separate dev request from prod

app/models/journeys/base.rb
1993jamesgrant: Reverted this specific change and added a comment clarifying why this the design choice

rjlynch (approved):
LGTM 👍

PR #4887, opened 2026-08-19
[#313] Admin sign out also calls DfE Sign In

asmega:
# Context

- https://github.com/DFE-Digital/claim-issues/issues/313


rjlynch (approved):
LGTM 👍

PR #4905, opened 2026-08-25
276 add award amount to claim model

kenfodder:


asmega:
this is no longer needed. we are currently dual writing to both columns at the moment. so that should cover any work in these changes. there still needs to be some work done to unwire the old column

PR #4909, opened 2026-08-26
[#209] Prepare Claim#award_amount migration

asmega:
# Context

- epic: https://github.com/DFE-Digital/claim-issues/issues/209
- ticket - https://github.com/DFE-Digital/claim-issues/issues/276
- this is preparation work before we backfill `Claim#award_amount` from `Eligibility#award_amount`

# Changes

- add blank column `Claim#award_amount`
- changes to support the addition of this blank column
- `Claim#award_amount` method reads the blank column otherwise if `nil` delegates to original data `Eligibility#award_amount`
- change getters/readers to use `Claim#award_amount` method above
- we double write. So on claim submission and claim amendment we set both `Claim#award_amount` and `Eligibility#award_amount`

# Next steps

- backfill data
- pretend to delete `Eligibility#award_amount` as to unwire everything, then we can wire everything to `Claim#award_amount`


app/forms/journeys/early_years_teachers_financial_incentive_payments/check_your_answers_form.rb
```diff
@@ -67,6 +67,8 @@ def build_claim
             end
           end
 
+          claim.write_attribute(:award_amount, Policies::EarlyYearsTeachersFinancialIncentivePayments.award_amount)
```
rjlynch: How come we're using write attribute here?

rjlynch (approved):
LGTM

> pretend to delete Eligibility#award_amount as to unwire everything

I think this should catch most things but would check the payroll run calculations separately as they query the eligibility#award_amount in sql

app/forms/journeys/early_years_teachers_financial_incentive_payments/check_your_answers_form.rb
asmega: my mistake, have now changed to use setter method `claim.award_amount = ...`

PR #4915, opened 2026-09-02
Fix formatting for provider_verification_chase_email_last_sent_at

1993jamesgrant:
Remove the word to avoid confusion


rjlynch (approved):

rjlynch:
think the config linter might complain if there's no description though

PR #4916, opened 2026-09-03
[#401] Journey configs uses uuid pkey over routing name

asmega:
# Context

- https://github.com/DFE-Digital/claim-issues/issues/401
- alter table `journey_configurations` so it is now backed by a `uuid` column named `id` as primary key rather than `routing_name` which is a string
- this will enable `routing_name` to contain duplicates which is not possible now


rjlynch (approved):
👍 - would have expected this to be more painful!

PR #4920, opened 2026-09-03
[#368] Add toggle for new STRI journey

asmega:
# Context

- https://github.com/DFE-Digital/claim-issues/issues/368
- Add new STRI journey that can be enabled or disabled with feature flag `new_stri`
- New journey continues to use existing policy `Policies::TargetedRetentionIncentivePayments`
- When claim is submitted, tasks created depend on the policy and academic year. at the moment it does not create any tasks
- The new journey creates some randomly generated dummy data, so claim can be viewed in admin back office area


app/controllers/base_public_controller.rb
```diff
@@ -12,6 +12,8 @@ def clear_claim_session
   end
 
   def add_view_paths
+    return unless journey
```
rjlynch, line 15: How do we hit this without a journey?

config/routes.rb
```diff
@@ -57,7 +57,7 @@ def matches?(request)
   end
 
   # Define the generic routes that aren't specific to any given journey
-  scope path: ":journey", constraints: {journey: %r{#{Journeys.all_routing_names.join("|")}}} do
+  scope path: ":journey", constraints: {journey: %r{#{Journeys.all.map(&:routing_name).join("|")}}} do
```
rjlynch, line 60: `Journeys.available.map(&:routing_name)`
Is there any reason to generate routes for unavailable journeys?

app/models/policies/targeted_retention_incentive_payments/claim_checking_tasks.rb
```diff
@@ -4,6 +4,16 @@ module Policies
   module TargetedRetentionIncentivePayments
     class ClaimCheckingTasks < Policies::ClaimCheckingTasks
       def applicable_task_names
+        if claim.academic_year <= AcademicYear.new("2025")
+          task_names_2025_2026_and_before
```
rjlynch, line 8: Not sure if we're there yet, and checking is prob out of scope of this ticket, but now we've persisted and back filled claim tasks could this just be `claim.tasks`?

rjlynch (approved):
LGTM - left some comments but don't think they're blockers

config/routes.rb
asmega: Yes, my understanding is that routes will be cached at boot time.  Therefore if the toggle is off it won't generate the routes and when the feature is toggled on there will be missing routes. So we just add all the routes. But I've added a guard on the controller so at runtime we can protect against unavailable journeys.

The alternative is we have to implement some kind of mechanism to reload routes which we would need to sync across all web apps.

app/models/policies/targeted_retention_incentive_payments/claim_checking_tasks.rb
asmega: I can do that check. Should save a whole load of complexity if that work is complete

app/controllers/base_public_controller.rb
asmega: all routes are generated regardless of whether they are available or not. but if a journey is toggled off then in won't be in the `Journeys.available` list, so would be `nil` here

app/models/policies/targeted_retention_incentive_payments/claim_checking_tasks.rb
asmega: Numbers don't match so leaving this as is as a workaround. We can work to get those to match at a later date

PR #4925, opened 2026-09-08
Enable DfE Analytics in review environments

asatwal:
#Context
- We need to test if DfE Analytics works in a review environment
- This is in preparation for Airbyte testing.

### DIFF
--- a/config/initializers/dfe_analytics.rb
+++ b/config/initializers/dfe_analytics.rb
@@ -2,7 +2,7 @@ DfE::Analytics.configure do |config|
   # Whether to log events instead of sending them to BigQuery.
   #
   # config.log_only = true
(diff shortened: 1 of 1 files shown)


config/initializers/dfe_analytics.rb
```diff
@@ -41,7 +41,7 @@
   # enable analytics. You might want to hook this up to a feature flag or
   # environment variable.
   #
-  config.enable_analytics = proc { Rails.env.production? }
+  config.enable_analytics = proc { Rails.env.production? || Rails.env.review? }
```
asmega, line 44: this will likely not work as desired. all deployed apps boot in production mode. the bit deleted from line 5 probably wants to move into this line?

PR #4927, opened 2026-09-10
[#445] Add academic year to claim filtering

asmega:
# Context

- https://github.com/DFE-Digital/claim-issues/issues/445

### DIFF
--- a/app/forms/admin/claims_filter_form.rb
+++ b/app/forms/admin/claims_filter_form.rb
@@ -14,23 +14,37 @@ class Admin::ClaimsFilterForm
   end
 
   def filters_changed?
-    if filters[:team_member].blank? ||
-        filters[:policy].blank? ||
-        filters[:status].blank?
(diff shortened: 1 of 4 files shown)


app/forms/admin/claims_filter_form.rb
```diff
@@ -33,6 +35,18 @@ def team_member
     @team_member ||= filters[:team_member] || session[:filter]["team_member"] || "all"
   end
 
+  def academic_year
+    return AcademicYear.previous if reset?
```
rjlynch:
```suggestion
    return default_academic_year if reset?
```
Should we reset to the default here?

rjlynch (approved):
LGTM - left one question but not a blocker if it was intended

app/forms/admin/claims_filter_form.rb
asmega: good spot. sorted now

PR #4930, opened 2026-09-11
Allow filtering tasks by academic year

rjlynch:
Adds an academic year select to the admin task list page so ops can continue to work on the previous AY's claims.


rjlynch:
Considering not nesting the params under the `admin_task_list_form` key as the list of query params is getting up to 3K chars. If we start getting url too long errors on prod dropping that from params is probably the easiest fix.

asmega (approved):

PR #4931, opened 2026-09-11
[#449] Admin dashboard year

asmega:
# Context

- https://github.com/DFE-Digital/claim-issues/issues/449

### DIFF
--- a/app/models/admin/dashboard.rb
+++ b/app/models/admin/dashboard.rb
@@ -1,0 +1,64 @@
+module Admin
+  class Dashboard
+    attr_reader :academic_year
+
+    def initialize(academic_year: AcademicYear.current)
+      @academic_year = academic_year
+    end
+
+    def claims_received_for_year
+      @claims_received_for_year ||= Claim
+        .where(academic_year: academic_year)
+        .count
+    end
+
+    def claims_approved_for_year
+      @claims_approved_for_year ||= Claim
+        .where(academic_year: academic_year)
+        .approved
+        .count
+    end
+
+    def claims_rejected_for_year
+      @claims_rejected_for_year ||= Claim
+        .where(academic_year: academic_year)
+        .rejected
+        .count
+    end
+
+    def all_claims_received
+      @all_claims_received ||= Claim.count
+    end
+
+    def claims_approaching_deadline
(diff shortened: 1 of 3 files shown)


app/views/admin/page/index.html.erb
```diff
@@ -2,30 +2,38 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-one-half">
-    <h2 class="govuk-heading-m">Claims received (<%= AcademicYear.current %>)</h2>
-    <p class="govuk-body govuk-!-font-size-36"><%= @claims_received %></p>
+    <div class="govuk-body">
+      <%= @dashboard.academic_year == AcademicYear.previous ? AcademicYear.previous : govuk_link_to(AcademicYear.previous, url_for(academic_year: AcademicYear.previous.start_year)) %> | <%= @dashboard.academic_year == AcademicYear.current ? AcademicYear.current : govuk_link_to(AcademicYear.current, url_for) %>
```
rjlynch, line 6:
```suggestion
      <%= @dashboard.academic_year == AcademicYear.previous ? AcademicYear.previous : govuk_link_to(AcademicYear.previous, url_for(academic_year: AcademicYear.previous)) %> | <%= @dashboard.academic_year == AcademicYear.current ? AcademicYear.current : govuk_link_to(AcademicYear.current, url_for) %>
```

Do we need to link to the start year? If so do we need to link to the start year for AcademicYear.current too?

app/models/admin/dashboard.rb
```diff
@@ -0,0 +1,64 @@
...
+              eligibility_type: "Policies::FurtherEducationPayments::Eligibility",
+              academic_year: academic_year
+            }
+          )
+          .passed_automatically
+          .where(name: "provider_verification")
+          .count
+    end
+  end
+end
```
rjlynch, line 64: Way nicer 👍

rjlynch (approved):
LGTM - left one comment not sure if it's an issue or not

app/views/admin/page/index.html.erb
asmega: if its blank it defaults to current academic year. so on first page load it will default to current academic year. otherwise if its present, it'll use that year. i used the start year just so its encodes a bit more cleanly in the url query string

PR #4950, opened 2026-09-17
2929 - Amends WAF policy settings

steves-mobilise:
## Context

Front Door rate limiting settings should be applied and standardised across all services.

- Check whether the service has a rate limit set (either via `rate_limit` or `rate_limit_max`)
- Investigate if `allow_aks` is set on each domain of the service
- Ensure `block_ip` is also set as a disabled rule

## Changes proposed in this pull request

I propose to add `rate_limit_max` to both the **staging** and **test** environments - the former at 100 requests per IP every 5 minutes, and the latter at 50.

This change will also add both `allow_aks` and `block_ip` to the configuration, so that the requested WAF policies are applied.

**PLEASE NOTE:** this is not yet being applied to Production.

## Guidance to review

Simply run both `make staging domains-plan` and `make test domains-plan` to ensure that the three custom rules `all` (`RateLimitRule`), `AKS` and `blockIP` are applied.

Then also just run `make production domains-plan CONFIRM_PRODUCTION=yes`, to ensure the changes are not applied to the prod environment.

## Link to Trello card

[Link](https://trello.com/c/Ct1TypdW/2929-rate-limiting-for-all-services)

## Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history
- [x] Applied DevOps label


hopep-mobilise (approved):
Plans produced the expected resource creation including the azurerm_cdn_frontdoor_security_policy. No changes occurred in production. I know you've checked the appropriate rate limiting values and discussed with the service.

